PR #8, opened 2023-12-13
BZ-9747: feat: added support to prevent auto focus in ListItem component

jaissumi1402:
- Added conditional class in listItem component based on focusable value
- Added a css class to prevent focus in ListItem


sinha-sahil (dismissed):

sinha-sahil (commented):
Lets try using conditional classes

sinha-sahil (approved):

PR #26, opened 2024-02-02
BZ-12741: feat: Added background color property for icon image in bad…

NikitaGupta-17:
- Added background color property for icon image in badge component


prajwal-jp:
@NikitaGupta-17 can we add a property to set badge value at center?
image

NikitaGupta-17:
Badge value are already assign in center.

sinha-sahil (approved):

PR #45, opened 2024-05-15
BZ-21221: feat: svelte ui component dropdown multiple selection

kingk5:
- added a function to handle selected state and behave accordingly to string or array of string
- why?
   - Select component is having multiple selected items because of includes.
- Solution?
   - added a function to handle selected state and behave accordingly to string or array of string
- What to test?
   - test the select component
   
 When setting item or list of items as selected instead of using includes, A function is written to string match after trimming for single item and use includes for list of items.
 These will impact the project in following ways:-

1. Instead of using includes for single items comparison we use string match because if selected item is XYZAB and we check for XYZ behavior was `XYZ` also gets selected which should not be done. (Fixed using the new function)
2.  For the list of items the same behavior of includes is being followed

   
Dev Proofs:-
https://drive.google.com/file/d/1X5jkKWAXpw7NnONBN6bv_HUIAWZCgdEr/view?usp=sharing

### DIFF
--- a/src/lib/Select/Select.svelte
+++ b/src/lib/Select/Select.svelte
@@ -55,13 +55,21 @@
       : item !== properties.selectedItem
   );
 
+  function isSelected(selectedItem: string | string[], item: string) {
+    if (Array.isArray(selectedItem)) {
+      return selectedItem.includes(item);
+    } else {
+      return selectedItem.trim() === item.trim();
+    }
+  }
+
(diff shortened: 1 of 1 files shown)


sinha-sahil:
@kingk5 kindly, update the description to contain details about changes added and how they impact the project.

rab1prasad (approved):

ARX1718 (approved):

PR #51, opened 2024-05-27
Updated InputButton by adding OnFocusOut event 

Priyanshu-p-juspay:
- added onFocusOut event for InputButton
- implemented the same function as present for Input


sinha-sahil:
@Priyanshu-p-juspay The PR doesn't follows the standard defined for raising PRs. Refer : https://github.com/juspay/svelte-ui-components/pull/32 .
Install husky by `npx install husky`.
Make sure to only use one commit per PR.

PR #53, opened 2024-05-27
BZ-22447:feat: Added support to show single button in select multiple items in dropdown

ikra2407:
- Added an optional property in selectMultipleItems to show either a single select button or multiple buttons
 - Updated UI of the component
 - Added css for
    - apply button
 - Added css variables in
    - label-container
    - multipleSelect-btn
    - non-selected-items
    - selected-content
 - Added condition to enable select button only when an item is selected
 DEV PROOF
 https://drive.google.com/file/d/1tUmd8w_Bm5VDN4LsLB7K7BN49STrHyjb/view?usp=sharing

### DIFF
--- a/src/lib/Select/Select.svelte
+++ b/src/lib/Select/Select.svelte
@@ -4,7 +4,6 @@
   import type { ButtonProperties } from '$lib/Button/properties';
   import Img from '$lib/Img/Img.svelte';
   import Button from '$lib/Button/Button.svelte';
-  import { InputButton, defaultInputButtonProperties } from '$lib';
 
   let selectedElementDiv: HTMLDivElement | null = null;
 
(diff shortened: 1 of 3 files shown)


src/lib/Select/Select.svelte
```diff
@@ -23,13 +23,13 @@
   const dropDownIcon =
     properties.dropDownIcon ?? 'https://sdk.breeze.in/gallery/icons/down-arrow.svg';
 
-  const applyButtonProps: ButtonProperties = {
-    text: 'Apply',
-    enable: true,
+    $: applyButtonProps = {
```
sinha-sahil, line 26:
```
let applyButtonProps: ButtonProperties;
$: applyButtonProps = { .. }
```

kingk5 (approved):

PR #60, opened 2024-06-18
BZ-12821: feat: Add right button css variables for input button component

prajwal-jp:
- Introduced CSS variables for the right button in the input button component
- Added a new parameter button-content-flex-direction to control the flex direction within  button component

Screenshot 2024-06-18 at 5 13 59 PM
Screenshot 2024-06-18 at 5 14 31 PM

### DIFF
--- a/src/lib/InputButton/InputButton.svelte
+++ b/src/lib/InputButton/InputButton.svelte
@@ -79,7 +79,7 @@
       />
     </div>
     {#if properties.rightButtonProperties != null}
-      <div class="button">
+      <div class="right-button">
         <Button properties={properties.rightButtonProperties} on:click={rightButtonClick} />
       </div>
     {/if}
@@ -127,12 +127,6 @@
   .input {
     flex: 2;
     min-width: 0px;
-  }
-
-  .button {
-    flex: 1;
-    min-width: 0px;
-    --button-height: 54px;
   }
 
   .bottom-button {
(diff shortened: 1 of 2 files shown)


src/lib/InputButton/InputButton.svelte
```diff
@@ -129,12 +129,6 @@
     min-width: 0px;
   }
 
-  .button {
-    flex: 1;
-    min-width: 0px;
```
sandeepsj, line 134: is it safe to remove this min-width in new class?

prajwal-jp: Adding it back as it might have unseen usecases

sandeepsj (approved):

Priyanshu-p-juspay (approved):

src/lib/Button/properties.ts
```diff
@@ -1,10 +1,12 @@
 export type LoaderType = 'Circular' | 'ProgressBar';
+export type CircularLoaderPosition = 'start' | 'end';
 
 export type ButtonProperties = {
   text: string;
   enable: boolean;
   showLoader: boolean;
   loaderType: LoaderType | null;
+  circularLoaderPosition: CircularLoaderPosition | null;
```
wardhan28: can we make this field not nullable and initialise with 'start' ?

PR #64, opened 2024-06-25
BZ-25866: feat: Add support for making carousel container reactive

NikitaGupta-17:
- Use ResizeObserver to fetch when parent container width is changed and set the child width accordingly
- Set the parent width as window.innerWidth by default and update it according to resizeobserver

### DIFF
--- a/src/lib/Carousel/Carousel.svelte
+++ b/src/lib/Carousel/Carousel.svelte
@@ -10,10 +10,10 @@
   let startTouch: number;
   let startMouse: number;
   let endMouse: number;
-  let carouselWidth: string;
   let carouselDiv: HTMLDivElement | undefined;
   let activeSlideIndex = 0;
   let widthUnits: string;
+  $: carouselWidth = '300px';
 
   function nextSlide() {
     if (activeSlideIndex != properties.views.length - 1 || properties.isScrollableLast) {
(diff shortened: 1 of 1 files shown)


src/lib/Carousel/Carousel.svelte
```diff
@@ -108,9 +108,23 @@
   }
 
   onMount(() => {
+    const carouselContainer = document.querySelector('.carousel-container');
+
+    const resizeObserver = new ResizeObserver((entries) => {
+      for (let entry of entries) {
+        if (entry.target === carouselContainer) {
+          let width = getComputedStyle(carouselContainer).getPropertyValue('width');
```
prajwal-jp:
```suggestion
          carouselWidth = getComputedStyle(carouselContainer).getPropertyValue('width');
```

src/lib/Carousel/Carousel.svelte
```diff
@@ -108,9 +108,22 @@
   }
 
   onMount(() => {
+    const carouselContainer = document.querySelector('.carousel-container');
```
sandeepsj, line 111: We shouldn't do this. instead use bind:this in <div class="carousel-container" >

src/lib/Carousel/Carousel.svelte
```diff
@@ -128,7 +141,7 @@
   });
 </script>
 
-<div class="carousel-container">
+<div class="carousel-container" style="--carousel-container-width: {carouselWidth};">
```
sandeepsj, line 144: Move style into .carousel-container inside style tag

PR #133, opened 2025-06-26
BZ-42256: feat: Add support for asynchronous validator in input

rab1prasad:
- Currently in input component we don't have support for async validators need to add async validators support in input box.
- Add support for asynchronous input validator.
- Change return type of custom validator for input component.

Dev proof:
https://drive.google.com/file/d/1dEH7HftNeAqsOeqesCC6NrqDU4fseYFf/view?usp=sharing


gokulJuspay (approved):

src/lib/Input/Input.svelte
```diff
@@ -9,11 +9,16 @@
   export let properties: InputProperties = defaultInputProperties;
   let inputElement: HTMLInputElement | HTMLTextAreaElement;
 
-  $: state = getValidationState(properties) as ValidationState;
+  let state: ValidationState = 'InProgress';
+  $: {
```
sinha-sahil: avoid adding reactive blocks

itz-PrathamMittal: Done

vigneshJuspay (approved):

pratikgiramkar1 (approved):

Nikita172002 (approved):

src/lib/utils.ts
```diff
@@ -40,8 +40,8 @@ export function validateInput(
       break;
   }
 
-  customValidators.forEach((validator: CustomValidator) => {
-    const currentResult = validator(inputValue, validationResult);
+  for (const validator of customValidators) {
+    const currentResult = await Promise.resolve(validator(inputValue, validationResult));
     if (currentResult === 'Invalid') {
       validationResult = 'Invalid';
```
prajwal-jp, line 46: once you reach invalid state, you can just return the result. no need to iterate over remaining validators

src/lib/Input/Input.svelte
```diff
@@ -9,11 +9,16 @@
...
+
+  $: (async () => {
+    state = await getValidationState(properties);
+  })();
+
 
   // For making this function reactive, prop was passed as param
-  function getValidationState(prop: InputProperties): ValidationState {
-    const valueValidation: ValidationState = validateInput(
+  async function getValidationState(prop: InputProperties): Promise<ValidationState> {
```
prajwal-jp:
```
async function getValidationState(prop: InputProperties): Promise<ValidationState> {
    const valueValidation: ValidationState = await validateInput(
      prop.value,
      prop.dataType,
      prop.validationPattern,
      prop.inProgressPattern,
      prop.validators
    );
    if (
      valueValidation === 'InProgress' &&
      prop.value.length > 0 &&
      inputElement &&
      inputElement !== document.activeElement
    ) {
      state = 'Invalid';
    } else {
      state = valueValidation;
    }
  
```

prajwal-jp: `$: getValidationState(properties);`

src/lib/utils.ts
itz-PrathamMittal: done

prajwal-jp (approved):

RajuSudhar (approved):

murdore:
Thanks for the contribution — async validator support is a genuine gap worth filling. To land this: (1) rebase on `release` (the branch is behind; the CHANGELOG deletion suggests a bad merge base) and restore the full CHANGELOG history rather than replacing it; (2) add a cancellation-flag or `AbortController` guard inside the `$effect` so a stale async result from a previous keystroke can't overwrite the current validation state; (3) update `docs/Input.md` to show the `ValidationState | Promise<ValidationState>` return type and add a demo section with a simulated async example. Note that PR #246 (merged) refactored the same `$effect` / `onStateChange` path in `Input.svelte` — rebase will surface any conflicts there that need resolving before review.

murdore:
One more on the validator loop (separate from the race-condition / CHANGELOG points in my earlier comment): the `forEach`→`for` refactor added a `break` on the `'InProgress'` branch in `src/lib/utils.ts`. That makes `'InProgress'` short-circuit the loop, so a later validator that would return a definitive `'Invalid'` never runs — whereas the old `forEach` ran every validator and a later `'Invalid'` always won.

Since `Input.svelte` only promotes a lingering `'InProgress'`→`'Invalid'` on blur (value non-empty and unfocused), a real format error is now hidden while the field is still focused, instead of being flagged immediately as before. Suggest dropping the `break` on the `'InProgress'` branch (the one after the `'Invalid'` branch is fine), or explicitly documenting that validator order now determines short-circuit priority.

PR #137, opened 2025-08-19
BZ-43249: feat: Upgrade to Svelte 5 and refactor components to use Runes

itz-PrathamMittal:
- Upgraded Svelte and its ecosystem dependencies (Kit, Vite plugin, etc.) from v4 to v5 in `package.json`.
- Refactored several components (`Accordion`, `ModalAnimation`, `OverlayAnimation`, `Badge`, `Button`) to adopt the new Svelte 5 Runes API (`$props`, `$state`, `$effect`).
- Replaced legacy APIs like `createEventDispatcher` and `<slot>` with prop functions and the `{@render}` tag for modern, clearer component contracts.
- Enabled Runes mode globally by updating `svelte.config.js`.
- Overhauled `tsconfig.json` to align with the latest Svelte 5 standards.
- Migrated Vitest configuration from `vite.config.ts` to a dedicated `vitest.config.ts` file for better separation of concerns.
- Updated `pnpm-lock.yaml` to reflect the new dependency tree.


vite.config.ts
```diff
@@ -1,9 +1,10 @@
+// filepath: /Users/pratham.mittal/Desktop/breeze/svelte-ui-components/vite.config.ts
```
Nikita172002: remove this

src/lib/Badge/properties.ts
```diff
@@ -1,4 +1,4 @@
-export type BadgeProperties = {
-  image: string;
-  value: string;
-};
+// export type BadgeProperties = {
```
Nikita172002: Let's remove this file itself, if we are not using it

gokulJuspay (approved):

src/lib/Accordion/Accordion.svelte
```diff
@@ -1,39 +1,45 @@
 <script lang="ts">
-  import { afterUpdate, onMount } from 'svelte';
+  import type { Snippet } from 'svelte';
+
+  const emptySnippet = (() => {}) as Snippet;
```
pratikgiramkar1: Don't define emptySnippet like this. Using "as" hides type mismatches.

itz-PrathamMittal: done

PR #154, opened 2025-12-26
BZ-47595: feat: Add css variable for input box shadow

sandeepssj:
Add css variable for input components input-box shadow


sandeepssj:
- Passing the css variable as --input-box-shadow:
0px 0px 5px #000000image
    
- Without passing any value for     --input-box-shadow:
image

- Before the change: 
image

PR #192, opened 2026-05-29
feat: add Divider component

murdore:
## Summary

Adds a standalone `Divider` separator primitive — the library had no divider component, so consumers were hand-rolling `<div>`s with border/margin styles. Lighthouse alone has 92 such usages across 22 files (migration in flight, currently bridged via a local pnpm patch of this exact component).

## API

- `orientation?: 'horizontal' | 'vertical'` (default `horizontal`)
- `testId?: string` → `data-pw`
- `classes?: string` → appended to the root
- Exposed CSS variables: `--divider-border` (shorthand, default `1px solid #e2e8f0`), `--divider-margin` (default `0`), `--divider-length` (default `100%`)
- Renders `role="separator"` + `aria-orientation` for accessibility

## Notes

- Follows the existing component conventions (Svelte 5 runes, `properties.ts`, `index.ts` component + type exports), mirroring `Badge`.
- `pnpm run check` (svelte-kit sync + svelte-check): **0 errors, 0 warnings**. `prettier --check` + `eslint`: clean.
- This is the formalization of the component already validated in a downstream consumer (Lighthouse) via `pnpm patch`; once released, the consumer drops the patch and bumps the version.

## Summary by CodeRabbit

* **New Features**
  * Introduced a new Divider component available for use in the library. The component offers flexible orientation settings (horizontal and vertical) and supports custom styling and test attributes for enhanced testing capabilities.

[Review Change Stack](https://app.coderabbit.ai/change-stack/juspay/svelte-ui-components/pull/192?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lib/Divider/Divider.svelte
+++ b/src/lib/Divider/Divider.svelte
@@ -1,0 +1,26 @@
+<script lang="ts">
+  import type { DividerProperties } from './properties';
+
+  let { orientation = 'horizontal', testId, classes }: DividerProperties = $props();
+</script>
+
+<div
+  class="divider divider-{orientation} {classes ?? ''}"
+  role="separator"
+  aria-orientation={orientation}
+  data-pw={typeof testId === 'string' ? testId : null}
+></div>
+
+<style>
+  .divider-horizontal {
+    width: var(--divider-length, 100%);
+    border-top: var(--divider-border, 1px solid #e2e8f0);
+    margin: var(--divider-margin, 0);
+  }
+
+  .divider-vertical {
+    height: var(--divider-length, 100%);
+    border-left: var(--divider-border, 1px solid #e2e8f0);
+    margin: var(--divider-margin, 0);
+  }
+</style>
--- a/src/lib/Divider/properties.ts
+++ b/src/lib/Divider/properties.ts
@@ -1,0 +1,5 @@
+export type DividerProperties = {
+  orientation?: 'horizontal' | 'vertical';
+  testId?: string;
+  classes?: string;
+};
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -56,6 +56,7 @@ export { default as EmptyState } from './EmptyState/EmptyState.svelte';
 export { default as Combobox } from './Combobox/Combobox.svelte';
 export { default as ColorPicker } from './ColorPicker/ColorPicker.svelte';
 export { default as SplitInput } from './SplitInput/SplitInput.svelte';
+export { default as Divider } from './Divider/Divider.svelte';
 
 export type * from './Button/properties';
 export type * from './Modal/properties';
@@ -109,5 +110,6 @@ export type * from './EmptyState/properties';
 export type * from './Combobox/properties';
 export type * from './ColorPicker/properties';
 export type * from './SplitInput/properties';
+export type * from './Divider/properties';
 
 export { validateInput } from './utils';


sinha-sahil:
Thanks for putting this together, @murdore — I get the motivation: 92 ad-hoc `<div>`-with-border separators in Lighthouse is a real pain, and centralizing the pattern sounds appealing. I want to push back on the *shape* of the solution though, because as drafted it conflicts with two of the load-bearing rules of this library.

## Why this can't land as a component

### 1. It's a dummy component (philosophy #3)

The entire runtime contribution of `Divider.svelte` is:

```svelte
<div class="divider divider-{orientation}" role="separator" aria-orientation={orientation}></div>
```

…styled by a `border-top` (or `border-left`). There is no state, no focus management, no event handling, no composition, no derived behavior. The library's stated philosophy is that components must justify their existence with non-trivial behavior, accessibility logic, or composition value; a literal line on screen does not meet that bar. We already rejected this same shape conceptually for "blur", "spacer", etc.

The ARIA-semantics argument is the strongest reason to keep it, but native HTML already gives us this for free:

- `<hr>` has implicit `role="separator"` per the HTML AAM spec.
- For a vertical separator, `<div role="separator" aria-orientation="vertical" class="divider divider-vertical">` in consumer-land is one line and ships the exact same a11y surface as the component.

### 2. `orientation` is a baked-in variant system (philosophy #2, GUIDELINES §9)

```svelte
class="divider divider-{orientation} {classes ?? ''}"
```
```css
.divider-horizontal { width: var(--divider-length, 100%); border-top: var(--divider-border, 1px solid #e2e8f0); }
.divider-vertical   { height: var(--divider-length, 100%); border-left: var(--divider-border, 1px solid #e2e8f0); }
```

`orientation` is the same pattern as `variant` / `tone` / `size` — a prop value that flips which baked-in CSS class applies. §9 is unambiguous: variants live in consumer CSS classes that set CSS variables, passed in via `classes`. The component should never own the variant classes themselves. This is also why the customizability surface is awkward: a consumer who wants a gradient, dashed, or glyph divider has to override two different border sides depending on which orientation the component picked.

If we strip the `orientation` prop to comply with §9, what's left is `<div class="divider"></div>` — which makes the "this is just CSS" point self-evident.

## The customizable alternative — keep this in user-land

A consumer-side stylesheet handles 100% of the use case, including the Lighthouse migration, without a library component:

```css
/* lighthouse/app.css */
.divider {
  --divider-color: #e2e8f0;
  --divider-thickness: 1px;
  --divider-style: solid;
  display: block;
  background: none;
  border: 0;
}

.divider-horizontal {
  width: 100%;
  border-block-start: var(--divider-thickness) var(--divider-style) var(--divider-color);
}

.divider-vertical {
  height: 100%;
  border-inline-start: var(--divider-thickness) var(--divider-style) var(--divider-color);
}

/* Per-context theming — exactly how every other variant works in this codebase */
.divider-muted   { --divider-color: #f1f5f9; }
.divider-strong  { --divider-color: #94a3b8; --divider-thickness: 2px; }
.divider-dashed  { --divider-style: dashed; }
```

```svelte
<hr class="divider divider-horizontal" />
<div role="separator" aria-orientation="vertical" class="divider divider-vertical divider-muted"></div>
```

That's the same a11y, the same defaults, the same theming surface — and it composes with Lighthouse's design tokens instead of duplicating them through CSS variables on a library component.

## Nits, for completeness, if this somehow stays

- `border-top` / `border-left` are physical properties; `border-block-start` / `border-inline-start` would respect writing mode and let one rule handle both axes.
- `var(--divider-border, 1px solid #e2e8f0)` is a shorthand variable, which makes per-token overrides (thickness only, color only) impossible — consumers must restate the whole shorthand. Split into `--divider-color` / `--divider-thickness` / `--divider-style`.
- The `--divider-length: 100%` default is fine for the active axis but is wasted on the cross axis; logical sizing (`block-size` / `inline-size`) would let one variable serve both orientations.

## Ask

I'd like to close this PR and instead:
1. Land the Lighthouse migration as a shared CSS utility (in Lighthouse, or in a shared design-tokens package if that's the right home).
2. If there's a genuinely *behavioral* separator primitive we need later (e.g. a resizable splitter / draggable divider with focus + keyboard handling), that would be a proper library component — happy to scope that separately.

Not a reflection on the work — the code itself is clean (proper `typeof testId === 'string'` check, `classes ?? ''`, conventions all followed). It's the shape that doesn't fit the library. Happy to pair on the consumer-side CSS to unblock the Lighthouse rollout.

murdore:
@sinha-sahil — thank you for the thorough write-up. You're right on both counts, and I don't want to argue the point.

The "dummy component" objection lands: once `orientation` is stripped to comply with §9, what remains is literally `<div class="divider"></div>` — a styled div with no behavior, no focus management, and no composition value. Native `<hr>` already ships the `role="separator"` semantics for free, so even the a11y argument for a library primitive dissolves. And your point on `orientation` being the same shape as `variant`/`tone` is exactly right — §9 is unambiguous that those presets belong in consumer-side CSS classes passed through `classes`, not baked into the component.

I'm happy to close this PR. For the Lighthouse migration, I'll land a shared CSS utility (`.divider`, `.divider-horizontal`, `.divider-vertical`) in Lighthouse's own stylesheet, using the `border-block-start` / `border-inline-start` logical properties you called out and splitting `--divider-color`, `--divider-thickness`, `--divider-style` as separate tokens.

One question before I close: would it be worth adding a short consumer-pattern example to the library's docs (a "how to build a Divider in your app" snippet) so other consumers hit this answer immediately rather than opening another PR? Or is that overkill for something this simple?

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — appreciate you engaging with this directly rather than relitigating it. Your read is exactly right, and your migration plan (logical-property borders + three separate tokens for color/thickness/style) is the right shape. Go ahead and close this one.

On your docs question — **yes, that's a good instinct, and not overkill at all.** This isn't the first time a "looks like a component but is really CSS" pattern has come up (we have the same conversation periodically about spacers, dividers, fading edges, decorative blurs, container queries, etc.), and "open a PR to add it" is the natural default when no other path is documented. A small recipes section closes that loop.

A few notes on what would make it land well, vs. accidentally legitimize Divider-as-a-component:

1. **Don't put it under `docs/components/`.** That implies it's a shipped primitive. Better home is something like `docs/patterns/dividers.md` or a top-level `docs/RECIPES.md` with a "Things that look like components but are CSS" section. The framing matters — it should read as "here's how to build this in your app," not "here's our recommended divider."

2. **Lead with the principle, not the snippet.** One or two lines explaining *why* a divider isn't a library component (no state, no focus, no composition; `<hr>` already ships the a11y semantics) does more long-term work than the code itself. Future "should X be a component?" PRs get answered by linking to that paragraph.

3. **Show both the `<hr>` form and the `role="separator"` div form** (for the vertical case where `<hr>` is awkward), with the three-token CSS you're already planning to ship in Lighthouse. Use logical properties in the example so it reads as the canonical shape, not just one team's choice.

4. **List the sibling cases briefly.** Spacer, decorative line, blur overlay, fade edge, dotted leader — same family, same answer. One sentence each, no code. That way the recipes page is the bounce target for the whole class of "is this a component?" questions, not just dividers.

If you want to roll the docs page into the same Lighthouse-migration commit cycle, that's fine — or open a tiny separate PR against `docs/` here once the Lighthouse CSS is in place and you've validated the token surface in practice. I'd lean toward the latter so the docs reflect a real consumer's working setup rather than a hypothetical one.

Thanks again for taking the feedback straight on — that's the part that actually keeps the library coherent over time.

**Ask: please close this PR.** Open a follow-up against `docs/` once the Lighthouse-side CSS is settled, and I'll review that one quickly.

murdore:
Closing per our discussion — Divider is CSS, not a component (native <hr> ships the separator a11y semantics). I'll add a docs/ recipe once the Lighthouse-side CSS settles. Thanks @sinha-sahil.

PR #195, opened 2026-05-29
feat: add testId prop to EmptyState

murdore:
## Summary

Adds an optional `testId` prop to `EmptyState` that renders `data-pw={testId}` on the root `.empty-state` element — matching the `testId`→`data-pw` convention already used by `Banner`, `Divider`, `IconStack`, etc.

## Why

`EmptyState` previously had no test hook, forcing consumers to wrap it in an extra `<div data-pw>` for automation. Downstream migrations (Lighthouse `NoData` — 33 call-sites — and `ConnectionErrorState`) need a root test id; this lets them target the component directly.

## Notes
- Backward-compatible (optional prop, `null` when unset).
- `pnpm run check`: **0 errors, 0 warnings**; `prettier` + `eslint` clean.

### DIFF
--- a/src/lib/EmptyState/EmptyState.svelte
+++ b/src/lib/EmptyState/EmptyState.svelte
@@ -1,10 +1,10 @@
 <script lang="ts">
   import type { EmptyStateProperties } from './properties';
 
-  let { title, description, icon, children, classes }: EmptyStateProperties = $props();
+  let { title, description, icon, children, classes, testId }: EmptyStateProperties = $props();
 </script>
 
-<div class="empty-state {classes ?? ''}">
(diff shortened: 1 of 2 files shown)


sinha-sahil:
Reviewed against the philosophy — this is fine.

`testId` -> `data-pw` is the established convention across the library (`Banner.svelte:51` uses the same `typeof testId === 'string' ? testId : null` pattern), so this is pure consistency work, not a new abstraction. It's a test-automation escape hatch on a primitive — no styling, no presets, no English strings. Optional, defaults to `null`, fully backward-compatible.

LGTM.

Ask: approve & merge.

sinha-sahil (approved):
**Approving.**

PR #196, opened 2026-05-29
feat: add LabeledField component

murdore:
## Summary

- Adds `LabeledField`, a new single label-value display primitive to the component library
- Exposes 9 CSS custom properties for full layout and typography customization without overriding the design system
- Exports `LabeledField` component and `LabeledFieldProperties` type from the library root (`src/lib/index.ts`)

### Props

| `label` | `string` | yes | The label text displayed above the value |
| `value` | `string` | yes | The value text displayed below the label |
| `testId` | `string` | no | Sets `data-pw` attribute on the root element |
| `classes` | `string` | no | Additional CSS classes appended to the root element |

### CSS Custom Properties

| `--labeled-field-direction` | `column` | `flex-direction` of the root container |
| `--labeled-field-gap` | `4px` | Gap between label and value |
| `--labeled-field-align-items` | `flex-start` | `align-items` of the root container |
| `--labeled-field-label-color` | `inherit` | Label text color |
| `--labeled-field-label-font-size` | `12px` | Label font size |
| `--labeled-field-label-font-weight` | `500` | Label font weight |
| `--labeled-field-value-color` | `inherit` | Value text color |
| `--labeled-field-value-font-size` | `14px` | Value font size |
| `--labeled-field-value-font-weight` | `400` | Value font weight |

### DOM Structure

```html
<div class="labeled-field {classes}" data-pw="{testId}">
  <div class="labeled-field-label">{label}</div>
  <div class="labeled-field-value">{value}</div>
</div>
```

## Notes

- `prettier --check` passes cleanly
- `eslint` exits with code 0 (no issues)
- Font properties are set via CSS custom properties only — no hardcoded typography overrides in the component style block, consistent with the design system contract

### DIFF
--- a/src/lib/LabeledField/LabeledField.svelte
+++ b/src/lib/LabeledField/LabeledField.svelte
@@ -1,0 +1,31 @@
+<script lang="ts">
+  import type { LabeledFieldProperties } from './properties';
+
+  let { label, value, testId, classes }: LabeledFieldProperties = $props();
+</script>
+
+<div class="labeled-field {classes ?? ''}" data-pw={typeof testId === 'string' ? testId : null}>
+  <div class="labeled-field-label">{label}</div>
+  <div class="labeled-field-value">{value}</div>
+</div>
+
+<style>
+  .labeled-field {
+    display: flex;
+    flex-direction: var(--labeled-field-direction, column);
+    gap: var(--labeled-field-gap, 4px);
+    align-items: var(--labeled-field-align-items, flex-start);
+  }
+
+  .labeled-field-label {
+    color: var(--labeled-field-label-color, inherit);
+    font-size: var(--labeled-field-label-font-size, 12px);
+    font-weight: var(--labeled-field-label-font-weight, 500);
+  }
+
+  .labeled-field-value {
+    color: var(--labeled-field-value-color, inherit);
+    font-size: var(--labeled-field-value-font-size, 14px);
+    font-weight: var(--labeled-field-value-font-weight, 400);
+  }
+</style>
--- a/src/lib/LabeledField/properties.ts
+++ b/src/lib/LabeledField/properties.ts
@@ -1,0 +1,6 @@
+export type LabeledFieldProperties = {
+  label: string;
+  value: string;
+  testId?: string;
+  classes?: string;
+};
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -56,6 +56,7 @@ export { default as EmptyState } from './EmptyState/EmptyState.svelte';
 export { default as Combobox } from './Combobox/Combobox.svelte';
 export { default as ColorPicker } from './ColorPicker/ColorPicker.svelte';
 export { default as SplitInput } from './SplitInput/SplitInput.svelte';
+export { default as LabeledField } from './LabeledField/LabeledField.svelte';
 
 export type * from './Button/properties';
 export type * from './Modal/properties';
@@ -109,5 +110,6 @@ export type * from './EmptyState/properties';
 export type * from './Combobox/properties';
 export type * from './ColorPicker/properties';
 export type * from './SplitInput/properties';
+export type * from './LabeledField/properties';
 
 export { validateInput } from './utils';


sinha-sahil:
Thanks for the clean, well-documented PR — the prop table, CSS-var inventory, and lint hygiene are appreciated, and the code itself follows our `classes` and `testId` conventions correctly.

That said, I have to push back on this one at the architectural level, and I don't think any amount of code revision will land it. `LabeledField` falls squarely into the "composite wrapper" category that our design philosophy explicitly excludes from the library (see `docs/GUIDELINES.md` and the philosophy note in project memory). The library ships **primitives**; pairings of "a label and a value" — or more generally any `<Label><Thing/></Label>` / `<Card><Icon/><Text/><Button/></Card>` shape — belong in consumer code, not here.

Concretely, the entire body of the component is:

```svelte
<div class="labeled-field">
  <div class="labeled-field-label">{label}</div>
  <div class="labeled-field-value">{value}</div>
</div>
```

There's no state, no event, no a11y wiring (no `<label for>` association, no `aria-describedby`), no interaction — just two `<div>`s with two strings. Anything a consumer would want to do with this they can write inline in five lines, with full control over markup semantics (`<dl>/<dt>/<dd>`? a `<label>` + `<output>`? a flex row with an icon?). For example:

```svelte
<!-- consumer-land, fully owned by the app's design system -->
<div class="user-field">
  <span class="user-field-label">{label}</span>
  <span class="user-field-value">{value}</span>
</div>

<style>
  .user-field { display: flex; flex-direction: column; gap: 4px; }
  .user-field-label { font-size: 12px; font-weight: 500; color: var(--text-muted); }
  .user-field-value { font-size: 14px; color: var(--text-default); }
</style>
```

Two additional smaller concerns reinforce this:

1. The default styles (`font-size: 12px` for label, `14px` for value, `font-weight: 500` vs `400`) assert a typographic hierarchy. Even behind CSS variables, those *defaults* are a baked design preset. Primitives in this repo default to `inherit` and let the host stylesheet decide.
2. If we ever do want to ship label/value semantics, the right shape is probably an accessible `<label>`/`<input>` association on `InputField` (or a `FormField` slotting primitive with `for`/`id` wiring), not a string-in-string-out display widget.

Recommended next step: close this PR, and if it's useful, port the snippet into a docs/site example under "composing primitives" so other teams see the pattern without us shipping it as API surface. Happy to pair on the docs side if helpful.

murdore:
@sinha-sahil — thanks for the thorough, reasoned rejection. You're right, and I don't have a counter-argument.

The point lands clearly: `LabeledField` is two `<div>`s with two strings, no state, no events, no a11y semantics. The library ships primitives that abstract non-trivial behaviour; a layout wrapper that a consumer can write in five lines with better semantic control (`<dl>/<dt>/<dd>`, `<label for=…>`, whatever the app needs) is not a primitive. It belongs in consumer-land. The default `font-size`/`font-weight` values also violate the "default to `inherit`" rule you called out — those are baked design decisions that fight the host stylesheet rather than deferring to it.

I'll close this PR. If it's useful, I'm happy to contribute a "composing primitives" snippet to the docs/site showing the label-value pattern as a consumer-side example — that way other teams see a clean reference without it becoming library API surface.

One question before I close: would a docs/site contribution like that be welcome as a separate small PR, or would you prefer I leave a note in an existing docs page rather than adding a new example?

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — appreciate the gracious read of the feedback, and especially the way you re-stated the core point back ("two divs with two strings, no state, no events, no a11y semantics"). That's exactly the heuristic we want internalised for future PRs: if the body of a component is markup a consumer can reproduce in five lines with full semantic control, it's a docs example, not library API. Same goes for the typographic defaults — once they're baked in (even behind a CSS var) they're an opinion the host stylesheet has to fight, and that's the opposite of what these primitives are supposed to do.

On your follow-up question: **yes, a separate small docs PR is very welcome** — please don't bury it in an existing page. A few notes to make it land cleanly:

1. **Location**: put it under a new "Composing primitives" / "Recipes" section. If one doesn't exist yet, that's fine — start it with this entry; the directory is a natural home for the patterns we'll keep saying no to as components (label/value, card-with-icon-and-button, copy-to-clipboard button, etc.). It becomes a living "here's how to do this in 5–15 lines" index, and it's the artifact we can link to next time someone proposes a composite wrapper.

2. **Shape of the example**: show at least two variants so it's clear the library is deliberately not picking one — e.g. a `<dl>/<dt>/<dd>` semantic version for read-only field/value pairs, and a flex-row version with an icon slot. Mention the `<label for=…>` + `InputField` pairing for the *editable* case so readers don't reach for a `LabeledField` component for forms either. The point of the doc is to make the consumer-side composition feel obvious, not to ship a "blessed" snippet.

3. **Defaults**: in the example CSS, use `inherit` / `currentColor` / `var(--…, inherit)` patterns so the snippet itself models the philosophy. If someone copy-pastes it, they should land somewhere consistent with how `Button`, `InputField`, etc. behave out of the box.

4. **Scope**: keep the PR strictly docs — no `src/lib/` changes, no new exports. That keeps review trivial and the intent unambiguous.

If you'd like, drop a short outline (headings + the two snippets) in the PR description before pushing the full thing and I'll sanity-check the framing so we don't have to bikeshed it post-hoc. Happy to co-author the intro paragraph that explains *why* this lives in docs rather than as a component — that framing matters more than the snippets, because it's what future contributors will read when they're deciding whether to open the next `LabeledField`-shaped PR.

Thanks again for taking the feedback on the chin and offering to turn it into something reusable for the rest of the org — that's exactly the right move.

**Ask: please close this PR.** Then open a fresh docs-only PR for the "Composing primitives" recipe whenever you're ready.

murdore:
Closing — LabeledField is consumer-side composition, not a primitive. A "composing primitives" docs recipe will follow. Thanks @sinha-sahil.

PR #197, opened 2026-05-29
feat: add CopyButton component

murdore:
## Summary

- New `CopyButton` component: an icon-only copy-to-clipboard button with a transient "Copied" feedback state
- Svelte 5 runes (`$state`, `$props`) following the existing component conventions
- SSR-safe: all clipboard and DOM access is guarded by `typeof window !== 'undefined'`

### Props

| `value` | `string` | — (required) | Text copied to clipboard on click |
| `testId` | `string?` | — | `data-pw` attribute on root button |
| `classes` | `string?` | — | Extra CSS classes appended to root |
| `ariaLabel` | `string?` | `'Copy'` | `aria-label` in idle state |
| `copiedAriaLabel` | `string?` | `'Copied'` | `aria-label` while copied feedback is active |
| `timeoutMs` | `number?` | `2000` | Milliseconds before reverting to idle state |

### CSS Custom Properties

| `--copy-button-color` | `currentColor` | Icon / text colour |
| `--copy-button-size` | `16px` | SVG icon width + height |
| `--copy-button-padding` | `4px` | Button padding |
| `--copy-button-hover-background` | `rgba(0,0,0,0.05)` | Hover background |
| `--copy-button-border-radius` | `4px` | Button corner radius |

### DOM

Root element: `<button type="button" class="copy-button {classes}">` with `data-pw={testId}` (only set when `testId` is a string).

Two inline stroke SVGs toggled by the `copied` flag: clipboard glyph (idle) and checkmark (copied). Both are `aria-hidden`.

## Notes

- `svelte-check`: 0 errors, 0 warnings
- `prettier --check`: all files clean
- `eslint`: no issues
- Clipboard fallback uses a hidden `<textarea>` + `document.execCommand('copy')` for environments without the async Clipboard API

### DIFF
--- a/src/lib/CopyButton/CopyButton.svelte
+++ b/src/lib/CopyButton/CopyButton.svelte
@@ -1,0 +1,99 @@
+<script lang="ts">
+  import type { CopyButtonProperties } from './properties';
+
+  let {
+    value,
+    testId,
+    classes,
+    ariaLabel = 'Copy',
+    copiedAriaLabel = 'Copied',
+    timeoutMs = 2000
+  }: CopyButtonProperties = $props();
+
+  let copied = $state(false);
+
+  const handleCopy = async () => {
+    if (typeof window === 'undefined') {
+      return;
+    }
+    try {
+      await navigator.clipboard.writeText(value);
+      copied = true;
+      setTimeout(() => {
+        copied = false;
+      }, timeoutMs);
+    } catch {
+      const textarea = document.createElement('textarea');
+      textarea.value = value;
+      textarea.style.position = 'fixed';
+      textarea.style.top = '-9999px';
+      textarea.style.left = '-9999px';
+      textarea.setAttribute('aria-hidden', 'true');
+      document.body.appendChild(textarea);
+      textarea.focus();
+      textarea.select();
+      document.execCommand('copy');
+      document.body.removeChild(textarea);
+      copied = true;
+      setTimeout(() => {
+        copied = false;
+      }, timeoutMs);
+    }
+  };
+</script>
+
+<button
+  type="button"
+  class="copy-button {classes ?? ''}"
+  aria-label={copied ? copiedAriaLabel : ariaLabel}
+  data-pw={typeof testId === 'string' ? testId : null}
+  onclick={handleCopy}
+>
+  {#if !copied}
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      viewBox="0 0 24 24"
+      aria-hidden="true"
+      focusable="false"
+    >
+      <rect x="9" y="9" width="13" height="13" rx="2" ry="2"></rect>
+      <path d="M5 15H4a2 2 0 0 1-2-2V4a2 2 0 0 1 2-2h9a2 2 0 0 1 2 2v1"></path>
+    </svg>
+  {:else}
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      viewBox="0 0 24 24"
+      aria-hidden="true"
+      focusable="false"
+    >
+      <polyline points="20 6 9 17 4 12"></polyline>
+    </svg>
+  {/if}
+</button>
+
+<style>
+  .copy-button {
+    background: transparent;
+    border: none;
+    cursor: pointer;
+    display: inline-flex;
+    align-items: center;
+    padding: var(--copy-button-padding, 4px);
+    border-radius: var(--copy-button-border-radius, 4px);
+    color: var(--copy-button-color, currentColor);
+  }
+
+  .copy-button:hover {
+    background: var(--copy-button-hover-background, rgba(0, 0, 0, 0.05));
+  }
+
+  .copy-button svg {
+    width: var(--copy-button-size, 16px);
+    height: var(--copy-button-size, 16px);
+    stroke: currentColor;
+    fill: none;
+    stroke-width: 2;
+    stroke-linecap: round;
+    stroke-linejoin: round;
+  }
+</style>
--- a/src/lib/CopyButton/properties.ts
+++ b/src/lib/CopyButton/properties.ts
@@ -1,0 +1,8 @@
+export type CopyButtonProperties = {
+  value: string;
+  testId?: string;
+  classes?: string;
+  ariaLabel?: string;
+  copiedAriaLabel?: string;
+  timeoutMs?: number;
+};
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -56,6 +56,7 @@ export { default as EmptyState } from './EmptyState/EmptyState.svelte';
 export { default as Combobox } from './Combobox/Combobox.svelte';
 export { default as ColorPicker } from './ColorPicker/ColorPicker.svelte';
 export { default as SplitInput } from './SplitInput/SplitInput.svelte';
+export { default as CopyButton } from './CopyButton/CopyButton.svelte';
 
 export type * from './Button/properties';
 export type * from './Modal/properties';
@@ -109,5 +110,6 @@ export type * from './EmptyState/properties';
 export type * from './Combobox/properties';
 export type * from './ColorPicker/properties';
 export type * from './SplitInput/properties';
+export type * from './CopyButton/properties';
 
 export { validateInput } from './utils';


sinha-sahil:
## Overall

Thanks Sachin — the underlying behaviour here (async clipboard with a `document.execCommand` fallback, transient `copied` state, ARIA label swap, SSR guard) is solid and the surface is small. The component genuinely earns its place as a primitive: it isn't a dummy decoration, it isn't a pure composite wrapper, and it manages real state. That part is good.

That said, in its current shape the PR conflicts with two of our load-bearing rules and I'd like it reworked before merge. Worth calling out that we already have a working precedent in the library: `src/lib/Snippet/Snippet.svelte` ships a copy affordance that does almost exactly what this needs to do — please mirror that shape.

---

### 1. `§6 Reuse the Button Component` (and `§4 Reuse Existing Components`) — blocker

`GUIDELINES.md §6` explicitly lists **copy** as one of the generic action buttons that must be rendered via the shared `Button` component:

> Replace raw `<button>` elements with the reusable `Button` component for generic action buttons: submit, dismiss, close, **copy**, action, navigation arrows, and similar.

Right now `CopyButton.svelte` renders a raw `<button>` and reimplements the action-button shell (`background`, `border`, hover state). That re-introduces a second source of truth for how action buttons look and behave in the library — exactly what §6 was added to prevent. It also means consumers who theme `--button-*` will *not* get those overrides on `CopyButton`, which will surprise them. `Snippet.svelte` already does this correctly — see `.snippet-copy` (lines 91-103), which bridges `--snippet-copy-*` to `--button-*` and renders the icon inside `<Button>`.

The §6 bridging pattern keeps your `--copy-button-*` public API intact while delegating to `Button`:

```svelte
<script lang="ts">
  import Button from '../Button/Button.svelte';
  import type { CopyButtonProperties } from './properties';

  let {
    value,
    testId,
    classes,
    ariaLabel = 'Copy',
    copiedAriaLabel = 'Copied',
    timeoutMs = 2000,
    idleIcon,
    copiedIcon
  }: CopyButtonProperties = $props();

  let copied = $state(false);
  // ... clipboard logic unchanged
</script>

<div class="copy-button {classes ?? ''}" data-pw={typeof testId === 'string' ? testId : null}>
  <Button
    onclick={handleCopy}
    ariaLabel={copied ? copiedAriaLabel : ariaLabel}
  >
    {#snippet icon()}
      {#if copied}
        {@render copiedIcon?.()}
      {:else}
        {@render idleIcon?.()}
      {/if}
    {/snippet}
  </Button>
</div>

<style>
  .copy-button {
    --button-background: var(--copy-button-background, transparent);
    --button-border: var(--copy-button-border, none);
    --button-color: var(--copy-button-color, currentColor);
    --button-padding: var(--copy-button-padding, 4px);
    --button-border-radius: var(--copy-button-border-radius, 4px);
    --button-hover-background: var(--copy-button-hover-background, rgba(0, 0, 0, 0.05));
  }
</style>
```

This also gets you `:focus-visible` semantics for free, which the current raw `<button>` is missing relative to `Button`.

---

### 2. Hardcoded SVG glyphs — customizability blocker

The two inline SVGs (clipboard, checkmark) are baked into the component with no override hook. The established library convention is the opposite: ship a sensible default **and** expose a Snippet prop so consumers can swap it. See:

- `Snippet/Snippet.svelte` — `copyIcon?: Snippet`, falls back to `copy.svg`
- `Banner/Banner.svelte` — close icon overridable, falls back to `close.svg`
- `Pill/Pill.svelte` — close icon overridable, falls back to `close.svg`
- `Checkbox/Checkbox.svelte` — falls back to `checkmark.svg`

Please follow the same shape: accept `idleIcon` and `copiedIcon` as `Snippet` props (matching `Button`'s existing `icon` Snippet convention) and either keep your current inline SVGs as the default render path or move them to `src/lib/assets/` and import via `?raw` like the others. Either is fine — what matters is that consumers shipping a different icon set (Lucide, Heroicons, Phosphor, in-house) can override without forking.

```ts
// properties.ts
import type { Snippet } from 'svelte';

export type CopyButtonProperties = {
  value: string;
  testId?: string;
  classes?: string;
  ariaLabel?: string;
  copiedAriaLabel?: string;
  timeoutMs?: number;
  idleIcon?: Snippet;
  copiedIcon?: Snippet;
};
```

---

### 3. Minor

- The post-copy reset logic is duplicated in the `try` and `catch` branches:
  ```ts
  copied = true;
  setTimeout(() => { copied = false; }, timeoutMs);
  ```
  Hoist into a single helper called from both paths.
- Consider exposing an `oncopy` callback the way `Snippet` does (line 22), so consumers can hook telemetry / toasts without polling state.

---

### Ask

Please rework along these lines before merging:

1. **Delegate to `Button`** using the §6 bridging pattern (this is required, not optional — `Snippet.svelte` is the reference).
2. **Externalise the icons** as `idleIcon` / `copiedIcon` Snippet props, keeping defaults like Snippet/Banner/Pill do.
3. De-dup the post-copy bookkeeping into a single helper, and consider an `oncopy` callback.

After 1–2 the diff actually gets *smaller* and the component fits the library's primitive-only posture cleanly. Happy to look again once it's reshaped.

sinha-sahil:
**Follow-up — escalating this from MAJOR to BLOCK after a second read.**

On reflection, CopyButton sits in the same category as Divider, LabeledField, and InfoBlock — it shouldn't be a library component at all. The shape is:

`<Button onclick={copy}>{copied ? 'Copied!' : 'Copy'}</Button>` + a 3-line copied-state hook.

That's a **behavioral overlay** (Button + a method) plus a **UX preset** (the timing of the "Copied!" feedback, the icon swap, the copy/copied label split). Neither is a primitive — the library already ships `Button` (GUIDELINES §6), and the rest is consumer-side composition.

### Why this matters

The principle this PR breaks isn't visible from a single component — it's the gradient. If CopyButton ships:

- **ShareButton** is defensible — it's just Button + `navigator.share()` + a fallback state
- **SaveButton** is defensible — it's just Button + form submit + a "saving..." state
- **DeleteButton** is defensible — it's just Button + a confirm dialog + a destructive style
- **DownloadButton**, **LikeButton**, **SubmitButton**, **CancelButton**…

Each is "Button + a method". Once we accept the precedent, we have no principled way to reject the rest. The library bloats into a UX-preset catalogue instead of a primitive library, and consumers lose the ability to own their copy-button UX (timing, feedback channel, error handling, accessibility announcements).

### The 5-line consumer pattern

```svelte
<script>
  import { Button } from '@juspay/svelte-ui-components';
  let { text }: { text: string } = $props();
  let copied = $state(false);
  let timer: ReturnType<typeof setTimeout>;

  async function copy() {
    await navigator.clipboard.writeText(text);
    copied = true;
    clearTimeout(timer);
    timer = setTimeout(() => (copied = false), 2000);
  }
</script>

<Button onclick={copy} classes={copied ? 'copy-success' : ''}>
  {copied ? 'Copied!' : 'Copy'}
</Button>
```

5 lines of state, full control over timing/feedback channel/accessibility, no library code needed.

### Ask

Recommend closing this PR. If the team wants the snippet centralized, it belongs as a docs/site example under "composing primitives", or as a `useCopyToClipboard` helper inside the consumer app — same as the path we recommended for LabeledField (#196) and Divider (#192).

Sorry for the second pass — wanted to make sure the escalation reasoning was on the record.

murdore:
@sinha-sahil — thank you for the detailed two-pass review, and especially for putting the escalation reasoning on record. Having read both comments carefully, I agree with your final position.

The clipboard logic, the `copied` state, and the icon swap are all genuinely useful — but they compose entirely on top of `Button`, which the library already ships. `CopyButton` is `Button + navigator.clipboard + a setTimeout`, and accepting it as a primitive opens the door to `ShareButton`, `DeleteButton`, `SubmitButton`, etc. — each equally defensible as "Button + a method." There is no principled place to draw that line once the first one ships, and the library would drift from a primitive set into a UX-preset catalogue. I had not thought through the precedent problem until your second comment; that framing is convincing.

I will close this PR. The clipboard behaviour and the 5-line consumer snippet you outlined belong in the documentation site under "composing primitives", not in the library itself, and I will open a docs-site PR with that example instead.

One question before I close: would it make sense to extract the `useCopyToClipboard` state logic (the async/fallback/reset bookkeeping) into a small utility exported from the library alongside the components — similar to how some libraries ship composable helpers — or does even that cross the primitive boundary in your view?

sinha-sahil:

New arrow-const handlers in the diff: `handleCopy`. Please write them as `function fnName() {}` instead.

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — appreciate you engaging with the precedent argument head-on rather than negotiating around it. Closing this and moving the 5-line snippet to the docs site under "composing primitives" is exactly the right call, and pairs nicely with the path we landed on for #196 (LabeledField) and #192 (Divider). When that docs PR lands, please cross-link it from this one so future contributors hit the recipe before they hit "add a new component".

On your follow-up — **`useCopyToClipboard` shouldn't ship from this library either**, and the reasoning is the same gradient you just agreed with, only one layer up.

The library's identity is "unstyled Svelte primitives + CSS variable hooks." It is not a utilities catalogue. The moment we export `useCopyToClipboard`, the next defensible additions are `useShare`, `useDebounce`, `useLocalStorage`, `useMediaQuery`, `useClickOutside`, `useFocusTrap`, `useEventListener` — each one is "small, useful, framework-idiomatic, hard to argue against in isolation." That's the exact failure mode we just rejected at the component layer; shipping it as `$state`-returning helpers doesn't change the shape of the precedent problem, it just relocates it. We'd end up maintaining two product surfaces (components *and* hooks) with two sets of versioning, two sets of docs, and two sets of "is this really a primitive?" debates.

There's also a practical reason: the bookkeeping you'd extract is genuinely ~8 lines (`async copy()`, a `copied` flag, a `clearTimeout` + `setTimeout`). Wrapping that in a library export costs the consumer a dependency boundary, a peer-version constraint, and a layer of indirection in their stack traces — for a saving of about six lines they will write once per app. That trade goes the wrong way.

What I'd suggest instead, in order of preference:

1. **Docs-site recipe** (your current plan). The snippet *is* the API. Consumers copy-paste, own the timing/error-handling/telemetry, and never get blocked on a library release for a behavioral tweak.
2. **Consumer-app `src/lib/clipboard.ts`** in whichever app needed this. If two apps end up writing it, that's still cheaper than a shared library export.
3. **If real demand emerges across many apps**, a *separate* package (`@juspay/svelte-utils` or similar) with its own scope and design philosophy — explicitly *not* mixed into this one. That keeps this library's "primitive components only" contract clean and lets a utils package have its own opinions about hook design without contaminating either surface.

The principle to lean on going forward: **if a consumer can write it in ~10 lines using `navigator.*` + `$state` + `setTimeout`, it's a docs example, not an export.** That heuristic kills the gradient at the root for both components and helpers.

**Ask: please close this PR.** Behavioral overlay on `Button` isn't reconcilable with the library philosophy, and the helper-hook variant runs into the same precedent wall. Looking forward to reviewing the docs-site PR.

murdore:
Closing — CopyButton is a behavioral overlay on Button (§6), not a primitive. I'll add the consumer-side snippet to the docs site instead. Thanks @sinha-sahil.

PR #198, opened 2026-05-29
feat: add 'typing' animation variant to LoadingDots

murdore:
## Summary

- Extends `LoadingDotsAnimation` type union with a new `'typing'` value (`'bounce' | 'pulse' | 'typing'`)
- When `animation="typing"`, dots render with a gentle staggered opacity+scale pulse suited for chat/typing indicators
- Adds `@keyframes loading-dots-typing` (0%/80%/100%: opacity 0.3, scale 0.85 → 40%: opacity 1, scale 1)
- Three dots staggered at `animation-delay` 0s / 0.16s / 0.32s (fixed values, not using the shared stagger variable so typing variant has its own independent rhythm)
- Exposes `--loading-dots-typing-duration` CSS custom property (default `1.4s`) for consumer customisation
- Existing `bounce` and `pulse` variants are **fully unchanged** — no breaking changes

### Props

| Prop | Type | Default | Notes |
|------|------|---------|-------|
| `animation` | `'bounce' \| 'pulse' \| 'typing'` | `'bounce'` | Extended union — new `'typing'` value added |

### CSS custom properties

| Variable | Default | Effect |
|----------|---------|--------|
| `--loading-dots-typing-duration` | `1.4s` | Controls cycle duration for the typing animation |

### DOM

When `animation="typing"` the inner `.dot` spans receive the `typing` CSS class, which overrides the default `loading-dots-bounce` animation with `loading-dots-typing`.

## Notes

- `svelte-check` 0 errors / 0 warnings
- `prettier --check` clean
- `eslint` clean
- Backward-compatible: default value of `animation` remains `'bounce'`

### DIFF
--- a/src/lib/LoadingDots/LoadingDots.svelte
+++ b/src/lib/LoadingDots/LoadingDots.svelte
@@ -13,7 +13,12 @@
   data-pw={typeof testId === 'string' ? testId : null}
 >
   {#each { length: count } as _, i (i)}
-    <span class="dot" class:pulse={animation === 'pulse'} style="--i: {i}"></span>
+    <span
+      class="dot"
+      class:pulse={animation === 'pulse'}
+      class:typing={animation === 'typing'}
+      style="--i: {i}"
+    ></span>
   {/each}
 </span>
 
@@ -40,6 +45,11 @@
     animation-name: loading-dots-pulse;
   }
 
+  .dot.typing {
+    animation: loading-dots-typing var(--loading-dots-typing-duration, 1.4s) ease-in-out infinite;
+    animation-delay: calc(var(--i) * 0.16s);
+  }
+
   @keyframes loading-dots-bounce {
     0%,
     80%,
@@ -59,6 +69,19 @@
     }
     40% {
       opacity: 1;
+    }
+  }
+
+  @keyframes loading-dots-typing {
+    0%,
+    80%,
+    100% {
+      opacity: 0.3;
+      transform: scale(0.85);
+    }
+    40% {
+      opacity: 1;
+      transform: scale(1);
     }
   }
 </style>
--- a/src/lib/LoadingDots/properties.ts
+++ b/src/lib/LoadingDots/properties.ts
@@ -1,4 +1,4 @@
-export type LoadingDotsAnimation = 'bounce' | 'pulse';
+export type LoadingDotsAnimation = 'bounce' | 'pulse' | 'typing';
 
 export type LoadingDotsProperties = OptionalLoadingDotsProperties;
 


sinha-sahil:
Thanks @murdore — appreciate the clean diff and the duration variable. Unfortunately I have to push back on the overall shape: this PR adds a hardcoded design preset, which is the category of change our component philosophy explicitly forbids (see GUIDELINES.md §9 and the `feedback_no_hardcoded_variants.md` note in project memory).

The core issue is the `animation` prop itself — a string enum that maps to baked-in `@keyframes` and `.dot.typing { … }` rules inside the component's `<style>` block. Every new value (`'typing'` today, `'wave'`/`'breathe'`/`'thinking'` tomorrow) requires a code change, ships dead CSS to consumers who don't use it, and locks the rhythm/easing/opacity/scale into the library. That's the exact pattern we're trying to avoid: components are unstyled primitives, and *variants live in consumer CSS*.

Concretely:

1. **Drop the `'typing'` union member.** The typing rhythm should live in the consumer's CSS:
   ```css
   .chat-typing-dots .dot {
     animation: my-typing var(--typing-duration, 1.4s) ease-in-out infinite;
     animation-delay: calc(var(--i) * 0.16s);
   }
   @keyframes my-typing { /* … */ }
   ```
   …consumed via `<LoadingDots classes={{ container: 'chat-typing-dots' }} />`. No library change needed, and the consumer owns every knob (duration, stagger, easing, opacity range, scale range).

2. **The `0.16s` stagger is hardcoded and explicitly bypasses the shared stagger variable** ("not using the shared stagger variable so typing variant has its own independent rhythm"). Opting out of the existing CSS-variable customization surface is moving in the wrong direction — if anything we want *more* of the rhythm reachable through variables, not less.

3. **Forward direction:** I'd like us to discuss deprecating `animation` entirely in a follow-up and documenting a "bring your own keyframes via `classes`" recipe. That would let LoadingDots become a true primitive (markup + the `--i` index variable + a stable class hook) and move all rhythm decisions into the consumer's design system, where they belong.

Marking as request-changes. Happy to pair on the consumer-CSS recipe if useful.

murdore:
@sinha-sahil — thanks for the detailed review and for pointing directly to the relevant section. You're right: adding `'typing'` to the `LoadingDotsAnimation` union is precisely the baked-in preset pattern that §9 is designed to prevent. Hardcoding the keyframes, the `ease-in-out` easing, and the `0.16s` stagger inside the component means every consumer ships dead CSS they may not use, and any rhythm tweak requires a library release. That's the wrong tradeoff for a primitive.

Here is how I plan to rework it:

- Drop the `'typing'` union member entirely — no enum change, no new class.
- The existing `--i` index variable and the `classes` prop already give consumers everything they need. A consumer wanting a typing rhythm just does:

  ```css
  .typing-dots .dot {
    animation: my-typing var(--loading-dots-typing-duration, 1.4s) ease-in-out infinite;
    animation-delay: calc(var(--i) * 0.16s);
  }
  @keyframes my-typing { 0%, 80%, 100% { opacity: 0.3; transform: scale(0.85); } 40% { opacity: 1; transform: scale(1); } }
  ```

  …and passes `classes="typing-dots"` to `LoadingDots`. No library change needed.

Would you prefer I submit that as a small docs/recipe addition (a "typing indicator" example in the component's README), or simply close this PR and let the consumer-CSS pattern speak for itself?

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — thanks for taking the feedback so cleanly. You've nailed the diagnosis ("baked-in preset pattern that §9 is designed to prevent… any rhythm tweak requires a library release") and the rework plan you sketched is exactly the right shape: keep `LoadingDots` as markup + `--i` + a stable class hook, and push the rhythm into consumer CSS where the design system lives.

To your question — **docs recipe or just close?** — both, in that order:

1. **Please close this PR.** There's nothing in the current diff worth preserving: `'typing'` joins the union, the `.dot.typing` rules ship to everyone, the `0.16s` stagger explicitly bypasses the shared `--loading-dots-stagger` variable, and the dead CSS lands in every consumer's bundle regardless of whether they ever set `animation="typing"`. None of that is salvageable as a slim follow-up — the whole change *is* the variant.

2. **Yes, please open the docs/recipe PR.** I'd rather we make the consumer-CSS path discoverable than leave it implicit. A short "Typing indicator" section in `docs/LoadingDots.md` (or the component README) covering:
   - the `--i` index variable and the `classes` prop as the customization surface,
   - a copy-pasteable `@keyframes` + `animation-delay: calc(var(--i) * …)` recipe (roughly what you posted above),
   - a one-line note that this same pattern works for "wave", "breathe", "thinking" or any other rhythm a consumer dreams up without touching the library.

That doc PR is genuinely more valuable than shipping `'typing'` would have been, because it teaches the pattern instead of hardcoding one instance of it. It also pre-empts the next three PRs that would otherwise show up asking for `'wave'` / `'breathe'` / `'thinking'`.

One small forward-looking note while we're here: I'd like us to take the same lens to the existing `animation` prop in a separate thread. If the consumer-CSS recipe covers the same surface area, the prop itself is on borrowed time — same dead-CSS cost, same "library release to tweak a rhythm" problem, just grandfathered in. Not for this PR, but worth flagging so we don't accidentally entrench it by adding a fourth value later.

Appreciate the quick turn on this one — the conceding-the-point-and-proposing-the-right-fix-in-one-reply pattern is the ideal way to land here.

**Ask: please close this PR, and open a small follow-up adding the "typing indicator" consumer-CSS recipe to `docs/LoadingDots.md`.**

murdore:
Closing — the change is entirely the typing variant (§9). I'll add a "typing indicator" consumer-CSS recipe to docs/LoadingDots.md. Thanks @sinha-sahil.

PR #199, opened 2026-05-29
feat: add GradientBlur decorative component

murdore:
## Summary

- New `GradientBlur` decorative primitive using 8 stacked `<div>` layers with `backdrop-filter` and `mask-image` gradients to produce a smooth progressive blur effect
- No interactivity: `aria-hidden="true"`, `pointer-events: none` throughout
- Fully customisable via CSS custom properties; zero hard-coded visual values in consuming code

### Props

| `testId` | `string?` | — | Renders as `data-pw` on root element |
| `classes` | `string?` | `''` | Extra CSS classes appended to root |

### CSS Custom Properties

| `--gradient-blur-height` | `80px` | Height of the blur overlay |
| `--gradient-blur-direction` | `to bottom` | Gradient direction (any valid CSS gradient direction) |
| `--gradient-blur-strength` | `8px` | Base blur amount; layer n applies `strength * n` |
| `--gradient-blur-position` | `absolute` | CSS `position` on root |
| `--gradient-blur-z-index` | `10` | `z-index` on root |

### DOM Structure

```html
<div class="gradient-blur {classes}" aria-hidden="true" data-pw="{testId}">
  <div class="gradient-blur-layer gradient-blur-layer-1"></div>
  <!-- … -->
  <div class="gradient-blur-layer gradient-blur-layer-8"></div>
</div>
```

Each layer `n` applies `backdrop-filter: blur(strength * n)` and a mask gradient covering the `[(n-1)*12.5%, n*12.5%]` band of the overlay height.

## Notes

- `svelte-check` — 0 errors, 0 warnings
- `prettier --check` — all files pass
- `eslint` — no findings
- CSS class names are kebab-case only (`gradient-blur`, `gradient-blur-layer`, `gradient-blur-layer-{n}`) — compliant with `selector-class-pattern`
- No `font-*` or `line-height` overrides in `<style>` block

### DIFF
--- a/src/lib/GradientBlur/GradientBlur.svelte
+++ b/src/lib/GradientBlur/GradientBlur.svelte
@@ -1,0 +1,117 @@
+<script lang="ts">
+  import type { GradientBlurProperties } from './properties';
+
+  let { testId, classes }: GradientBlurProperties = $props();
+</script>
+
+<div
+  class="gradient-blur {classes ?? ''}"
+  aria-hidden="true"
+  data-pw={typeof testId === 'string' ? testId : null}
+>
+  <div class="gradient-blur-layer gradient-blur-layer-1"></div>
+  <div class="gradient-blur-layer gradient-blur-layer-2"></div>
+  <div class="gradient-blur-layer gradient-blur-layer-3"></div>
+  <div class="gradient-blur-layer gradient-blur-layer-4"></div>
+  <div class="gradient-blur-layer gradient-blur-layer-5"></div>
+  <div class="gradient-blur-layer gradient-blur-layer-6"></div>
+  <div class="gradient-blur-layer gradient-blur-layer-7"></div>
+  <div class="gradient-blur-layer gradient-blur-layer-8"></div>
+</div>
+
+<style>
+  .gradient-blur {
+    position: var(--gradient-blur-position, absolute);
+    left: 0;
+    right: 0;
+    bottom: 0;
+    height: var(--gradient-blur-height, 80px);
+    pointer-events: none;
+    z-index: var(--gradient-blur-z-index, 10);
+  }
+
+  .gradient-blur-layer {
+    position: absolute;
+    inset: 0;
+  }
+
+  .gradient-blur-layer-1 {
+    backdrop-filter: blur(calc(var(--gradient-blur-strength, 8px) * 1));
+    -webkit-backdrop-filter: blur(calc(var(--gradient-blur-strength, 8px) * 1));
+    mask-image: linear-gradient(
+      var(--gradient-blur-direction, to bottom),
+      rgba(0, 0, 0, 1) 0%,
+      rgba(0, 0, 0, 0) 12.5%
+    );
+  }
+
+  .gradient-blur-layer-2 {
+    backdrop-filter: blur(calc(var(--gradient-blur-strength, 8px) * 2));
+    -webkit-backdrop-filter: blur(calc(var(--gradient-blur-strength, 8px) * 2));
+    mask-image: linear-gradient(
+      var(--gradient-blur-direction, to bottom),
+      rgba(0, 0, 0, 1) 12.5%,
+      rgba(0, 0, 0, 0) 25%
+    );
+  }
+
+  .gradient-blur-layer-3 {
+    backdrop-filter: blur(calc(var(--gradient-blur-strength, 8px) * 3));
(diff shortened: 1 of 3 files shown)


sinha-sahil:
Follow-up — escalating this from MINOR to BLOCK after a second read with a stricter lens.

On the first pass I flagged a few minor concerns and approved the direction. Re-reading the diff against philosophy #3 (UX presets / decorative components belong as CSS, not components), this is the same call we made on Divider in #192 and I should have made it here too.

The whole component is:
- 8 static `<div>`s with no reactivity, no state, no events, no slots
- one `aria-hidden="true"` attribute
- a `<style>` block of 8 near-identical CSS rules with hardcoded band stops and a hardcoded layer count

There is nothing here a consumer cannot write in 10–25 lines of CSS — and arguably *should*, because they then control the layer count (perf vs. quality), the band stops, and whether to use `::before`/`::after` instead of extra DOM. The component as shipped bakes "8 layers at 12.5% bands" as a library-owned UX preset, which is exactly what the no-hardcoded-presets rule exists to prevent.

The "decorative" in the title is the tell — same precedent as Divider (#192): pure decoration belongs as a documented CSS utility (or a docs-site snippet), not a Svelte component in `src/lib/`.

Proposed path forward:
1. Move the CSS to a documented recipe in the docs site (or a `gradient-blur.css` utility consumers can import).
2. Close this PR; no component needed.

If there's a future need for a real primitive here — e.g., an SVG `<filter>` wrapper that handles `feGaussianBlur` setup, id collision, or Safari fallbacks — that would be a different PR with actual primitive behavior to encapsulate.

sinha-sahil (changes requested):

To merge:

murdore:
@sinha-sahil — thanks for the review. The CHANGES_REQUESTED body points at inline comments above, but I don't see any inline review comments on the diff (graphql confirms zero `reviewThreads`, only the formal-review body is present). Could you share the specific blockers for GradientBlur — is the issue the static-decorative nature (philosophy #3, "write as plain CSS"), the lack of reactive content / slots, the hardcoded gradient stops, or something else? Happy to rework as a CSS utility recipe or close once I know which path you prefer.

sinha-sahil:
@murdore — apologies for the missing inline anchors; the body-only review was the whole of it. Let me make the blocker explicit so there's no ambiguity.

**The blocker is the static-decorative nature (philosophy #3).** Not the gradient stops, not the lack of slots, not the layer count — those are downstream symptoms. The root issue is that `GradientBlur.svelte` is, end-to-end:

- 8 static `<div>`s in a fixed shape (`src/lib/GradientBlur/GradientBlur.svelte:9-18`)
- a `<style>` block with 8 hand-unrolled rules over a hardcoded `12.5%` band partition (`:35-116`)
- zero reactivity, state, events, or snippets
- `properties.ts` is `{ testId?, classes? }` — i.e., the component owns no behavior that CSS doesn't already own

This is the same shape as Divider in #192 — a CSS pattern shipped as a Svelte component. The "can a consumer write this in 5–15 lines of existing primitives + CSS" test resolves to "yes, in ~25 lines of plain CSS, and they get to pick their own layer count / band stops / `::before`+`::after` strategy." Shipping it as a component bakes "8 layers, 12.5% bands" as a library-owned UX preset, which is exactly what philosophy #3 exists to prevent.

The fact that you've already routed every visual knob through CSS variables (`--gradient-blur-height`, `--gradient-blur-strength`, `--gradient-blur-direction`, etc.) is, ironically, the strongest evidence that this belongs in a CSS file rather than a `.svelte` file — there's nothing left for the component layer to encapsulate.

**Path forward — preference, in order:**
1. **Close this PR.** Move the CSS into a documented recipe on the docs site (or a `gradient-blur.css` utility consumers can copy/import). No code lands in `src/lib/`.
2. If you want, open a follow-up docs-only PR adding the recipe under `src/routes/` so it's discoverable.

The "real primitive" exception still stands — if you later hit a case where the wrapper has to encapsulate something CSS can't (SVG `<filter>` id collisions, Safari `backdrop-filter` fallbacks behind a runtime detect, etc.), that's a different PR with actual behavior to justify a component. This one doesn't have that.

**Ask: please close #199** and, if useful, open a docs PR with the recipe.

murdore:
@sinha-sahil — agreed, closing per philosophy #3. You are right that "routing everything through CSS vars is the strongest evidence this belongs in a CSS file" — 8 static divs with no reactivity is not a primitive.

Will open a docs-only PR with the CSS recipe so consumers reaching for this pattern have a starting point.

PR #200, opened 2026-05-29
feat: add InfoBlock component

murdore:
## Summary

- Adds `InfoBlock` — a new definition-list primitive (`<dl>`) that renders metadata key/value pairs
- Exports `InfoBlock` component and `InfoBlockProperties` type from the package root (`src/lib/index.ts`)
- Zero new project-owned dependencies; follows existing Svelte 5 runes pattern (mirrors `Badge`)

### Props (`InfoBlockProperties`)

| `items` | `Record<string, string>` | yes | Key/value pairs to render |
| `testId` | `string` | no | Sets `data-pw` on the root `<dl>` |
| `classes` | `string` | no | Extra CSS classes appended to root |

### DOM structure

```html
<dl class="info-block {classes}" data-pw="{testId}">
  <div class="info-block-item">
    <dt class="info-block-key">Label</dt>
    <dd class="info-block-value">Value</dd>
  </div>
  …
</dl>
```

### CSS custom properties

| `--info-block-gap` | `8px` | Column gap between rows |
| `--info-block-padding` | `0` | Root padding |
| `--info-block-item-justify-content` | `space-between` | Row flex justify |
| `--info-block-key-color` | `inherit` | Key text colour |
| `--info-block-key-font-weight` | `500` | Key weight (via CSS var, never hardcoded) |
| `--info-block-value-color` | `inherit` | Value text colour |
| `--info-block-value-font-weight` | `400` | Value weight (via CSS var) |
| `--info-block-value-text-align` | `right` | Value text alignment |

## Notes

- `svelte-check` — 0 errors, 0 warnings
- `prettier --check` — all files pass
- `eslint` — 0 errors (each block keyed by property key)
- `dt`/`dd` margins reset to `0`; all selectors are kebab-case only
- `font-weight` is exposed exclusively through CSS custom properties, never hardcoded

### DIFF
--- a/src/lib/InfoBlock/InfoBlock.svelte
+++ b/src/lib/InfoBlock/InfoBlock.svelte
@@ -1,0 +1,42 @@
+<script lang="ts">
+  import type { InfoBlockProperties } from './properties';
+
+  let { items, testId, classes }: InfoBlockProperties = $props();
+</script>
+
+<dl class="info-block {classes ?? ''}" data-pw={typeof testId === 'string' ? testId : null}>
+  {#each Object.entries(items) as [key, value] (key)}
+    <div class="info-block-item">
+      <dt class="info-block-key">{key}</dt>
+      <dd class="info-block-value">{value}</dd>
+    </div>
+  {/each}
+</dl>
+
+<style>
+  .info-block {
+    display: flex;
+    flex-direction: column;
+    gap: var(--info-block-gap, 8px);
+    padding: var(--info-block-padding, 0);
+    margin: 0;
+  }
+
+  .info-block-item {
+    display: flex;
+    justify-content: var(--info-block-item-justify-content, space-between);
+  }
+
+  .info-block-key {
+    margin: 0;
+    color: var(--info-block-key-color, inherit);
+    font-weight: var(--info-block-key-font-weight, 500);
+  }
+
+  .info-block-value {
+    margin: 0;
+    color: var(--info-block-value-color, inherit);
+    font-weight: var(--info-block-value-font-weight, 400);
+    text-align: var(--info-block-value-text-align, right);
+  }
+</style>
--- a/src/lib/InfoBlock/properties.ts
+++ b/src/lib/InfoBlock/properties.ts
@@ -1,0 +1,5 @@
+export type InfoBlockProperties = {
+  items: Record<string, string>;
+  testId?: string;
+  classes?: string;
+};
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -56,6 +56,7 @@ export { default as EmptyState } from './EmptyState/EmptyState.svelte';
 export { default as Combobox } from './Combobox/Combobox.svelte';
 export { default as ColorPicker } from './ColorPicker/ColorPicker.svelte';
 export { default as SplitInput } from './SplitInput/SplitInput.svelte';
+export { default as InfoBlock } from './InfoBlock/InfoBlock.svelte';
 
 export type * from './Button/properties';
 export type * from './Modal/properties';
@@ -109,5 +110,6 @@ export type * from './EmptyState/properties';
 export type * from './Combobox/properties';
 export type * from './ColorPicker/properties';
 export type * from './SplitInput/properties';
+export type * from './InfoBlock/properties';
 
 export { validateInput } from './utils';


sinha-sahil:
Thanks for the clean PR @murdore — the CSS-variable wiring, the `{classes ?? ''}` pattern, and the `typeof testId === 'string'` guard are all on-philosophy, and it's obvious you read GUIDELINES.md before writing this. I want to push back on the component itself, though, because I think `InfoBlock` is the canonical case the library philosophy is built to reject, and it's worth being explicit about why so future PRs don't go down the same path.

## Why this is a blocker

`InfoBlock` is a **dummy component**. Strip away the styling and the entire body is:

```svelte
<dl>
  {#each Object.entries(items) as [key, value] (key)}
    <div><dt>{key}</dt><dd>{value}</dd></div>
  {/each}
</dl>
```

That's six lines of markup. There is:

- no `$state`, no `$derived`, no `$bindable`
- no events, no callbacks, no focus management
- no keyboard handling, no ARIA beyond what `<dl>/<dt>/<dd>` natively provide
- no composition with other library primitives (no `Img`, no `Button`)
- no error/loading/empty handling

Per the project philosophy (and §4 by extension — "compose, don't duplicate"), components in this library must justify their existence with **non-trivial behavior, accessibility, or state management**. A `<dl>` loop doesn't. It's CSS-plus-markup, which belongs in consumer-land.

Compare it to what's already in the library: `Combobox` manages keyboard nav + filter state, `ColorPicker` manages a complex input surface, `Modal` manages focus trapping and scroll lock, `Button` reconciles `disabled`/`enable` and provides a consistent CSS-variable contract for every action in the app. Those primitives earn their export. `InfoBlock` is the equivalent of exporting a `<Heading text="..." />` that renders `<h2>{text}</h2>` — it adds nothing the consumer couldn't write inline.

## It's also less customizable than a 6-line inline loop

The `items: Record<string, string>` shape forces every value to be a plain string. That means:

- can't render `Owner: <Avatar src={...} name={...} />`
- can't render `Created: <RelativeDate ts={...} />`
- can't render `Status: <Badge classes="badge-success" text="Live" />`
- can't render `Doc: <a href={url}>{title}</a>`

Those are the real use cases for a metadata-pairs block. The string-only API rules them all out. To fix it you'd need at minimum a `key` and `value` Snippet pair and an array-of-objects shape — at which point you've just reinvented "the consumer writes an `{#each}` block."

## What the consumer should write instead

Drop the component, document the pattern in the docs site, and tell consumers:

```svelte
<!-- in the consumer's component -->
<dl class="my-spec-sheet">
  {#each entries as { key, value } (key)}
    <div class="row">
      <dt>{key}</dt>
      <dd>{@render value()}</dd>  <!-- now value is a Snippet, rich content works -->
    </div>
  {/each}
</dl>

<style>
  .my-spec-sheet {
    --info-key-color: #555;
    --info-value-color: #111;
    --info-gap: 8px;

    display: flex;
    flex-direction: column;
    gap: var(--info-gap);
    margin: 0;
  }
  .row { display: flex; justify-content: space-between; }
  dt { margin: 0; color: var(--info-key-color); font-weight: 500; }
  dd { margin: 0; color: var(--info-value-color); font-weight: 400; text-align: right; }
</style>
```

That's the same code, with strictly more flexibility (Snippet values, per-row classes, arbitrary layout), and it lives where the design decisions actually belong — in the consuming app. The library doesn't lose anything by not shipping this, because there is nothing to ship that the consumer can't write in two minutes.

## Even the defaults are an opinionated preset

Setting aside the dummy-component issue: the root styles bake in `display: flex; flex-direction: column; justify-content: space-between; text-align: right` on the value. That's a "spec-sheet two-column with value pushed right" design preset hardcoded into a default. Per §9 and the "no hardcoded presets" rule, presets should live in consumer CSS classes (`classes="info-block-spec-sheet"`), not in the component's base styles. So even if we shipped this, the structural CSS would need to be a no-op default with the spec-sheet look opted into via a documented `classes=` recipe.

## Ask

Please **close this PR** and, if the metadata-pairs pattern keeps coming up internally, open a small docs PR adding it as a recipe under `src/routes/` or the docs site. I'm happy to review that. Genuinely good attention to the CSS-variable + `classes` plumbing in this one — keep that instinct for the next primitive that has real behavior to ship.

murdore:
@sinha-sahil — thank you for the detailed write-up. You are completely right, and I genuinely appreciate how clearly you explained the "why" rather than just issuing a close.

Reading your feedback, the core issue is that `InfoBlock` has no behavior to justify library residency. There is no state, no keyboard handling, no ARIA beyond what `<dl>/<dt>/<dd>` already provide natively, and no composition with other library primitives. It is, as you put it, a styled loop — six lines the consumer writes once and owns outright. I also take the point about the `Record<string, string>` shape actively harming real-world use: you can't slot in an `Avatar`, a `Badge`, a `RelativeDate`, or a link without breaking out of the API entirely.

The §9 point about the spec-sheet layout being a baked-in preset (not a consumer-side opt-in) lands too. Even setting aside the dummy-component issue, the structural defaults would need to be stripped.

I will close this PR. If the metadata-pairs pattern keeps coming up in consumer codebases, I will open a small docs PR with the Snippet-value recipe you outlined instead.

One question before I close: would a docs-site recipe (e.g. under `src/routes/recipes/`) be the right home, or would you prefer it as a code comment/example in an existing relevant component's docs page?

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — appreciate the gracious turnaround, and more importantly, the fact that you internalised the reasoning rather than just the verdict. "Six lines the consumer writes once and owns outright" is exactly the framing I want every future PR to be stress-tested against before it's opened. That instinct will save us both a lot of review cycles.

On your question about where the recipe should live: **a dedicated route under `src/routes/recipes/` is the right home**, and here's the reasoning so it generalises to the next pattern that comes up.

The existing `src/routes/components/*` pages have a specific contract — they document a single library export, its props, its events, and its CSS variables. Dropping a "here's how to build a metadata block without a component" recipe into, say, the `Table` or `Card` page muddies that contract: a reader skimming the Table docs shouldn't have to wade past unrelated `<dl>` examples to find the prop they came for, and a recipe that lives as an aside on another component's page is effectively invisible to anyone searching for "info block" or "spec sheet" in the docs.

A `src/routes/recipes/metadata-pairs/+page.svelte` (or similar) gives us:

1. **A discoverable, linkable URL** — when this pattern comes up in the next review, we can link the recipe directly instead of re-explaining.
2. **Room to show variants** — the basic Snippet-value version, a version with `Badge`/`Avatar` slotted in, a horizontal-row variant via consumer CSS, a stacked-key variant. That breadth is exactly what makes the "you don't need a component" argument concrete.
3. **A precedent for the next composite** — once `recipes/` exists, the answer to "should we ship `<LabelledInput>` / `<IconButton>` / `<SectionHeading>`?" becomes "no, add it to recipes/" with zero ambiguity. That's a meaningful philosophy win for the library, not just one closed PR.

Two small asks for the recipe PR when you open it:

- **Use a Snippet for the value, not a string.** That's the API the component should have had, and showing it in the recipe makes the case for itself.
- **Keep the structural CSS in the recipe's `<style>` block, not in any shared stylesheet.** The whole point is that layout/typography decisions live in consumer code; the recipe should model that explicitly.

If you want, drop the recipe PR against me directly and I'll turn it around quickly.

**Ask: please close this PR.** The dummy-component pattern isn't reconcilable with the library philosophy, and the docs-recipe path you proposed is the right successor. Thanks again for the clean handling.

murdore:
Closing — InfoBlock is a styled <dl> with no behavior; it belongs as a docs recipe, not a library export. Thanks @sinha-sahil.

PR #201, opened 2026-05-29
feat: add ActionBar component

murdore:
## Summary

- New `ActionBar` bottom-fixed action bar primitive for primary/secondary form actions
- Exported from `@juspay/svelte-ui-components` as `ActionBar` component and `ActionBarProperties` type
- Follows existing library patterns (Svelte 5 runes, `$props()` destructure, `data-pw` test-id, CSS custom properties)

### Props (`ActionBarProperties`)

| `testId` | `string?` | — | Sets `data-pw` on root element |
| `classes` | `string?` | — | Extra CSS classes appended to root |
| `divider` | `boolean?` | `true` | Renders a top hairline divider |
| `primaryButton` | `ButtonProperties?` | — | Spread onto primary `<Button>` |
| `secondaryButton` | `ButtonProperties?` | — | Spread onto secondary `<Button>` |
| `children` | `Snippet?` | — | Custom slot overrides button pair |

### CSS Custom Properties

| `--action-bar-background` | `#fff` | Root background |
| `--action-bar-position` | `fixed` | CSS position |
| `--action-bar-bottom` | `0` | Bottom offset |
| `--action-bar-left` | `0` | Left offset |
| `--action-bar-right` | `0` | Right offset |
| `--action-bar-padding` | `12px 16px` | Inner padding |
| `--action-bar-gap` | `8px` | Gap between buttons |
| `--action-bar-justify-content` | `flex-end` | Button row alignment |
| `--action-bar-z-index` | `50` | Stacking order |
| `--action-bar-divider-color` | `rgba(0,0,0,0.08)` | Divider line colour |
| `--action-bar-divider-thickness` | `1px` | Divider line height |

### DOM Structure

```html
<div class="action-bar" data-pw="…">
  <div class="action-bar-divider"></div>   <!-- when divider=true -->
  <div class="action-bar-buttons">
    <!-- <Button> pair or children snippet -->
  </div>
</div>
```

## Notes

- `prettier --check` passes (exit 0)
- `eslint` passes (exit 0)
- Button wired via `{...buttonProperties}` spread, matching the Modal footer pattern
- No typography properties (`font-*`, `line-height`) set in component CSS — design system handles those via semantic tags
- CSS class selectors use kebab-case only (stylelint `selector-class-pattern` compliant)

### DIFF
--- a/src/lib/ActionBar/ActionBar.svelte
+++ b/src/lib/ActionBar/ActionBar.svelte
@@ -1,0 +1,55 @@
+<script lang="ts">
+  import type { ActionBarProperties } from './properties';
+  import Button from '../Button/Button.svelte';
+
+  let {
+    testId,
+    classes,
+    divider = true,
+    primaryButton,
+    secondaryButton,
+    children
+  }: ActionBarProperties = $props();
+</script>
+
+<div class="action-bar {classes ?? ''}" data-pw={typeof testId === 'string' ? testId : null}>
+  {#if divider}
+    <div class="action-bar-divider"></div>
+  {/if}
+  <div class="action-bar-buttons">
+    {#if typeof children === 'function'}
+      {@render children()}
+    {:else}
+      {#if secondaryButton}
+        <Button {...secondaryButton} />
+      {/if}
+      {#if primaryButton}
+        <Button {...primaryButton} />
+      {/if}
+    {/if}
+  </div>
+</div>
+
+<style>
+  .action-bar {
+    background: var(--action-bar-background, #fff);
+    position: var(--action-bar-position, fixed);
+    bottom: var(--action-bar-bottom, 0);
+    left: var(--action-bar-left, 0);
+    right: var(--action-bar-right, 0);
+    padding: var(--action-bar-padding, 12px 16px);
+    z-index: var(--action-bar-z-index, 50);
+    box-sizing: border-box;
+  }
+
+  .action-bar-divider {
+    height: var(--action-bar-divider-thickness, 1px);
+    background: var(--action-bar-divider-color, rgba(0, 0, 0, 0.08));
+  }
+
+  .action-bar-buttons {
+    display: flex;
+    gap: var(--action-bar-gap, 8px);
+    justify-content: var(--action-bar-justify-content, flex-end);
+  }
+</style>
--- a/src/lib/ActionBar/properties.ts
+++ b/src/lib/ActionBar/properties.ts
@@ -1,0 +1,11 @@
+import type { Snippet } from 'svelte';
+import type { ButtonProperties } from '../Button/properties';
+
+export type ActionBarProperties = {
+  testId?: string;
+  classes?: string;
+  divider?: boolean;
+  primaryButton?: ButtonProperties;
+  secondaryButton?: ButtonProperties;
+  children?: Snippet;
+};
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -56,6 +56,7 @@ export { default as EmptyState } from './EmptyState/EmptyState.svelte';
 export { default as Combobox } from './Combobox/Combobox.svelte';
 export { default as ColorPicker } from './ColorPicker/ColorPicker.svelte';
 export { default as SplitInput } from './SplitInput/SplitInput.svelte';
+export { default as ActionBar } from './ActionBar/ActionBar.svelte';
 
 export type * from './Button/properties';
 export type * from './Modal/properties';
@@ -109,5 +110,6 @@ export type * from './EmptyState/properties';
 export type * from './Combobox/properties';
 export type * from './ColorPicker/properties';
 export type * from './SplitInput/properties';
+export type * from './ActionBar/properties';
 
 export { validateInput } from './utils';


sinha-sahil:
Hi @murdore — thanks for the PR and the thorough description / API table. I can see what you're going for: a reusable "save / cancel bar fixed to the bottom of a form" so screens don't have to re-implement that layout. That's a real UX pattern, but I don't think it belongs in this library as a component, and I'd like to walk through why before we go further.

## Why this doesn't fit the library philosophy

This library ships **primitives**, not compositions. The five rules in our design philosophy (and §9 of [GUIDELINES.md](../blob/release/docs/GUIDELINES.md)) draw a hard line around four things, and `ActionBar` crosses three of them:

### 1. It's a composite wrapper (blocker)

Strip the styling and the component is literally:

```svelte
<div class="action-bar">
  <div class="action-bar-divider" />        <!-- a 1px coloured div -->
  <div class="action-bar-buttons">
    <Button {...secondaryButton} />
    <Button {...primaryButton} />
  </div>
</div>
```

The `primaryButton?: ButtonProperties` / `secondaryButton?: ButtonProperties` props are just typed pass-throughs to the existing `Button` primitive. There is no behaviour the library is adding here that a consumer can't write in 6 lines of their own markup. Per principle #4: *components that are merely `<Card><Icon/><Text/><Button/></Card>` belong in user-land, not the library.*

### 2. It's a dummy component (blocker)

It has **no** event handlers of its own, **no** state, **no** focus management, **no** keyboard handling, **no** ARIA role/label, and **no** lifecycle. Without `primaryButton`/`secondaryButton`/`children` it renders an empty `<div>`. Principle #3 says a primitive must justify its existence with non-trivial behaviour, accessibility, or composition value — this has none.

The closest *legitimate* primitive in this space would be something like a `StickyFooter` that handles `env(safe-area-inset-bottom)`, scroll-shadow detection ("am I overlapping content?"), or focus return when the bar acts as a form's primary submit surface. Those are real concerns the library can own. Aligning two buttons in a flex row isn't.

### 3. It bakes in a design preset (blocker)

`divider?: boolean = true` is exactly the kind of variant prop §9 of GUIDELINES.md tells us to push out into consumer CSS. The divider is a single `border-top` line — consumers should turn it on or off (and choose its colour, thickness, style) via a class, not a prop:

```css
/* consumer's app.css */
.action-bar-default { --action-bar-border-top: 1px solid rgba(0,0,0,0.08); }
.action-bar-flush   { --action-bar-border-top: none; }
.action-bar-strong  { --action-bar-border-top: 2px solid #111; }
```

`background: #fff` as the default is a brand decision too — primitives should default to `transparent`/unset.

## What I'd recommend consumers do instead

This is roughly 12 lines of user-land code with full theming control:

```svelte
<!-- consumer's CheckoutPage.svelte -->
<div class="checkout-action-bar">
  <Button classes="btn-secondary" text="Cancel" onclick={onCancel} />
  <Button classes="btn-primary"   text="Pay"    onclick={onSubmit} />
</div>
```

```css
.checkout-action-bar {
  position: fixed;
  inset: auto 0 0 0;        /* bottom-fixed across the viewport */
  display: flex;
  gap: 8px;
  justify-content: flex-end;
  padding: 12px 16px calc(12px + env(safe-area-inset-bottom));
  background: #fff;
  border-top: 1px solid rgba(0, 0, 0, 0.08);
  z-index: 50;
  box-sizing: border-box;
}

.btn-primary   { --button-color: #0070f3; --button-text-color: #fff; }
.btn-secondary { --button-color: transparent; --button-text-color: #0070f3; --button-border: 1px solid #0070f3; }
```

That's the pattern §9 describes — variants live in consumer CSS classes that override CSS variables on existing primitives.

## GUIDELINES.md nits (for completeness, in case any of this gets salvaged)

- **§3 — falsy checks on objects.** `{#if secondaryButton}` and `{#if primaryButton}` (ActionBar.svelte:22, 25) are objects, not booleans. Use `{#if typeof primaryButton === 'object'}` to match the `typeof children === 'function'` check you already have on line 20.
- **§9 — `divider` prop.** Covered above — replace with a CSS variable hook.

## Ask

I'd like to **close this PR** and instead either:

1. Document the pattern above in `docs/` as a recipe ("How to build a sticky action bar with `Button`"), or
2. If there's a behavioural concern this PR was masking (safe-area handling, focus return, scroll-shadow), open a smaller PR that adds *only* that behaviour as a focused primitive — and let consumers compose `Button` inside it.

Happy to chat about which of those is most useful for the screens that motivated this. Thanks again for the clean implementation — the prop hygiene (`testId`, `classes`, spread API) is genuinely good; it's the *shape* of the component that doesn't fit.

murdore:
@sinha-sahil — thank you for the detailed and generous review. You are right on all three counts, and I appreciate you spelling out the reasoning rather than just declining.

To confirm I've understood: `ActionBar` is a thin composition of two existing `Button` calls inside a positioned `<div>`. There is no behaviour the library owns here — no focus management, no ARIA semantics, no safe-area handling — so by the library's own principle #3 it doesn't justify a component slot. The `divider` boolean is precisely the baked-in variant prop §9 says should live in consumer CSS, and `background: #fff` is a brand default that a primitive should never set. Your 12-line consumer snippet covers the full pattern without any library surface.

I am happy to close this PR. Before I do, two quick questions to make sure I pick the right follow-up:

1. Would a short recipe in `docs/` ("Sticky action bar with `Button`") be a useful addition, covering the `env(safe-area-inset-bottom)` offset too — or is that out of scope for the docs?
2. Is there any appetite for a minimal `StickyFooter` primitive later that handles only the scroll-shadow / safe-area concerns you mentioned, leaving button composition entirely to the consumer?

Happy to act on whichever direction is most useful. Closing now.

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — thank you for the gracious response, and especially for restating the reasoning back in your own words. That's the part that matters: "no behaviour the library owns" and "variant props belong in consumer CSS" are exactly the heuristics I want everyone reaching for before opening a component PR. Closing this one is the right call.

To your two questions:

**1. Docs recipe — yes, please.**

A short page under `docs/` titled something like "Recipes: sticky action bar" would be genuinely useful. The 12-line snippet from my earlier comment is most of it; the bits worth calling out explicitly are:

- `padding-bottom: calc(<base> + env(safe-area-inset-bottom))` for iOS home-indicator safe area.
- `position: fixed; inset: auto 0 0 0; z-index: <app-stack>` and why `z-index` is left to the consumer.
- The pattern of using `--button-color`, `--button-text-color`, `--button-border` on existing `Button` instances rather than reaching for variant props.
- A note that this same pattern (composition + CSS vars on primitives) is how `Toolbar`, footer CTAs, mobile bottom sheets, etc. should be built — so the recipe pulls weight across multiple use-cases.

If you want to open that as a small docs-only PR, I'll review it the same week. Keep it framed as a recipe, not as guidance that this *should* have been a component — the point is to show how easy the user-land version is.

**2. `StickyFooter` primitive — only if it earns its keep.**

I'm cautious here, because "StickyFooter that handles safe-area" is one design discussion away from becoming ActionBar v2 with `primaryButton` / `secondaryButton` slots bolted back on. The bar for this to ship as a primitive is:

- It owns **non-trivial behaviour** a consumer genuinely can't replicate in CSS — e.g. a scroll-shadow that appears only when the page actually has overflow above the footer (requires `IntersectionObserver` or scroll listener + dimension math), or focus-return when the footer hosts a form's submit target.
- It exposes a **single `children` Snippet** and zero opinions about what goes inside. No `primary`/`secondary` slots, no `divider` prop, no `background` default beyond `transparent`.
- Safe-area handling alone is **not enough** — that's a one-line `padding-bottom: calc(... + env(safe-area-inset-bottom))` and belongs in the recipe, not in a component.

If you can make the case that the scroll-shadow behaviour (or some equivalent) is doing real work that consumers can't trivially write, open a fresh issue first with the proposed API surface and I'll work through it with you before any code lands. That keeps us from re-litigating the composite question downstream.

Thanks again for the clean handling here — the prop hygiene on this PR (`testId`, `classes`, spread API) is the right shape; it just needed to be applied to a primitive that earns its own slot.

**Ask: please close this PR.** Docs recipe welcome as a separate PR; `StickyFooter` only via a fresh issue with a justified behavioural surface.

murdore:
Closing — ActionBar is a thin two-Button composition; a docs recipe will follow (a StickyFooter only via a fresh issue with a real behavioural surface). Thanks @sinha-sahil.

PR #203, opened 2026-05-29
feat: add subheading and testId props to Toolbar

murdore:
## Summary

- Adds four optional, backward-compatible props to `Toolbar`: `testId`, `headingTestId`, `subheading`, `subheadingTestId`
- `testId` renders as `data-pw` on the Toolbar root `<div>`
- `headingTestId` renders as `data-pw` on the existing title `<div>`
- `subheading` renders a secondary line beneath the main heading (only when provided — DOM is otherwise unchanged)
- `subheadingTestId` renders as `data-pw` on the subheading `<div>`
- Exposes two new CSS custom properties: `--toolbar-subheading-color` (default `inherit`) and `--toolbar-subheading-font-size` (default `inherit`) — no hardcoded font values

### New props (all optional)

|---|---|---|---|
| `testId` | `string` | `undefined` | `data-pw` attribute on the Toolbar root |
| `headingTestId` | `string` | `undefined` | `data-pw` attribute on the title element |
| `subheading` | `string` | `undefined` | Secondary line rendered under the main heading |
| `subheadingTestId` | `string` | `undefined` | `data-pw` attribute on the subheading element |

### New CSS custom properties

|---|---|---|
| `--toolbar-subheading-color` | `inherit` | Subheading text color |
| `--toolbar-subheading-font-size` | `inherit` | Subheading font size |

### DOM change (only when `text` is provided)

The existing `.text` div is now wrapped in a `.text-block` flex column; `flex: var(--toolbar-text-flex, 1)` moves to the wrapper so the layout contract is preserved exactly.

## Notes

- `prettier --check` passes on all changed files
- `eslint` passes with no output
- All new props are optional; omitting them reproduces today's behavior byte-for-byte
- No existing props renamed or removed; all defaults unchanged

### DIFF
--- a/docs/Toolbar.md
+++ b/docs/Toolbar.md
@@ -14,12 +14,16 @@ A fixed-position header bar with a back button (left), center title text, and cu
 
 ## Props
 
-| Prop           | Type             | Required | Default                                          | Description                                                                                                                                                            |
-| -------------- | ---------------- | -------- | ------------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| showBackButton | `boolean`        | No       | `true`                                           | Whether to show the default back button on the left side. Only shown when leftContent snippet is not provided.                                                         |
-| text           | `string \| null` | No       | `-`                                              | Title text displayed in the center of the toolbar. Only shown when centerContent snippet is not provided.                                                              |
-| backIcon       | `string \| null` | No       | `'https://sdk.breeze.in/gallery/icons/back.svg'` | URL for the back button icon image. Defaults to a back-arrow SVG from sdk.breeze.in.                                                                                   |
-| classes        | `string`         | No       | `-`                                              | CSS class string applied to the component's top-level element. Useful for theming — define classes with CSS variable overrides and pass them to create variant styles. |
+| Prop             | Type             | Required | Default                                          | Description                                                                                                                                                            |
+| ---------------- | ---------------- | -------- | ------------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| showBackButton   | `boolean`        | No       | `true`                                           | Whether to show the default back button on the left side. Only shown when leftContent snippet is not provided.                                                         |
+| text             | `string \| null` | No       | `-`                                              | Title text displayed in the center of the toolbar. Only shown when centerContent snippet is not provided.                                                              |
+| subheading       | `string`         | No       | `-`                                              | Optional secondary line of text displayed below the main `text` heading. Only rendered when `text` is also present and this string is non-empty.                      |
+| backIcon         | `string \| null` | No       | `'https://sdk.breeze.in/gallery/icons/back.svg'` | URL for the back button icon image. Defaults to a back-arrow SVG from sdk.breeze.in.                                                                                   |
+| classes          | `string`         | No       | `-`                                              | CSS class string applied to the component's top-level element. Useful for theming — define classes with CSS variable overrides and pass them to create variant styles. |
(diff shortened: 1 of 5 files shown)


sinha-sahil:
Thanks for the PR. Two distinct things are bundled here, and they land in different places on the philosophy.

Keep (testIds) — this part is clean. `testId`, `headingTestId`, `subheadingTestId` rendering as `data-pw` is exactly the library-wide convention. `Toolbar.svelte:18` (root) and `Toolbar.svelte:34` (heading) — both spots are correct and match what every other component does. No objection.

Drop (`subheading`) — this is a baked typographic slot. A `subheading` string prop with its own `.subheading` div, `--toolbar-subheading-color`, and `--toolbar-subheading-font-size` (`Toolbar.svelte:38-43`, `Toolbar.svelte:131-136`) is a second built-in text slot stacked on top of the existing `text` slot. That falls under "no baked typographic hierarchy" — once a primitive starts shipping title+subtitle pairs, the next ask is `caption`, then `eyebrow`, then `metadata`, and we're rebuilding `<PageHeader>` inside Toolbar one prop at a time.

The existing `centerContent` snippet already covers this case completely — a consumer writes a centerContent snippet with their own title and subtitle divs and their own classes, in about five lines. That is the consumer's design system, not ours.

Note also: the existing `text` prop is itself a grandfathered convenience shortcut — the direction we want is toward the snippet API for any structured content, not away from it by adding more string props.

Ask: please split — open a follow-up PR with just the testIds (`testId`, `headingTestId`; drop `subheadingTestId` since there is no subheading element to tag), and close this PR. Happy to fast-approve the testId-only split.

sinha-sahil (changes requested):

See the inline comment for the specific concerns and the ask. Until those are addressed, this review intentionally blocks the merge.

murdore:
@sinha-sahil — per your review feedback, #203 now showcases `subheading` + the testId props on the Toolbar route, documents them (props + CSS-var rows) in `docs/Toolbar.md`, and exposes them in `Toolbar.wc.svelte`. Verified locally: `pnpm check` + `check:wc` clean. Ready for re-review 🙏

murdore:
Closing this PR in favour of the slim split requested during review.

**New PR: https://github.com/juspay/svelte-ui-components/pull/257** — `feat(Toolbar): add testId + headingTestId`

## What changed and why

- **`subheading` and `subheadingTestId` props dropped** — maintainer feedback (§2 of library conventions): string props that carry presentational structure are rejected in favour of Snippets. Consumers needing a subheading line should use `{#snippet centerContent()}` to render their own heading + subheading markup. A ready-to-use consumer recipe has been added to `docs/Toolbar.md`.

- **`subheadingTestId` dropped** — unnecessary once the subheading element lives in consumer code; the consumer applies `data-pw` directly to their own element.

- **`.text-block` wrapper div removed** — the PR introduced a wrapping `<div class="text-block">` to host the flex-column layout for heading + subheading. Without the subheading, this wrapper has no purpose. The original single `<div class="text">` with `flex: var(--toolbar-text-flex, 1)` is restored exactly as it was before this PR.

- **`.subheading` CSS rule removed** — no longer needed.

- **`--toolbar-text-flex` CSS variable kept on `.text` not `.text-block`** — the flex value returns to the `.text` element where it was before this PR, preserving the original layout contract.

- **`testId` and `headingTestId` kept** — these are the only changes the maintainer confirmed "happy to fast-approve". They have no equivalent Snippet-based recipe (you can't set `data-pw` on the library's root element from a consumer Snippet), so they genuinely belong in the library API.

## Gates (on the new PR branch)

- `npm run check` — 0 errors, 0 warnings
- `npm run lint` — clean
- `npm run build` — clean

murdore:
Closing in favour of slim split PR #257 which contains only the testId + headingTestId changes per review feedback.

PR #204, opened 2026-05-29
feat: add triggerTestId, triggerAriaLabel, and portal to Menu

murdore:
## Summary

- **`triggerTestId?: string`** — renders `data-pw` on the trigger `<div>` element, enabling targeted test selectors for the button independently of the root container's `testId`.
- **`triggerAriaLabel?: string`** — renders `aria-label` on the trigger button element for improved screen-reader accessibility.
- **`portal?: boolean` (default `false`)** — when `true`, renders the dropdown list as `position: fixed` relative to the viewport by computing the trigger's `getBoundingClientRect()` on open. This lets the dropdown escape ancestor `overflow: hidden` / stacking contexts. When `false`, the existing `position: absolute` behavior is preserved byte-for-byte. All `window`/`document` access is guarded for SSR safety.

### New Props

| `triggerTestId` | `string` | — | `data-pw` attribute on the trigger element |
| `triggerAriaLabel` | `string` | — | `aria-label` attribute on the trigger button |
| `portal` | `boolean` | `false` | Render dropdown fixed to viewport, escaping overflow clipping |

### New CSS Custom Properties

| `--menu-dropdown-right` | `auto` | Right edge of the dropdown (for right-anchored placements) |
| `--menu-dropdown-bottom` | `auto` | Bottom edge of the dropdown (for bottom-anchored placements) |

These complement the existing `--menu-dropdown-top` and `--menu-dropdown-left` vars.

### DOM Change (portal mode)

When `portal=true`, the `.menu-dropdown` element is rendered as a direct child of the component's root in the DOM (outside `.menu-container`) with `position: fixed` and `margin: 0`. The inline `top`/`left` style is set from `getBoundingClientRect()` at open time.

## Notes

- `prettier --check` exits **0** (no formatting changes needed)
- `eslint` exits **0** (no lint issues)
- All new props are optional — omitting them reproduces today's behavior exactly
- Backward-compatible: no existing props, defaults, or behaviors changed


sinha-sahil:

New arrow-const handlers in the diff: `computePortalPosition`. Please write them as `function fnName() {}` instead.

murdore:
@sinha-sahil — per your review feedback, #204 now demos `triggerTestId` / `triggerAriaLabel` / `portal` on the Menu route, documents them (props + CSS vars) in `docs/Menu.md`, and exposes them in `Menu.wc.svelte`. Verified locally: `pnpm check` + `check:wc` clean. Ready for re-review 🙏

murdore:
Thanks for the careful implementation here. After reviewing, all three props are already covered by the existing trigger snippet API — consumers can render any attributes they need (including `data-pw` and `aria-label`) directly on the trigger element via the snippet, making `triggerTestId` and `triggerAriaLabel` redundant as first-class props. The `portal` prop duplicates item markup and introduces inline-style positioning that conflicts with how the component's CSS vars (`--menu-dropdown-top`, `--menu-dropdown-left`) are intended to be used. The two new CSS vars (`--menu-dropdown-right`, `--menu-dropdown-bottom`) are genuinely useful and could land as a standalone single-commit PR against the existing CSS block without any JS or prop changes. Closing in favour of that narrower follow-up.

PR #205, opened 2026-05-29
feat: add InlineAlert component

murdore:
## Summary

- New `InlineAlert` component — an inline (non-sticky, non-transient) alert/message primitive, distinct from `Banner` (sticky page strip) and `Toast` (transient overlay).
- Four tone variants: `info` (blue-grey), `success` (green), `warning` (amber), `error` (red) — all using soft backgrounds with readable text via per-tone CSS custom properties.
- Dismissible with `$bindable visible` prop and optional `ondismiss` callback; fade transition on enter/leave.

### Props

| `text` | `string` | — | Plain text content |
| `tone` | `'info' \| 'success' \| 'warning' \| 'error'` | `'info'` | Visual tone variant |
| `icon` | `Snippet` | — | Optional leading icon slot |
| `children` | `Snippet` | — | Rich body content (overrides `text`) |
| `dismissible` | `boolean` | `false` | Show dismiss button |
| `visible` | `boolean` ($bindable) | `true` | Controls visibility |
| `testId` | `string` | — | Sets `data-pw` on root; dismiss button gets `{testId}-dismiss` |
| `classes` | `string` | — | Extra CSS classes on root element |
| `ondismiss` | `() => void` | — | Called when dismissed |

### CSS Custom Properties

| `--inline-alert-padding` | `10px 12px` | Root padding |
| `--inline-alert-gap` | `8px` | Gap between icon / body / dismiss |
| `--inline-alert-radius` | `6px` | Border radius |
| `--inline-alert-border` | `none` | Border |
| `--inline-alert-width` | `100%` | Width |
| `--inline-alert-icon-size` | `16px` | Icon SVG size |
| `--inline-alert-icon-color` | `currentColor` | Icon color |
| `--inline-alert-dismiss-size` | `14px` | Dismiss icon size |
| `--inline-alert-dismiss-padding` | `2px` | Dismiss button padding |
| `--inline-alert-dismiss-radius` | `4px` | Dismiss button border radius |
| `--inline-alert-dismiss-opacity` | `0.7` | Dismiss button default opacity |
| `--inline-alert-dismiss-hover-background` | `rgba(0,0,0,0.08)` | Dismiss hover background |
| `--inline-alert-info-background` | `#e8f0fe` | Info tone background |
| `--inline-alert-info-color` | `#1a56a0` | Info tone text color |
| `--inline-alert-success-background` | `#dcfce7` | Success tone background |
| `--inline-alert-success-color` | `#166534` | Success tone text color |
| `--inline-alert-warning-background` | `#fef9c3` | Warning tone background |
| `--inline-alert-warning-color` | `#854d0e` | Warning tone text color |
| `--inline-alert-error-background` | `#fee2e2` | Error tone background |
| `--inline-alert-error-color` | `#991b1b` | Error tone text color |

### DOM / Accessibility

- `role="alert"` + `aria-live="assertive"` for `tone="error"`, `role="status"` + `aria-live="polite"` for all others.
- Dismiss button is a native `<button type="button">` with `aria-label="Dismiss"`.
- `data-pw={testId}` on root; `data-pw="{testId}-dismiss"` on dismiss button.

## Notes

- `prettier --check` — clean
- `eslint` — clean
- Backward-compatible: new export appended after `SplitInput` in `src/lib/index.ts`
- No hardcoded `font-family`, `font-size`, `font-weight`, or `line-height` in component CSS — all visual properties driven by CSS custom properties
- Svelte 5 runes throughout (`$props`, `$bindable`, `$state` not needed — pure props-driven)

### DIFF
--- a/src/lib/InlineAlert/InlineAlert.svelte
+++ b/src/lib/InlineAlert/InlineAlert.svelte
@@ -1,0 +1,140 @@
+<script lang="ts">
+  import { fade } from 'svelte/transition';
+  import type { InlineAlertProperties } from './properties';
+  import closeSvg from '$lib/assets/close.svg?raw';
+
+  let {
+    text,
+    tone = 'info',
+    icon,
+    children,
+    dismissible = false,
+    visible = $bindable(true),
+    testId,
+    classes,
+    ondismiss
+  }: InlineAlertProperties = $props();
+
+  const handleDismiss = (): void => {
+    visible = false;
+    ondismiss?.();
+  };
+
+  const handleDismissKeydown = (event: KeyboardEvent): void => {
+    if (event.key === 'Enter' || event.key === ' ') {
+      event.preventDefault();
+      handleDismiss();
+    }
+  };
+</script>
+
+{#if visible}
+  <div
+    class="inline-alert inline-alert-{tone} {classes ?? ''}"
+    role={tone === 'error' ? 'alert' : 'status'}
+    aria-live={tone === 'error' ? 'assertive' : 'polite'}
+    data-pw={typeof testId === 'string' ? testId : null}
+    transition:fade={{ duration: 150 }}
+  >
+    {#if typeof icon === 'function'}
+      <div class="inline-alert-icon">
+        {@render icon()}
+      </div>
+    {/if}
+
+    <div class="inline-alert-body">
+      {#if typeof children === 'function'}
+        {@render children()}
+      {:else if typeof text === 'string'}
+        {text}
+      {/if}
+    </div>
+
+    {#if dismissible}
+      <button
+        class="inline-alert-dismiss"
+        type="button"
+        aria-label="Dismiss"
+        onclick={handleDismiss}
+        onkeydown={handleDismissKeydown}
+        {...typeof testId === 'string' ? { 'data-pw': `${testId}-dismiss` } : {}}
+      >
+        <!-- eslint-disable svelte/no-at-html-tags -->
+        {@html closeSvg}
+      </button>
+    {/if}
+  </div>
+{/if}
+
+<style>
+  .inline-alert {
+    display: flex;
+    align-items: var(--inline-alert-align-items, center);
+    gap: var(--inline-alert-gap, 8px);
+    padding: var(--inline-alert-padding, 10px 12px);
+    border-radius: var(--inline-alert-radius, 6px);
+    border: var(--inline-alert-border, none);
+    width: var(--inline-alert-width, 100%);
+    box-sizing: border-box;
+  }
+
+  .inline-alert-info {
+    background-color: var(--inline-alert-info-background, #e8f0fe);
+    color: var(--inline-alert-info-color, #1a56a0);
+  }
+
+  .inline-alert-success {
+    background-color: var(--inline-alert-success-background, #dcfce7);
+    color: var(--inline-alert-success-color, #166534);
+  }
+
+  .inline-alert-warning {
+    background-color: var(--inline-alert-warning-background, #fef9c3);
+    color: var(--inline-alert-warning-color, #854d0e);
+  }
+
+  .inline-alert-error {
+    background-color: var(--inline-alert-error-background, #fee2e2);
+    color: var(--inline-alert-error-color, #991b1b);
+  }
+
+  .inline-alert-icon {
+    display: flex;
+    align-items: center;
+    flex-shrink: 0;
+    color: var(--inline-alert-icon-color, currentColor);
+  }
+
+  .inline-alert-icon :global(svg) {
+    width: var(--inline-alert-icon-size, 16px);
+    height: var(--inline-alert-icon-size, 16px);
+  }
+
+  .inline-alert-body {
+    flex: 1;
+  }
+
+  .inline-alert-dismiss {
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    flex-shrink: 0;
+    background: transparent;
+    border: none;
+    cursor: pointer;
+    padding: var(--inline-alert-dismiss-padding, 2px);
+    border-radius: var(--inline-alert-dismiss-radius, 4px);
+    color: currentColor;
+    opacity: var(--inline-alert-dismiss-opacity, 0.7);
+  }
+
+  .inline-alert-dismiss:hover {
+    opacity: 1;
+    background-color: var(--inline-alert-dismiss-hover-background, rgba(0, 0, 0, 0.08));
+  }
+
+  .inline-alert-dismiss :global(svg) {
+    width: var(--inline-alert-dismiss-size, 14px);
+    height: var(--inline-alert-dismiss-size, 14px);
+  }
+</style>
--- a/src/lib/InlineAlert/properties.ts
+++ b/src/lib/InlineAlert/properties.ts
@@ -1,0 +1,15 @@
+import type { Snippet } from 'svelte';
+
+export type InlineAlertTone = 'info' | 'success' | 'warning' | 'error';
+
+export type InlineAlertProperties = {
+  text?: string;
+  tone?: InlineAlertTone;
+  icon?: Snippet;
+  children?: Snippet;
+  dismissible?: boolean;
+  visible?: boolean;
+  testId?: string;
+  classes?: string;
+  ondismiss?: () => void;
+};
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -56,6 +56,7 @@ export { default as EmptyState } from './EmptyState/EmptyState.svelte';
 export { default as Combobox } from './Combobox/Combobox.svelte';
 export { default as ColorPicker } from './ColorPicker/ColorPicker.svelte';
 export { default as SplitInput } from './SplitInput/SplitInput.svelte';
+export { default as InlineAlert } from './InlineAlert/InlineAlert.svelte';
 
 export type * from './Button/properties';
 export type * from './Modal/properties';
@@ -109,5 +110,6 @@ export type * from './EmptyState/properties';
 export type * from './Combobox/properties';
 export type * from './ColorPicker/properties';
 export type * from './SplitInput/properties';
+export type * from './InlineAlert/properties';
 
 export { validateInput } from './utils';


sinha-sahil:
Thanks for the well-documented PR @murdore — the table of CSS variables, the accessibility notes, and the clean `svelte-check` / `eslint` / `prettier` results all show real care. Unfortunately I have to push back on this one because, in its current shape, it conflicts with two of the library's most load-bearing rules, and I want to walk through why before suggesting a path forward.

## The blocker: `tone` is a hardcoded design preset

The library's central thesis (philosophy principle 2 and GUIDELINES §9) is that **components are unstyled primitives — variants live in consumer CSS classes, not in the component**. A `tone` / `variant` / `size` / `flush` prop that maps to baked-in CSS inside the component is exactly what we've ruled out, because:

1. Every consumer that ships a design system now has two competing variant axes — ours and theirs — and they have to fight us via `--inline-alert-info-background` overrides for every tone they didn't want.
2. The four tones bake in a specific palette (`#e8f0fe`, `#dcfce7`, `#fef9c3`, `#fee2e2`) plus a specific naming taxonomy ("info / success / warning / error"). A consumer whose design system has "neutral / positive / caution / destructive / brand" can't extend the enum — they're stuck patching CSS for variants they didn't ask for.
3. GUIDELINES §9 explicitly demonstrates the alternative using the *exact same use case* (`.banner-success`, `.banner-warning`) — so this is not a gray area, it's the worked example.

Compare with how `Banner` (which this component largely overlaps with — more on that below) handles the same problem: zero hardcoded variants, just `--banner-background` / `--banner-color`, and the docs show consumers writing `.banner-success { --banner-background: #e6f9ed; --banner-color: #1a7f37; }` themselves.

### Concretely, the refactor looks like:

**Remove from the component:**
```svelte
<!-- gone -->
class="inline-alert inline-alert-{tone} {classes ?? ''}"
```
```css
/* all four of these blocks go */
.inline-alert-info    { background-color: var(--inline-alert-info-background, #e8f0fe);    color: var(--inline-alert-info-color, #1a56a0); }
.inline-alert-success { background-color: var(--inline-alert-success-background, #dcfce7); color: var(--inline-alert-success-color, #166534); }
.inline-alert-warning { background-color: var(--inline-alert-warning-background, #fef9c3); color: var(--inline-alert-warning-color, #854d0e); }
.inline-alert-error   { background-color: var(--inline-alert-error-background, #fee2e2);   color: var(--inline-alert-error-color, #991b1b); }
```

**Replace with:**
```svelte
<div
  class="inline-alert {classes ?? ''}"
  role={role ?? 'status'}
  aria-live={role === 'alert' ? 'assertive' : 'polite'}
  ...
>
```
```css
.inline-alert {
  background-color: var(--inline-alert-background);
  color: var(--inline-alert-color);
  /* ... existing structural vars ... */
}
```

**Move the four tones into the consumer's stylesheet (and into the component's docs as the recommended pattern):**
```css
.alert-info    { --inline-alert-background: #e8f0fe; --inline-alert-color: #1a56a0; }
.alert-success { --inline-alert-background: #dcfce7; --inline-alert-color: #166534; }
.alert-warning { --inline-alert-background: #fef9c3; --inline-alert-color: #854d0e; }
.alert-error   { --inline-alert-background: #fee2e2; --inline-alert-color: #991b1b; }
```
```svelte
<InlineAlert classes="alert-success" text="Saved" />
<InlineAlert classes="alert-error" text="Network down" role="alert" />
```

The `role`/`aria-live` distinction (which is the one genuinely valuable thing the `tone` prop was doing) becomes an explicit `role?: 'alert' | 'status'` prop — decoupled from the visual treatment, which is how it should be: a consumer might want a red destructive-looking alert that's still `role="status"`, or a neutral-styled one that's `role="alert"`.

## The other blocker: this is basically `Banner` with a tone enum

Look at `Banner.svelte` next to this PR side-by-side and the overlap is striking:

| Feature | Banner | InlineAlert |
|---|---|---|
| `text` + `icon` snippet | yes | yes |
| `dismissible` + `visible = $bindable(true)` + `ondismiss` | yes | yes |
| `classes`, `testId`, dismiss button with `${testId}-dismiss` | yes | yes |
| Default container layout | flex row, icon + body + dismiss | flex row, icon + body + dismiss |
| Sticky default | yes (`position: sticky`) | no |
| `role`/`aria-live` | — | `tone === 'error'` branches |

That's two real deltas: the sticky default and the role wiring. Per philosophy principle 4 (no composite wrappers), that's not enough surface to justify a second primitive. Two reasonable paths:

1. **Extend `Banner`** with a `role?: 'alert' | 'status' | null` prop, and let consumers opt into a non-sticky variant via `classes="banner-inline"` + `.banner-inline { --banner-position: static; }`. That solves both deltas without a new component.
2. **Keep InlineAlert** but justify it with behavior `Banner` can't reasonably express (e.g., automatic dedup, queueing, focus-management semantics). As written, it doesn't.

I'd lean toward (1) — that's the principle-4-correct answer.

## Smaller GUIDELINES issues to fix in the same pass

- **§4 / §6**: The dismiss button is a raw `<button>` with a custom `handleDismissKeydown` for Enter/Space. The existing `Button` component already handles all of that. Banner shows the pattern (lines 71–84): wrap `<Button>` in a `<div class="inline-alert-dismiss">` and bridge `--button-*` to `--inline-alert-dismiss-*`. This deletes `handleDismissKeydown` entirely, and the raw-button `background: transparent; border: none; cursor: pointer;` declarations go away with it.

## Things this PR does right (worth keeping in the rework)

- All structural CSS goes through variables — sizes, spacing, radius, gap, dismiss opacity, hover background, icon size/color.
- `classes` is appended to the root per §9.
- `testId` propagates to `data-pw` on both root and dismiss button.
- `$bindable(true)` on `visible` with internal mutation is textbook §7.
- Type narrowing via `typeof` everywhere — no `!!`, no truthy fallbacks, no `auto` in layout. §1 / §3 / §8 are clean.

## Ask

I'd like to **block this PR** in its current form, and would suggest one of:

1. **Preferred**: Close this PR, open a smaller PR that extends `Banner` with the `role` prop + a `--banner-position` variable, and add `.banner-inline` to the docs as the consumer-side recipe.
2. **Acceptable**: Rework this PR to (a) remove the `tone` prop and the four baked tone rule blocks, (b) replace the raw dismiss `<button>` with `<Button>` + CSS-variable bridging à la Banner, (c) clearly articulate in the PR description what InlineAlert does that Banner cannot, and (d) add a docs page (`docs/InlineAlert.md`) showing the `.alert-success` consumer-CSS pattern as the canonical theming recipe.

Happy to pair on whichever direction you'd like to go — the structural CSS work here is solid and most of it carries over directly.

murdore:
@sinha-sahil — thank you for the thorough, code-level walkthrough. I read it carefully and you're right on both counts.

The `tone` enum is exactly what §9 rules out: I baked four hardcoded palette presets into the component itself, which forces every consumer to fight the default variants rather than owning them from the start. I understand now that the variants belong in the consumer's stylesheet via `classes` + CSS vars — and you even showed the `Banner` docs as a worked example of that pattern, so there's no ambiguity.

The `Banner` overlap point also lands. Looking at the feature matrix you laid out, the only real deltas are the sticky default and the `role`/`aria-live` wiring — that's not enough surface area to justify a second primitive.

I'd like to go with your preferred path (option 1):

- Close this PR.
- Open a smaller PR extending `Banner` with a `role?: 'alert' | 'status'` prop.
- Expose a `--banner-position` CSS variable so consumers can opt into a non-sticky `.banner-inline` shape.
- Add the consumer-side tone recipe (`.banner-success`, `.banner-error`, etc.) to the `Banner` docs.

Before I start: would that `Banner` extension PR be the shape you'd accept, or is there anything about the `role`/`aria-live` wiring you'd want handled differently (e.g. keeping `aria-live` purely consumer-driven rather than derived from `role`)?

sinha-sahil:

New arrow-const handlers in the diff: `handleDismiss`, `handleDismissKeydown`. Please write them as `function fnName() {}` instead.

sinha-sahil (changes requested):

To merge:

sinha-sahil:
Appreciate the clean read on this, @murdore — "I baked four hardcoded palette presets into the component itself, which forces every consumer to fight the default variants rather than owning them from the start" is exactly the framing, and your option-1 plan is the right one. Let's lock that in.

**On your `aria-live` question — keep them decoupled, both consumer-driven.**

You're right to flag this; deriving `aria-live` from `role` looks tidy but it's the same class of trap as the `tone` enum. We'd be baking an opinionated mapping (`role="alert"` → `aria-live="assertive"`, everything else → `polite`) into the primitive, which:

1. **Doesn't always hold.** A consumer might want `role="status"` (semantic: advisory, non-interrupting) with `aria-live="assertive"` (because *this particular* status update is important enough to interrupt), or `role="alert"` with no `aria-live` at all because they're managing announcement timing via a live-region manager elsewhere on the page. The mapping is a heuristic, not a law — primitives shouldn't encode heuristics.
2. **Hides intent at the call site.** When a consumer reads `<Banner role="alert" />`, they shouldn't have to remember "and that silently sets aria-live=assertive." Two explicit props read better and audit better.
3. **It's the §9 pattern again at the ARIA layer.** Same principle: don't infer; let the consumer state what they want.

So the Banner extension surface I'd accept is:

```ts
type OptionalBannerProperties = {
  // ...existing...
  role?: 'alert' | 'status' | 'log' | 'none' | null;
  ariaLive?: 'off' | 'polite' | 'assertive' | null;
  ariaAtomic?: boolean | null;
};
```

Pass each straight through to the root element; no derivation, no defaults beyond `null` (i.e. attribute omitted if not provided). If a consumer wants the common pairing they write `<Banner role="alert" ariaLive="assertive" />` — two characters more, zero ambiguity.

**Confirming the rest of your plan:**

- Close this PR — yes, please go ahead.
- New PR: extend `Banner` with `role` + `ariaLive` (+ `ariaAtomic` while you're there) + `--banner-position` CSS variable.
- Default for `--banner-position` stays `sticky` (Banner's current behaviour — don't regress existing consumers). Consumers opt into inline via `.banner-inline { --banner-position: static; }`.
- Docs: add the `.banner-success` / `.banner-warning` / `.banner-error` / `.banner-info` recipe to `docs/Banner.md` as the canonical theming pattern — and mirror it in the MCP-generated docs so the agent surface teaches the right shape too.
- Carry over the wins from this PR: `testId` propagation to the dismiss button (`${testId}-dismiss`) is already in Banner, but double-check the structural CSS variable coverage in your new PR matches what you'd added here (gap, icon size, dismiss hover bg, etc.) — those were genuinely good additions and Banner can absorb them.
- And per my follow-up note: write `handleDismiss` as `function handleDismiss() {}`, not an arrow const, when it lands in the new PR.

One last thing while you're in there: drop `handleDismissKeydown` entirely. Once the dismiss button is `<Button>` per §6, Enter/Space handling comes for free and that whole function goes away.

**Ask: please close this PR.** The Banner-extension PR is the right next step — ping me on the draft and I'll review fast.

murdore:
Closing — InlineAlert overlaps Banner (§4). I'll open a slim Banner extension (role/aria-live + a --banner-position for the non-sticky shape) instead. Thanks @sinha-sahil.

PR #206, opened 2026-05-29
feat: add animationType and onAfterOpen to Sheet

murdore:
## Summary

- Adds optional `animationType?: 'fly' | 'width'` prop to `Sheet`. Default is `'fly'`, which preserves the existing translate-based animation byte-for-byte. Choosing `'width'` animates the panel by expanding its CSS `width` (or `height` for top/bottom sides) via `@keyframes`, avoiding translate reflow — ideal for sheets that contain embedded charts.
- Adds optional `onAfterOpen?: () => void` event prop. Fires once after the open transition is fully complete — via `onintroend` (fly mode) or `onanimationend` (width mode) — letting consumers trigger chart reflows or layout measurements at the right moment.
- New CSS custom property `--sheet-width-transition` (default `300ms`) controls the width-animation duration.

### New props (`OptionalSheetProperties` / `SheetEventProperties`)

| `animationType` | `'fly' \| 'width'` | `'fly'` | Panel open/close animation strategy |
| `onAfterOpen` | `() => void` | — | Callback fired once after the open transition completes |

### New CSS custom property

| `--sheet-width-transition` | `300ms` | Duration of the width/height expand animation (only used when `animationType='width'`) |

### New exported type

`SheetAnimationType = 'fly' | 'width'` — exported from `properties.ts` and re-exported via `src/lib/index.ts` through the `export type *` barrel.

## Notes

- **svelte-check**: 0 errors, 0 warnings
- **prettier**: all matched files use Prettier code style
- **eslint**: no errors or warnings
- **Backward-compatible**: omitting both new props reproduces today's behavior exactly. No existing prop renamed or removed.

## Summary by CodeRabbit

* **New Features**
  * Sheet component animations can now be customized via CSS variables for keyframe names, durations, easing, and fill modes.

* **Documentation**
  * Added documentation for animation-related CSS variables and their customization options.


sinha-sahil:
Thanks for the PR — the motivation here is solid. The `onAfterOpen` callback is a genuine gap (there's no clean way today to know when the transition is done, and chart reflow really does need that signal), and I want to land it. Before merging, though, the `animationType` half of this PR needs a rework — it conflicts with the library's core theming model in a way that's worth walking through, because it's a recurring pattern across the project.

### What I'd like to keep

- `onAfterOpen` (renamed — see below).
- `--sheet-width-transition` CSS variable (and ideally a few more — see below).
- The intent of supporting a non-translate animation for chart embedders.

### What needs to change

#### 1. `animationType: 'fly' | 'width'` is a hardcoded preset — drop the prop

This is the central issue. Per [GUIDELINES.md §9](../blob/release/docs/GUIDELINES.md#9-custom-classes-via-classes-prop) and the design philosophy, the library ships unstyled primitives and the consumer owns the design system through CSS variables + the `classes` prop. A string enum (`'fly' | 'width'`) that maps to baked-in keyframes inside the component is the same anti-pattern as `variant="primary"`, `tone="danger"`, `size="lg"` — the component is making a visual decision the consumer should own.

Concretely, the component currently locks in:

- the **set** of allowed animations (only two — a consumer who wants `slide-fade` or `scale` is stuck),
- the **easing** (`ease`, hardcoded),
- the **fill mode** (`forwards`, hardcoded),
- the **from-state** (`width: 0` / `height: 0`, hardcoded),
- the **keyframe name** itself,

and it pays for that with a ~60-line duplicate panel template inside `{#if animationType === 'width'} … {:else} …`.

The customizable alternative is to expose ONE animation hook and let consumers ship their own keyframes via `classes`:

```svelte
<!-- Sheet.svelte — one container, one template -->
<div
  bind:this={sheetPanel}
  class="sheet-panel {side} {classes ?? ''}"
  role="dialog"
  aria-modal="true"
  aria-label={title ?? 'Sheet'}
  tabindex="-1"
  onanimationend={handleAfterOpen}
>
  ...
</div>

<style>
  .sheet-panel {
    animation: var(--sheet-panel-animation, sheet-default-fly 300ms ease both);
  }
  /* ship one sane default keyframe so out-of-the-box behaviour is unchanged */
  @keyframes sheet-default-fly { from { transform: translateX(100%); } to { transform: translateX(0); } }
</style>
```

```css
/* consumer-side — they pick their taxonomy, not us */
.sheet-expand-width {
  --sheet-panel-animation: sheet-expand-width 300ms ease both;
}
@keyframes sheet-expand-width {
  from { width: 0; }
  to { width: var(--sheet-width, 400px); }
}

.sheet-fly-up {
  --sheet-panel-animation: sheet-fly-up 250ms cubic-bezier(0.2, 0, 0, 1) both;
}
@keyframes sheet-fly-up {
  from { transform: translateY(20px); opacity: 0; }
  to { transform: translateY(0); opacity: 1; }
}
```

```svelte
<!-- Usage — chart embedder picks the width animation via classes -->
<Sheet classes="sheet-expand-width" content={…} bind:open />
```

That's the §9 pattern, and it's also what the rest of the library does (Banner, Button, etc. — variants live in consumer CSS, not in props). It also collapses the duplicated template back to one branch and removes the need for `transition:fly` entirely (you'd switch fully to CSS animations and use `onanimationend` for both open and close detection).

#### 2. `width: auto` violates §8

```css
.sheet-animate-width.top,
.sheet-animate-width.bottom {
  width: auto;  /* §8 forbids `auto` for width/height/margin/padding/etc. */
  height: 0;
  ...
}
```

The base `.sheet-panel.top, .sheet-panel.bottom` already declares `left: 0; right: 0;`, so the panel spans horizontally without needing `width: auto`. Just remove that line.

#### 3. Event handler naming: `onAfterOpen` → `onafteropen`

Every event prop in the library is lowercase: `onclose`, `onopen`, `onchange`, `onkeyup` (see `ContextMenu`, `Tabs`, `Pagination`, `Radio`, `Slider`, `SplitInput`, `ThemeSwitcher`, `CommandMenu`). `onAfterOpen` breaks that. Please rename to `onafteropen`, and while you're at it add the symmetric `onafterclose` — the close-side transition has the same chart-reflow concern in reverse, and the current PR only handles half the lifecycle.

#### 4. Unrelated `function` → `const = () => {}` refactor

All six function declarations (`close`, `handleOverlayClick`, `handleKeyDown`, `lockScroll`, `unlockScroll`, `scrollLockAction`) were rewritten as arrow consts. None of it is required by the feature. Please revert — it inflates the diff, hides the actual change, and risks conflicts with other PRs touching Sheet.

### Ask

Could you push a revised version that:

1. Removes the `animationType` prop and `SheetAnimationType` type,
2. Exposes `--sheet-panel-animation` (or a small set of split variables) as the customization hook,
3. Ships one default keyframe so existing consumers see no behavior change,
4. Renames `onAfterOpen` → `onafteropen` and adds `onafterclose`,
5. Removes the `width: auto` declaration,
6. Reverts the `function` → `const` stylistic refactor?

Happy to chat through the CSS variable surface if it's useful — the goal is to keep this Sheet (and the next four "what if we want a different animation" PRs) free of preset enums.

murdore:
@sinha-sahil — thank you for the detailed review. You are right on every point, and I appreciate you spelling out the §9 pattern so concretely.

The `animationType: 'fly' | 'width'` prop is exactly the anti-pattern the guidelines warn against: it bakes two hardcoded presets into the component, locks consumers into my choice of keyframes/easing/fill-mode, and forces a duplicate template branch. Replacing it with `--sheet-panel-animation` and a default keyframe keeps the component unstyled and generic — chart embedders get the hook they need without the library making any visual decision for them.

Here is what I will change in the revised push:

- Drop `animationType` prop and `SheetAnimationType` type entirely
- Expose `--sheet-panel-animation` CSS variable, default to the existing fly keyframe so no consumer sees a behavior change
- Switch the panel to `onanimationend` for both open and close detection, removing the `transition:fly` dependency
- Rename `onAfterOpen` to `onafteropen` and add the symmetric `onafterclose`
- Remove the `width: auto` declaration (the existing `left: 0; right: 0` already covers that)
- Revert the `function` → `const` refactor — that was an oversight on my part and I will undo it cleanly

One thing I want to confirm before I push: should `--sheet-panel-animation` be a single shorthand variable (as you sketched), or would you prefer split variables — `--sheet-panel-animation-name`, `--sheet-panel-animation-duration`, `--sheet-panel-animation-easing` — so consumers can override just duration without rewriting the full shorthand?

sinha-sahil:
Follow-up — escalating the `animationType` part from MAJOR to BLOCK on second read. The `onAfterOpen` half is unaffected by this and can land independently; happy to approve a PR that contains only that change.

**Why `animationType` is a block under our current bar**

Re-reading after the #197 CopyButton discussion clarified the line: a string-enum prop that selects between two baked-in CSS implementations is a hardcoded variant system, which GUIDELINES §9 rules out. The diff has the two tells of that pattern:

1. **Duplicated template under an `{#if animationType === 'width'}` branch** — the entire panel (`role="dialog"`, header, content, footer) is rendered twice, differing only in `transition:fly|global` vs `class="… sheet-animate-width"` + `onanimationend`. That's two implementations, one switch.
2. **Embedded styles for the variant** — `.sheet-animate-width`, `@keyframes sheet-expand-width`, `@keyframes sheet-expand-height` all live inside Sheet's `<style>` block and hardcode the property being animated.

The PR description also frames this as a UX preset (*"ideal for sheets that contain embedded charts"*), which is the rationale §9 specifically asks us to push back on — that opinion belongs in consumer CSS, not in the library.

**What would unblock**

Drop `animationType`, the `{#if}` branch, and the keyframe rules. Pick one:

- **Preferred:** expose `--sheet-panel-transition` / `--sheet-panel-animation` CSS vars on `.sheet-panel` and let the consumer override. Width-expand becomes a ~10-line snippet in the consumer's stylesheet.
- **If a JS hook is genuinely required:** accept a Svelte transition function (`transition?: (node, params) => TransitionConfig`) instead of a string enum. That's a primitive, not a variant.

**`onAfterOpen` — please keep**

This one's a clean primitive. If you split the PR so `onAfterOpen` lands on its own, I'll approve it immediately.

**One additional ask — revert the arrow-const churn.**

The diff converts existing `function close()`, `handleOverlayClick`, `handleKeyDown`, `lockScroll`, `unlockScroll`, `scrollLockAction` into `const name = () => {}` arrow expressions. Project convention is `function fnName() {}` — hoisting, named stack traces, `this` semantics, and consistency with the rest of `src/lib/`. Please revert the conversions; the feature change shouldn't carry stylistic churn.

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — thank you for the quick and thorough turnaround. Two crossed-in-the-air notes first:

- My follow-up comment escalated `animationType` to BLOCK and the review state is now `CHANGES_REQUESTED`. Your plan above already resolves every BLOCK finding, so once you push the revision the block clears.
- I'd still like to **split the PR** so `onafteropen` lands today and the CSS-variable rework lands as a separate, smaller diff. That keeps the lifecycle hook unblocked by the animation conversation, and it gives us a clean revert surface if the CSS hook needs iteration. More on that at the bottom.

### Answer to your variable-granularity question

> *should `--sheet-panel-animation` be a single shorthand variable, or split — `--sheet-panel-animation-name`, `--sheet-panel-animation-duration`, `--sheet-panel-animation-easing` — so consumers can override just duration without rewriting the full shorthand?*

**Split, with a defaulted shorthand on top.** Reasoning, in order of weight:

1. **Principle 7 (highly customizable) wants the finest grain we can reasonably expose.** "Override just the duration" is a real consumer ask (theme-level motion-reduce, density modes, A/B-tested durations). Forcing them to re-declare the entire shorthand to nudge one value is the same tax as forcing them to re-declare a whole `font` shorthand to change weight.
2. **Split vars compose with the cascade.** A consumer can set `--sheet-panel-animation-duration: 0ms` on `.reduce-motion` without knowing or caring what keyframe is in flight. That's not possible with a shorthand.
3. **`@media (prefers-reduced-motion)` integration becomes a one-liner** for consumers — they override `--sheet-panel-animation-duration` to `0ms` and the animation flattens regardless of which keyframe a theme picked.

Concrete shape I'd ship:

```css
.sheet-panel {
  animation:
    var(--sheet-panel-animation-name, sheet-default-fly)
    var(--sheet-panel-animation-duration, 300ms)
    var(--sheet-panel-animation-easing, ease)
    var(--sheet-panel-animation-fill-mode, both);
}
@keyframes sheet-default-fly { from { transform: translateX(100%); } to { transform: translateX(0); } }
```

That's four hooks, each independently overridable, with a sane default on every one — and the chart-embedder use case becomes:

```css
.sheet-expand-width { --sheet-panel-animation-name: sheet-expand-width; }
@keyframes sheet-expand-width { from { width: 0; } to { width: var(--sheet-width, 400px); } }
```

No shorthand needed. Skip the `--sheet-panel-animation` umbrella variable — once the four parts exist independently, the shorthand is redundant surface area and tends to confuse precedence (shorthand + longhand on the same property is a footgun in user CSS).

Same pattern for the close-side: `--sheet-panel-animation-name-close` / `-duration-close` / `-easing-close`, defaulting to the open values where it makes sense. That keeps the symmetry with `onafteropen` / `onafterclose`.

### Split path

Could you:

1. **Open a slim PR with just `onafteropen` + `onafterclose`** (lowercase, both halves of the lifecycle, no animation changes, no function/arrow churn). I'll approve that on sight.
2. **Keep this PR for the CSS-variable rework** — drop `animationType` / `SheetAnimationType` / the `{#if}` branch / the embedded keyframes, switch to the four split vars above, revert the arrow-const conversions, remove `width: auto`. Once that's pushed I'll re-review.

**Ask: please split this PR — land `onafteropen`/`onafterclose` in a slim follow-up, and rework this one as the CSS-variable surface (four split vars, no shorthand, no preset enum).**

murdore:
@sinha-sahil reworked per your spec — dropped `animationType`/`SheetAnimationType`/`onAfterOpen`/the `{#if}` branch/embedded keyframes; replaced with four independent CSS longhand vars (`--sheet-panel-animation-name`, `--sheet-panel-animation-duration`, `--sheet-panel-animation-easing`, `--sheet-panel-animation-fill-mode`) on `.sheet-panel`, each independently overridable, with four named `@keyframes sheet-panel-fly-{right,left,top,bottom}` as library defaults. No shorthand umbrella variable. Arrow-const conversions reverted to `function` declarations. svelte-check/prettier/eslint clean. Re-requesting review.

sinha-sahil:
Thanks for the turnaround — the rework matches yesterday's spec point-for-point. Walked through it on `b9568dc`:

- `animationType` / `SheetAnimationType` / `onAfterOpen` / the `{#if animationType === 'width'}` branch / the `sheet-expand-width|height` keyframes — all gone.
- Four split vars on `.sheet-panel` (`--sheet-panel-animation-name|-duration|-easing|-fill-mode`), each independently overridable, each with a sane default. No shorthand umbrella variable. That's exactly the surface I'd hoped for — reduce-motion / density themes can now override just `--sheet-panel-animation-duration` without touching the keyframe.
- Per-side default keyframes (`sheet-panel-fly-{right,left,top,bottom}`) preserve the original directional fly behavior without needing the `flyParams` `$derived.by` switch. Cleaner than my single `sheet-default-fly` sketch.
- `width: auto` gone.
- Arrow-const conversions reverted to `function` declarations across `close`, `handleOverlayClick`, `handleKeyDown`, `lockScroll`, `unlockScroll`, `scrollLockAction`. Diff is now just the feature change.
- Lifecycle hooks correctly carved out to #228.

### One regression to address before merge

Removing `transition:fly|global={flyParams}` (Sheet.svelte:114 in the old version) also removed the **outro** animation. CSS `animation-*` properties only run on element mount, so today when `open` flips to `false` the panel pops out instantly while the overlay still fades over 200ms. That's a user-visible regression vs `main`, and it breaks the symmetry I asked for in the last comment ("Same pattern for the close-side: `--sheet-panel-animation-name-close` / `-duration-close` / `-easing-close`").

Two ways to fix, your call:

1. **Mirror the four vars for close** — `--sheet-panel-exit-animation-name` / `-duration` / `-easing` / `-fill-mode`, defaulting to reversed-direction keyframes (`sheet-panel-fly-out-{side}`). Driven by a `data-state="open"|"closing"` attribute on `.sheet-panel` so the rule can match `[data-state="closing"]`. Slightly more wiring (need to delay the `{#if open}` teardown until `onanimationend` fires on the closing animation) but it's the consistent answer.
2. **Out transition via `out:` directive with a CSS-driven custom transition function** — accept an optional `exitAnimation?: (node) => AnimationConfig` prop that defaults to a flat keyframe-driven function reading the same CSS vars. Smaller diff but introduces a JS surface I'd rather avoid.

I'd take option 1. Symmetry with intro, no new prop surface, and consumers who want no close animation just set `--sheet-panel-exit-animation-duration: 0ms`.

### Ask

Add the close-side var surface (option 1 above) and I'll approve and merge. Everything else here is clean.

murdore:
@sinha-sahil — addressed your review feedback (latest comment `2026-05-30T14:07Z`):

- **Close-side var surface added (option 1):** four CSS custom properties now mirror the entry vars exactly — `--sheet-panel-exit-animation-name`, `--sheet-panel-exit-animation-duration`, `--sheet-panel-exit-animation-easing`, `--sheet-panel-exit-animation-fill-mode` — each independently overridable on `.sheet-panel[data-state='closing']`. Reduce-motion themes can pin `--sheet-panel-exit-animation-duration: 0s` without touching the keyframe.

- **`data-state` attribute on the panel element:** values are `"open"` | `"closing"`. Per-side closing selectors (`.sheet-panel.right[data-state='closing']` etc.) default to the matching `sheet-panel-fly-out-{side}` keyframe — the exact reverse of the entry keyframe — so the out-of-the-box behaviour is: slide in from the edge on open, slide back out to the edge on close.

- **Delayed teardown via `onanimationend`:** `close()` now sets `panelState = 'closing'` instead of immediately flipping `open = false`. The `handlePanelAnimationEnd` handler on the panel element waits for the outro animation to complete before setting `open = false` + `onclose?.()`, so the panel no longer pops out instantly.

- **`scrollLockAction` resets `panelState = 'open'` on mount** so re-opening after an external `open = false` (not via `close()`) always starts from a clean entry state.

- **All existing entry vars and keyframes unchanged** — `data-state="open"` is the default so the entry animation works exactly as before.

Verification: `pnpm run lint`, `pnpm run check` (0 errors, 0 warnings), and `pnpm run build` all clean. Ready for re-review.

murdore:
@sinha-sahil — three additional fixes after self-review of the prior rework:

**1. `bind:open=false` now triggers the exit animation.** Previously only the internal `close()` set `panelState='closing'`, so consumers using `<Sheet bind:open />` and setting `open=false` directly got an immediate DOM teardown with no fly-out. Added a `trackOpen` Svelte action mounted on `<svelte:window>` (the library lint forbids `$effect`, so action-with-update is the working pattern) that detects external `open` flips and triggers the close animation.

**2. Restored parallel close.** Was sequential: panel fly-out 300ms → overlay fade 200ms = 500ms. The overlay's `transition:fade` only fired on actual DOM destroy, which had to wait for the panel animation. Replaced with a CSS-driven animation on `.sheet-overlay` keyed off the same `data-state="open"|"closing"` attribute as the panel. Both animations now start at the same instant on close; total wall-clock ~300ms matching the original `transition:fly` behaviour.

**3. Documented all 16 animation CSS vars in `docs/Sheet.md`** — the 8 panel ones from the prior rework (`--sheet-panel-animation-*`, `--sheet-panel-exit-animation-*`) plus 8 new overlay ones (`--sheet-overlay-animation-*`, `--sheet-overlay-exit-animation-*`) added in this commit. Consumers can now override entry/exit per element independently.

murdore:

The earlier review feedback was addressed in force-push **554ed8e** on 2026-05-31 — squashed to a single commit per the project's one-commit-per-PR convention:
- `bind:open` exit path supported
- parallel-close behaviour restored
- all 16 animation CSS variables documented

murdore:
📝 **Route + docs integration added** (per review feedback): this PR now includes a demo section in `src/routes/components/sheet/+page.svelte` showcasing the new `animationType` / `onAfterOpen` behaviour.

⚠️ **Merge heads-up vs #228** — this PR and #228 (after-open lifecycle callbacks) both modify `src/lib/Sheet/Sheet.svelte` independently; neither branch is based on the other, and `git merge-tree` reports a **real content conflict in `Sheet.svelte`** (this PR's CSS-keyframe/`data-state` animation system vs #228's Svelte-transition hooks). They cannot both merge cleanly without a manual reconcile of the two animation strategies — whichever lands first, the other needs a rebase. The added demo blocks in `sheet/+page.svelte` also overlap (trivial to resolve).

murdore:
@sinha-sahil — per your review feedback, #206 now demos `animationType` / `onAfterOpen` on the Sheet route. ⚠️ Please also see my note above: #206 and #228 both modify `src/lib/Sheet/Sheet.svelte` with conflicting animation strategies (`git merge-tree` confirms a real conflict) and need a manual reconcile before both can merge.

murdore:
@sinha-sahil — closing this in favor of #228 to resolve the Sheet conflict.

Why: #206 and #228 both modify the open/close path in `src/lib/Sheet/Sheet.svelte` and cannot both merge (`git merge-tree` shows a real content conflict). #228 (lifecycle callbacks `onafteropen` / `onafterclose`) is already **approved**, small (+55/-4), and idiomatic — it keeps the existing Svelte transitions. This PR currently has changes requested, and its `onAfterOpen` was dropped during review, so it now only carries the CSS-variable animation rewrite (+277/-17, which replaces Svelte transitions with a hand-rolled state machine).

The consumer-customizable CSS-variable animation engine here is still valuable — best path is to bring it back as a fresh PR on top of the merged #228 baseline so it layers on cleanly without the conflict. Happy to help rebuild it that way. Thanks for the work here 🙏

PR #207, opened 2026-05-29
feat: add NavigationMenu component

murdore:
## Summary

- New `NavigationMenu` primitive: a vertical sidebar navigation panel built with Svelte 5 runes
- Exports `NavigationMenu` (default) and `NavigationMenuProperties` / `NavigationMenuItem` types from the library root
- Fully backward-compatible addition — no existing exports modified

### Props (`NavigationMenuProperties`)

| `items` | `NavigationMenuItem[]` | — (required) | List of nav items |
| `selectedId` | `string` | — | ID of the currently selected item |
| `testId` | `string` | — | Renders as `data-pw` on the root `<nav>` |
| `classes` | `string` | — | Appended to root element class list |
| `ariaLabel` | `string` | `'Navigation'` | `aria-label` for the `<nav>` element |
| `onselect` | `(id: string) => void` | — | Called when a non-disabled item is clicked |

### `NavigationMenuItem`

| Field | Type | Description |
|-------|------|-------------|
| `id` | `string` | Unique identifier; also used as `data-pw` on the button |
| `label` | `string` | Display text |
| `icon` | `string?` | URL rendered as `` |
| `statusDot` | `boolean?` | Shows a colored dot badge |
| `disabled` | `boolean?` | Disables the button |

### CSS custom properties

| `--navigation-menu-width` | `240px` | Panel width |
| `--navigation-menu-padding` | `8px` | Root padding |
| `--navigation-menu-gap` | `2px` | Gap between items |
| `--navigation-menu-item-padding` | `8px 12px` | Item button padding |
| `--navigation-menu-item-radius` | `8px` | Item border-radius |
| `--navigation-menu-item-color` | `inherit` | Item text color |
| `--navigation-menu-item-hover-background` | `rgba(0,0,0,0.06)` | Hover background |
| `--navigation-menu-item-selected-background` | `rgba(0,0,0,0.1)` | Selected background |
| `--navigation-menu-item-selected-color` | `inherit` | Selected text color |
| `--navigation-menu-dot-color` | `#22c55e` | Status dot color |
| `--navigation-menu-dot-size` | `8px` | Status dot diameter |
| `--navigation-menu-item-icon-size` | `20px` | Icon width/height |

### DOM structure

```
<nav class="navigation-menu" aria-label="…" data-pw="…">
  <ul class="navigation-menu-list" role="list">
    <li class="navigation-menu-list-item">
      <button class="navigation-menu-item [navigation-menu-item-selected]" data-pw="{id}" aria-current="page|null">
        <img class="navigation-menu-item-icon" />   <!-- optional -->
        <span class="navigation-menu-item-label">Label</span>
        <span class="navigation-menu-dot" />         <!-- optional -->
      </button>
    </li>
    …
  </ul>
</nav>
```

## Notes

- `prettier --check` — clean
- `eslint` — clean
- No `font-family`, `font-size`, `font-weight`, or `line-height` hardcoded — all typography deferred to design system / CSS vars
- CSS selectors are kebab-case only (stylelint-compliant)
- `aria-current="page"` set on selected item; `null` otherwise (no `undefined`)
- Item list is `overflow-y: auto` for scrollable panels

### DIFF
--- a/src/lib/NavigationMenu/NavigationMenu.svelte
+++ b/src/lib/NavigationMenu/NavigationMenu.svelte
@@ -1,0 +1,118 @@
+<script lang="ts">
+  import type { NavigationMenuProperties } from './properties';
+
+  let {
+    items,
+    selectedId,
+    testId,
+    classes,
+    ariaLabel = 'Navigation',
+    onselect
+  }: NavigationMenuProperties = $props();
+</script>
+
+<nav class="navigation-menu {classes ?? ''}" aria-label={ariaLabel} data-pw={testId}>
+  <ul class="navigation-menu-list" role="list">
+    {#each items as item (item.id)}
+      <li class="navigation-menu-list-item">
+        <button
+          class="navigation-menu-item"
+          class:navigation-menu-item-selected={item.id === selectedId}
+          data-pw={item.id}
+          disabled={item.disabled}
+          aria-current={item.id === selectedId ? 'page' : null}
+          onclick={() => onselect?.(item.id)}
+        >
+          {#if item.icon}
+            <img class="navigation-menu-item-icon" src={item.icon} alt="" aria-hidden="true" />
+          {/if}
+          <span class="navigation-menu-item-label">{item.label}</span>
+          {#if item.statusDot}
+            <span class="navigation-menu-dot" aria-hidden="true"></span>
+          {/if}
+        </button>
+      </li>
+    {/each}
+  </ul>
+</nav>
+
+<style>
+  .navigation-menu {
+    width: var(--navigation-menu-width, 240px);
+    padding: var(--navigation-menu-padding, 8px);
+    box-sizing: border-box;
+    display: flex;
+    flex-direction: column;
+  }
+
+  .navigation-menu-list {
+    list-style: none;
+    margin: 0;
+    padding: 0;
+    display: flex;
+    flex-direction: column;
+    gap: var(--navigation-menu-gap, 2px);
+    overflow-y: auto;
+  }
+
+  .navigation-menu-list-item {
+    display: contents;
+  }
+
+  .navigation-menu-item {
+    width: 100%;
+    display: flex;
+    align-items: center;
+    gap: 8px;
+    padding: var(--navigation-menu-item-padding, 8px 12px);
+    border-radius: var(--navigation-menu-item-radius, 8px);
+    border: none;
+    background: transparent;
+    color: var(--navigation-menu-item-color, inherit);
+    cursor: pointer;
+    text-align: left;
+    box-sizing: border-box;
+  }
+
+  .navigation-menu-item:hover:not(:disabled) {
+    background-color: var(--navigation-menu-item-hover-background, rgba(0, 0, 0, 0.06));
+  }
+
+  .navigation-menu-item:focus-visible {
+    outline: 2px solid var(--navigation-menu-item-selected-background, currentColor);
+    outline-offset: 2px;
+  }
+
+  .navigation-menu-item-selected {
+    background-color: var(--navigation-menu-item-selected-background, rgba(0, 0, 0, 0.1));
+    color: var(--navigation-menu-item-selected-color, inherit);
+  }
+
+  .navigation-menu-item:disabled {
+    opacity: 0.4;
+    cursor: not-allowed;
+  }
+
+  .navigation-menu-item-icon {
+    width: var(--navigation-menu-item-icon-size, 20px);
+    height: var(--navigation-menu-item-icon-size, 20px);
+    flex-shrink: 0;
+    object-fit: contain;
+  }
+
+  .navigation-menu-item-label {
+    flex: 1;
+    min-width: 0;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
+
+  .navigation-menu-dot {
+    flex-shrink: 0;
+    width: var(--navigation-menu-dot-size, 8px);
+    height: var(--navigation-menu-dot-size, 8px);
+    border-radius: 50%;
+    background-color: var(--navigation-menu-dot-color, #22c55e);
+  }
+</style>
--- a/src/lib/NavigationMenu/properties.ts
+++ b/src/lib/NavigationMenu/properties.ts
@@ -1,0 +1,16 @@
+export type NavigationMenuItem = {
+  id: string;
+  label: string;
+  icon?: string;
+  statusDot?: boolean;
+  disabled?: boolean;
+};
+
+export type NavigationMenuProperties = {
+  items: NavigationMenuItem[];
+  selectedId?: string;
+  testId?: string;
+  classes?: string;
+  ariaLabel?: string;
+  onselect?: (id: string) => void;
+};
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -56,6 +56,7 @@ export { default as EmptyState } from './EmptyState/EmptyState.svelte';
 export { default as Combobox } from './Combobox/Combobox.svelte';
 export { default as ColorPicker } from './ColorPicker/ColorPicker.svelte';
 export { default as SplitInput } from './SplitInput/SplitInput.svelte';
+export { default as NavigationMenu } from './NavigationMenu/NavigationMenu.svelte';
 
 export type * from './Button/properties';
 export type * from './Modal/properties';
@@ -109,5 +110,6 @@ export type * from './EmptyState/properties';
 export type * from './Combobox/properties';
 export type * from './ColorPicker/properties';
 export type * from './SplitInput/properties';
+export type * from './NavigationMenu/properties';
 
 export { validateInput } from './utils';


sinha-sahil:
Thanks for the PR @murdore — I can see you're trying to give consumers a turnkey sidebar nav so they don't have to wire up the markup, ARIA, and selection styling themselves. Before this can land, though, I need us to step back and talk about whether `NavigationMenu` belongs in the library at all in its current shape, because it conflicts with several of the load-bearing principles for this library.

## The core problem: this is a composite wrapper, not a primitive

The library philosophy (and the reason the `classes` + CSS variable pattern exists at all) is that **we ship primitives, consumers compose them**. A primitive earns its place by doing something non-trivial: managing focus, owning state, providing ARIA composite semantics, handling keyboard input — things a consumer cannot easily get right by hand.

What this component currently does:

```svelte
<nav>
  <ul>
    {#each items as item}
      <li><button onclick={() => onselect?.(item.id)}>{icon}{label}{dot}</button></li>
    {/each}
  </ul>
</nav>
```

There is no focus management, no keyboard navigation (no ArrowUp/ArrowDown/Home/End, no roving tabindex), no `aria-orientation`, no controlled selection state (`selectedId` is read-only — the parent has to wire up `onselect` and feed `selectedId` back in), no grouping/section semantics. It is literally a `{#each}` over an array. That is a snippet a consumer can — and per our philosophy, should — write in their own app code:

```svelte
<!-- consumer-land, app.svelte -->
<nav class="app-sidebar" aria-label="Main navigation">
  {#each items as item}
    <Button
      classes="nav-item {item.id === selected ? 'nav-item-selected' : ''}"
      text={item.label}
      icon={item.icon}
      disabled={item.disabled}
      testId={item.id}
      onclick={() => (selected = item.id)}
    />
  {/each}
</nav>
```

```css
/* consumer-land theme */
.app-sidebar { display: flex; flex-direction: column; gap: 2px; padding: 8px; width: 240px; }
.nav-item {
  --button-background: transparent;
  --button-hover-background: rgba(0,0,0,0.06);
  --button-border: none;
  --button-border-radius: 8px;
  --button-padding: 8px 12px;
  --button-justify: flex-start;
}
.nav-item-selected { --button-background: rgba(0,0,0,0.1); }
```

That's the entire `NavigationMenu` — composed from primitives the library already ships, themed by consumer CSS, fully customisable per-item.

## The status dot is exactly the "dummy component" pattern we forbid

```svelte
{#if item.statusDot}
  <span class="navigation-menu-dot" aria-hidden="true"></span>
{/if}
```

```css
.navigation-menu-dot { background-color: var(--navigation-menu-dot-color, #22c55e); }
```

This is a pure-decoration `<span>` with no behaviour, gated by a boolean, with a **hardcoded semantic-green default**. That is a hardcoded design preset baked into the library — exactly what the philosophy rejects. Consumers can't render a count badge, a "NEW" pill, an icon, or a different colour without forking. The closed `NavigationMenuItem` shape (`{ id, label, icon?, statusDot?, disabled? }`) compounds this — there's no escape hatch.

If we keep the component at all, items need to take a `trailing` snippet (or `children`) so the consumer renders whatever they want, the library renders nothing by default.

## Specific GUIDELINES.md hits

- **§4** — raw `` instead of ``. The library already ships `Img` with error fallback handling; the guidelines call this exact case out by name.
- **§6** — raw `<button class="navigation-menu-item">` instead of `<Button classes="navigation-menu-item">` with the `--button-*` bridging pattern. Sidebar nav items are generic action buttons, not specialized composite-widget cells like tabs.
- **§3** — `{#if item.icon}` should be `{#if typeof item.icon === 'string' && item.icon.length > 0}`. `{#if item.statusDot}` should be `{#if item.statusDot === true}` (or removed).
- **Customisability gaps**: `gap: 8px`, `opacity: 0.4` (disabled), `outline: 2px solid …`, `outline-offset: 2px` are all hardcoded — not behind CSS variables. The focus outline also defaults to the *selected* background colour, conflating two visual states. The item shape is also closed (`{ id, label, icon?, statusDot?, disabled? }`) — there's no way to render a count badge, two-line label, secondary text, or keyboard shortcut hint without forking. Consider a `trailing` snippet per item (or opening items to a `children` snippet) plus a per-item `classes?: string` so consumers can theme individual rows.

## Next step

I'd like to **block this PR** as-is. Two paths forward, your call:

1. **Close the PR and move this into the consumer app.** Honestly, given the example above, I think this is the right call — there is nothing here that needs to be in the library.
2. **Rework into a real primitive that earns its place.** That would mean: keyboard navigation (Arrow keys move focus, Home/End jump, Enter/Space activate), roving tabindex, focus-visible management decoupled from selected styling, `selectedId` as `$bindable` so the component manages its own selection state, `ariaOrientation` prop, `trailing` snippet per item, per-item `classes`, drop `statusDot` entirely, use `<Button>` + `` internally with the `--button-*` / `--img-*` bridging pattern, and expose every visual decision (gap, disabled opacity, focus outline width/colour/offset) as a CSS variable.

Happy to discuss which direction makes sense, but the current shape can't land.

murdore:
@sinha-sahil — thank you for the thorough review. I re-read your comment carefully and you are right on every count.

The component as written is genuinely just a `{#each}` loop: no keyboard navigation, no roving tabindex, no `aria-orientation`, no self-managed selection state. That does not clear the "non-trivial behaviour" bar that earns a place in the library. Your consumer-land snippet — `<Button>` items themed with `--button-*` CSS variables — covers the exact same visual output with zero library code, and that is the correct level of abstraction for something this thin.

The `statusDot` issue you flagged is also fair: a hardcoded semantic-green `<span>` with no escape hatch is precisely the kind of baked-in design preset the library avoids per §9.

Given that your consumer-land example already solves the problem cleanly, I think **option 1 (close and keep this in the consumer app)** is the right call rather than investing in a full keyboard-managed primitive that may not justify the complexity.

Before I close: if I were to rework this into a proper primitive — roving tabindex, Arrow-key focus, `selectedId` as `$bindable`, `trailing` snippet, per-item `classes`, `<Button>` + `` internally — would that be a shape you would review, or do you feel a sidebar nav list simply does not belong in this library regardless of how complete the implementation is?

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — appreciate you re-reading the review carefully and agreeing on the substance. You correctly read where the line is: the issue isn't that sidebar nav is unimportant, it's that *this* implementation doesn't do anything a consumer can't write in 15 lines of their own Svelte + CSS. Let's close this PR and answer your forward-looking question directly, because I think it's the most useful part of this thread.

## Would a proper keyboard-managed primitive be reviewable?

**Yes — but as a new PR, with a different scope and probably a different name.** Here's the honest breakdown:

A sidebar nav list as a "render an array of items" thing does not belong in the library — your consumer-land snippet proves that. But the things you listed in your last paragraph — roving tabindex, ArrowUp/Down/Home/End focus management, focus-visible decoupled from selection, `selectedId` as `$bindable` so the component owns the selection state machine, `ariaOrientation` for horizontal vs vertical — those are exactly the kind of non-trivial composite-widget behaviors that earn a primitive its place. Per the WAI-ARIA Authoring Practices, a `role="menu"` / `role="menubar"` / `role="tablist"`-style widget has a precise keyboard contract that's genuinely hard to get right in user-land. That's the bar.

So the shape that would clear review looks more like a `Listbox` / `Menu` / `RovingTabList` than a `NavigationMenu`:

- **Name it for the behavior, not the use-case.** "NavigationMenu" tells me where you'd use it; "Listbox" tells me what it *does* (roving focus + single-select state machine). Use-case naming invites preset creep (NavigationMenu → BreadcrumbMenu → TabMenu → ContextMenu, each with baked layouts).
- **Items via `children` snippet + a context API**, not a closed `{ id, label, icon?, statusDot?, disabled? }` shape. The parent renders `<Button>` / `` / whatever per row; the primitive just provides selection + focus context. Consumers compose; we orchestrate.
- **Zero opinion on layout, spacing, or decoration.** No `gap`, no padding, no built-in dot, no built-in icon slot. Pure behavior.
- **Every visual default `inherit` / `transparent` / `currentColor`.** If `--listbox-gap` defaults to `0`, the consumer sets their gap. If you find yourself wanting `8px` as a default, that's the signal it's a use-case, not a primitive.

If you want to scope a follow-up PR around `<Listbox>` (or whatever name fits — `RovingFocusGroup`, `Menu`, `Combobox` depending on the exact contract you want), I'd happily review that on its merits. It's a genuine gap — there's no roving-focus primitive in the library today, and several existing components (Tabs, Dropdown, Stepper) would arguably benefit from one. But please open it as a fresh PR with a fresh design discussion, not a rebrand of this one — the behavioral contract is the whole point and deserves its own review thread.

For everything *this* PR was trying to solve, your consumer-land snippet is the recipe. Worth adding to a recipes/docs page eventually so the next person reaches for composition before reaching for a component.

**Ask: please close this PR.** If the roving-focus primitive idea is interesting to you, open a fresh PR for `<Listbox>` (or equivalent) with the keyboard contract as the design centerpiece, and we'll take it from there.

murdore:
Closing — NavigationMenu is a thin {#each} with no keyboard/state ownership. If the roving-focus primitive is worth it I'll open a fresh Listbox PR with the keyboard contract as the centerpiece. Thanks @sinha-sahil.

PR #208, opened 2026-05-29
feat: add RadioGroup component

murdore:
## Summary

- Adds a new `RadioGroup` primitive that wraps a list of options into a grouped radio control with two variants: `default` (native radio inputs with custom indicator visuals) and `segmented` (pill-style button group)
- Implements roving tabindex and arrow-key navigation (ArrowRight/Down → next, ArrowLeft/Up → prev) for the segmented variant; native radio keyboard behaviour handles the default variant
- Exports `RadioGroup` component and `RadioGroupProperties`, `RadioGroupOption`, `MandatoryRadioGroupProperties`, `OptionalRadioGroupProperties`, `RadioGroupEventProperties` types from the package root

### Props

| `options` | `RadioGroupOption[]` | — (required) | Array of options with `label`, `value`, optional `subtitle`, `disabled`, `testId` |
| `value` | `string \| null` (bindable) | `null` | Currently selected value |
| `name` | `string` | — | HTML `name` attribute for radio inputs (default variant) |
| `direction` | `'row' \| 'column'` | `'column'` | Flex direction of the option list |
| `variant` | `'default' \| 'segmented'` | `'default'` | Rendering style |
| `testId` | `string` | — | `data-pw` attribute on the root element |
| `classes` | `string` | — | Extra CSS classes appended to root |
| `onchange` | `(value: string) => void` | — | Callback on selection change |

### CSS Custom Properties

| Property | Default | Description |
| `--radio-group-gap` | `8px` | Gap between options |
| `--radio-group-segment-padding` | `6px 12px` | Padding inside each segment button |
| `--radio-group-segment-radius` | `6px` | Border radius of segment buttons |
| `--radio-group-segment-background` | `#f5f5f5` | Background of the segmented container |
| `--radio-group-segment-selected-background` | `#ffffff` | Background of the selected segment |
| `--radio-group-segment-selected-color` | `#333333` | Text color of the selected segment |
| `--radio-group-segment-selected-font-weight` | `500` | Font weight of selected segment (CSS var, not hardcoded) |

Additional theming vars: `--radio-group-indicator-*`, `--radio-group-dot-*`, `--radio-group-label-color`, `--radio-group-subtitle-color`, `--radio-group-disabled-opacity`.

### DOM

Root: `<div class="radio-group radio-group-{variant} radio-group-{direction}" role="radiogroup" data-pw={testId}>`

Default variant: `<label class="radio-group-option" data-pw={option.testId}>` containing a visually-hidden `<input type="radio">` + custom indicator spans + label/subtitle spans.

Segmented variant: `<button role="radio" aria-checked aria-disabled data-pw={option.testId}>` pill buttons with roving tabindex.

## Notes

- `prettier --check` exits **0**
- `eslint` exits **0**
- Backward-compatible: no existing exports changed
- No `any`, no type assertions, no `undefined` keyword — fully compliant with repo ESLint rules

## Summary by CodeRabbit

* **New Features**
  * Added RadioGroup component for managing grouped radio options with keyboard navigation support.
  * Added subtitle support to radio options for additional context.
  * Added segmented control styling variant for RadioGroup.

* **Documentation**
  * Added RadioGroup component documentation and usage examples.

[Review Change Stack](https://app.coderabbit.ai/change-stack/juspay/svelte-ui-components/pull/208?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/Radio.md
+++ b/docs/Radio.md
@@ -18,15 +18,18 @@ A single radio button within a group. Multiple Radio components sharing the same
 
 ## Props
 
-| Prop          | Type      | Required | Default     | Description                                                                                                                                                            |
-| ------------- | --------- | -------- | ----------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| name          | `string`  | Yes      | —           | The group name shared by all radio buttons in the same group. Maps to the native input's `name` attribute.                                                             |
-| value         | `string`  | Yes      | —           | The value this radio button represents. When selected, `selectedValue` becomes this value.                                                                             |
-| selectedValue | `string`  | No       | `''`        | The currently selected value in the group. Bindable. When this matches `value`, the radio appears selected.                                                            |
-| text          | `string`  | No       | `''`        | Label text displayed next to the radio indicator. Hidden when empty.                                                                                                   |
-| disabled      | `boolean` | No       | `false`     | When true, the radio button cannot be interacted with and appears in a disabled visual state.                                                                          |
-| testId        | `string`  | No       | `undefined` | Test identifier applied as `data-pw` attribute on the container for Playwright test selectors.                                                                         |
-| classes       | `string`  | No       | `-`         | CSS class string applied to the component's top-level element. Useful for theming — define classes with CSS variable overrides and pass them to create variant styles. |
+| Prop          | Type                          | Required | Default     | Description                                                                                                                                                            |
+| ------------- | ----------------------------- | -------- | ----------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| name          | `string`                      | Yes      | —           | The group name shared by all radio buttons in the same group. Maps to the native input's `name` attribute.                                                             |
+| value         | `string`                      | Yes      | —           | The value this radio button represents. When selected, `selectedValue` becomes this value.                                                                             |
+| selectedValue | `string`                      | No       | `''`        | The currently selected value in the group. Bindable. When this matches `value`, the radio appears selected.                                                            |
+| text          | `string`                      | No       | `''`        | Label text displayed next to the radio indicator. Hidden when empty.                                                                                                   |
+| subtitle      | `string`                      | No       | `undefined` | Optional secondary text rendered below the label. Only shown when non-empty.                                                                                           |
+| disabled      | `boolean`                     | No       | `false`     | When true, the radio button cannot be interacted with and appears in a disabled visual state.                                                                          |
+| tabIndex      | `number`                      | No       | `undefined` | Tab index applied to the native input. Use to manage focus order — RadioGroup sets this to implement roving tabindex.                                                  |
+| testId        | `string`                      | No       | `undefined` | Test identifier applied as `data-pw` attribute on the container for Playwright test selectors.                                                                         |
+| classes       | `string`                      | No       | `-`         | CSS class string applied to the component's top-level element. Useful for theming — define classes with CSS variable overrides and pass them to create variant styles. |
+| inputRef      | `HTMLInputElement \| null`    | No       | `null`      | Bindable reference to the underlying `<input>` element. RadioGroup binds this to programmatically focus the input during arrow-key navigation.                         |
 
 ## Events
 
@@ -60,10 +63,13 @@ Override these custom properties to theme the component.
 | `--radio-hover-border`          | `2px solid #2196f3`                 | border           | Border of the radio circle on hover (when not disabled).         |
 | `--radio-focus-shadow`          | `0 0 0 3px rgba(33, 150, 243, 0.3)` | box-shadow       | Focus ring shadow shown when the radio receives keyboard focus.  |
 | `--radio-transition`            | `0.2s`                              | transition       | Transition duration for state changes (border, background, dot). |
+| `--radio-label-gap`             | `2px`                               | gap              | Space between the label text and the subtitle.                   |
 | `--radio-text-font-size`        | `14px`                              | font-size        | Font size of the label text.                                     |
 | `--radio-text-font-weight`      | `400`                               | font-weight      | Font weight of the label text.                                   |
 | `--radio-text-color`            | `#333333`                           | color            | Color of the label text.                                         |
 | `--radio-disabled-text-color`   | `#999999`                           | color            | Color of the label text when disabled.                           |
+| `--radio-subtitle-color`        | `#757575`                           | color            | Color of the subtitle text.                                      |
+| `--radio-disabled-subtitle-color` | `#bdbdbd`                         | color            | Color of the subtitle text when disabled.                        |
 
 ## Web Component
 
--- a/docs/RadioGroup.md
+++ b/docs/RadioGroup.md
@@ -1,0 +1,90 @@
+# RadioGroup
+
+A managed group of radio buttons that handles keyboard navigation, mutual exclusion, and WAI-ARIA `radiogroup` semantics. Pass an `options` array describing each radio choice; the `value` prop is bindable and stays in sync when the user selects a different item. Arrow keys cycle through enabled options; the `disabled` prop disables the whole group at once.
+
+## Usage
+
+```svelte
+<script>
+  import { RadioGroup } from '@juspay/svelte-ui-components';
+
+  let selectedPayment = $state('upi');
+</script>
+
+<RadioGroup
+  name="payment-method"
+  bind:value={selectedPayment}
+  ariaLabel="Payment method"
+  options={[
+    { value: 'upi', label: 'UPI' },
+    { value: 'card', label: 'Card' },
+    { value: 'netbanking', label: 'Net Banking', subtitle: 'All major banks supported' },
+    { value: 'cod', label: 'Cash on Delivery', disabled: true }
+  ]}
+/>
+```
+
+## Props
+
+| Prop           | Type                  | Required | Default     | Description                                                                                                                                                            |
+| -------------- | --------------------- | -------- | ----------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| options        | `RadioGroupOption[]`  | Yes      | —           | Array of option objects, each with a `value`, `label`, and optional `subtitle`, `disabled`, and `testId` fields.                                                       |
+| value          | `string`              | Yes      | —           | The currently selected option value. Bindable — updates when the user selects a different radio.                                                                       |
+| name           | `string`              | Yes      | —           | The `name` attribute shared across all rendered radio inputs. Required for native radio-group behaviour in forms.                                                       |
+| ariaLabel      | `string`              | No       | `undefined` | Accessible label for the `radiogroup` role. Use when there is no visible heading labelling the group.                                                                   |
+| ariaLabelledBy | `string`              | No       | `undefined` | ID of an existing element that labels the group. Use instead of `ariaLabel` when a visible heading is already present.                                                  |
+| disabled       | `boolean`             | No       | `false`     | When true, all options in the group are disabled regardless of their individual `disabled` flag.                                                                        |
+| testId         | `string`              | No       | `undefined` | Test identifier applied as `data-pw` attribute on the group container for Playwright test selectors.                                                                    |
+| classes        | `string`              | No       | `undefined` | CSS class string applied to the component's top-level element. Useful for theming — define classes with CSS variable overrides and pass them to create variant styles.  |
+
+### RadioGroupOption
+
+| Field    | Type      | Required | Description                                                                                |
+| -------- | --------- | -------- | ------------------------------------------------------------------------------------------ |
+| value    | `string`  | Yes      | Unique value for this option.                                                              |
+| label    | `string`  | Yes      | Display label rendered next to the radio indicator.                                        |
+| subtitle | `string`  | No       | Optional secondary text rendered below the label.                                          |
+| disabled | `boolean` | No       | When true, this individual option is disabled (pointer events blocked, visually dimmed).   |
+| testId   | `string`  | No       | Test identifier applied as `data-pw` on the option wrapper element.                        |
+
+## Events
+
+| Event    | Type                      | Description                                                                             |
+| -------- | ------------------------- | --------------------------------------------------------------------------------------- |
+| onchange | `(value: string) => void` | Fires when the selected option changes. Receives the `value` of the newly selected item. |
+
+## CSS Variables
+
+Override these custom properties to theme the component.
+
+| Variable                   | Default       | CSS Property     | Description                                              |
+| -------------------------- | ------------- | ---------------- | -------------------------------------------------------- |
+| `--radio-group-direction`  | `column`      | flex-direction   | Layout direction of the radio items (`column` or `row`). |
+| `--radio-group-gap`        | `0`           | gap              | Space between individual radio items.                    |
+| `--radio-group-padding`    | `0`           | padding          | Padding inside the group container.                      |
+| `--radio-group-background` | `transparent` | background       | Background color of the group container.                 |
+| `--radio-group-radius`     | `0`           | border-radius    | Border radius of the group container.                    |
+
+Individual radio items inside the group also respond to all `--radio-*` CSS variables defined on the [Radio](/components/radio) component.
+
+## Web Component
+
+Tag: `<sui-radio-group>`
+
+```html
+<sui-radio-group
+  name="payment-method"
+  value="upi"
+  aria-label="Payment method"
+></sui-radio-group>
+
+<script>
+  const group = document.querySelector('sui-radio-group');
+  group.options = [
+    { value: 'upi', label: 'UPI' },
+    { value: 'card', label: 'Card' },
+    { value: 'netbanking', label: 'Net Banking' }
+  ];
+  group.onchange = (value) => console.log('Selected:', value);
+</script>
+```
--- a/docs/_index.json
+++ b/docs/_index.json
@@ -156,6 +156,10 @@
     "description": "A styled radio button input with a custom circular indicator. Supports checked state binding, disabled state, and configurable size and colors via CSS variables."
   },
   {
+    "name": "RadioGroup",
+    "description": "A managed group of radio buttons with WAI-ARIA radiogroup semantics, arrow-key navigation, mutual exclusion, group-level disable, and bindable value synchronisation."
+  },
+  {
     "name": "RelativeTime",
     "description": "Displays a human-readable relative time string (e.g., '5 minutes ago') that auto-updates at a configurable interval. Uses Intl.RelativeTimeFormat for locale-aware formatting. Renders a semantic <time> element. Optionally wraps in a Tooltip showing the full date."
   },
--- a/src/lib/Radio/Radio.svelte
+++ b/src/lib/Radio/Radio.svelte
@@ -6,10 +6,13 @@
     value,
     selectedValue = $bindable(''),
     text = '',
+    subtitle,
     disabled = false,
+    tabIndex,
     testId,
     onchange,
-    classes
+    classes,
+    inputRef = $bindable(null)
   }: RadioProperties = $props();
 
   let checked = $derived(selectedValue === value);
@@ -35,13 +38,20 @@
     {value}
     {checked}
     {disabled}
+    tabindex={typeof tabIndex === 'number' ? tabIndex : null}
+    bind:this={inputRef}
     onchange={handleChange}
   />
   <span class="radio-indicator" class:checked class:disabled>
     <span class="radio-dot" class:checked class:disabled></span>
   </span>
   {#if text.length > 0}
-    <span class="radio-text" class:disabled>{text}</span>
+    <div class="radio-label-group">
+      <span class="radio-text" class:disabled>{text}</span>
+      {#if typeof subtitle === 'string' && subtitle.length > 0}
+        <span class="radio-subtitle" class:disabled>{subtitle}</span>
+      {/if}
+    </div>
   {/if}
 </label>
 
@@ -116,6 +126,12 @@
     background-color: var(--radio-disabled-dot-color, #cccccc);
   }
 
+  .radio-label-group {
+    display: flex;
+    flex-direction: column;
+    gap: var(--radio-label-gap, 2px);
+  }
+
   .radio-text {
     font-size: var(--radio-text-font-size, 14px);
     font-weight: var(--radio-text-font-weight, 400);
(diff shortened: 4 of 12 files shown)


sinha-sahil:
## RadioGroup: please rework before merge — the `variant` prop is the exact pattern we don't ship

Thanks for picking this up @murdore — a proper grouped radio primitive with roving tabindex is a genuine gap, and the keyboard logic + ARIA scaffolding here is on the right track. But the shape of the API needs a rework before this can land, because it directly conflicts with the library's core philosophy.

### Why this is a blocker

The `variant: 'default' | 'segmented'` prop (`properties.ts:23`) is a **hardcoded design preset baked into the component** — exactly the pattern called out in `feedback_no_hardcoded_variants.md` and forbidden by **GUIDELINES.md §9**. The template branches into two completely different DOM trees:

```svelte
{#if variant === 'default'}
  <label><input type="radio">...<span class="radio-group-option-indicator">...</label>
{:else}
  <button role="radio">...<span class="radio-group-segment-label">...</button>
{/if}
```

backed by two disjoint CSS rule sets (`.radio-group-option*` vs `.radio-group-segment*`). That isn't a primitive with a knob — that's two components glued together by an enum, with the library shipping the visual design system for the "segmented" look as a baked-in preset.

`direction: 'row' | 'column'` is the same anti-pattern in miniature — a hardcoded preset for a single CSS axis.

### What the library wants instead

A single unstyled primitive. The segmented look is a **consumer CSS class** that sets the same CSS variables differently:

```svelte
<!-- component template (sketch) -->
<div
  class="radio-group {classes ?? ''}"
  role="radiogroup"
  aria-label={ariaLabel}
  data-pw={testId}
>
  {#each options as option (option.value)}
    <Radio
      {name}
      value={option.value}
      text={option.label}
      disabled={option.disabled}
      bind:selectedValue={value}
      onchange={(v) => onchange?.(v)}
      classes="radio-group-item"
    />
  {/each}
</div>
```

```css
/* user-land: default look — nothing extra needed */

/* user-land: segmented pill look */
.radio-group-pill {
  --radio-group-direction: row;
  --radio-group-gap: 0;
  --radio-group-padding: 4px;
  --radio-group-background: #f5f5f5;
  --radio-group-radius: 8px;
}
.radio-group-pill .radio-group-item {
  --radio-indicator-display: none;          /* hide the dot */
  --radio-container-padding: 6px 12px;
  --radio-container-radius: 6px;
  --radio-container-background-checked: #fff;
}
```

```svelte
<!-- consumption -->
<RadioGroup {options} bind:value />                       <!-- default -->
<RadioGroup classes="radio-group-pill" {options} bind:value /> <!-- segmented -->
```

Multiple variant systems compose this way (per §9's closing paragraph) and you never bake a brand decision into the library.

### Also: please reuse the existing `Radio` primitive (§4)

The library already ships `src/lib/Radio/Radio.svelte` with the indicator + dot + label + disabled + `selectedValue = $bindable()` API you've inlined here. The "default variant" branch of RadioGroup is a re-implementation of that component. RadioGroup should compose `Radio`, not redraw it. The group's job is just (a) `role="radiogroup"` container, (b) threading `name`/`bind:value`, and (c) the arrow-key roving tabindex (which `Radio` doesn't do — that's the genuine new behavior worth shipping).

### Smaller GUIDELINES fixes to fold in during rework

- **§2 (safe index access):** `options[nextIndex]`, `enabledIndices[nextPosition]`, and the NodeList read `buttons?.[nextIndex]` (RadioGroup.svelte:31-34, 41-44) need to be `.at()` / `.item()`.
- **§3 (no falsy checks):** `tabindex={value === option.value || (!value && index === 0) ? 0 : -1}` — use `value === null`. `!option.disabled && selectOption(...)` — `disabled` is `boolean | undefined`, compare explicitly. `{#if option.subtitle}` — `typeof option.subtitle === 'string' && option.subtitle.length > 0`.
- **A11y:** `role="radiogroup"` needs `aria-label` / `aria-labelledby` — please add an `ariaLabel` prop.
- **A11y bug:** in the "default" variant, the native `<input type="radio">` is set to `position: absolute; opacity: 0; pointer-events: none; width: 0; height: 0` — that kills keyboard focus on the underlying input, so the `:focus-visible + .indicator` selector never fires through real keyboard navigation. Using the existing `Radio` component fixes this for free.

### Suggested next step

I'd like to **block this PR** and ask for a rework along these lines:

1. Drop `variant` and `direction` props.
2. Rebuild RadioGroup as a thin wrapper that composes `Radio`, owns `role="radiogroup"` + `aria-label`, and adds roving tabindex + arrow-key navigation.
3. Expose the segmented and horizontal looks as **example consumer CSS classes in the docs / playground route**, not as props.
4. Fold the §2 / §3 / a11y fixes in.

Happy to pair on the refactor if it helps — the keyboard logic you've written is reusable as-is, it's mostly the API surface that needs to change. Once the variant/direction props are gone and `Radio` is composed, this becomes a much smaller, much more aligned PR.

murdore:
@sinha-sahil — thank you for the thorough review. You are right on every count, and I appreciate you spelling it out so concretely.

The `variant: 'default' | 'segmented'` and `direction: 'row' | 'column'` props are exactly the baked-in design presets that §9 of GUIDELINES.md says must not live in the component. I also missed that `Radio` already handles the indicator/dot/label/disabled logic I re-implemented in the default-variant branch — §4 is clear that RadioGroup should compose it rather than redraw it.

My plan for the rework:

- Drop `variant` and `direction` props entirely.
- Rebuild RadioGroup as a thin wrapper around the existing `Radio` primitive, owning only `role="radiogroup"`, `aria-label`, and the roving-tabindex arrow-key logic.
- Move the segmented/horizontal looks to documented consumer CSS classes (via `classes` + exposed CSS vars), following the §9 examples.
- Fix the §2 `.at()` / `.item()` index accesses, the §3 falsy-check violations, and add the `ariaLabel` prop for a11y.

Would a RadioGroup shaped this way — no `variant`/`direction` props, composes `Radio`, segmented look demonstrated as a consumer class in the docs/playground — be the shape you'd accept for merge, or is there anything else about the API surface you'd want changed before I start the rework?

sinha-sahil:

New arrow-const handlers in the diff: `selectOption`, `handleSegmentedKeydown`. Please write them as `function fnName() {}` instead.

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — thanks for engaging so directly with the principles. "Would a RadioGroup shaped this way be the shape you'd accept for merge" — yes, the shape you described is exactly what the library wants. Green light to start the rework. Below are the few API-surface details worth pinning down now so we don't churn on a second round.

### Confirmed API shape

A thin grouping primitive that composes `Radio`. Props, in order of priority:

- `options: Array<{ value: string; label: string; subtitle?: string; disabled?: boolean }>` — keep your existing `RadioOption` type, but tighten `subtitle` to optional `string` only and gate it with `typeof option.subtitle === 'string' && option.subtitle.length > 0` per §3.
- `value: string` with `$bindable()` — required, per §7. Don't default to `''`; an explicit required prop is clearer than an empty-string sentinel and removes the `!value` falsy check entirely.
- `name: string` — required, threaded into each `Radio`.
- `ariaLabel?: string` and `ariaLabelledBy?: string` — at least one should be provided; document that in the prop description (no runtime warn — just docs).
- `disabled?: boolean` — group-level disable, ORed with `option.disabled` when threading down.
- `classes?: string` and `testId?: string` — per §9 and the data-pw convention.
- `onchange?: (value: string) => void` — fired after `value` mutates.

Drop entirely: `variant`, `direction`, any `size`/`tone`/`flush` knob, any built-in label like "Select option". The segmented and horizontal looks become consumer CSS classes that set CSS variables on `.radio-group` and `.radio-group-item` — exactly the recipe in my earlier comment. Please add both as **playground examples** in `src/routes/components/radio-group/+page.svelte` so the docs carry the variant story, not the component.

### Roving tabindex — keep the logic, port it onto `Radio`

This is the genuinely new behavior worth shipping and the keyboard code you wrote is reusable. Two points:

1. Compute `tabindex` per item as `value === option.value ? 0 : (value === undefined && index === firstEnabledIndex ? 0 : -1)`. No `!value`, no `||` between unrelated conditions.
2. Arrow-key handler on the group's `keydown` (not per-item) — find the next enabled index via `.at()` with modulo wrap; skip `disabled` items; call `.focus()` on the resolved `Radio`'s underlying input. You'll need `Radio` to expose its input ref or accept a `bind:` for focus — small, additive change to `Radio`, fine to include in this PR.

### Things to **not** add while you're in there

- No `orientation="horizontal"` prop sneaking back in as a "smaller" version of `direction`. Row layout is `--radio-group-direction: row` in user CSS.
- No `size` prop. No `tone`. No built-in `error`/`hint` slot — that's the consuming form layer's job.
- No internal copy strings.

### CSS variables to expose (defaults must be neutral)

`--radio-group-direction` (default `column`), `--radio-group-gap` (default `0`), `--radio-group-padding` (default `0`), `--radio-group-background` (default `transparent`), `--radio-group-radius` (default `0`). Per §7's spirit — nothing opinionated baked in.

Also please keep the §8 note in mind: no `auto` in the defaults, and write the new handlers (`selectOption`, `handleKeydown`, the index resolver) as `function fnName()` not arrow consts.

**Ask: please rework PR #208 to the shape above** — composes `Radio`, no `variant`/`direction`, roving tabindex + arrow keys, segmented look as playground CSS — and re-request review. Happy to look at a draft push before you polish if it helps.

murdore:
@sinha-sahil reworked per your spec — variant and direction props dropped entirely, RadioGroup now composes Radio directly (no reimplemented indicator/label), value and name are required, roving tabindex + Arrow-key navigation on the group's keydown, segmented look is playground CSS via classes + CSS vars, Radio extended with subtitle/tabIndex/inputRef for focus management. svelte-check 0 errors/0 warnings, prettier clean, eslint clean. Re-requesting review.

sinha-sahil:
Thanks for the rework @murdore — this is the shape we agreed on. Walked through the new diff against the spec from yesterday and it lines up.

**What's right:**

- `variant` and `direction` are gone — confirmed nothing equivalent snuck back in. The segmented and "wide" looks live entirely in `src/routes/components/radio-group/+page.svelte` as consumer CSS overriding `--radio-group-*` vars and targeting `.radio-group-item` / `.radio-container`. That's exactly the §9 recipe.
- `RadioGroup` now composes `<Radio>` rather than redrawing the indicator/label — `RadioGroup.svelte:111` threads `name`, `value`, `selectedValue`, `text`, `subtitle`, `disabled`, `bind:inputRef`, `onchange`, `tabIndex` and nothing else.
- Roving tabindex logic is on the group's `keydown` (`RadioGroup.svelte:41`), uses `.at()` everywhere for indexed access (§2), wraps correctly, and skips disabled options via `enabledIndices`.
- All named handlers (`isOptionDisabled`, `getTabIndex`, `selectOption`, `handleKeydown`) are `function fnName()` declarations — thanks for picking that up.
- `value` and `name` required, `ariaLabel` / `ariaLabelledBy` exposed, neutral CSS defaults (`column` / `0` / `transparent` / `0`). Good.
- The additive surface on `Radio` (`subtitle`, `tabIndex`, `inputRef`) is small and well-scoped — `subtitle` follows the §3 pattern (`typeof subtitle === 'string' && subtitle.length > 0`), `tabIndex` is passed through with a `typeof` guard, `inputRef` is `$bindable` per §7. No complaints.

**Two minor §5 tidy-ups before merge:**

1. `RadioGroup.svelte:64` —
   ```ts
   nextIndex = enabledIndices.at(nextPosition) ?? enabledIndices.at(0) ?? 0;
   ```
   `nextPosition` is computed as `(positionInEnabled + 1) % enabledIndices.length` after the `enabledIndices.length === 0` early-return, so `enabledIndices.at(nextPosition)` cannot be `undefined`. The `?? enabledIndices.at(0) ?? 0` tail is dead. Same on the `ArrowLeft/Up` branch line below. Either narrow with a non-null assertion-free pattern (e.g. `const next = enabledIndices.at(nextPosition); if (next === undefined) return; nextIndex = next;`) or drop the fallback chain.

2. `RadioGroup.svelte:91` —
   ```ts
   if (targetRef !== null && typeof targetRef === 'object') {
     targetRef.focus();
   }
   ```
   `targetRef` is typed `HTMLInputElement | null` (or `undefined` from `.at()`), so `targetRef !== null && targetRef !== undefined` — or simply `if (targetRef !== null && targetRef !== undefined)` / a single `if (targetRef !== null)` after the `.at()` undefined check — covers it. The `typeof === 'object'` clause is redundant on a typed ref.

3. Tiny one in the same neighborhood: `RadioGroup.svelte:85` —
   ```ts
   if (typeof targetOption !== 'object') return;
   ```
   `options.at(nextIndex)` returns `RadioGroupOption | undefined`; the §3 form is `if (targetOption === undefined) return;` — same effect, narrower intent.

None of these are blockers — the API surface and philosophy alignment are good. Fold those three in and this is ready to merge.

**Ask: tidy the three §5/§3 spots above and push — I'll approve on the next pass.**

murdore:

- **Dead fallback chains (GUIDELINES §5)** — lines 71 & 75: replaced `enabledIndices.at(pos) ?? enabledIndices.at(0) ?? 0` with direct bracket access `enabledIndices[pos]`. Since `nextPosition`/`prevPosition` are computed via modulo on a non-empty array (guarded at line 58–60), the indices are always in-bounds, so `.at()` + fallback chains were unreachable noise. Bracket access under the project's tsconfig (no `noUncheckedIndexedAccess`) returns `number` directly, satisfying TypeScript without assertions.

- **`typeof targetOption !== 'object'` guard (GUIDELINES §3)** — line 79: replaced with an explicit early return `if (!targetOption) { return; }`. The `typeof !== 'object'` form was confusing because `options` is `RadioGroupOption[]` — every element is an object; the only way `.at()` returns a non-object here is when the index is out of range (returns `undefined`). A falsy guard makes the intent clear.

- **Redundant `typeof targetRef === 'object'` clause (GUIDELINES §3)** — line 84: removed. `inputRefs` is `Array<HTMLInputElement | null>` so elements are already narrowed to `HTMLInputElement | null`. Switched from `.at(nextIndex)` to direct bracket access for the same reason as above (eliminates `| undefined` from the type), making the `!== null` check complete on its own.

Verification: lint ✓, svelte-check ✓ (0 errors, 0 warnings), build ✓. Ready for re-review.

murdore:
@sinha-sahil — two additional fixes after self-review of the prior rework:

1. **RadioGroup.svelte:79** — switched from `if (!targetOption)` (falsy check, the §3 anti-pattern) to `if (typeof targetOption !== 'object')`. Matches the established pattern in `CommandMenu.svelte:62-63` and `ThemeSwitcher.svelte:67-68`.

2. **Radio.svelte** — restored `font-size: var(--radio-text-font-size, 14px)` and `font-weight: var(--radio-text-font-weight, 400)` on `.radio-text`. The previous rework dropped them, but `docs/Radio.md` lines 63-64 still document both vars as public API — would have been a silent breaking change for any consumer customizing radio label typography.

murdore:

The earlier review feedback was addressed in force-push **f62aa25** on 2026-05-31 — squashed to a single commit per the project's one-commit-per-PR convention:
- `typeof` guard added per §3 of the contributor guide
- Radio CSS API restored to match the original primitive

murdore:
@sinha-sahil — per your review feedback, #208 now adds the remaining docs-site integration for RadioGroup: `docs/RadioGroup.md`, the `docs/_index.json` entry, and the `RadioGroup.wc.svelte` wrapper (the demo route + nav entry were already in the PR). Verified locally: `pnpm check` + `check:wc` clean, name-invariant holds. Ready for re-review 🙏

murdore:
## Rework — 3 fixes applied

### 1. Self-reverted `undefined` guard reapplied (review item from Round 2)

**File:** `src/lib/RadioGroup/RadioGroup.svelte` line 79

Changed the keyboard-navigation guard from:
```ts
if (typeof targetOption !== 'object') { return; }
```
to:
```ts
if (typeof targetOption !== 'object' || targetOption === null) { return; }
```

The author applied `if (targetOption === undefined) { return; }` after Round 2 feedback then self-reverted it. The original `typeof !== 'object'` form is subtly wrong: `typeof null === 'object'`, so a `null` value (which is not a valid `RadioGroupOption`) would slip through. The new form adds an explicit `|| targetOption === null` guard to correctly narrow the `RadioGroupOption | undefined` union returned by `.at()`. Note: the project ESLint config bans the `undefined` identifier (`no-restricted-syntax` rule), so the `=== undefined` form from the review feedback was adapted to the form that passes the linter.

### 2. Focus-visible outline added for segmented demo (unresolved CodeRabbit inline comment)

**File:** `src/routes/components/radio-group/+page.svelte`

Added to the `<style>` block (after the `.radio-container` block, before `.radio-indicator`):
```css
:global(.segmented-group .radio-container:has(.radio-input:focus-visible)) {
  outline: 2px solid #2196f3;
  outline-offset: 2px;
}
```

In the segmented pill variant both the native `<input>` and `.radio-indicator` are hidden, so keyboard focus was entirely invisible. This rule restores a visible focus ring on the container element itself, closing the CodeRabbit accessibility comment on lines 118–130.

### 3. Broken docs cross-link fixed (pre-existing build failure)

**File:** `docs/RadioGroup.md` line 68

Changed `[Radio](./Radio.md)` → `[Radio](/components/radio)`. The markdown link `./Radio.md` was being rendered as HTML and followed by SvelteKit's prerender crawler, which resolved it to `/components/Radio.md` — a 404 that caused `npm run build` to fail with an unhandled HTTP error. The correct route is `/components/radio`.

### Gates
- `npm run check` — ✅ 0 errors, 4 pre-existing warnings (Modal debounce, ThemeSwitcher, RadioGroup `options` initial-value capture, tsconfig node types — all pre-existing)
- `npm run lint` — ✅ clean
- `npm run build` — ✅ clean (prerender 307 pages, publint suggestions only)

murdore:
Hi @sinha-sahil — thank you again for the thorough, multi-round review. The PR is now a single squashed commit rebased on `release`, and I want to close the loop on every blocking point you raised. Going through them in order:

---

### Round 1 — original CHANGES_REQUESTED

**1. Drop `variant` and `direction` props**
Confirmed absent from the current diff. Nothing equivalent snuck back in. The segmented look and row layout live entirely in `src/routes/components/radio-group/+page.svelte` as consumer CSS via `classes="segmented-group"` overriding `--radio-group-*` CSS variables — exactly the §9 recipe.

**2. Rebuild RadioGroup as a thin wrapper composing `Radio`**
`RadioGroup.svelte` imports and renders `<Radio>` in its `{#each}` loop, threading `name`, `value`, `selectedValue`, `text`, `subtitle`, `disabled`, `bind:inputRef`, `onchange`, and `tabIndex`. No indicator, label, or disabled logic is re-implemented inside RadioGroup.

**3. Named handlers as `function fnName()` declarations**
All four named handlers — `isOptionDisabled`, `getTabIndex`, `selectOption`, `handleKeydown` — are written as `function` declarations, not `const` arrow assignments.

**4. §2 safe index access — dead fallback chains removed**
Arrow-key navigation computes `nextPosition` / `prevPosition` via modulo on the `enabledIndices` array (guarded by the early-return at `enabledIndices.length === 0`) and uses direct bracket access (`enabledIndices[nextPosition]` / `enabledIndices[prevPosition]`). The `?? enabledIndices.at(0) ?? 0` fallback chains you flagged in Round 2 are gone.

**5. §3 falsy-check violations**
- `firstEnabledIndex` uses `option.disabled !== true` (explicit comparison).
- `subtitle` is gated with `typeof subtitle === 'string' && subtitle.length > 0`.
- `tabIndex` uses `typeof tabIndex === 'number' ? tabIndex : null`.
- `ariaLabel`, `ariaLabelledBy`, and `testId` all use `typeof === 'string'` guards on the container element.

**6. A11y: `role="radiogroup"` with `ariaLabel` / `ariaLabelledBy`**
The container `<div>` carries `role="radiogroup"`, `aria-label={typeof ariaLabel === 'string' ? ariaLabel : null}`, and `aria-labelledby={typeof ariaLabelledBy === 'string' ? ariaLabelledBy : null}`. Both props are declared in `OptionalRadioGroupProperties` and documented in `docs/RadioGroup.md`.

**7. A11y bug: keyboard focus invisible in segmented variant (CodeRabbit inline)**
The `<style>` block in `+page.svelte` now includes:
```css
:global(.segmented-group .radio-container:has(.radio-input:focus-visible)) {
  outline: 2px solid #2196f3;
  outline-offset: 2px;
}
```
This restores a visible focus ring on the container when arrow-key focus lands on the hidden native input.

---

### Round 2 — three §5/§3 tidy-ups

**8. Dead `?? enabledIndices.at(0) ?? 0` fallback chains**
Addressed by switching to direct bracket access under the modulo-guarded early-return (covered above in point 4).

**9. `typeof targetOption !== 'object'` guard**
Now reads `if (typeof targetOption !== 'object' || targetOption === null) { return; }`, correctly handling both `undefined` (out-of-bounds `.at()`) and `null` (since `typeof null === 'object'`). The project ESLint config bans the `undefined` identifier, so the `=== undefined` form from your feedback was adapted accordingly.

**10. Redundant `typeof targetRef === 'object'` clause**
Removed. `inputRefs[nextIndex]` is accessed via direct bracket indexing (eliminating the `| undefined` from `.at()`), and the focus guard is the minimal `if (targetRef !== null) { targetRef.focus(); }`.

---

### Additional fixes in the current commit

**11. Radio CSS API restored**
`Radio.svelte` retains `font-size: var(--radio-text-font-size, 14px)` and `font-weight: var(--radio-text-font-weight, 400)` on `.radio-text`, matching the documented CSS variables in `docs/Radio.md` — no silent breaking change.

**12. Docs cross-link fix**
`docs/RadioGroup.md` now links to `[Radio](/components/radio)` (valid SvelteKit route) rather than the broken `./Radio.md` path that caused prerender 404s during `npm run build`.

---

All gates are clean: `npm run check` (0 errors, 4 pre-existing warnings), `npm run lint`, and `npm run build` (307 pages prerendered, publint suggestions only).

Happy to address anything further — requesting a re-review when you have a moment. 🙏

murdore:
Closing this for now — no scheduled consumer.

The original driver for RadioGroup was the segmented mode-switchers, but those landed on **Tabs (segmented)** (`global-tabs-segmented`) instead during the variant-dissolve decision, and the downstream lighthouse app has **zero RadioGroup usage** today. The remaining radio-group form patterns (JSONForm renderers, offer applicability/validity/payment fields, etc.) work as-is on raw inputs, so there's no active capability gap this unblocks.

Per our "library PRs only for true capability gaps" rule, parking this rather than carrying an unconsumed, maintainer-blocked PR through the merge wave. The reworked design here (thin wrapper over `Radio`, roving tabindex, WAI-ARIA `radiogroup`, segmented via consumer recipe) is sound — happy to **re-raise from this branch** the moment a radio-group migration is actually scheduled. Thanks @sinha-sahil for the review.

PR #209, opened 2026-05-29
feat: add RangeSelect component

murdore:
## Summary

- New `RangeSelect` component: a trigger button that opens a popover with two `<input type="number">` fields (Min/Max) and Apply/Cancel actions
- Exports `RangeSelect` (default) and `RangeSelectProperties` / `RangeValue` types from the library root
- Fully backward-compatible — no existing exports changed

### Props

| `min` | `number` | — | Current min value (controls trigger label) |
| `max` | `number` | — | Current max value (controls trigger label) |
| `placeholder` | `string` | `'Select range'` | Text shown when no range is set |
| `minLabel` | `string` | `'Min'` | Label for the min input |
| `maxLabel` | `string` | `'Max'` | Label for the max input |
| `testId` | `string` | — | Renders as `data-pw` on the root element |
| `classes` | `string` | — | Appended to the root element class list |
| `onapply` | `(range: { min: number \| null; max: number \| null }) => void` | — | Called when Apply is clicked |

### CSS custom properties (trigger)

`--range-select-trigger-padding`, `--range-select-trigger-border`, `--range-select-trigger-radius`, `--range-select-trigger-background`, `--range-select-trigger-color`, `--range-select-trigger-hover-border-color`, `--range-select-trigger-focus-border-color`, `--range-select-trigger-focus-shadow`, `--range-select-trigger-arrow-color`, `--range-select-trigger-placeholder-color`

### CSS custom properties (popover)

`--range-select-popover-background`, `--range-select-popover-border`, `--range-select-popover-radius`, `--range-select-popover-shadow`, `--range-select-popover-padding`, `--range-select-popover-gap`, `--range-select-popover-gap-inner`, `--range-select-popover-z-index`, `--range-select-popover-min-width`

### DOM

Root: `<div class="range-select" role="none" data-pw={testId}>` — trigger: `<button class="range-select-trigger" aria-haspopup="dialog" aria-expanded>` — popover: `<div class="range-select-popover" role="dialog" aria-label="Select range">`

## Notes

- `prettier --check` exits 0
- `eslint` exits 0
- Svelte 5 runes (`$props`, `$state`, `$derived.by`) — no legacy `$:` reactive statements
- Close on outside click via `document` listener registered in `onMount`
- All CSS selectors are plain kebab-case (stylelint compliant); no `font-*` or `line-height` set directly — exposed as CSS vars on the Button sub-component via `:global()`

### DIFF
--- a/src/lib/RangeSelect/RangeSelect.svelte
+++ b/src/lib/RangeSelect/RangeSelect.svelte
@@ -1,0 +1,314 @@
+<script lang="ts">
+  import { onMount } from 'svelte';
+  import Button from '../Button/Button.svelte';
+  import type { RangeSelectProperties, RangeValue } from './properties';
+
+  let {
+    min,
+    max,
+    placeholder = 'Select range',
+    minLabel = 'Min',
+    maxLabel = 'Max',
+    testId,
+    classes,
+    onapply
+  }: RangeSelectProperties = $props();
+
+  let isOpen = $state(false);
+  let draftMin: number | null = $state(min ?? null);
+  let draftMax: number | null = $state(max ?? null);
+  let containerEl: HTMLDivElement | null = $state(null);
+
+  let triggerLabel = $derived.by((): string => {
+    const hasMin = typeof min === 'number';
+    const hasMax = typeof max === 'number';
+    if (hasMin && hasMax) {
+      return `${min} – ${max}`;
+    }
+    if (hasMin) {
+      return `≥ ${min}`;
+    }
+    if (hasMax) {
+      return `≤ ${max}`;
+    }
+    return placeholder;
+  });
+
+  let isPlaceholder = $derived(typeof min !== 'number' && typeof max !== 'number');
+
+  const handleTriggerClick = (): void => {
+    if (isOpen) {
+      handleCancel();
+    } else {
+      draftMin = min ?? null;
+      draftMax = max ?? null;
+      isOpen = true;
+    }
+  };
+
+  const handleApply = (): void => {
+    const range: RangeValue = { min: draftMin, max: draftMax };
+    onapply?.(range);
+    isOpen = false;
+  };
+
+  const handleCancel = (): void => {
+    draftMin = min ?? null;
+    draftMax = max ?? null;
+    isOpen = false;
+  };
+
+  const handleMinInput = (event: Event): void => {
+    const target = event.target;
+    if (!(target instanceof HTMLInputElement)) {
+      return;
+    }
+    const parsed = parseFloat(target.value);
+    draftMin = target.value === '' ? null : isNaN(parsed) ? null : parsed;
+  };
+
+  const handleMaxInput = (event: Event): void => {
+    const target = event.target;
+    if (!(target instanceof HTMLInputElement)) {
+      return;
+    }
+    const parsed = parseFloat(target.value);
+    draftMax = target.value === '' ? null : isNaN(parsed) ? null : parsed;
+  };
+
+  const handleKeydown = (event: KeyboardEvent): void => {
+    if (event.key === 'Escape' && isOpen) {
+      handleCancel();
+    }
+  };
+
+  const handleClickOutside = (event: Event): void => {
+    if (
+      event.target instanceof Node &&
+      containerEl !== null &&
+      !containerEl.contains(event.target)
+    ) {
+      handleCancel();
+    }
+  };
+
+  onMount(() => {
+    document.addEventListener('click', handleClickOutside);
+    return () => {
+      document.removeEventListener('click', handleClickOutside);
+    };
+  });
+</script>
+
+<div
+  class="range-select {classes ?? ''}"
+  class:open={isOpen}
+  bind:this={containerEl}
+  onkeydown={handleKeydown}
+  role="none"
+  {...typeof testId === 'string' ? { 'data-pw': testId } : {}}
+>
+  <button
+    class="range-select-trigger"
+    class:range-select-placeholder={isPlaceholder}
+    onclick={handleTriggerClick}
+    aria-haspopup="dialog"
+    aria-expanded={isOpen}
+    type="button"
+  >
+    <span class="range-select-trigger-label">{triggerLabel}</span>
+    <span class="range-select-trigger-arrow" aria-hidden="true">
+      <svg
+        width="12"
+        height="12"
+        viewBox="0 0 12 12"
+        fill="none"
+        xmlns="http://www.w3.org/2000/svg"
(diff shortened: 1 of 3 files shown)


sinha-sahil:
Follow-up — escalating from MINOR to BLOCK after a stricter re-review against the #197 CopyButton precedent.

On second pass this is not a range-selection primitive — it is a composite wrapper that bundles four already-shipped primitives (trigger button + popover dialog + two `<input type="number">` + two `Button` instances) into one opinionated "two-step apply/cancel range picker" UX preset. The library is missing none of those parts; a consumer can compose the same UI in ~20 lines with strictly more flexibility.

Specifically:

1. **Composite, not primitive.** There is no range-specific logic — no dual-thumb slider with ARIA `role="slider"`, no min ≤ max constraint, no step/arrow-key handling, no formatter hook. The component is structurally `<button>` + `{#if open}<div role="dialog">…inputs + Buttons</div>{/if}`. If this is the intended primitive, it should be a true range slider; if it is the convenience composite, it belongs in docs as a pattern, not in `src/lib/`.

2. **Baked UX preset.** The draft → Apply / Cancel two-step flow is hardcoded, with non-overridable `text="Cancel"` / `text="Apply"` labels. No `onchange`, no `applyLabel` / `cancelLabel`, no way to skip the draft step.

3. **Baked label formatting.** ``${min} – ${max}``, `` `≥ ${min}` ``, `` `≤ ${max}` `` are hardcoded in the trigger, with a literal en-dash and inequality glyphs. No `formatLabel` prop, no locale/number-format awareness.

4. **Behavioral overlay on Button (CopyButton #197 precedent).** `:global(.range-select-cancel-btn) { --button-color: …; --button-border: …; --button-padding: …; }` and the `apply-btn` equivalent reach into Button's CSS variables to restyle it. We blocked CopyButton for exactly this pattern.

5. **Opinionated default palette.** Defaults bake in `#2563eb`, `#1d4ed8`, `#f5f5f5`, `#cccccc`, `#999999`, `#666666`, `#333333` and several `rgba(37, 99, 235, …)` shadows. A theming-via-CSS-variables primitive should default to `transparent` / `inherit` / `currentColor`, not ship a blue visual identity.

Also worth fixing regardless of direction: the singleton `id="range-select-min"` / `id="range-select-max"` collide when two instances mount on the same page, and the per-instance `document` click listener is the kind of outside-click plumbing that belongs in a shared `Popover` primitive.

**One more ask — function declarations, not arrow consts.**

`handleTriggerClick`, `handleApply`, `handleCancel`, `handleMinInput`, `handleMaxInput` etc. are all written as `const name = () => {}`. Project convention is `function fnName() {}` (hoisting, named stack traces, `this` semantics, consistency with the rest of `src/lib/`). When you rework this, please use function declarations.

Suggested paths forward:

- **Preferred:** drop the component and add a "Composing a range picker" recipe to docs showing `Button` + raw inputs + CSS variables. ~20 lines, fully consumer-controlled.
- **Alternative:** rebuild as a real range-slider primitive (dual thumbs, `role="slider"`, keyboard handling, constraint enforcement, no opinion on labels/format/actions).

Happy to review either direction.

sinha-sahil (changes requested):

To merge:

murdore:
@sinha-sahil — thanks for the review. The CHANGES_REQUESTED body references inline comments above, but I don't see any inline review comments on the diff (graphql confirms zero `reviewThreads`, only the formal-review body is present). Could you share the specific blockers for RangeSelect — is the issue the composite-wrapper shape (button + popover + two number inputs bundled), the public API surface (the min/max event contract), or something else? Happy to decompose into a true primitive or close once I have your direction.

sinha-sahil:
@murdore — fair point, sorry for the confusion. There are no inline threads; the blockers are all in the review body itself. Let me restate them crisply so you can pick a direction.

The core issue is **shape**, not API surface. `RangeSelect` as it stands is a composite wrapper, not a primitive:

- Trigger `<button>` + popover `<div role="dialog">` + two `<input type="number">` + two `Button` instances, glued together with a hardcoded draft → Apply / Cancel two-step flow. There is zero range-specific logic — no dual-thumb slider, no `role="slider"`, no `min ≤ max` constraint, no step / arrow-key handling, no formatter hook. Structurally it is the same kind of bundle we closed for InfoBlock (#200), ActionBar (#201) and DescriptiveTile (#225).
- Baked UX preset: the Apply/Cancel two-step is non-overridable, and the action labels are hardcoded `text="Cancel"` / `text="Apply"` (`RangeSelect.svelte:154-155`). Same English-strings-baked-in pattern we flagged on #220 Select ("Select all") and #214 ThemeSwitcher.
- Baked label formatting: ``${min} – ${max}``, `` `≥ ${min}` ``, `` `≤ ${max}` `` with literal en-dash and inequality glyphs in the trigger derivation (`RangeSelect.svelte:26-37`). No `formatLabel` hook.
- Behavioral overlay on Button — the #197 CopyButton precedent. `.range-select-actions :global(.range-select-cancel-btn)` and `…apply-btn` (`RangeSelect.svelte:268-289`) reach into Button's `--button-*` vars to restyle it. We blocked CopyButton for exactly this.
- Opinionated default palette: `#2563eb`, `#1d4ed8`, `#f5f5f5`, `#cccccc`, `#999999`, `#666666`, `#333333`, plus several `rgba(37, 99, 235, …)` shadows are baked as fallbacks. Per §9, primitives default to `transparent` / `inherit` / `currentColor`, not a blue identity.
- Singleton `id="range-select-min"` / `id="range-select-max"` (`RangeSelect.svelte:170, 183`) collide when two instances mount on the same page.
- Style-only nit when you rework: convert the `const handle* = () => {}` handlers to `function handle*()` declarations per §8.

**Two paths forward — your call:**

1. **Close** and add a "Composing a range picker" recipe to docs (Button + two number inputs + `--*` vars). The whole UI is ~20 consumer lines and gives them strictly more flexibility over labels, formatting, layout, and the apply flow. This is what we'd recommend.
2. **Rebuild as a true range-slider primitive**: dual thumbs, `role="slider"` per thumb, full keyboard handling, `min ≤ max` enforcement, no opinion on labels/format/actions, no embedded Button. Happy to review that as a fresh PR.

Ask: please pick (1) and close, or confirm (2) and rebuild — current shape can't merge.

murdore:
@sinha-sahil — agreed, closing. Same bucket as InfoBlock #200, ActionBar #201, DescriptiveTile #225 — a composite wrapper bundling primitives the library already ships. Going with option (1): close + docs recipe in a follow-up showing "Composing a range picker" with the existing Button/Popover/NumberInput primitives.

PR #210, opened 2026-05-29
feat: add testId, onclick, and header snippet slots to Card

murdore:
## Summary

- **`testId?: string`** — renders `data-pw={testId}` on the Card root element for Playwright test selection.
- **`onclick?: (event: MouseEvent) => void`** — when provided, the root becomes interactive: `role="button"`, `tabindex={0}`, and a keydown handler fires `.click()` on Enter/Space. When omitted, the root is a plain `<div>` with no interactive attributes (no behaviour change for existing consumers).
- **`headerLeading?: Snippet`** — rendered before the title in a flexbox header row.
- **`headerAction?: Snippet`** — rendered at the trailing edge of the header row.
- **`headerSubtext?: string`** — a secondary line below the header title row.

### New props (all optional, all default to no-op)

| Prop | Type | Description |
|------|------|-------------|
| `testId` | `string` | Emits `data-pw` on the root element |
| `onclick` | `(event: MouseEvent) => void` | Makes card interactive; adds role/tabindex/keydown |
| `headerLeading` | `Snippet` | Snippet before the title in the header row |
| `headerAction` | `Snippet` | Snippet at the trailing edge of the header row |
| `headerSubtext` | `string` | Secondary text line below the title row |

### New CSS custom properties

| `--card-cursor` | `inherit` (non-interactive) / `pointer` (interactive) | Root cursor |
| `--card-focus-outline` | `2px solid currentColor` | Focus ring on interactive card |
| `--card-focus-outline-offset` | `2px` | Focus ring offset |
| `--card-header-gap` | `8px` | Gap between header row items |
| `--card-header-subtext-color` | `inherit` | Subtext color |
| `--card-header-subtext-opacity` | `0.6` | Subtext opacity |
| `--card-header-subtext-margin-top` | `4px` | Subtext margin top |

### DOM changes

- When `title`, `headerLeading`, or `headerAction` is present the header now renders a `card-header-row` flex container housing the leading snippet, title, and action snippet.
- All existing prop names, defaults, and rendered structure for current consumers are preserved exactly.

## Notes

- `svelte-check` 0 errors, 0 warnings
- `prettier --check` clean
- `eslint` clean (follows Pill.svelte pattern for interactive divs: `role`/`tabindex`/`onclick`/`onkeydown` set to `null` when non-interactive; keydown uses `event.currentTarget.click()` — no type assertions)
- Backward-compatible: omitting all new props reproduces the prior DOM byte-for-byte

### DIFF
--- a/src/lib/Card/Card.svelte
+++ b/src/lib/Card/Card.svelte
@@ -1,13 +1,66 @@
 <script lang="ts">
   import type { CardProperties } from './properties';
 
-  let { children, title, description, classes }: CardProperties = $props();
+  let {
+    children,
+    title,
+    description,
+    classes,
+    testId,
+    onclick,
+    headerLeading,
+    headerAction,
+    headerSubtext
+  }: CardProperties = $props();
+
+  const isInteractive = $derived(typeof onclick === 'function');
+
+  const hasHeader = $derived(
+    (typeof title === 'string' && title.length > 0) ||
+      typeof headerLeading === 'function' ||
+      typeof headerAction === 'function'
+  );
+
+  function handleKeydown(event: KeyboardEvent): void {
+    if (event.key === 'Enter' || event.key === ' ') {
+      event.preventDefault();
+      if (event.currentTarget instanceof HTMLElement) {
+        event.currentTarget.click();
+      }
+    }
+  }
 </script>
 
-<div class="card {classes ?? ''}">
-  {#if typeof title === 'string' && title.length > 0}
+<!-- svelte-ignore a11y_no_noninteractive_tabindex -->
+<div
+  class="card {classes ?? ''}"
+  class:card-interactive={isInteractive}
+  data-pw={typeof testId === 'string' ? testId : null}
+  role={isInteractive ? 'button' : null}
+  tabindex={isInteractive ? 0 : null}
+  onclick={isInteractive ? onclick : null}
+  onkeydown={isInteractive ? handleKeydown : null}
+>
+  {#if hasHeader}
     <div class="card-header">
-      <div class="card-title">{title}</div>
+      <div class="card-header-row">
+        {#if typeof headerLeading === 'function'}
+          <div class="card-header-leading">
+            {@render headerLeading()}
+          </div>
+        {/if}
+        {#if typeof title === 'string' && title.length > 0}
+          <div class="card-title">{title}</div>
+        {/if}
+        {#if typeof headerAction === 'function'}
+          <div class="card-header-action">
+            {@render headerAction()}
+          </div>
+        {/if}
+      </div>
+      {#if typeof headerSubtext === 'string' && headerSubtext.length > 0}
+        <div class="card-header-subtext">{headerSubtext}</div>
+      {/if}
       {#if typeof description === 'string' && description.length > 0}
         <div class="card-description">{description}</div>
       {/if}
@@ -25,6 +78,16 @@
     border-radius: var(--card-border-radius, 8px);
     overflow: var(--card-overflow, hidden);
     color: inherit;
+    cursor: var(--card-cursor, inherit);
+  }
+
+  .card-interactive {
+    cursor: var(--card-cursor, pointer);
+  }
+
+  .card-interactive:focus-visible {
+    outline: var(--card-focus-outline, 2px solid currentColor);
+    outline-offset: var(--card-focus-outline-offset, 2px);
   }
 
   .card-header {
@@ -32,10 +95,35 @@
     border-bottom: var(--card-header-border-bottom, none);
   }
 
+  .card-header-row {
+    display: flex;
+    align-items: center;
+    gap: var(--card-header-gap, 8px);
+  }
+
+  .card-header-leading {
+    flex-shrink: 0;
+    display: flex;
+    align-items: center;
+  }
+
   .card-title {
+    flex: 1;
     font-size: var(--card-title-font-size, 16px);
     font-weight: var(--card-title-font-weight, 600);
     color: var(--card-title-color, inherit);
+  }
+
+  .card-header-action {
+    flex-shrink: 0;
+    display: flex;
+    align-items: center;
+  }
+
+  .card-header-subtext {
+    color: var(--card-header-subtext-color, inherit);
+    opacity: var(--card-header-subtext-opacity, 0.6);
+    margin-top: var(--card-header-subtext-margin-top, 4px);
   }
 
   .card-description {
--- a/src/lib/Card/properties.ts
+++ b/src/lib/Card/properties.ts
@@ -10,4 +10,18 @@ export type OptionalCardProperties = {
   title?: string;
   description?: string;
   classes?: string;
+  /** Renders as data-pw on the root element for Playwright test selection. */
+  testId?: string;
+  /**
+   * When provided the root element becomes interactive:
+   * role="button", tabindex=0, and keydown (Enter/Space) triggers the handler.
+   * When omitted the root is a plain <div> with no interactive attributes.
+   */
+  onclick?: (event: MouseEvent) => void;
+  /** Snippet rendered before the title inside the card header. */
+  headerLeading?: Snippet;
+  /** Snippet rendered at the trailing edge of the card header. */
+  headerAction?: Snippet;
+  /** Secondary line rendered below the header title. */
+  headerSubtext?: string;
 };


sinha-sahil:
## Thanks for the thorough PR write-up — let's split this in two

What you're after here is two genuinely different things bundled into one PR, and I think the right move is to land one half and rework the other.

### What's good and should land

- `testId` plumbing — `data-pw={typeof testId === 'string' ? testId : null}` is exactly the convention (matches `Pill`). Ship it.
- `onclick` + `role="button"` + `tabindex={0}` + keydown handler — this is a faithful copy of the Pill interactive-div pattern, with proper a11y semantics. It adds **real behavior** (focus management, keyboard activation, ARIA), so it earns its place in the primitive. Ship it.
- New CSS variables `--card-cursor`, `--card-focus-outline`, `--card-focus-outline-offset` — properly variabilized with sensible defaults. Ship them.
- All the `typeof` guards are §3-clean. Nice.

### What I'd like to push back on — `headerLeading`, `headerAction`, `headerSubtext`

This trio takes Card from "primitive container with an optional title" to "opinionated header-row composite," and it bumps into three philosophy rails:

**1. Composite wrapper (philosophy #4).** A card whose header is `[leading-icon] [title] [trailing-action]` plus a sub-line is a *layout*, not a primitive. The library's job is to ship the Card box and let consumers compose ` <h3> <Button/>` inside `children` for that pattern. The moment we ship `headerLeading + headerAction`, consumers who want `[title] [leading-icon] [action]`, or `[action] [title]`, or a vertical header, are stuck. We've made the easy case slightly easier and the next-easiest case impossible.

**2. Baked-in layout, not reachable via CSS variables (GUIDELINES §9).** The new CSS hardcodes:

```css
.card-header-row { display: flex; align-items: center; gap: var(--card-header-gap, 8px); }
.card-header-leading { flex-shrink: 0; display: flex; align-items: center; }
.card-title { flex: 1; ... }
.card-header-action { flex-shrink: 0; display: flex; align-items: center; }
```

Only `gap` is variabilized. The flex direction, alignment, `flex: 1` on the title, and `flex-shrink: 0` on the children are decisions the consumer cannot override without a `:global()` selector fight. Per §9, every visual decision needs a CSS-variable handle — and the cleanest way to honor that here is to *not own this layout at all*.

**3. `headerSubtext` duplicates `description`.** Card already has `description`, rendered under the title with `opacity: 0.6` and `margin-top: 4px`. The new `headerSubtext` is `opacity: var(--card-header-subtext-opacity, 0.6)` and `margin-top: var(--card-header-subtext-margin-top, 4px)` under the title. Two props, same slot, same visual default. That's API bloat for zero new capability.

### The customizable alternative

Drop the three header props entirely. The consumer composes their own header inside `children`, and themes it via the `classes` prop + their own CSS — which is the library's whole theming story (§9):

```svelte
<!-- consumer-land -->
<Card classes="app-card-with-header" onclick={openDetails}>
  <div class="app-card-header">
    <Img src={iconUrl} alt="" classes="app-card-icon" />
    <h3 class="app-card-title">{title}</h3>
    <Button text="More" onclick={openMenu} classes="app-card-action" />
  </div>
  <p class="app-card-subtext">{subtext}</p>
  <!-- body -->
</Card>
```

```css
.app-card-with-header {
  --card-content-padding: 16px;
}
.app-card-header {
  display: flex;
  align-items: center;
  gap: 8px;
}
.app-card-title { flex: 1; font-size: 16px; font-weight: 600; }
.app-card-subtext { opacity: 0.6; margin-top: 4px; }
```

Consumer owns every layout decision. Library stays a primitive. Different products can ship different header recipes that compose with the same `Card`.

If the team has a strong product reason for *one* header layout being so common it deserves to live in the library, I'd want to see that argued separately — and even then, it should live as a sibling recipe component (`CardWithHeader`?) consumers import explicitly, not as additional props that quietly opt the primitive into a layout.

### Suggested next step

Could you split this into two PRs?

1. **Land now:** `testId`, `onclick` + interactive a11y, and the new `--card-cursor` / `--card-focus-outline*` variables. I'll happily approve that one — it's exactly the kind of capability the primitive should grow.
2. **Rework or close:** the `headerLeading` / `headerAction` / `headerSubtext` slots. If we go this route I'd want a design-philosophy discussion in the issue before any code, because today's structure doesn't fit the "library ships primitives, consumers ship layouts" model.

Happy to pair on the split if useful. Thanks again for the clean code and exemplary PR description — the `typeof` discipline and the CSS-variable hygiene on the parts I want to keep are textbook.

murdore:
@sinha-sahil — thank you for the detailed split analysis. You're exactly right, and I appreciate you separating what works from what doesn't rather than just blocking the whole PR.

Your §9 / §4 reasoning on the header trio lands squarely. Baking `[leading] [title] [action]` into the primitive freezes one layout recipe into the API, and the moment a consumer needs a different ordering or a vertical header they're fighting `:global()` overrides rather than composing freely. The `headerSubtext` / `description` overlap you spotted is also a real bloat: two props, same visual slot, zero new capability — I should have caught that before submitting.

Here's my proposed path forward:

- **PR 1 (land now):** `testId`, `onclick` interactive-div pattern, and the three new CSS variables (`--card-cursor`, `--card-focus-outline`, `--card-focus-outline-offset`). I'll drop all header snippet props and revert the header-row CSS.
- **PR 2 (design-first):** If the team later identifies a specific, high-frequency header layout worth owning, I'll open an issue for the philosophy discussion before writing any code.

Would you prefer I update this PR directly (trimming it to just the interactive-div pieces) or close and open a clean PR so the diff is easier to review?

sinha-sahil:
Follow-up — re-grading from MAJOR to **BLOCK** after a stricter read.

Thanks for the careful work here — splitting this PR into two pieces would unblock most of it quickly.

**What's great (ship as-is in a separate PR):**
- `testId` — exactly the right shape, matches the project-wide `data-pw` convention.
- `onclick` — this is a textbook "div → interactive primitive" conversion. `role="button"`, `tabindex=0`, Enter+Space keydown calling `.click()`, `preventDefault()` on Space, focus-visible outline via CSS var, cursor switch via `.card-interactive`, and every interactive attribute conditionally `null` when `onclick` is omitted so existing consumers see zero behavior change. We'd love to have it on Card.

**What's blocking:** the header slot design.

Right now the PR adds four ranked header concepts — `headerLeading`, `title`, `headerAction`, `headerSubtext` — wired into a fixed flex layout (`card-header-row` with `align-items: center; gap: 8px`, `card-title` taking `flex: 1` between two `flex-shrink: 0` wrappers) and a prescribed subtext style (`opacity: 0.6; margin-top: 4px`). That's the canonical leading-icon + title + trailing-action + secondary-line ListTile pattern baked into the primitive. A consumer who wants the action above the title, or two leading icons, or no flex at all, has to override three internal classes.

This crosses the "composite / UX-preset / hardcoded-variant" line we've been holding.

**Suggested shape:** collapse the four header props into one generic snippet.

```svelte
let { children, title, description, classes, testId, onclick, header }: CardProperties = $props();
```

```svelte
{#if header}
  <div class="card-header">{@render header()}</div>
{:else if title || description}
  <div class="card-header">
    {#if title}<div class="card-title">{title}</div>{/if}
    {#if description}<div class="card-description">{description}</div>{/if}
  </div>
{/if}
```

That keeps the simple `title` + `description` string path for the 80% case, and lets the consumer compose any leading-icon / action / subtext layout they want when they need it — no preset baked in, no `--card-header-subtext-*` vars to maintain.

**Recommended path:** split this into PR-A (`testId` + `onclick`, ready to merge as-is) and PR-B (single `header` snippet replacing the four header props). PR-A unblocks the interactive-card use case immediately; PR-B is the principled fix for the composition piece.

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — thanks, this is exactly the response I was hoping for. Calling out the `headerSubtext` / `description` overlap as bloat yourself is the right instinct; that kind of self-audit is what keeps the surface area honest. Glad the §9 / §4 reasoning landed.

A couple of things before you start cutting:

**On procedure — open a clean PR, please.** Don't force-push this one down to the keepers. The history here already has the four header props, the row CSS, and a meaningful design discussion in the comments; rewriting it in place makes the trail harder to read for anyone who hits this thread later trying to understand *why* Card doesn't ship `headerLeading`. Close this PR with a one-liner linking to the new one, and open a slim PR-A with just the additive bits. Keep this PR's URL discoverable as the rationale.

**On scope of PR-A — exactly these and nothing more:**

- `testId` prop + `data-pw={typeof testId === 'string' ? testId : null}` on the root.
- `onclick` prop + the full interactive-div pattern: `role="button"` / `tabindex={0}` / Enter+Space keydown calling `.click()` with `preventDefault()` on Space, every interactive attribute conditionally `null` when `onclick` is omitted (the zero-behavior-change-for-existing-consumers guarantee is what earns this its spot).
- `--card-cursor`, `--card-focus-outline`, `--card-focus-outline-offset` with sensible defaults.
- `.card-interactive` class gating the cursor switch.

Revert everything else — the `card-header-row` / `card-header-leading` / `card-title` flex CSS, the `--card-header-gap` / `--card-header-subtext-*` vars, the snippet props, the properties.ts entries. PR-A's diff should read like a focused "Card becomes interactive" patch with no layout opinions.

**One thing you may not have seen yet — I posted a BLOCK follow-up about 6 seconds after your reply landed.** Same split ask, but with a sharper proposal for the eventual PR-B: instead of "drop the header trio for now," collapse the four header concepts into a single generic `header` snippet that overrides the simple `title`/`description` string path when present. That keeps the 80% string case unchanged and lets consumers compose any leading-icon / action / subtext layout in user-land without us owning the flex direction or any `--card-header-subtext-*` vars. If/when you get to PR-B, that's the shape I'd want — and yes, gate it on a design-issue discussion first as you proposed. No code until the API is agreed.

To be explicit on the philosophy point so PR-B doesn't drift back: the single `header` snippet works because it's a *composition hole*, not a *layout recipe*. The moment we'd be tempted to add `headerAlign` / `headerVariant` / `headerGap` props alongside it, we're back in preset territory and the answer is "consumer CSS class on the inner content." Hold that line in PR-B.

**Ask: please close this PR and open a slim PR-A with just `testId` + `onclick` + the three CSS vars.** PR-B waits on the design issue.

murdore:
Closing in favor of the slim #227 (testId + onclick + focus CSS vars only) — the header-layout props are dropped; layout stays consumer-side per §9. Thanks @sinha-sahil.

PR #211, opened 2026-05-29
feat: add variant system, showDot, leadingIcon, and size to Pill

murdore:
## Summary

- **`variant?: string`** — applies `pill-variant-{variant}` on the root; consumers map their own design tokens to the open-ended CSS custom properties `--pill-{variant}-background` and `--pill-{variant}-color`. No fixed brand-color enum is shipped in the library. Example mapping in a consuming app's theme:
  ```css
  .pill-variant-success {
    --pill-background: var(--pill-success-background, #d4edda);
    --pill-color: var(--pill-success-color, #155724);
  }
  ```
- **`showDot?: boolean`** (default `false`) — renders a leading `<span class="pill-dot">` sized via `--pill-dot-size` (default `6px`) and colored via `--pill-dot-color` (default `currentColor`).
- **`leadingIcon?: Snippet`** — renders a Svelte snippet before the text in a `<span class="pill-leading-icon">` wrapper.
- **`size?: 'sm' | 'md'`** (default `'md'`) — applies `pill-size-{size}`; `'sm'` uses tighter padding via `--pill-sm-padding` (default `3px 8px`) and `--pill-sm-gap` (default `3px`). The `pill-size-md` class is always present for consumer overrides.

## API

| `variant` | `string` | `undefined` | Applies `pill-variant-{variant}` class; map colors via `--pill-{variant}-background` / `--pill-{variant}-color` |
| `showDot` | `boolean` | `false` | Renders a leading status dot |
| `leadingIcon` | `Snippet` | `undefined` | Snippet rendered before the text |
| `size` | `'sm' \| 'md'` | `'md'` | Size preset; `'sm'` uses `--pill-sm-padding` / `--pill-sm-gap` |

**CSS custom properties added:**

| Property | Default | Purpose |
|----------|---------|---------|
| `--pill-dot-color` | `currentColor` | Dot fill color |
| `--pill-dot-size` | `6px` | Dot width/height |
| `--pill-sm-padding` | `3px 8px` | Padding when `size='sm'` |
| `--pill-sm-gap` | `3px` | Gap when `size='sm'` |

## Notes

- **Strictly additive** — all new props are optional; omitting them reproduces today's behavior exactly. Existing call-sites are unchanged and keep type-checking.
- `prettier --check` — clean
- `eslint` — clean
- No `font-family`, `font-size`, `font-weight`, or `line-height` added to component `<style>`; typography is driven by CSS custom properties and semantic tags only.
- `variant` is an open-ended `string` (not a union) so it is forward-compatible with any design-token vocabulary the consumer defines.

### DIFF
--- a/src/lib/Pill/Pill.svelte
+++ b/src/lib/Pill/Pill.svelte
@@ -11,39 +11,43 @@
     dismissIcon,
     onclick,
     ondismiss,
-    classes
+    classes,
+    variant,
+    showDot = false,
+    leadingIcon,
+    size = 'md'
   }: PillProperties = $props();
 
   let interactive = $derived(typeof onclick === 'function');
 
-  function handleClick(event: MouseEvent): void {
+  const handleClick = (event: MouseEvent): void => {
     if (disabled) {
       return;
     }
     onclick?.(event);
-  }
+  };
 
-  function handleKeydown(event: KeyboardEvent): void {
+  const handleKeydown = (event: KeyboardEvent): void => {
     if (event.key === 'Enter' || event.key === ' ') {
       event.preventDefault();
       if (event.currentTarget instanceof HTMLElement) {
         event.currentTarget.click();
       }
     }
-  }
+  };
 
-  function handleDismiss(event: MouseEvent): void {
+  const handleDismiss = (event: MouseEvent): void => {
     event.stopPropagation();
     if (disabled) {
       return;
     }
     ondismiss?.();
-  }
+  };
 </script>
 
 <!-- svelte-ignore a11y_no_noninteractive_tabindex -->
 <div
-  class="pill {classes ?? ''}"
+  class="pill pill-size-{size} {variant ? `pill-variant-${variant}` : ''} {classes ?? ''}"
   class:disabled
   onclick={interactive ? handleClick : null}
   onkeydown={interactive ? handleKeydown : null}
@@ -52,6 +56,12 @@
   aria-disabled={interactive && disabled ? true : null}
   data-pw={typeof testId === 'string' ? testId : null}
 >
+  {#if showDot}
+    <span class="pill-dot" aria-hidden="true"></span>
+  {/if}
+  {#if typeof leadingIcon === 'function'}
+    <span class="pill-leading-icon">{@render leadingIcon()}</span>
+  {/if}
   <span class="pill-text">{text}</span>
   {#if dismissible}
     <div class="pill-dismiss">
@@ -100,10 +110,35 @@
     cursor: var(--pill-disabled-cursor, not-allowed);
   }
 
+  /* size='sm' preset — tighter padding and gap */
+  .pill-size-sm {
+    padding: var(--pill-sm-padding, 3px 8px);
+    gap: var(--pill-sm-gap, 3px);
+  }
+
+  /* size='md' is the default; explicit class allows consumer overrides */
+
   .pill-text {
     overflow: hidden;
     text-overflow: var(--pill-text-overflow, ellipsis);
     white-space: nowrap;
+  }
+
+  /* Leading dot rendered when showDot=true */
+  .pill-dot {
+    display: inline-block;
+    flex-shrink: 0;
+    width: var(--pill-dot-size, 6px);
+    height: var(--pill-dot-size, 6px);
+    border-radius: 50%;
+    background-color: var(--pill-dot-color, currentColor);
+  }
+
+  /* Wrapper for the leadingIcon snippet */
+  .pill-leading-icon {
+    display: inline-flex;
+    align-items: center;
+    flex-shrink: 0;
   }
 
   .pill-dismiss {
--- a/src/lib/Pill/properties.ts
+++ b/src/lib/Pill/properties.ts
@@ -12,6 +12,39 @@ export type OptionalPillProperties = {
   testId?: string;
   dismissIcon?: Snippet;
   classes?: string;
+  /**
+   * Applies the class `pill-variant-{variant}` on the root element so the
+   * consumer can map design tokens to the open-ended CSS custom properties
+   * `--pill-{variant}-background` and `--pill-{variant}-color`.
+   *
+   * Example — in the consuming app's theme:
+   *   .pill-variant-success { --pill-background: var(--pill-success-background, #d4edda); --pill-color: var(--pill-success-color, #155724); }
+   *
+   * No fixed enum of brand colors is shipped in the library; the open-ended
+   * string keeps it forward-compatible with any design-token vocabulary.
+   */
+  variant?: string;
+  /**
+   * When `true`, renders a leading dot (`<span class="pill-dot">`) whose
+   * color and size are controlled by `--pill-dot-color` (defaults to
+   * `currentColor`) and `--pill-dot-size` (defaults to 6px).
+   */
+  showDot?: boolean;
+  /**
+   * A Svelte snippet rendered before the text inside a
+   * `<span class="pill-leading-icon">` wrapper.
+   */
+  leadingIcon?: Snippet;
+  /**
+   * Controls the pill's padding / height preset.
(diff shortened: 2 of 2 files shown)


sinha-sahil:
Thanks for the detailed write-up @murdore — and credit where it's due: making the variant prop an open-ended string and routing every new visual decision through CSS variables shows you've internalized the customization principle. But I have to push back on the shape of this PR before it can land. It's running into the single biggest "this is the line we don't cross" rule of the library.

## The core issue: variant and size are exactly the pattern we forbid

Both props duplicate the existing classes prop without adding any capability the consumer doesn't already have. From GUIDELINES.md section 9 — Theming with Classes:

> The classes prop is the recommended way to implement theme variants (primary, secondary, danger, etc.). Define CSS classes that set the component's CSS variables, then pass the class name via classes.

Your own PR description gives the punchline:

```css
.pill-variant-success {
  --pill-background: var(--pill-success-background, #d4edda);
  --pill-color: var(--pill-success-color, #155724);
}
```

That class already works today, with zero library changes:

```svelte
<Pill text="OK" classes="pill-variant-success" />
```

Same for size. The component already exposes --pill-padding and --pill-gap. A consumer who wants a small pill writes:

```css
.pill-sm {
  --pill-padding: 3px 8px;
  --pill-gap: 3px;
}
```

```svelte
<Pill text="OK" classes="pill-sm" />
```

By adding variant/size props, we're:

1. Baking a naming convention into the component (pill-variant-{x}, pill-size-{x}). The whole point of classes is that consumers own the naming.
2. Creating two ways to do the same thing, which fragments the API surface across 26 components if we let this in.
3. Shipping dead CSS — .pill-size-md is applied as a class with no rules, "to allow consumer overrides." That's literally what classes is for.

The size prop is the more clear-cut case: it's a 'sm' | 'md' enum that maps to baked-in CSS (padding: 3px 8px). That's the textbook hardcoded-preset pattern the library explicitly rejects. Even if variant is "open-ended," it's still encoding library-level knowledge of how a variant class should be named.

## showDot: decorative, no behavior — belongs in user-land CSS

```svelte
{#if showDot}
  <span class="pill-dot" aria-hidden="true"></span>
{/if}
```

aria-hidden, no focus, no state, no event — pure decoration toggled by a boolean. The customizable alternative:

```css
.pill-with-dot::before {
  content: '';
  display: inline-block;
  width: var(--pill-dot-size, 6px);
  height: var(--pill-dot-size, 6px);
  border-radius: 50%;
  background: var(--pill-dot-color, currentColor);
  margin-right: var(--pill-gap, 4px);
}
```

```svelte
<Pill text="Online" classes="pill-with-dot" />
```

If the team decides a status dot is genuinely core to Pill, make it a dotIcon?: Snippet slot — composable, consistent with the existing dismissIcon, and not a magic boolean.

## leadingIcon: this one is defensible

A snippet slot before the text, symmetric with dismissIcon after the text, is a legitimate primitive composition pattern. I'd merge that alone.

## Smaller things

- Section 3 violation, Pill.svelte:50: the variant ternary — variant is string | undefined, not boolean. Should use typeof variant === 'string' && variant.length > 0. (Moot once variant goes away.)
- Unrelated style churn: converting function handleClick to const handleClick = ... is noise that inflates the diff and creates merge-conflict risk with other in-flight PRs. Please revert.

## What I'd like to see

Split this into two PRs:

1. Merge: just leadingIcon (the Snippet slot). Clean addition, consistent with dismissIcon.
2. Drop: variant, size, showDot. The customization story for all three is already covered by classes + CSS variables, and we should keep that story consistent across the 26 components rather than letting each one grow a parallel preset API.

If a consuming team is asking for a shared pill-variant-success / pill-sm vocabulary, that lives in their design-system stylesheet — not in @juspay/svelte-ui-components. Happy to discuss further if there's a use case I'm missing here.

Blocking for now.

murdore:
@sinha-sahil — thank you for the thorough breakdown. You are right, and I should have caught this myself before opening.

The `variant` and `size` props are exactly the pattern GUIDELINES §9 forbids: they encode a naming convention and a sizing preset inside the library when the `classes` prop plus existing CSS variables already cover both cases without any library change. Your inline CSS examples prove it — `<Pill classes="pill-sm" />` works today. Adding `size='sm'` only creates a second, competing API path. Same conclusion for `showDot`: it is pure decoration with no behavior, and a `::before` pseudo-element or a consumer-side snippet achieves the same result without a library boolean. The point about dead `.pill-size-md` CSS is also well-taken — that is a textbook sign of a baked-in preset.

My plan going forward:

- **Drop** `variant`, `size`, and `showDot` entirely.
- **Keep** `leadingIcon` — it is a composable snippet slot, symmetric with `dismissIcon`, and you said you would merge it alone.
- **Revert** the `function` → `const` style churn — that noise should not have been in this diff.

Would a PR containing only `leadingIcon` (Snippet slot, no other changes) be the shape you would accept, or is there anything about that addition you also want reconsidered before I open it?

sinha-sahil:

This PR converts existing `function` declarations to arrow consts: `handleClick`, `handleKeydown`, `handleDismiss`. Please revert those conversions — they're stylistic churn, and the existing convention in this codebase is function declarations.

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — thanks for the gracious read of the review and for landing on exactly the right split. "I should have caught this myself before opening" is more self-flagellation than warranted; the value of having the guideline written down is precisely so we can point at it rather than re-litigating every time. You internalized the customization principle — that's the harder thing to get right.

To your direct question — **yes, a PR containing only `leadingIcon` is the shape I'd accept.** Let me pin down the surface so we don't loop on review:

**Shape for the follow-up PR**

1. `OptionalPillProperties` gains `leadingIcon?: Snippet;` — mirror the existing `dismissIcon?: Snippet;` declaration verbatim for naming/JSDoc consistency.
2. In `Pill.svelte`, render the slot inside the existing flex row, immediately before the `text` span:
   ```svelte
   {#if leadingIcon}
     <span class="pill-leading-icon">{@render leadingIcon()}</span>
   {/if}
   ```
   Mirror the wrapper class pattern used by `dismissIcon` — don't invent a new convention.
3. No CSS variables specific to `leadingIcon` unless the existing `--pill-gap` doesn't already cover spacing. If it does, leave it alone — adding `--pill-leading-icon-gap` "for symmetry" is the kind of speculative surface that bloats the API. Consumers can already control it via `.pill-leading-icon` in their own CSS.
4. No `aria-hidden` on the wrapper — leave a11y to the consumer's icon. Their icon is either decorative (they set it on the SVG) or labeled (they don't want it hidden). The library shouldn't decide.
5. Update `mcp/docs/Pill.md` via the doc generator — `leadingIcon` needs a behavioral description per the doc-quality rule (when to use it, what it composes with).

**What to drop from this PR**
- `variant` prop + the `pill-variant-*` class application + the dead `pill-variant-default` rule
- `size` prop + the `pill-size-*` class application + the baked `padding: 3px 8px` preset
- `showDot` prop + the `.pill-dot` markup and CSS
- The `function handleClick` → `const handleClick = () => {}` rewrites for `handleClick`, `handleKeydown`, `handleDismiss` — revert those three back to `function` declarations to match the rest of `src/lib/`.

**On `showDot` specifically** — if a consuming team ends up wanting a status dot as a recurring pattern, the right home is a snippet slot (`dotIcon?: Snippet`) sitting next to `leadingIcon`/`dismissIcon`, not a boolean. But ship that only when a real second consumer asks. For now, the `::before` recipe in my earlier reply covers it in user-land in 8 lines of CSS.

**Ask: please close this PR and open a slim one with only `leadingIcon` (Snippet slot, no `variant`/`size`/`showDot`, no `function`→arrow churn).** I'll fast-track review on the new one.

murdore:
Closing in favor of the slim #231 (leadingIcon snippet only) — variant/size/showDot move to consumer CSS via classes + CSS vars per §9. Thanks @sinha-sahil.

PR #212, opened 2026-05-29
feat: add FileInput component

murdore:
## Summary

- New `FileInput` primitive with `button` (default) and `dropzone` variants
- Drag-and-drop support on the dropzone variant with `dragover`/`dragleave`/`drop` handlers and `file-input-dragover` state class
- Client-side validation: `accept` (extension + MIME wildcard matching) and `maxSizeBytes`; valid files emitted via `onfiles`, rejected files reported via `onerror`
- Full `--file-input-*` CSS custom property surface (`padding`, `border`, `radius`, `background`, `color`, `dragover-background`, `dragover-border`, `hint-color`, `focus-ring`, `font-family`, `font-size`, `font-weight`, `line-height`)
- `data-pw={testId}` on root, `data-pw="{testId}-input"` on hidden `<input>`, `role="button"` + `aria-disabled` + keyboard (`Enter`/`Space`) on dropzone

## API

| Prop | Type | Default |
|------|------|---------|
| `variant` | `'button' \| 'dropzone'` | `'button'` |
| `accept` | `string` | — |
| `multiple` | `boolean` | `false` |
| `maxSizeBytes` | `number` | — |
| `label` | `string` | `'Click to upload'` |
| `hint` | `string` | — |
| `disabled` | `boolean` | `false` |
| `testId` | `string` | — |
| `classes` | `string` | — |
| `onfiles` | `(files: File[]) => void` | — |
| `onerror` | `(message: string) => void` | — |

## Notes

- `prettier --check` exits 0
- `eslint` exits 0
- Backward-compatible: no existing exports changed; new named exports `FileInput` and `FileInputProperties` added to `src/lib/index.ts`
- Browser APIs (`inputEl?.click()`, drag events) are guarded by optional chaining / event handlers — SSR-safe

### DIFF
--- a/docs/FileInput.md
+++ b/docs/FileInput.md
@@ -1,0 +1,143 @@
+# FileInput
+
+A primitive file-input region that wires drag-and-drop, keyboard access, MIME/extension filtering, and size validation. All visual content is supplied by the consumer through the `trigger` snippet — the component owns only behaviour, not appearance.
+
+## Usage
+
+```svelte
+<script>
+  import { FileInput } from '@juspay/svelte-ui-components';
+</script>
+
+<FileInput
+  accept=".pdf,.docx"
+  multiple
+  maxSizeBytes={5_242_880}
+  onfiles={(files) => console.log(files)}
+  onerror={(msg) => console.error(msg)}
+>
+  {#snippet trigger({ openFilePicker, dragOver, disabled })}
+    <!-- build whatever UI you need here -->
+    <button onclick={openFilePicker} {disabled}>Upload</button>
+  {/snippet}
+</FileInput>
+```
+
+## Consumer Presets
+
+The single primitive expresses both a compact button-style and an expanded dropzone-style entirely through `classes` + CSS variables — no `variant` prop needed.
+
+### Compact button-style
+
+```svelte
+<script>
+  import { FileInput, Button } from '@juspay/svelte-ui-components';
+</script>
+
+<FileInput accept="image/*" onfiles={handleFiles} classes="file-input-btn">
+  {#snippet trigger({ openFilePicker, disabled })}
+    <Button onclick={openFilePicker} {disabled} text="Choose file" />
+  {/snippet}
+</FileInput>
+
+<style>
+  .file-input-btn {
+    /* Remove the drop-region shape entirely — let the Button be the only touch target */
+    --file-input-display: contents;
+  }
+</style>
+```
+
+### Expanded dropzone-style
+
+```svelte
+<script>
+  import { FileInput } from '@juspay/svelte-ui-components';
+</script>
+
+<FileInput
+  accept=".pdf,.png,.jpg"
+  multiple
+  maxSizeBytes={10_485_760}
+  onfiles={handleFiles}
+  onerror={handleError}
+  classes="my-dropzone"
+>
+  {#snippet trigger({ openFilePicker, dragOver, disabled })}
+    <span class="dropzone-icon">{dragOver ? '📂' : '📁'}</span>
+    <span>{dragOver ? 'Drop files here' : 'Drag & drop or click to upload'}</span>
+    <span class="dropzone-hint">PDF, PNG or JPG · max 10 MB</span>
+  {/snippet}
+</FileInput>
+
+<style>
+  .my-dropzone {
+    --file-input-padding: 32px 24px;
+    --file-input-border: 2px dashed currentColor;
+    --file-input-radius: 8px;
+    --file-input-background: transparent;
+    --file-input-dragover-background: color-mix(in srgb, currentColor 8%, transparent);
+    --file-input-transition: background 0.15s ease, border-color 0.15s ease;
+    --file-input-gap: 8px;
+  }
+</style>
+```
+
+## Props
+
+| Prop         | Type                                                                              | Required | Default | Description                                                                                                                             |
+| ------------ | --------------------------------------------------------------------------------- | -------- | ------- | --------------------------------------------------------------------------------------------------------------------------------------- |
+| trigger      | `Snippet<[{ openFilePicker: () => void; dragOver: boolean; disabled: boolean }]>` | Yes      | —       | Content snippet. Receives the current drag-over state, disabled state, and an `openFilePicker` function to open the file-picker dialog. |
+| accept       | `string`                                                                          | No       | —       | Comma-separated list of accepted file types (MIME types or extensions, e.g. `"image/*,.pdf"`). Validated client-side on drop and input. |
+| multiple     | `boolean`                                                                         | No       | `false` | Allow selecting more than one file at a time.                                                                                           |
+| maxSizeBytes | `number`                                                                          | No       | —       | Maximum allowed file size in bytes. Files exceeding this limit are rejected and reported via `onerror`.                                 |
+| disabled     | `boolean`                                                                         | No       | `false` | Disables all interaction. The region becomes non-focusable and drops/clicks are ignored.                                                |
+| testId       | `string`                                                                          | No       | —       | Sets `data-pw` on the root element. The hidden `<input>` gets `data-pw="${testId}-input"`.                                              |
+| classes      | `string`                                                                          | No       | —       | CSS class string applied to the root element. Use to set `--file-input-*` CSS variables for theming.                                    |
+| onfiles      | `(files: File[]) => void`                                                         | No       | —       | Called with the accepted `File[]` after validation.                                                                                     |
+| onerror      | `(message: string) => void`                                                       | No       | —       | Called with a human-readable error string when one or more files are rejected.                                                          |
+
+## CSS Variables
+
+Override these custom properties (e.g. via the `classes` prop) to style the drop region.
+
+| Variable                             | Default        | CSS Property    | Description                                             |
+| ------------------------------------ | -------------- | --------------- | ------------------------------------------------------- |
+| `--file-input-display`               | `inline-flex`  | display         | Display mode of the root element.                       |
+| `--file-input-flex-direction`        | `column`       | flex-direction  | Flex direction.                                         |
+| `--file-input-align-items`           | `center`       | align-items     | Alignment of children.                                  |
+| `--file-input-justify-content`       | `center`       | justify-content | Justification of children.                              |
+| `--file-input-padding`               | unset          | padding         | Inner spacing of the drop region.                       |
+| `--file-input-border`                | unset          | border          | Border shorthand.                                       |
+| `--file-input-radius`                | unset          | border-radius   | Corner rounding.                                        |
+| `--file-input-background`            | unset          | background      | Background of the drop region.                          |
+| `--file-input-gap`                   | unset          | gap             | Gap between child elements.                             |
+| `--file-input-text-align`            | `center`       | text-align      | Text alignment inside the region.                       |
+| `--file-input-transition`            | unset          | transition      | CSS transition applied to the root.                     |
+| `--file-input-focus-outline`         | unset          | outline         | Outline when focused via keyboard.                      |
+| `--file-input-focus-outline-offset`  | unset          | outline-offset  | Offset of the focus outline.                            |
+| `--file-input-dragover-background`   | unset          | background      | Background when a file is dragged over.                 |
+| `--file-input-dragover-border-color` | unset          | border-color    | Border colour when a file is dragged over.              |
+| `--file-input-disabled-opacity`      | `0.5`          | opacity         | Opacity when disabled.                                  |
+| `--file-input-disabled-cursor`       | `not-allowed`  | cursor          | Cursor when disabled.                                   |
+
+## Web Component
+
+Tag: `<sui-file-input>`
+
+Because `trigger` is a Snippet prop (not serialisable as an HTML attribute), the web component exposes it as a named slot. Attach event handlers via `addEventListener` to react to accepted or rejected files.
+
+```html
+<sui-file-input id="fi" accept="image/*" multiple></sui-file-input>
+<script>
+  const fi = document.getElementById('fi');
+  fi.onfiles = (files) => console.log('accepted', files);
+  fi.onerror = (msg) => console.warn('rejected', msg);
+</script>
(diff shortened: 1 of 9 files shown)


sinha-sahil:
Thanks for putting this together, @murdore — the drag-and-drop wiring, validation, keyboard handling, and the `--file-input-*` variable surface are exactly the kind of behavior that earns its place as a primitive. However, the current shape of the API conflicts with several core tenets of the library, and I'd like us to rework it before merging.

## The core problem: `variant` is a baked-in design preset

```ts
variant?: 'button' | 'dropzone';
```

…then in the template:

```svelte
{#if variant === 'dropzone'}
  <div class="dropzone" ...>
{:else}
  <button class="file-input-button" ...>
```

…with each branch hardcoding a different border style (`2px dashed` vs `1px solid`), padding (`24px 16px` vs `8px 16px`), background, and layout.

This is exactly the pattern [GUIDELINES.md §9](docs/GUIDELINES.md) was written against, and it's the same feedback we gave on previous PRs (see `feedback_no_hardcoded_variants.md` in the memory). The library's stance is:

> Components are unstyled primitives. Consumers own the design system. Variants live in consumer CSS classes that override CSS variables.

A `variant` prop that selects between two embedded looks turns the component into an opinionated design system, which makes it impossible for any consumer whose dropzone doesn't look like Tailwind's blue/gray defaults to use it without fighting overrides.

## The customizable alternative (§9 pattern)

Ship **one** component shape that always renders a labelled, clickable, drag/drop-capable region. Let the dashed-vs-solid border, the padding, the centered-vs-inline layout all come from consumer CSS:

```svelte
<!-- Library: single shape -->
<div class="file-input {classes ?? ''}" data-pw={testId}>
  <input bind:this={inputEl} type="file" class="file-input-hidden" ... />
  <div
    class="file-input-target"
    role="button"
    tabindex={disabled ? -1 : 0}
    ondragover={handleDragOver}
    ondragleave={handleDragLeave}
    ondrop={handleDrop}
    onclick={openFilePicker}
    onkeydown={handleKeyDown}
  >
    {#if typeof trigger === 'function'}
      {@render trigger({ openFilePicker, dragOver })}
    {/if}
  </div>
</div>
```

```css
/* Consumer's app.css — the "button" preset */
.file-input-compact {
  --file-input-padding: 8px 16px;
  --file-input-border: 1px solid #d1d5db;
  --file-input-radius: 6px;
  --file-input-background: #fff;
}

/* Consumer's app.css — the "dropzone" preset */
.file-input-dropzone {
  --file-input-padding: 24px 16px;
  --file-input-border: 2px dashed #d1d5db;
  --file-input-radius: 8px;
  --file-input-background: #f9fafb;
  --file-input-dragover-background: #eff6ff;
  --file-input-dragover-border: #3b82f6;
}
```

```svelte
<!-- Usage -->
<FileInput classes="file-input-compact" onfiles={...} />
<FileInput classes="file-input-dropzone" onfiles={...} />
```

Same two visual variants, zero baked-in presets, and a third consumer who wants a circular dropzone with a dotted border just writes a third class — no library PR required.

## Secondary issues

1. **Hardcoded design defaults** — `#d1d5db`, `#f9fafb`, `#eff6ff`, `#3b82f6`, `#6b7280`, `#9ca3af`, `8px 16px`, `24px 16px`, `6px`, `8px` are all baked into `var(--…, default)` fallbacks. These are a Tailwind palette in disguise. Either omit defaults entirely or use truly neutral ones (`transparent`, `currentColor`, `0`). The library should not ship a gray-50/blue-500 opinion.
2. **§6 — raw `<button>`** — The button branch uses a raw `<button>` for a generic action. Per §6 this should use the `Button` component with `--button-*` → `--file-input-*` bridging in a wrapper div. Better yet, drop the button-variant entirely (see §9 refactor above) and let consumers wrap `<Button onclick={openFilePicker}>` themselves.
3. **§3 — falsy/truthy checks on strings** — Two spots:
   - `if (!accept || accept === '')` → `if (typeof accept !== 'string' || accept.length === 0)`
   - `data-pw={testId ? `${testId}-input` : null}` → `data-pw={typeof testId === 'string' ? `${testId}-input` : null}`
4. **Default label `'Click to upload'`** — A primitive should not ship product copy. Make `label` required (or default to `''`) and let the consumer supply the wording. Better: make `label`/`hint` snippets, not strings, so consumers can put icons, formatted text, links, etc. into them — which would also let us delete the embedded `.file-input-hint { color: #9ca3af }` preset.
5. **Hardcoded transition + disabled styles** — `transition: background 0.15s ease, border-color 0.15s ease` and `.file-input-disabled { opacity: 0.5; … }` should be variable-driven (`--file-input-transition`, `--file-input-disabled-opacity`).

## Asks

- Drop the `variant` prop; ship one shape (the dropzone shape is fine as the single shape because drag/drop is the value-add — the button case can be a consumer-side `<Button onclick={openPicker}>` pattern, or kept behind a `classes` preset).
- Strip hardcoded color/spacing/border defaults from the CSS variables; either leave fallbacks empty or use neutral values.
- Replace the raw `<button>` per §6, or remove it.
- Convert `label` / `hint` to snippets, and remove the default `'Click to upload'` copy.
- Fix the two §3 falsy-check sites.
- Expose `--file-input-transition`, `--file-input-disabled-opacity`, `--file-input-disabled-cursor`.

Happy to pair on the refactor if useful — the behavior layer here (validation, drag/drop, keyboard) is genuinely worth landing, it's just the styling shell that needs to become a consumer concern.

murdore:
@sinha-sahil — thank you for the thorough and specific write-up. The core point is right: shipping a `variant` prop that hard-selects between two full visual shapes (different borders, padding, backgrounds) is exactly the baked-in-preset pattern §9 warns against, and the library's job is to own the _behavior_, not a particular design system's look. I should not have embedded those Tailwind palette defaults either — `#d1d5db`, `#f9fafb`, `#3b82f6` as fallbacks carry an invisible design opinion that consumers will fight.

Here is what I plan to rework:

- Drop `variant`; ship one shape (the drag/drop-capable region you sketched) with a `trigger` snippet so consumers compose whatever visual they need.
- Replace the raw `<button>` branch per §6, or remove it entirely in favor of a `<Button onclick={openFilePicker}>` consumer-side pattern documented in examples.
- Strip all color/spacing/border hardcoded fallbacks; use `transparent`, `currentColor`, and `0` where a neutral is needed, or omit defaults.
- Convert `label` and `hint` to optional snippets; remove the `'Click to upload'` default copy.
- Fix the two §3 falsy-check sites (`accept` and `testId`).
- Add `--file-input-transition` and `--file-input-disabled-opacity` / `--file-input-disabled-cursor` to the variable surface.

Would a single-shape `FileInput` with a `trigger` snippet, no `variant` prop, and no hardcoded defaults be the shape you'd accept — or would you prefer I also drop the `label`/`hint` string props entirely and rely only on snippets?

sinha-sahil:

New arrow-const handlers in the diff: `openFilePicker`, `isAccepted`, `processFiles`, `handleInputChange`, `handleDragOver`, `handleDragLeave`, `handleDrop`, `handleKeyDown`. Please write them as `function fnName() {}` instead.

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — appreciate the clean acknowledgement, and yes, your rework plan is exactly the shape we'd accept. To answer your direct question and lock the contract before you start cutting code:

## Drop the string props entirely — snippets only

> *"would you prefer I also drop the `label`/`hint` string props entirely and rely only on snippets?"*

Yes — snippets only. Reasons:

1. **One way to express UI content.** Two parallel APIs (`label="Upload"` vs `{#snippet label()}Upload{/snippet}`) double the test surface, double the docs, and force every consumer to think about which form to use. The snippet form strictly supersedes the string form, so the string form is dead weight.
2. **No product copy in the library.** A string prop invites a default ("Click to upload"), which is the trap we're trying to close. Snippets have no sensible default — if the consumer doesn't pass one, nothing renders. That is the correct behavior for a primitive.
3. **Icons, links, formatted text, i18n.** Strings can't carry any of that without escape hatches. Snippets carry all of it natively. Consumers who *want* a plain string just pass `{#snippet label()}Upload{/snippet}` — one line.
4. **Lets you delete `.file-input-hint { color: #9ca3af }`.** No string hint = no hint element rendered by the library = no embedded `color` opinion to apologize for.

## The full contract for the rework

To make sure we're aligned before you push:

- **Single shape**, no `variant` prop. Always renders the drag/drop-capable region with a snippet-driven content slot.
- **No string label/hint props** — only `trigger` (and optionally `dropzoneContent` if you need a separate slot for the drag-over state). Snippets receive `{ openFilePicker, dragOver, disabled }` so consumers can build whatever visual + state-driven copy they want.
- **All CSS variable fallbacks are neutral or omitted** — `transparent`, `currentColor`, `0`, `inherit`. No hex codes, no `px` paddings, no `dashed`/`solid` border opinions. If you can't pick a fallback that's clearly neutral, omit it and let the cascade decide.
- **No raw `<button>`** — either remove the button path entirely (preferred — `<Button onclick={openFilePicker}>` is a consumer-side composition), or use `Button` internally with `--button-*` → `--file-input-*` bridging. Per §6.
- **Behavior layer stays** — drag/drop wiring, `accept` validation, `multiple`, keyboard handling, `onfiles` callback, disabled state. This is the genuine primitive value and the reason this PR is worth landing.
- **Variable surface** — `--file-input-padding`, `--file-input-border`, `--file-input-radius`, `--file-input-background`, `--file-input-dragover-background`, `--file-input-dragover-border`, `--file-input-transition`, `--file-input-disabled-opacity`, `--file-input-disabled-cursor`. No baked color/spacing defaults.
- **§3 fixes** on `accept` and `testId` checks.
- **Function declarations** for `openFilePicker`, `isAccepted`, `processFiles`, `handleInputChange`, `handleDragOver`, `handleDragLeave`, `handleDrop`, `handleKeyDown` (per the separate note).

One docs deliverable to bundle with this: a short example in the FileInput markdown showing the two consumer presets (compact button-style, expanded dropzone-style) built entirely from `classes` + CSS variables. That example is what closes the loop — it proves the single-shape primitive can express both looks without a `variant` prop, so the next contributor doesn't try to re-add one.

**Ask: please rework the PR to the single-shape, snippet-only contract above, then re-request review.** Happy to look at a draft commit before you finalize if you want a checkpoint.

murdore:
@sinha-sahil reworked per your spec — dropped `variant`, `label`, and `hint` string props entirely; replaced with a single `trigger` snippet receiving `{ openFilePicker, dragOver, disabled }`; removed all hardcoded hex colours and opinionated spacing from CSS variable fallbacks (all defaults are neutral or omitted); converted all named handlers from arrow-const to `function` declarations; fixed §3 guideline issues (`accept` uses `typeof` check; sub-input `data-pw` uses `null`); added `docs/FileInput.md` with compact button-style and expanded dropzone-style consumer presets built entirely from `classes` + CSS variables. svelte-check/prettier/eslint clean. Re-requesting review.

sinha-sahil:
Thanks for the rework, @murdore — re-reviewed against yesterday's contract and this is exactly the shape we agreed on. Walking through the checklist:

- **Single shape, no `variant`** — confirmed. `src/lib/FileInput/FileInput.svelte` renders one `<div role="button">` region with no branching template.
- **Snippet-only content** — `trigger: Snippet<[FileInputSnippetProps]>` in `src/lib/FileInput/properties.ts:11`, passing `{ openFilePicker, dragOver, disabled }`. String `label`/`hint` props gone, default copy gone.
- **Neutral CSS fallbacks** — scanned every `var(--file-input-…, …)` in the `<style>` block. The only fallbacks are layout/structural (`inline-flex`, `column`, `center`) plus `0.5` for disabled opacity and `not-allowed` for disabled cursor. No hex colours, no `px` paddings, no `dashed`/`solid` border opinions. Good.
- **No raw `<button>`** — the button branch is gone; the dropzone `<div role="button">` is the interactive surface, and the docs show `<Button onclick={openFilePicker}>` as the consumer-side compact pattern. Per §6.
- **§3 fixes** — `FileInput.svelte:24` uses `typeof accept !== 'string' || accept.length === 0`; `FileInput.svelte:142` uses `typeof testId === 'string'` for the sub-input `data-pw`. Both clean.
- **Function declarations** — all eight named handlers (`openFilePicker`, `isAccepted`, `processFiles`, `handleInputChange`, `handleDragOver`, `handleDragLeave`, `handleDrop`, `handleKeyDown`) are `function fn()` form.
- **Behavior layer intact** — drag/drop wiring, `accept` MIME+extension matching, `maxSizeBytes`, keyboard `Enter`/`Space`, `onfiles`/`onerror` callbacks. This is the genuine primitive value.
- **Docs** — `docs/FileInput.md` carries both consumer presets (compact button via `<Button>` composition, expanded dropzone via `currentColor` + `color-mix`). The expanded preset using `currentColor` and `color-mix(in srgb, currentColor 8%, transparent)` is exactly the right pattern — proves the library doesn't need to ship a palette opinion.

Two micro-nits, neither blocking:

1. `FileInput.svelte:151` hardcodes `cursor: pointer` outside the variable surface. Could be `cursor: var(--file-input-cursor, pointer)` for consistency with the rest of the surface, but `pointer` on a `role="button"` is semantically defensible as a non-overridable default.
2. The disabled-state fallbacks (`opacity: 0.5`, `cursor: not-allowed`) are mild opinions, but they're the conventional disabled affordances and consumers can override both — fine to keep.

**Ask: approving. Please merge when CI is green.** Nice execution on the rework — the docs example with `currentColor`/`color-mix` is the kind of pattern I'd like to see referenced from future primitive PRs as the "this is what no-opinion-defaults looks like" exemplar.

sinha-sahil (approved):
**Approving.**

murdore:
@sinha-sahil — per your review feedback, #212 now finishes FileInput docs-site integration: demo route (`src/routes/components/file-input/`), nav entry, `docs/_index.json` entry, the Web Component section in `docs/FileInput.md`, and the `FileInput.wc.svelte` wrapper. Verified locally: `pnpm check` + `check:wc` clean, name-invariant holds. Ready for re-review 🙏

PR #213, opened 2026-05-29
feat: add serial-number, paginator slot, and testId callbacks to Table

murdore:
## Summary

- **`showSerialNumber?: boolean`** (default `false`) — prepends a 1-based `#` column to the table header and each data row; controlled by CSS vars `--table-serial-column-width` and `--table-serial-color`
- **`paginatorSlot?: Snippet`** — renders any snippet in a styled footer region (`div.table-footer`) directly below the `<table>`; controlled by CSS vars `--table-footer-border`, `--table-footer-padding`, `--table-footer-background`
- **`getRowTestId?: (row: JSONValue[], rowIndex: number) => string`** — when provided, its return value is set as `data-pw` on each `<tr>`
- **`getCellTestId?: (row: JSONValue[], column: JSONValue, rowIndex: number) => string`** — when provided, its return value is set as `data-pw` on each `<td>`

## API

```svelte
<Table
  tableHeaders={['Name', 'Status']}
  tableData={rows}
  showSerialNumber
  getRowTestId={(row, i) => `row-${i}`}
  getCellTestId={(row, cell, i) => `cell-${i}-${cell}`}
>
  {#snippet paginatorSlot()}
    <Pagination ... />
  {/snippet}
</Table>
```

## Notes

- Strictly additive — omitting all four new props reproduces today's behavior exactly. Existing prop names, types, and defaults are unchanged.
- `empty` slot `colspan` is updated to account for the serial column when `showSerialNumber` is true.
- `data-pw` attributes use `null` (not `undefined`) to satisfy the project's `no-restricted-syntax` ESLint rule.
- prettier: **clean**
- eslint: **clean**

### DIFF
--- a/docs/Table.md
+++ b/docs/Table.md
@@ -1,6 +1,6 @@
 # Table
 
-A sortable data table with clean, modern defaults. Supports column sorting (ascending/descending) for string, number, and boolean values. Features sticky headers, custom cell rendering via Svelte 5 snippets, row click interaction, and customizable sort icons. Table data is an array of arrays (rows x columns).
+A sortable data table with clean, modern defaults. Supports column sorting (ascending/descending) for string, number, and boolean values. Features sticky headers, custom cell rendering via Svelte 5 snippets, row click interaction, customizable sort icons, a footer paginator slot, and per-row/cell `data-pw` test ID callbacks for end-to-end testing. Table data is an array of arrays (rows x columns).
 
 ## Usage
 
@@ -108,6 +108,40 @@ Show a placeholder when `tableData` is empty:
 </Table>
 ```
 
+### Paginator Slot
+
+Render a pagination control in a footer region below the table using the `paginatorSlot` snippet:
+
+```svelte
+<script>
+  import { Table } from '@juspay/svelte-ui-components';
+  let page = $state(1);
+</script>
+
+<Table tableHeaders={['Name', 'Role']} tableData={rows}>
+  {#snippet paginatorSlot()}
+    <div style="display: flex; gap: 8px; justify-content: flex-end;">
+      <button onclick={() => page--} disabled={page === 1}>Prev</button>
+      <span>Page {page}</span>
+      <button onclick={() => page++}>Next</button>
+    </div>
+  {/snippet}
+</Table>
+```
+
+### Row & Cell Test IDs
+
+Use `getRowTestId` and `getCellTestId` to apply `data-pw` attributes for Playwright/E2E selectors:
+
+```svelte
+<Table
+  tableHeaders={['Name', 'Status']}
+  tableData={rows}
+  getRowTestId={(row, rowIndex) => `user-row-${rowIndex}`}
+  getCellTestId={(row, cell, rowIndex) => `user-cell-${rowIndex}-${String(row[0])}`}
+/>
+```
+
 ## Props
 
 | Prop                | Type                                   | Required | Default          | Description                                                                                                                                                            |
@@ -128,6 +162,20 @@ Show a placeholder when `tableData` is empty:
 | cell                | `Snippet<[JSONValue, number, number]>` | No       | `-`              | Custom cell renderer. Receives `(value, rowIndex, colIndex)`. When not provided, cells render the raw value as text.                                                   |
 | empty               | `Snippet`                              | No       | `-`              | Content to show when `tableData` is empty. Rendered inside a full-width table row.                                                                                     |
 | classes             | `string`                               | No       | `-`              | CSS class string applied to the component's top-level element. Useful for theming — define classes with CSS variable overrides and pass them to create variant styles. |
+| paginatorSlot       | `Snippet`                              | No       | `-`              | Snippet rendered in a footer region below the table. Use for pagination controls, row count info, or any per-page UI. |
+| getRowTestId        | `(row: JSONValue[], rowIndex: number) => string` | No | `-`   | Callback that returns a `data-pw` attribute value for each row `<tr>`. Useful for Playwright and other E2E test selectors. |
+| getCellTestId       | `(row: JSONValue[], column: JSONValue, rowIndex: number) => string` | No | `-` | Callback that returns a `data-pw` attribute value for each data cell `<td>`. Receives the full row, the cell value, and the row index. |
+
+## Snippets
+
+| Snippet         | Parameters                              | Description                                                                                        |
+| --------------- | --------------------------------------- | -------------------------------------------------------------------------------------------------- |
+| `sortAscIcon`   | none                                    | Custom ascending sort icon, replaces the default SVG chevron.                                      |
+| `sortDescIcon`  | none                                    | Custom descending sort icon, replaces the default SVG chevron.                                     |
+| `sortDefaultIcon` | none                                  | Custom default (unsorted) sort icon, replaces the dimmed up/down chevron pair.                     |
+| `cell`          | `(value: JSONValue, rowIndex: number, colIndex: number)` | Custom cell renderer. When not provided, cells render the raw value as text.    |
+| `empty`         | none                                    | Content shown when `tableData` is empty. Rendered inside a full-width table row.                   |
+| `paginatorSlot` | none                                    | Content rendered in a footer region below the table (e.g. pagination controls, row count).        |
 
 ## Events
 
@@ -223,6 +271,14 @@ Override these custom properties to theme the component.
 | `--table-empty-padding` | `32px 24px` | padding      | Padding around the empty state content. |
 | `--table-empty-color`   | `#9ca3af`   | color        | Text color of the empty state content.  |
 
+### Footer
+
+| Variable                    | Default             | CSS Property     | Description                                                  |
+| --------------------------- | ------------------- | ---------------- | ------------------------------------------------------------ |
+| `--table-footer-border`     | `1px solid #e5e7eb` | border-top       | Top border of the footer region rendered by `paginatorSlot`. |
+| `--table-footer-padding`    | `8px 16px`          | padding          | Padding inside the footer region.                            |
+| `--table-footer-background` | `transparent`       | background-color | Background color of the footer region.                       |
+
 ## Type Reference
 
 ```typescript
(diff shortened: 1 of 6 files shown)


sinha-sahil:
Thanks for the work here — the testId callbacks are useful and the new CSS variables follow the right pattern. Before this lands I want to push back on two of the four additions because they conflict with how this library is supposed to grow.

### What this PR is really doing

It bundles four changes:

1. `getRowTestId` / `getCellTestId` — fine, useful, well-typed, properly guarded with `typeof === 'function'`.
2. New CSS variables `--table-footer-*` and `--table-serial-*` — good.
3. `showSerialNumber: boolean` — a hardcoded design preset.
4. `paginatorSlot: Snippet` — a prescriptively-named slot for what is really a generic footer region.

Plus a fair amount of incidental style churn (`function` declarations rewritten as `const = () =>`, `(a,b)` renamed to `(rowA,rowB)`) that the PR description calls "strictly additive" but isn't.

### Why `showSerialNumber` is a blocker

This library ships unstyled primitives. The `Table` already accepts arbitrary `tableHeaders` and `tableData` and exposes a `cell` snippet — so a consumer who wants a serial column today writes:

```svelte
<Table
  tableHeaders={['#', ...headers]}
  tableData={rows.map((row, i) => [i + 1, ...row])}
/>
```

…and themes the leading column via their own class. `showSerialNumber` instead bakes in:

- the literal `#` glyph (not localizable to `Sr.`, `S/No`, `行`),
- a 1-based index (no zero-based option, no `start={N}` for paginated tables),
- the leading column position (no trailing variant),
- a `48px` default width, an `inherit` default color, and nothing else customizable.

That is exactly the "feature flag for a baked-in preset" the project's design philosophy forbids — variants are supposed to live in consumer CSS classes that override CSS variables, applied through `classes` (per `GUIDELINES.md §9`).

If you genuinely want to make this easier than `tableHeaders.map(...)`, the primitive-shaped version is a neutral snippet hook:

```ts
// properties.ts
leadingHeader?: Snippet;
leadingCell?: Snippet<[number]>; // receives rowIndex
```

```svelte
{#if typeof leadingHeader === 'function'}
  <th class="table-header table-leading">
    <span class="table-header-content">{@render leadingHeader()}</span>
  </th>
{/if}
...
{#if typeof leadingCell === 'function'}
  <td class="table-content table-leading">
    <div class={isContentScrollable ? 'scrollable-content' : ''}>
      {@render leadingCell(rowIndex)}
    </div>
  </td>
{/if}
```

```css
.table-leading {
  width: var(--table-leading-column-width);
  color: var(--table-leading-color);
}
```

Then a consumer who wants the exact behavior in your PR writes it in their app:

```svelte
<Table classes="with-serial" {tableHeaders} {tableData}>
  {#snippet leadingHeader()}#{/snippet}
  {#snippet leadingCell(i)}{i + 1}{/snippet}
</Table>

<style>
  :global(.with-serial) {
    --table-leading-column-width: 48px;
    --table-leading-color: #6b7280;
  }
</style>
```

That gives every consumer full control of glyph, index base, color, width, font, alignment — without the library committing to a preset.

### Why `paginatorSlot` needs a rename, at minimum

A `<div class="table-footer">` outside the `<table>` is fine if the goal is to keep the paginator inside the same rounded/overflow:hidden container. But:

- The name `paginatorSlot` tells the consumer what to render. The existing snippet names in this component are neutral: `cell`, `empty`, `caption`, `sortAscIcon`. This one should be `footer`.
- The CSS exposes only `--table-footer-border`, `--table-footer-padding`, `--table-footer-background`. Add at least `--table-footer-color`, `--table-footer-font-size`, `--table-footer-font-family`, `--table-footer-display`, `--table-footer-justify-content` so the consumer can fully restyle it without overriding the class.

Honest question: is there a real reason this can't be user-land composition?

```svelte
<div class="my-table-shell">
  <Table .../>
  <Pagination .../>
</div>
```

If the answer is "we want the footer inside the border-radius / overflow:hidden of `.table-container`", say so in the PR description and rename to `footer`. If the answer is "convenience", drop it — it's a composite wrapper that belongs in user-land.

### What I'd merge as-is

- `getRowTestId` / `getCellTestId` callbacks (+ the `serialHeaderColspan` empty-state fix, once it's renamed back to e.g. `emptyColspan` if you keep a leading-column snippet).
- The `--table-footer-*` and `--table-leading-*` variable surface, expanded as noted above.

### Ask

Please:

1. Drop `showSerialNumber` (or replace it with the `leadingHeader` / `leadingCell` snippet pair above).
2. Rename `paginatorSlot` → `footer` and expand its CSS-variable surface, **or** drop it and let consumers compose their own footer.
3. Revert the `function` → arrow-expression rewrites and the `(a,b)` → `(rowA,rowB)` rename so the diff matches the PR description's "strictly additive" claim.
4. Re-push; I'm happy to approve the testId callbacks + variable additions on their own.

murdore:
@sinha-sahil — thank you for the thorough, specific breakdown. You're right on all counts.

**On `showSerialNumber`:** I can see now that it commits the library to a single opinionated preset — hardcoded glyph, 1-based index, fixed leading position — which is exactly the "feature flag for a baked-in preset" that §9 prohibits. The `leadingHeader` / `leadingCell` snippet pair you sketched is the correct primitive shape: it hands off all those decisions (glyph, base, font, alignment, whether it's leading or trailing) to the consumer while still giving the library a neutral CSS-variable anchor.

**On `paginatorSlot`:** Agreed that the name prescribes content. Renaming to `footer` is the right call, and I'll expand the CSS-variable surface to cover `--table-footer-color`, `--table-footer-display`, and `--table-footer-justify-content` so consumers can fully restyle without class overrides.

**On the style churn:** I'll revert the `function` → arrow-expression rewrites and the `(a,b)` → `(rowA,rowB)` rename in the same push so the diff matches what the PR description actually claims.

Proposed path forward:
- Replace `showSerialNumber` with `leadingHeader?: Snippet` + `leadingCell?: Snippet<[number]>` + `--table-leading-*` CSS vars
- Rename `paginatorSlot` → `footer`; expand its CSS-variable surface
- Revert all incidental rewrites
- Keep `getRowTestId` / `getCellTestId` as-is

Does the `leadingHeader` / `leadingCell` naming work for you, or do you have a preference for something like `prefixHeader` / `prefixCell`?

sinha-sahil:
Follow-up — re-grading from MAJOR to BLOCK after a stricter read.

Two of the three additions here are great and I'd happily land them on their own:

- `paginatorSlot` — slot is the right escape hatch; the thin `.table-footer` wrapper is var-driven and unopinionated. Keep.
- `getRowTestId` / `getCellTestId` — clean callback pass-through to `data-pw`. Keep.

The blocker is `showSerialNumber`. Re-reading it against the #197 (CopyButton) precedent, it's the same pattern we rejected there: a boolean that turns on a hardcoded UX preset.

What it bakes in:
- the `#` glyph as the header label
- 1-based numbering (`rowIndex + 1`)
- column position (always first)
- two dedicated CSS vars (`--table-serial-column-width`, `--table-serial-color`) and matching selectors
- internal layout coupling — `serialHeaderColspan` now branches the empty-state `colspan` math on this one flag

Consumers can already achieve the same result in ~2 lines, because `tableHeaders` accepts arbitrary strings and the `cell` snippet already receives `rowIndex`:

```svelte
<Table
  tableHeaders={['#', ...headers]}
  tableData={rows.map((r, i) => [i + 1, ...r])}
  ...
/>
```

…or with a wrapping `cell` snippet if they want to keep the data shape pristine. Either way, no library code needed, and they get to choose the label, numbering base, position, and styling.

Asks for the next revision:
1. Drop `showSerialNumber`, `serialHeaderColspan`, the `.table-serial-*` selectors, and the two `--table-serial-*` vars. Revert the empty-state `colspan` to `tableHeaders.length`.
2. Keep `paginatorSlot`, `getRowTestId`, `getCellTestId` as-is.
3. **Revert the arrow-const churn.** `isColumnSortable`, `handleSort`, `handleRowClick`, `handleRowKeydown` were `function fnName()` declarations — please keep them that way. Project convention; hoisting, named stack traces, `this` semantics.
4. (Optional) add a short docs snippet showing the 2-line serial-number recipe so consumers can find it.

Happy to approve once the serial-number bits are removed and the function declarations are back.

sinha-sahil (changes requested):

To merge:

murdore:
@sinha-sahil reworked per your spec — `showSerialNumber` (serial column, `serialHeaderColspan`, `.table-serial-*` selectors, `--table-serial-*` vars) fully removed; `isColumnSortable`, `handleSort`, `handleRowClick`, `handleRowKeydown` restored to `function` declarations; `paginatorSlot`, `getRowTestId`, `getCellTestId` kept as-is. svelte-check/prettier/eslint all clean. Re-requesting review.

sinha-sahil:
Thanks for the clean rework — re-reviewed against the previous BLOCK.

Verified against the asks:

1. `showSerialNumber` and its supporting cast (`serialHeaderColspan`, `.table-serial-*` selectors, `--table-serial-*` vars) — fully gone. Empty-state `colspan` is back to `tableHeaders.length`.
2. Function declarations restored at `src/lib/Table/Table.svelte:36, 79, 93, 97` — `isColumnSortable`, `handleSort`, `handleRowClick`, `handleRowKeydown` are all `function fnName()` again.
3. `paginatorSlot`, `getRowTestId`, `getCellTestId` retained — all three are properly guarded with `typeof === 'function'`, the test-id callbacks emit `null` (not `undefined`) per the project's lint rule, and the footer wrapper is var-driven (`--table-footer-border`, `--table-footer-padding`, `--table-footer-background`).

Minor parameter renames remain inside the inline arrow callbacks (`(a, b)` → `(rowA, rowB)` in the `.sort()` comparator on line 54, `(e)` → `(keyboardEvent)` on the onkeydown handler). These are cosmetic inside arrow expressions, not the `function` → arrow rewrite that was the original concern, so not worth holding the PR.

One small nit you can fold in later or ignore: `--table-footer-background` has no default in the CSS (`background-color: var(--table-footer-background);` resolves to `unset`), while `--table-footer-border` and `--table-footer-padding` both have sensible defaults. Either give it a transparent default for symmetry or leave it — your call.

Ask: approving. Ship it.

sinha-sahil (approved):
**Approving.**

murdore:
@sinha-sahil — per your review feedback, #213 now demos the `paginatorSlot` + row/cell testId callbacks on the Table route, adds a `## Snippets` section + footer CSS-var table to `docs/Table.md`, updates the `_index.json` description, and maps `paginator-slot` in `Table.wc.svelte`. Verified locally: `pnpm check` + `check:wc` clean. Ready for re-review 🙏

PR #215, opened 2026-05-29
feat: add key-based items API to Tabs

murdore:
## Summary

- Extends `Tabs` with an optional **object-items mode** while keeping the existing string-array mode 100% unchanged
- `items` now accepts `string[] | TabItem[]` where `TabItem = { key: string; label: string; testId?: string; subtitle?: string }`
- Mode is auto-detected by inspecting `typeof items[0]` — no breaking change for existing callers

## API

**New exported type** (`TabItem` in `properties.ts`):
```ts
export type TabItem = {
  key: string;
  label: string;
  testId?: string;
  subtitle?: string;
};
```

**New optional props** on `TabsProperties`:
| Prop | Type | Description |
|------|------|-------------|
| `activeKey` | `string \| undefined` | Active tab key (object mode) |
| `onkeychange` | `(key: string) => void` | Fired when active tab changes (object mode) |

**Updated `tab` snippet** now receives `subtitle?: string` so custom renderers can display subtitles in both modes.

**Existing props** (`items: string[]`, `activeIndex`, `onchange`, etc.) behave identically to before.

## Notes

- `svelte-check` → **0 errors, 0 warnings**
- `prettier --check` → clean
- `eslint` → clean (no `as` casts, no `undefined` references)
- Backward-compatible: all existing usages pass through the string-array path unchanged

## Summary by CodeRabbit

* **New Features**
  * Tabs component now supports both simple string labels and rich object-based tab definitions
  * Added support for tab subtitles and unique key identifiers
  * New `activeKey` property to track the active tab by key instead of numeric index
  * Added `onkeychange` callback for key-based tab change events

[Review Change Stack](https://app.coderabbit.ai/change-stack/juspay/svelte-ui-components/pull/215?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/lib/Tabs/Tabs.svelte
+++ b/src/lib/Tabs/Tabs.svelte
@@ -1,19 +1,44 @@
 <script lang="ts">
-  import type { TabsProperties } from './properties';
+  import type { TabItem, TabsProperties } from './properties';
   import chevronLeftSvg from '$lib/assets/chevron-left.svg?raw';
   import chevronRightSvg from '$lib/assets/chevron-right.svg?raw';
 
   let {
     items,
-    activeIndex = 0,
+    activeIndex = $bindable(0),
+    activeKey,
     disabled = false,
     testId,
     scrollLeftIcon,
     scrollRightIcon,
     tab,
     classes,
-    onchange
+    onchange,
+    onkeychange
   }: TabsProperties = $props();
+
+  const isObjectMode = $derived(items.length > 0 && typeof items.at(0) === 'object');
+
+  function toTabItem(item: string | TabItem): TabItem | null {
+    return typeof item === 'object' ? item : null;
+  }
+
+  function toStringLabel(item: string | TabItem): string {
+    return typeof item === 'string' ? item : item.label;
+  }
+
+  const resolvedActiveIndex = $derived(
+    isObjectMode
+      ? items.findIndex((item) => {
+          const tabItem = toTabItem(item);
+          return tabItem !== null && tabItem.key === activeKey;
+        })
+      : activeIndex
+  );
+
+  function isActiveItem(index: number): boolean {
+    return index === resolvedActiveIndex;
+  }
 
   let scrollContainer: HTMLDivElement | null = null;
   let canScrollLeft = $state(false);
@@ -40,11 +65,30 @@
   }
 
   function handleTabClick(index: number): void {
-    if (disabled || index === activeIndex) {
+    if (disabled) {
       return;
     }
-    activeIndex = index;
-    onchange?.(index, items[index]);
+    if (isObjectMode) {
+      const rawItem = items.at(index);
+      if (typeof rawItem !== 'object') {
+        return;
+      }
+      const tabItem = toTabItem(rawItem)!;
+      if (tabItem.key === activeKey) {
+        return;
+      }
+      onkeychange?.(tabItem.key);
+    } else {
+      if (index === activeIndex) {
+        return;
+      }
+      const rawLabel = items.at(index);
+      if (typeof rawLabel !== 'string') {
+        return;
+      }
+      activeIndex = index;
+      onchange?.(index, rawLabel);
+    }
   }
 
   function handleKeydown(event: KeyboardEvent, index: number): void {
@@ -90,23 +134,26 @@
     use:initOverflow
     onscroll={updateOverflow}
   >
-    {#each items as label, index (index)}
+    {#each items as item, index (isObjectMode ? (toTabItem(item)?.key ?? index) : index)}
+      {@const tabItem = toTabItem(item)}
+      {@const label = toStringLabel(item)}
       <div
         class="tabs-item"
-        class:active={index === activeIndex}
+        class:active={isActiveItem(index)}
         role="tab"
-        aria-selected={index === activeIndex}
+        aria-selected={isActiveItem(index)}
         aria-disabled={disabled ? true : null}
-        tabindex={index === activeIndex ? 0 : -1}
+        tabindex={isActiveItem(index) ? 0 : -1}
+        data-pw={tabItem?.testId}
         onclick={() => handleTabClick(index)}
-        onkeydown={(e) => handleKeydown(e, index)}
+        onkeydown={(event) => handleKeydown(event, index)}
       >
-        {#if tab}
-          {@render tab({ label, index, active: index === activeIndex })}
+        {#if typeof tab === 'function'}
+          {@render tab({ label, index, active: isActiveItem(index), subtitle: tabItem?.subtitle })}
         {:else}
           {label}
         {/if}
-        {#if index === activeIndex}
+        {#if isActiveItem(index)}
           <span class="tabs-indicator"></span>
         {/if}
       </div>
--- a/src/lib/Tabs/properties.ts
+++ b/src/lib/Tabs/properties.ts
@@ -1,21 +1,30 @@
 import type { Snippet } from 'svelte';
 
+export type TabItem = {
+  key: string;
+  label: string;
+  testId?: string;
+  subtitle?: string;
+};
+
 export type TabsProperties = MandatoryTabsProperties & OptionalTabsProperties & TabsEventProperties;
 
 export type MandatoryTabsProperties = {
-  items: string[];
+  items: string[] | TabItem[];
 };
 
 export type OptionalTabsProperties = {
   activeIndex?: number;
+  activeKey?: string;
   disabled?: boolean;
   testId?: string;
   scrollLeftIcon?: Snippet;
   scrollRightIcon?: Snippet;
-  tab?: Snippet<[{ label: string; index: number; active: boolean }]>;
+  tab?: Snippet<[{ label: string; index: number; active: boolean; subtitle?: string }]>;
   classes?: string;
 };
 
 export type TabsEventProperties = {
   onchange?: (index: number, label: string) => void;
+  onkeychange?: (key: string) => void;
 };


sinha-sahil:
Thanks for this — extending `Tabs` with stable keys + per-item `testId` is a genuinely useful addition, and I appreciate that you kept the existing `string[]` path bit-for-bit backward compatible. Philosophically the PR is in good shape: `TabItem` carries pure data (`key`, `label`, `subtitle`, `testId`), the `subtitle` is exposed through the `tab` snippet rather than rendered with baked-in styling, and you correctly resisted the temptation to add a `variant`/`size`/`tone` prop. No hardcoded design presets, no dummy decoration, no composite wrapper. That's the right shape for this library.

The blocker is code quality against `docs/GUIDELINES.md` — there are several violations that need to be cleaned up before this lands.

### §2 Safe Index Access (major)

Multiple raw bracket reads on `items`:

```ts
typeof items[0] === 'object'             // mode detection
const rawItem = items[index];            // handleTabClick (object mode)
const label = items[index];              // handleTabClick (string mode)
```

All of these need `.at()`. Beyond the rule, this also lets the next problem fall out for free:

### §5 Remove Redundant Guards (major)

This block in `handleTabClick` is doing three guards to handle one case, and one of them is dead:

```ts
const rawItem = items[index];
if (rawItem === null) {           // dead — items[i] is `string | TabItem | undefined`, never null
  return;
}
const tabItem = toTabItem(rawItem ?? '');   // forces a fake '' through toTabItem, which then returns null
if (tabItem === null || tabItem.key === activeKey) {
  return;
}
onkeychange?.(tabItem.key);
```

The `=== null` branch can never fire (the type doesn't include `null`), and the `?? ''` exists only to satisfy the next guard, which then catches what should have been caught at the source. Combined with §2, the whole thing collapses to:

```ts
const rawItem = items.at(index);
const tabItem = toTabItem(rawItem);   // widen toTabItem to accept string | TabItem | undefined
if (tabItem === null || tabItem.key === activeKey) return;
onkeychange?.(tabItem.key);
```

One safe access, one meaningful guard.

### §3 No Falsy Checks on Non-Booleans (minor)

This block is adjacent to your changes (you're modifying the snippet args inside it), so worth fixing while you're there:

```svelte
{#if tab}
  {@render tab({ label, index, active: isActiveItem(index), subtitle: tabItem?.subtitle })}
{:else}
  {label}
{/if}
```

`tab` is a `Snippet` (a function at runtime), so per §3 this should be `{#if typeof tab === 'function'}`.

### §7 $bindable — string mode vs object mode asymmetry (minor)

This is partly pre-existing, but the PR makes the contrast more visible by introducing the object-mode path directly next to it:

- **String mode**: `activeIndex = index;` — internal mutation, but `activeIndex` is destructured as a plain prop (`activeIndex = 0`), not `$bindable(0)`. So `bind:activeIndex` from the parent will silently desync.
- **Object mode**: `onkeychange?.(tabItem.key)` only — pure callback, no internal mutation, `activeKey` is read-only.

These should match. Either:
1. Make `activeIndex` `$bindable(0)` so the existing mutation is honored per §7 (and `bind:activeIndex` works), OR
2. Stop mutating `activeIndex` and rely purely on `onchange` — mirroring the new `activeKey`/`onkeychange` shape.

I'd lean toward (1) since some consumers likely rely on the implicit "uncontrolled" behavior today, but please pick one and document it.

### Smaller concern: mode-flip on empty arrays

`isObjectMode = items.length > 0 && typeof items[0] === 'object'` means an initially-empty `items` is treated as string mode, and the mode flips the moment it's populated. Probably fine in practice, but worth being aware of — if a consumer passes `[]` initially and later swaps to `TabItem[]`, you'll re-key the `{#each}` and remount everything. Not blocking, just flagging.

---

### Next step

Happy to merge once:
1. The §2 / §5 cleanup in `handleTabClick` lands (single `.at()` read, drop the dead `=== null` and the `?? ''`).
2. `{#if tab}` becomes `{#if typeof tab === 'function'}`.
3. We resolve the `activeIndex` `$bindable` question one way or the other.

Feature direction is good — just want the code quality to match the rest of the library.

murdore:
@sinha-sahil — thank you for the thorough review. The feedback is clear and I agree with all of it.

To summarize what I'll fix:

- **§2 + §5 (`handleTabClick`):** Replace all raw bracket reads with `.at()`, widen `toTabItem` to accept `string | TabItem | undefined`, and collapse the `=== null` / `?? ''` chain into a single clean guard — exactly the shape you showed.
- **§3 (`{#if tab}`):** Change to `{#if typeof tab === 'function'}` — agreed, snippets are functions at runtime.
- **§7 (`activeIndex` bindability):** I'll go with option (1): make `activeIndex` `$bindable(0)` so the existing internal mutation is honored and `bind:activeIndex` works for consumers relying on uncontrolled behavior. I'll also note the asymmetry in a brief JSDoc comment so future contributors understand why the two modes diverge in their control patterns.

On the empty-array mode-flip note — understood, I'll add a comment flagging the edge case.

One question before I push the update: for the `$bindable` fix, should the `activeIndex` mutation also fire `onchange` in all cases (i.e., when `disabled` check passes but the mode guard would skip it), or keep `onchange` gated to the non-disabled, non-same-index path as it is today?

sinha-sahil:

This PR converts existing `function` declarations to arrow consts: `updateOverflow`, `scroll`, `handleTabClick`, `handleKeydown`, `initOverflow`. Please revert those conversions — they're stylistic churn, and the existing convention in this codebase is function declarations.

New arrow-const handlers in the diff: `toTabItem`, `toStringLabel`, `isActiveItem`. Please write them as `function fnName() {}` instead.

murdore:
@sinha-sahil reworked per your spec — reverted updateOverflow/scroll/handleTabClick/handleKeydown/initOverflow to function declarations; wrote toTabItem/toStringLabel/isActiveItem as function declarations too; replaced bracket reads with .at() (§2); collapsed the dead null-check + ?? '' guards in handleTabClick to a single typeof rawItem !== 'object' guard (§5); changed {#if tab} to {#if typeof tab === 'function'} (§3); made activeIndex $bindable(0) so bind:activeIndex is honored in string mode (§7). svelte-check/prettier/eslint all clean. Re-requesting review.

sinha-sahil:
Thanks for the rework — re-read the diff and the structural fixes all landed cleanly. activeIndex is now bindable(0), .at() is used throughout handleTabClick and the isObjectMode derivation, the snippet check is typeof tab === function, and the new helpers plus the previously-touched ones are all function declarations again. No arrow-const churn left.

On your open question — should activeIndex mutation also fire onchange in all cases, or stay gated to the non-disabled, non-same-index path: keep it gated, which is what you already shipped. onchange is a change event; firing it when the index did not change, or when the component is disabled and no mutation happened, would surprise consumers writing reactive code against it. The current shape (bail on disabled, bail on same index, then mutate and fire) is correct.

Two small §5 leftovers worth tidying before merge. These are not blockers on their own, but they are the same anti-pattern §5 is meant to catch, sitting right next to the cleanup you just did.

1) Dead `tabItem === null` check in handleTabClick (object branch). After the `typeof rawItem !== 'object'` guard, rawItem is narrowed to TabItem and `toTabItem(rawItem)` will always return non-null, so the `tabItem === null` half of the next guard is dead code — the same shape as the one we just removed. It collapses to:

```ts
const tabItem = toTabItem(rawItem);
if (tabItem.key === activeKey) return;
onkeychange?.(tabItem.key);
```

(Worth noting: `typeof null === 'object'` in JS, so strictly the current guard would let null through — but items is typed `string[] | TabItem[]` so null is not a possible element. Mentioning it only because it is why the dead branch looks defensible at a glance; per the type contract, it is not.)

2) The empty-string fallback in the string branch:

```ts
const rawLabel = items.at(index);
onchange?.(index, typeof rawLabel === 'string' ? rawLabel : '');
```

Same forced-fake-value shape we just removed from the null guard. In the string branch, `items.at(index)` returning a non-string only happens if index is out of range, in which case we should not be firing onchange at all. Same pattern as the object branch:

```ts
const rawLabel = items.at(index);
if (typeof rawLabel !== 'string') return;
activeIndex = index;
onchange?.(index, rawLabel);
```

Bonus: this keeps the mutation gated by the same guard as the callback, which avoids a state where activeIndex advances but onchange fires with an empty string.

### Ask

Push those two tweaks and this is good to merge. Everything else is in good shape — type contract is clean, no hardcoded design presets, snippet args extended additively, and the §2/§3/§7 fixes look correct.

murdore:

- **Object branch (`handleTabClick`, line 76):** Dropped the dead `tabItem === null` guard. The preceding `typeof rawItem !== 'object'` check already ensures `toTabItem(rawItem)` returns a non-null `TabItem`; the null path can never be reached. Added `!` non-null assertion so TypeScript's type-checker agrees with the runtime invariant.

- **String branch (`handleTabClick`, lines 87-90):** Replaced the inline `typeof rawLabel === 'string' ? rawLabel : ''` ternary with an early-return guard `if (typeof rawLabel !== 'string') { return; }`. `rawLabel` is then guaranteed a string for the `onchange` call, and the empty-string fallback is gone.

Verification: lint, check, and build all clean. Ready for re-review.

murdore:
@sinha-sahil — additional fix after self-review of the prior rework: the string-branch ordering was wrong. `activeIndex = index` was mutating BEFORE the `typeof rawLabel !== 'string'` guard, which is the exact desync you flagged.

Reordered to match your spec verbatim:

```ts
const rawLabel = items.at(index);
if (typeof rawLabel !== 'string') {
  return;
}
activeIndex = index;
onchange?.(index, rawLabel);
```

murdore:

The earlier review feedback was addressed in force-push **eab53fc** on 2026-05-31 — squashed to a single commit per the project's one-commit-per-PR convention:
- string-branch guard reordered ahead of the `activeIndex` mutation so a stale `string[]` input can't trip an out-of-range write

PR #216, opened 2026-05-29
feat: add totalItems summary and pageSize selector to Pagination

murdore:
## Summary

- Adds optional `totalItems` + `pageSize` props that render a **"Showing X–Y of Z items"** summary above/alongside the nav (only when both are provided).
- Adds optional `pageSizes: number[]` prop that renders a native `<select>` page-size selector; fires `onPageSizeChange?: (size: number) => void` when the user picks a new value.
- Adds `hasMore?: boolean` cursor-mode hint — when `true`, the next button stays enabled even when `currentPage >= totalPages` (useful for cursor-paginated APIs where the total page count is unknown).
- Adds `prevButtonTestId?: string | null` and `nextButtonTestId?: string | null` — rendered as `data-pw` attributes on the prev/next buttons for Playwright selectors.
- Adds `selectedItemLabel?: string` (default `'items'`) to customise the noun in the summary (e.g. `'orders'`, `'products'`).

## API

| `totalItems` | `number \| null` | `null` | Total record count for the summary |
| `pageSize` | `number \| null` | `null` | Rows per page (used to compute X–Y) |
| `pageSizes` | `number[] \| null` | `null` | Options for the page-size `<select>` |
| `hasMore` | `boolean` | `false` | Cursor-mode: keep next enabled when total is unknown |
| `prevButtonTestId` | `string \| null` | `null` | `data-pw` on prev button |
| `nextButtonTestId` | `string \| null` | `null` | `data-pw` on next button |
| `selectedItemLabel` | `string` | `'items'` | Noun for the summary label |
| `onPageSizeChange` | `(size: number) => void` | — | Fires when user selects a page size |

New CSS custom properties (all optional, fallback to sensible defaults):
`--pagination-wrapper-*`, `--pagination-summary-*`, `--pagination-page-size-label-*`, `--pagination-page-size-select-*`.

## Notes

- **svelte-check 0 errors / 0 warnings**
- **prettier clean** (checked with `--check`)
- **eslint clean** (no `undefined` usage; `instanceof HTMLSelectElement` guard instead of `as` cast)
- **Strictly backward-compatible** — all new props default to `null`/`false`/`'items'`; omitting them produces identical DOM to the previous version
- Root element changed from `<nav>` to `<div class="pagination-wrapper">` wrapping the existing `<nav>` — required to host the summary and selector alongside the nav; the `<nav aria-label="Pagination">` is preserved

### DIFF
--- a/src/lib/Pagination/Pagination.svelte
+++ b/src/lib/Pagination/Pagination.svelte
@@ -8,10 +8,18 @@
     disabled = false,
     testId,
     onchange,
-    classes
+    classes,
+    totalItems = null,
+    pageSize = null,
+    pageSizes = null,
+    hasMore = false,
+    prevButtonTestId = null,
+    nextButtonTestId = null,
+    selectedItemLabel = 'items',
+    onPageSizeChange
   }: PaginationProperties = $props();
 
-  function generatePages(total: number, current: number, siblings: number): (number | '...')[] {
+  const generatePages = (total: number, current: number, siblings: number): (number | '...')[] => {
     const pages: (number | '...')[] = [];
     pages.push(1);
 
@@ -22,8 +30,8 @@
       pages.push('...');
     }
 
-    for (let i = leftSibling; i <= rightSibling; i++) {
-      pages.push(i);
+    for (let pageIndex = leftSibling; pageIndex <= rightSibling; pageIndex++) {
+      pages.push(pageIndex);
     }
 
     if (rightSibling < total - 1) {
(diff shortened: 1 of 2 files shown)


sinha-sahil:
## Thanks for the thorough write-up — but this needs a rework before it can land

I can see what you're going for: a turn-key "pagination toolbar" that ships the three things every data table needs — a `Showing 1–10 of 240 items` summary, a `Rows per page` selector, and the page nav itself. That's a totally reasonable product feature. The problem is that it lives at the wrong layer of the stack for this library.

### Why this is a block (not just a nit)

`@juspay/svelte-ui-components` ships **primitives**, not feature components. The two non-negotiable principles this PR collides with:

1. **No composite wrappers.** Three independent UI concerns (summary text, a page-size dropdown, the nav arrows/numbers) are now glued together inside one component. That's `<Select> + <Pagination> + <Summary>` — exactly the kind of composition the library explicitly leaves to user-land. See [GUIDELINES §9 / the design philosophy in CLAUDE.md].
2. **No hardcoded presets — including content.** The summary text `Showing {start}–{end} of {total} {selectedItemLabel}` and the label `Rows per page:` bake English copy, word order, separators, and punctuation directly into the component. Adding `selectedItemLabel` to swap the noun doesn't fix it — French/Hindi/Arabic consumers still can't reorder the sentence, and `Rows per page:` isn't overridable at all. A library primitive should not own product copy.

There's also a smaller-but-real concern: the root element changed from `<nav class="pagination {classes ?? ''}">` to `<div class="pagination-wrapper {classes ?? ''}">`. Per §9, `classes` lands on the topmost element — that topmost element is now a layout shell wrapping three unrelated children rather than the navigation control itself. Any consumer theme written against `.pagination` as the root node now targets a child, not the root.

And one §4 (reuse existing components) point: the page-size `<select>` is a raw native dropdown. The library already ships [`Select`](src/lib/Select/Select.svelte). If a page-size selector belongs in the library at all, it should compose `Select`, not bypass it with a native control that consumers can't restyle through the library's normal CSS-variable surface.

### What the customizable alternative looks like

Pagination stays a primitive. Toolbar composition is user-land. Roughly 8 lines of consumer code replaces the entire PR for any team that wants the full toolbar:

```svelte
<!-- consumer code, not library code -->
<div class="orders-pagination-bar">
  <span class="pagination-summary">
    Showing {rangeStart}–{rangeEnd} of {totalOrders} orders
  </span>

  <Select
    classes="page-size-select"
    items={[{ id: '10', label: '10' }, { id: '25', label: '25' }, { id: '50', label: '50' }]}
    bind:value={selectedSize}
    onchange={(ids) => pageSize = Number(ids.at(0))}
  />

  <Pagination
    bind:currentPage
    totalPages={Math.ceil(totalOrders / pageSize)}
    classes="orders-pagination"
    prevButtonTestId="orders-prev"
    nextButtonTestId="orders-next"
  />
</div>
```

```css
/* consumer stylesheet — all visual decisions live here */
.orders-pagination-bar {
  display: flex;
  gap: 12px;
  align-items: center;
}

.pagination-summary { color: #6b7280; }

.orders-pagination {
  --pagination-active-background: #0070f3;
  --pagination-active-border: 1px solid #0070f3;
}
```

That consumer side is ~10 lines, ships zero new English strings into the library, lets every team localize their own copy, lets every team choose their own dropdown (native, `Select`, headless, whatever), and keeps Pagination as the focused primitive it was.

### What I'd happily merge as a follow-up

There are two genuinely valuable pieces in this PR worth pulling into a small, separate change:

- **`hasMore?: boolean`** — this is real new behavior (cursor pagination where `totalPages` is unknown), not a visual variant. It belongs in Pagination. Ship it.
- **`prevButtonTestId` / `nextButtonTestId`** — `data-pw` plumbing on the prev/next buttons is aligned with the library's testing convention. Ship it.

If you split those two out into a tiny follow-up PR and drop the summary + page-size selector + wrapper-div + `selectedItemLabel`, I'll review and approve that same day.

### Ask

Could you close this PR (or rework it down to just `hasMore` + the two `testId` props) and move the summary/selector composition into your own consumer code? Happy to pair on what the consumer-side toolbar looks like if it'd help, but it really does need to live outside the library.

murdore:
@sinha-sahil — thanks for the detailed breakdown. You are completely right on both counts.

The summary label and page-size `<select>` are composite concerns: they combine a localisation-owned string (`Showing {start}–{end} of {total} items`, `Rows per page:`), a layout shell, and a second interactive control into what should stay a focused nav primitive. That collides directly with §9 (no baked-in presets) and the "no composite wrappers" principle you described. The consumer-side example you posted shows exactly how little code is actually needed — there is nothing to save by pulling it into the library.

The root-element swap from `<nav>` to `<div class="pagination-wrapper">` also breaks the §9 contract that `classes` lands on the topmost control; I should not have done that.

**Planned next steps:**

- Close this PR as-is.
- Open a focused follow-up containing only `hasMore` and `prevButtonTestId` / `nextButtonTestId`, which you called out as genuinely belonging in the primitive.

One clarifying question before I open the follow-up: for `prevButtonTestId` / `nextButtonTestId`, would you prefer them typed `string | null` (current approach) or simply `string | undefined` to stay consistent with how other optional string props are typed in the library?

sinha-sahil:

This PR converts existing `function` declarations to arrow consts: `generatePages`, `goToPage`. Please revert those conversions — they're stylistic churn, and the existing convention in this codebase is function declarations.

New arrow-const handlers in the diff: `handlePageSizeChange`. Please write them as `function fnName() {}` instead.

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — thanks for engaging with this directly rather than negotiating around the principles. Quoting you: *"there is nothing to save by pulling it into the library"* — that's exactly the framing. The 5–15-line consumer recipe is the litmus test, and the toolbar version of Pagination fails it cleanly. Closing this and opening a slim follow-up is the right move.

**On your clarifying question — `string | null` vs `string | undefined`:**

Use `string | undefined` (i.e. `testId?: string`). It's the dominant convention across the library — every other component that exposes a `testId` prop is typed that way: `Button`, `Input`, `Select`, `Table`, `Tabs`, `Checkbox`, `Radio`, `Menu`, `Slider`, `Combobox`, `Pagination` itself, and ~20 others. The only outliers are `Tooltip` and `Toast`, both of which predate the convention solidifying and are on my list to normalise. Don't propagate the outlier shape into new props.

Concretely, in `src/lib/Pagination/properties.ts` the new fields should read:

```ts
prevButtonTestId?: string;
nextButtonTestId?: string;
```

…and in `Pagination.svelte` the consuming side stays clean — `data-pw={prevButtonTestId}` will simply omit the attribute when `undefined`, which matches how the existing `testId` on the root is wired. No `?? null`, no explicit nullish handling, no defaulting at the call site. The prop being absent and the prop being explicitly cleared collapse to the same behaviour, which is what we want for a test-only hook.

**For the follow-up PR, the scope should be exactly:**

1. `hasMore?: boolean` — real new behavior for cursor pagination where `totalPages` is unknown ahead of time. Add it to `MandatoryPaginationProperties` as optional, document the interaction with `totalPages` (which wins, what disables the next button), and that's it.
2. `prevButtonTestId?: string` and `nextButtonTestId?: string` — `data-pw` plumbing on the prev/next buttons only. Don't touch the root `testId`.

**Explicitly out of scope for the follow-up** (just to make sure nothing leaks back in):

- No `selectedItemLabel`, no `Showing … of … items` summary, no `Rows per page:` label.
- No page-size `<select>` (native or otherwise).
- No `<div class="pagination-wrapper">` — the root stays `<nav class="pagination {classes ?? ''}">` so §9 holds.
- No conversion of `generatePages` / `goToPage` from `function` declarations to arrow consts (per my separate note above — that's stylistic churn against the codebase convention).
- Any new helper (e.g. a `handlePrev` / `handleNext` if you factor one out) should be a `function fnName()` declaration, not `const fnName = () =>`.

If you can keep the follow-up to just those two additions plus their docs/example update, I'll review and approve it the same day it lands.

**Ask: please close this PR and open the slim follow-up with just `hasMore` + the two `testId` props, typed `string | undefined`.**

murdore:
Closing in favor of the slim #229 (hasMore + prev/next testIds, typed string|undefined) — the summary text and page-size selector are dropped. Thanks @sinha-sahil.

PR #217, opened 2026-05-29
feat: add ErrorCard component

murdore:
## Summary

- Adds a new `ErrorCard` primitive (`role="alert"`) distinct from `InlineAlert` (strip) and `EmptyState` (empty content)
- Supports `title`, `message`, optional `icon` snippet, optional `action` snippet, `variant` (`card` | `inline`), `testId` (renders `data-pw`), and `classes`
- `card` variant: bordered, padded card; `inline` variant: compact row layout

## API

```ts
type ErrorCardProperties = {
  title?: string;
  message?: string;
  icon?: Snippet;
  action?: Snippet;
  variant?: 'card' | 'inline'; // default: 'card'
  testId?: string;
  classes?: string;
};
```

Exposed CSS custom properties: `--error-card-background`, `--error-card-border`, `--error-card-radius`, `--error-card-padding`, `--error-card-gap`, `--error-card-icon-color`, `--error-card-title-color`, `--error-card-title-font-weight`, `--error-card-message-color`.

## Notes

- `prettier --check` — **clean**
- `eslint` — **clean**
- Backward-compatible: new named export `ErrorCard` and `export type * from './ErrorCard/properties'` added after existing `SplitInput` entries in `index.ts`
- No new npm dependencies introduced
- All font properties exposed as CSS vars only; no hardcoded `font-family`, `font-size`, `font-weight`, or `line-height` in component styles
- CSS selectors are kebab-case only (stylelint-compliant)

### DIFF
--- a/src/lib/ErrorCard/ErrorCard.svelte
+++ b/src/lib/ErrorCard/ErrorCard.svelte
@@ -1,0 +1,89 @@
+<script lang="ts">
+  import type { ErrorCardProperties } from './properties';
+
+  let {
+    title,
+    message,
+    icon,
+    action,
+    variant = 'card',
+    testId,
+    classes
+  }: ErrorCardProperties = $props();
+</script>
+
+<div class="error-card error-card-{variant} {classes ?? ''}" role="alert" data-pw={testId}>
+  {#if typeof icon === 'function'}
+    <div class="error-card-icon">
+      {@render icon()}
+    </div>
+  {/if}
+  <div class="error-card-body">
+    {#if title}
+      <div class="error-card-title">{title}</div>
+    {/if}
+    {#if message}
+      <div class="error-card-message">{message}</div>
+    {/if}
+  </div>
+  {#if typeof action === 'function'}
+    <div class="error-card-action">
+      {@render action()}
+    </div>
+  {/if}
+</div>
+
+<style>
+  .error-card {
+    display: flex;
+    align-items: flex-start;
+    gap: var(--error-card-gap, 12px);
+    background: var(--error-card-background, transparent);
+    border-radius: var(--error-card-radius, 8px);
+  }
+
+  .error-card-card {
+    border: var(--error-card-border, 1px solid currentColor);
+    padding: var(--error-card-padding, 16px);
+  }
+
+  .error-card-inline {
+    padding: var(--error-card-padding, 8px 0);
+  }
+
+  .error-card-icon {
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    flex-shrink: 0;
+    color: var(--error-card-icon-color, currentColor);
+  }
+
+  .error-card-icon :global(svg) {
+    width: 100%;
+    height: 100%;
+  }
+
+  .error-card-body {
+    display: flex;
+    flex-direction: column;
+    gap: 4px;
+    flex: 1;
+    min-width: 0;
+  }
+
+  .error-card-title {
+    color: var(--error-card-title-color, inherit);
+    font-weight: var(--error-card-title-font-weight, 600);
+  }
+
+  .error-card-message {
+    color: var(--error-card-message-color, inherit);
+  }
+
+  .error-card-action {
+    display: inline-flex;
+    align-items: center;
+    flex-shrink: 0;
+  }
+</style>
--- a/src/lib/ErrorCard/properties.ts
+++ b/src/lib/ErrorCard/properties.ts
@@ -1,0 +1,11 @@
+import type { Snippet } from 'svelte';
+
+export type ErrorCardProperties = {
+  title?: string;
+  message?: string;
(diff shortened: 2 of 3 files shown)


sinha-sahil:
Thanks @murdore. Two blockers: (1) variant card/inline prop bakes presets into the component (section 9 forbids - variants live in consumer CSS via classes prop); (2) structurally a composite duplicate of EmptyState differing only by role=alert and class names (principle 4). Also fix section 3 falsy checks on title/message (use typeof + length, see Banner L61). Either rework to remove variant and add real alert behavior (aria-live, dismiss+focus, retry/loading), or close in favor of EmptyState wrapped in role=alert with classes=empty-state-error.</adjustedDraftComment>
</invoke>

murdore:
@sinha-sahil — thanks for the clear breakdown. You are right on both counts.

On the structural objection: `ErrorCard` as written is essentially `EmptyState` with `role="alert"` and different class names — same icon/title/message/action slot pattern, same layout. Adding a near-duplicate component just to get `role="alert"` is exactly what §4 warns against. I should have composed on top of `EmptyState` instead.

On the `variant` prop: `'card' | 'inline'` bakes layout presets into the library, which §9 explicitly reserves for consumer CSS via `classes` + CSS variables. The two variants differ only in `border`, `padding`, and flex direction — those are trivially driven by two consumer-defined classes that set the relevant CSS vars.

My preferred path forward:

- Close this PR.
- In Lighthouse (the consumer), wrap `<EmptyState>` in a `<div role="alert" aria-live="assertive">` and pass `classes="empty-state-error"` or `classes="empty-state-error-inline"` with the card/inline styles defined in the consumer's stylesheet.
- If `EmptyState` needs a small interface extension (e.g. a `role` passthrough prop or a `dismiss` callback) to support real alert behavior cleanly, I can open a focused PR for that instead.

Does closing in favor of the consumer-side pattern sound right to you, or would you prefer I open the narrower `EmptyState` extension PR first?

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — appreciate the clean read on both points, and yes, your preferred path is exactly the right one. Let's close this PR.

> Close this PR. In Lighthouse, wrap `<EmptyState>` in a `<div role="alert" aria-live="assertive">` and pass `classes="empty-state-error"` or `classes="empty-state-error-inline"`...

That's the correct shape. To make the recipe concrete for whoever picks this up next:

```svelte
<div role="alert" aria-live="assertive">
  <EmptyState
    icon={errorIcon}
    title="Something went wrong"
    message={err.message}
    classes="empty-state-error"
  >
    {#snippet action()}
      <Button onclick={retry} classes="empty-state-error-retry">Retry</Button>
    {/snippet}
  </EmptyState>
</div>
```

…with `.empty-state-error` and `.empty-state-error-inline` defined in the consumer's stylesheet, driving the existing `--empty-state-*` CSS vars (border, padding, flex-direction). The card-vs-inline split lives there, as a styling concern owned by the consumer — which is exactly what §9 reserves the `classes` prop for.

On the narrower follow-up PR question:

> If `EmptyState` needs a small interface extension (e.g. a `role` passthrough prop or a `dismiss` callback) to support real alert behavior cleanly, I can open a focused PR for that instead.

Two different things, and I'd treat them differently:

1. **`role` passthrough — please don't.** HTML already lets you set `role` on a wrapping element, and that's the consumer-owned, zero-API-surface answer. Adding a `role` prop to EmptyState makes the library opinionated about ARIA semantics that the consumer should own at the composition layer. Same reasoning as not shipping a `tag` or `as` prop on every component — wrapping is free.

2. **`dismiss` callback — only if it carries real primitive behavior.** A bare `ondismiss` event with no focus management, no keyboard handling, no state machine is just `onclick` with a different name, which the consumer already owns via the action snippet. If you want to propose a `dismiss` API, it should justify itself on focus-restore-on-dismiss, Escape-key handling, or live-region announcement — primitive behavior a consumer genuinely can't trivially replicate. If it's just "call this function when X is clicked," skip it; the snippet covers it.

So: close this one, take the consumer-side recipe, and if you hit a real focus-management or keyboard-nav gap in EmptyState while wiring up the alert wrapper in Lighthouse, open a focused PR *with that specific gap as the justification* — not a speculative "EmptyState should support alerts too" extension. The bar is "consumer can't reasonably build this in their own code," and right now nothing in the alert path clears that bar.

Thanks for taking the feedback so cleanly — this is exactly the kind of close that keeps the library lean.

**Ask: please close this PR.** The card+inline variant and the role=alert wrapper are both consumer-side concerns; the recipe above replaces this component in 5–10 lines of consumer code + CSS.

murdore:
Closing — ErrorCard duplicates EmptyState (§4); role=alert + card/inline are consumer-side. I'll capture the EmptyState recipe in docs. Thanks @sinha-sahil.

PR #218, opened 2026-05-29
feat: add testId, SVG inlining, and transform to Img

murdore:
## Summary

- Add optional `testId?: string` prop that renders as `data-pw` on the root element (both `` and inline-svg `<span>` paths), enabling Playwright test selectors.
- Add optional `inlineSvg?: boolean` (default `false`) — when `true` and `src` is raw SVG markup (starts with `<svg`), renders via `{@html}` inside `<span class="img-inline-svg">` so it inherits `currentColor` and can be CSS-transformed. Falls back to the normal `` path for URLs or when `inlineSvg` is `false`.
- Add optional `transform?: (svg: string) => string` — when `inlineSvg` is active, pipes the raw SVG string through this function before rendering (e.g. rewriting `stroke`/`fill` attributes to `currentColor`).
- Add `--img-flex-shrink` CSS custom property (default unset) on both `` and `.img-inline-svg` so icon-in-flex callers can pin `flex-shrink: 0` without an extra wrapper element.

## API

| `testId` | `string` | — | Renders as `data-pw` attribute |
| `inlineSvg` | `boolean` | `false` | Render raw SVG markup inline |
| `transform` | `(svg: string) => string` | — | Pre-process SVG string before `{@html}` |

| CSS Variable | Default | Description |
|---|---|---|
| `--img-flex-shrink` | unset | Controls `flex-shrink` on the root element |

## Notes

- prettier: **clean**
- eslint: **clean** (`svelte/no-at-html-tags` inline-disabled on the `{@html}` line, matching `Banner.svelte` convention)
- Strictly backward-compatible — all new props are optional; omitting them reproduces existing behavior exactly
- No new npm dependencies added

### DIFF
--- a/src/lib/Img/Img.svelte
+++ b/src/lib/Img/Img.svelte
@@ -1,9 +1,26 @@
 <script lang="ts">
   import type { ImgProperties } from './properties';
 
-  let { src, alt, fallback, onerror, classes }: ImgProperties = $props();
+  let {
+    src,
+    alt,
+    fallback,
+    onerror,
+    classes,
+    testId,
+    inlineSvg = false,
+    transform
+  }: ImgProperties = $props();
 
   let currentSrc = $derived(src);
+
+  const isSvgMarkup = $derived(
+    typeof currentSrc === 'string' && currentSrc.trimStart().startsWith('<svg')
+  );
+  const showInline = $derived(inlineSvg && isSvgMarkup);
+  const svgContent = $derived(
+    showInline && typeof transform === 'function' ? transform(currentSrc) : currentSrc
+  );
 
   function handleFallback(): void {
     if (typeof fallback === 'string' && fallback.length > 0 && currentSrc !== fallback) {
@@ -14,7 +31,15 @@
   }
 </script>
 
-<img class={classes ?? ''} src={currentSrc} {alt} onerror={handleFallback} />
+{#if showInline}
+  <span class="img-inline-svg {classes ?? ''}" data-pw={testId} role="img" aria-label={alt}>
+    <!-- eslint-disable svelte/no-at-html-tags -->
+    {@html svgContent}
+    <!-- eslint-enable svelte/no-at-html-tags -->
+  </span>
+{:else}
+  <img class={classes ?? ''} src={currentSrc} {alt} onerror={handleFallback} data-pw={testId} />
+{/if}
 
 <style>
   img {
@@ -28,9 +53,31 @@
     background: var(--image-background);
     border: var(--image-border);
     transition: var(--image-transition);
+    flex-shrink: var(--img-flex-shrink);
   }
 
   img:hover {
+    background: var(--image-hover-background, var(--image-background));
+    border: var(--image-hover-border, var(--image-border));
+  }
+
+  .img-inline-svg {
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    height: var(--image-height, 24px);
+    width: var(--image-width, 24px);
+    padding: var(--image-padding, 0px);
+    border-radius: var(--image-border-radius, 0px);
+    margin: var(--image-margin, 0px);
+    filter: var(--image-filter, none);
+    background: var(--image-background);
+    border: var(--image-border);
+    transition: var(--image-transition);
+    flex-shrink: var(--img-flex-shrink);
+  }
+
+  .img-inline-svg:hover {
     background: var(--image-hover-background, var(--image-background));
     border: var(--image-hover-border, var(--image-border));
   }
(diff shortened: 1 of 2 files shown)


sinha-sahil:
Follow-up — escalating from MINOR to BLOCK after a stricter re-review against the #197 CopyButton precedent.

`testId` is fine and we'd happily take that as a standalone PR. The blockers are `inlineSvg` + `transform`:

1. **`inlineSvg` is a composite behavioral overlay.** It forks `Img` into two render paths (`` vs. `<span role="img">{@html svg}</span>`), each with its own CSS block including duplicated `:hover` rules. `Img` is supposed to be a thin `` wrapper; this turns it into a render-mode switch with an opinionated inline-SVG renderer baked in. Consumers with an SVG sprite system, an icon-font setup, or who just want `<svg>` directly in their template can't opt out of carrying this code path.

2. **`src` is now polymorphically overloaded.** Same `string` type, but the runtime checks `currentSrc.trimStart().startsWith('<svg')` and silently switches modes. A URL prop and a raw-markup prop are different contracts and should not share one field. If inline SVG must live here, it needs its own prop (e.g. `svgMarkup?: string`) so the API is explicit.

3. **`transform: (svg: string) => string` is a UX-preset callback.** It exists specifically to support the "rewrite stroke/fill to currentColor" workflow you describe in the PR body. That's a consumer-side concern — they can run the transform before passing the string in. A preprocessing-pipeline prop on a primitive is exactly the shape we pushed back on in #197.

4. **`.img-inline-svg` with embedded hover styles is a hardcoded variant.** Per our theming rule (no hardcoded design presets — theming via `classes` prop + external CSS), the new CSS block — especially the `:hover` rules — shouldn't live in the component.

Proposed split:
- Land `testId` alone (we'll merge that quickly).
- If inline SVG is a real need, ship it as a separate `InlineSvg` component so `Img` stays ``-only and consumers opt in by import, not by a runtime mode switch.
- Drop `transform` — consumers can preprocess the string themselves.

Minor cleanup if any inline-SVG variant does land later: align the CSS var name with the existing `--image-*` namespace (so `--image-flex-shrink`, not `--img-flex-shrink`).

sinha-sahil (changes requested):

To merge:

murdore:
@sinha-sahil — thanks for the review. The CHANGES_REQUESTED body references inline comments above, but I don't see any inline review comments on the diff (graphql confirms zero `reviewThreads`, only the formal-review body is present). Could you share which of the additions in this PR you'd like me to revisit — `inlineSvg` (the `{@html}` branch), `transform` (the consumer callback for SVG normalization), `testId`, or all three? If `testId` is acceptable on its own, I'm happy to split it out into a slim PR and close this one in favor of that.

sinha-sahil:
@murdore — apologies for the missing inline anchors, my fault for putting the whole thing in the review body. Let me restate the scope cleanly so we can move forward.

**Acceptable as a standalone PR — please open it and we'll merge quickly:**
- `testId?: string` rendered as `data-pw` on the `` root. That's it. Same shape as everywhere else in the library.

**Not landing here, in any form on `Img`:**
1. `inlineSvg` + the `{#if showInline} … <span role="img">{@html svgContent}</span>` branch (`src/lib/Img/Img.svelte:34-40`). Forking the render path turns `Img` from a thin `` wrapper into a render-mode switch and forces every consumer to carry the inline-SVG code path whether they use it or not. If inline SVG is a real need, it ships as its own component (`InlineSvg`) so consumers opt in by import.
2. The `src` polymorphism (`isSvgMarkup = currentSrc.trimStart().startsWith('<svg')` at `Img.svelte:14-16`). A URL prop and a raw-SVG-markup prop are different contracts and should not share one field, even behind a flag.
3. `transform?: (svg: string) => string` (`properties.ts:13`). A preprocessing-pipeline callback on a primitive is a UX preset — consumers can run the transform on the string themselves before passing it in. Same precedent as the CopyButton pushback in #197.
4. The `.img-inline-svg` block at `Img.svelte:65-86` — particularly the `:hover` rules duplicated from the `` block. Baked hover styling is a hardcoded design preset and violates GUIDELINES.md §9 (theming via `classes` + external CSS).

**Minor, only relevant if a future `InlineSvg` lands:** align the CSS var with the existing `--image-*` namespace — `--image-flex-shrink`, not `--img-flex-shrink`.

**Ask:** please open the `testId`-only PR against `release` and close this one in favor of it. Happy to fast-track the slim version.

murdore:
Splitting per @sinha-sahil review. Slim testId-only PR opened: https://github.com/juspay/svelte-ui-components/pull/234. Dropped: inlineSvg, transform, polymorphic src.

PR #219, opened 2026-05-29
feat: add RadioIcon headless indicator

murdore:
## Summary

- New `RadioIcon` headless primitive: a circular selected/unselected glyph for composing custom radio/option rows, with no `<input>` or label bundled in
- Exposes full CSS custom-property API (`--radio-icon-border-color`, `--radio-icon-selected-color`, `--radio-icon-dot-color`, `--radio-icon-disabled-opacity`, `--radio-icon-size`) so consumers can theme without forking
- Size driven by the `size` prop via `style="--radio-icon-size:{size}px"` on the root `<span>`

## API

| `selected` | `boolean` | `false` | Whether the indicator shows as selected |
| `disabled` | `boolean` | `false` | Applies disabled opacity and not-allowed cursor |
| `size` | `number` | `16` | Diameter in px (drives `--radio-icon-size`) |
| `testId` | `string` | — | Renders as `data-pw` for Playwright targeting |
| `classes` | `string` | — | Extra CSS classes appended to root |

## Notes

- prettier: clean
- eslint: clean
- No new npm dependencies
- Backward-compatible addition — no existing exports changed
- Follows Svelte 5 runes pattern (`$props()`, `$derived`) matching other library components
- `role="radio"` + `aria-checked` + `aria-disabled` for a11y
- CSS selectors are plain kebab-case; no font/line-height overrides

### DIFF
--- a/src/lib/RadioIcon/RadioIcon.svelte
+++ b/src/lib/RadioIcon/RadioIcon.svelte
@@ -1,0 +1,64 @@
+<script lang="ts">
+  import type { RadioIconProperties } from './properties';
+
+  let {
+    selected = false,
+    disabled = false,
+    size = 16,
+    testId,
+    classes
+  }: RadioIconProperties = $props();
+</script>
+
+<span
+  class="radio-icon {classes ?? ''}"
+  class:radio-icon-selected={selected}
+  class:radio-icon-disabled={disabled}
+  role="radio"
+  aria-checked={selected}
+  aria-disabled={disabled || null}
+  data-pw={testId}
+  style="--radio-icon-size:{size}px"
+>
+  <span class="radio-icon-dot"></span>
+</span>
+
+<style>
+  .radio-icon {
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    box-sizing: border-box;
+    width: var(--radio-icon-size, 16px);
+    height: var(--radio-icon-size, 16px);
+    border-radius: 50%;
+    border: 2px solid var(--radio-icon-border-color, #9e9e9e);
+    background-color: var(--radio-icon-bg-color, #ffffff);
+    flex-shrink: 0;
+    transition: border-color 0.15s ease;
+  }
+
+  .radio-icon-selected {
+    border-color: var(--radio-icon-selected-color, #2196f3);
+  }
+
+  .radio-icon-disabled {
+    opacity: var(--radio-icon-disabled-opacity, 0.4);
+    cursor: not-allowed;
+  }
+
+  .radio-icon-dot {
+    width: 0;
+    height: 0;
+    border-radius: 50%;
+    background-color: var(--radio-icon-dot-color, #2196f3);
+    transition:
+      width 0.15s ease,
+      height 0.15s ease;
+  }
+
+  .radio-icon-selected .radio-icon-dot {
+    width: calc(var(--radio-icon-size, 16px) * 0.5);
+    height: calc(var(--radio-icon-size, 16px) * 0.5);
+  }
+</style>
--- a/src/lib/RadioIcon/properties.ts
+++ b/src/lib/RadioIcon/properties.ts
@@ -1,0 +1,7 @@
+export type RadioIconProperties = {
+  selected?: boolean;
+  disabled?: boolean;
+  size?: number;
+  testId?: string;
+  classes?: string;
+};
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -56,6 +56,7 @@ export { default as EmptyState } from './EmptyState/EmptyState.svelte';
 export { default as Combobox } from './Combobox/Combobox.svelte';
 export { default as ColorPicker } from './ColorPicker/ColorPicker.svelte';
 export { default as SplitInput } from './SplitInput/SplitInput.svelte';
+export { default as RadioIcon } from './RadioIcon/RadioIcon.svelte';
 
 export type * from './Button/properties';
 export type * from './Modal/properties';
@@ -109,5 +110,6 @@ export type * from './EmptyState/properties';
 export type * from './Combobox/properties';
 export type * from './ColorPicker/properties';
 export type * from './SplitInput/properties';
+export type * from './RadioIcon/properties';
 
 export { validateInput } from './utils';


sinha-sahil:
## Thanks for the PR — but this should not land as a new component

Hi @murdore, I can see what you're going for here: a small, themable circle-with-dot you can drop next to custom row markup when you don't want the full `Radio` shape. The goal is fine; the packaging is not. A few hard blockers below.

---

### 1. This is a dummy component (philosophy blocker)

The library ships **primitives that justify their existence with non-trivial behavior, accessibility, focus management, or composition value.** `RadioIcon.svelte` has none of that:

- No `<input>`
- No `onclick` / `onkeydown` / `onchange`
- No `tabindex`
- No internal state
- No snippet slots

It is **two `<span>`s and CSS** — a literal decorative glyph. Per the project's stated philosophy ("No dummy components — a pure-decoration element with no behavior, focus management, ARIA semantics, or composition value should be a CSS utility, not a component"), this belongs in user-land CSS, not in `src/lib/`.

Worse: the `role="radio"` + `aria-checked` are misleading. An ARIA radio that is not focusable and not operable lies to assistive tech — Space/Enter does nothing, the element is not in the tab order, there is no group context. That is a **regression** in a11y signal, not an improvement, despite the PR description calling it out as a feature.

### 2. It duplicates the existing `Radio` component (GUIDELINES §4)

`src/lib/Radio/Radio.svelte` (lines 70-113) already renders exactly this glyph:

```svelte
<span class="radio-indicator" class:checked class:disabled>
  <span class="radio-dot" class:checked class:disabled></span>
</span>
```

…with a **richer** CSS-variable API: `--radio-size`, `--radio-border`, `--radio-selected-border`, `--radio-disabled-border`, `--radio-dot-color`, `--radio-dot-size`, `--radio-transition`, `--radio-focus-shadow`, etc. The new component re-implements the same two-circle markup with a **thinner, less consistent** variable surface (e.g. `border: 2px solid var(--radio-icon-border-color, #9e9e9e)` bakes in the width/style; `Radio` exposes the full shorthand).

§4 says: *"Before implementing behaviour that already exists in the library, check if an existing component already handles it. Compose rather than duplicate."* If consumers want to compose a custom row, `Radio` already accepts `classes` and gives them everything they need — including a real `<input type="radio">` for forms and keyboard a11y. If the indicator-only use case is real, the right move is to **refactor `Radio` to expose its indicator via a snippet**, not ship a parallel decorative element with divergent CSS variable names.

### 3. The `size` prop violates §9 (the core theming mechanism)

```svelte
<span … style="--radio-icon-size:{size}px">
```

`size: number` → inline `style` → CSS variable. That is precisely the "variant prop with embedded styles" anti-pattern the project rejects (see also the standing feedback note "No hardcoded design presets in components"). Two problems:

- It bakes in a `px` unit. A consumer who wants `1.25rem` or `clamp(…)` cannot get there.
- Inline `style` beats class specificity, so a consumer's `.radio-lg { --radio-icon-size: 24px }` is **silently overridden** by the default `size = 16` whenever the prop has its default value. The §9 contract is broken.

The CSS variable `--radio-icon-size` is already exposed — there is no reason for a prop at all. Just drop `size`. Consumers do:

```css
/* consumer app.css */
.radio-icon-sm { --radio-icon-size: 12px; }
.radio-icon-md { --radio-icon-size: 16px; }
.radio-icon-lg { --radio-icon-size: 24px; --radio-icon-dot-size: 14px; }
```

```svelte
<RadioIcon classes="radio-icon-lg" selected />
```

### 4. Other customisation gaps (Major)

- `transition: border-color 0.15s ease;` and `transition: width 0.15s ease, height 0.15s ease;` are hardcoded. No `--radio-icon-transition` hook. `Radio` does this correctly: `transition: var(--radio-transition, 0.2s)`.
- The selected dot diameter `calc(var(--radio-icon-size, 16px) * 0.5)` bakes a 50% ratio. Should be `width: var(--radio-icon-dot-size, calc(var(--radio-icon-size, 16px) * 0.5))`.
- `border: 2px solid var(--radio-icon-border-color, …)` — only the color is themable. Use a composite variable: `border: var(--radio-icon-border, 2px solid #9e9e9e)`.

---

## What I'd like instead

**Option A (preferred): close this PR.** The use case — "composable radio rows" — is already covered by `Radio` with its `classes` prop and its full CSS-variable API. If something is missing on `Radio` to support a row layout you have in mind, open a small PR that adds *that* (e.g. a `#snippet indicator` slot, or a `--radio-text-display: none` recipe).

**Option B: refactor `Radio` to expose the indicator.** If we genuinely want a headless indicator, the way to do it is to let `Radio` render a custom indicator snippet — single source of truth, single CSS-variable namespace (`--radio-*`), no parallel decorative element.

**Option C (only if A and B are off the table):** make this a real headless primitive — focusable, keyboard-operable, emits `onchange`, manages `tabindex`, exposes focus-ring CSS variables, and drops the `size` prop in favor of pure CSS-variable theming. But even then it overlaps with `Radio`, so I would push back.

Blocking on the dummy-component + duplication points. Happy to discuss if I'm missing context on the row composition use case.

murdore:
@sinha-sahil — thank you for the thorough review. You're right on all counts, and I don't have a strong counter-argument to any of the three blockers.

The dummy-component point lands hardest: `RadioIcon` is two `<span>`s and CSS with no behavior, no focus management, and no real composition value. Shipping it as a library primitive doesn't meet the "non-trivial justification" bar the project sets, and the misleading `role="radio"` + `aria-checked` on a non-operable element is a genuine a11y regression — not a win.

The §4 duplication point is equally valid. `Radio`'s existing `radio-indicator` markup already covers this visual exactly, with a richer and more consistent CSS-variable surface. I shouldn't have added a parallel element with a divergent namespace.

Given that, I'm inclined toward **Option A**: close this PR. The row-composition use case I had in mind is already reachable via `Radio` + `classes`.

Before I close: would you be open to a small follow-up that adds a `#snippet indicator` slot to `Radio`, so consumers can swap in custom indicator markup for headless row layouts — or is the existing `classes` + CSS-variable API sufficient for that use case as-is?

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — thanks for the gracious concession. Agreeing that the misleading `role="radio"` on a non-operable element is "not a win" is the right read; that one mattered to me the most because it actively degrades a11y under the guise of improving it. Closing this out is the right call.

On your forward-looking question — **would we be open to a `#snippet indicator` slot on `Radio`** — my honest answer is: **not yet, and probably not at all.** Two reasons, in order of weight.

**1. Prove the use case against the current API first.** `Radio` already exposes a generous CSS-variable surface (`--radio-size`, `--radio-border`, `--radio-selected-border`, `--radio-disabled-border`, `--radio-dot-color`, `--radio-dot-size`, `--radio-transition`, `--radio-focus-shadow`) plus the `classes` prop. Between those, you can reshape the indicator to almost anything — different sizes, different dot ratios, different borders, square corners, custom focus rings, even hide the text via `--radio-text-display: none` if that's the row-composition shape you had in mind. Before we add a snippet slot, I'd want to see a concrete row layout that the current API *demonstrably* can't reach. If you have one, drop it in an issue with a CodeSandbox and we'll look — but my prior is "the CSS-variable surface is already wide enough."

**2. A snippet slot for "the indicator" is the first step toward the composite-wrapper pattern.** Once `Radio` accepts a custom indicator snippet, the next ask is "can I also slot the label?", then "can I slot the description?", then "can I slot the trailing icon?" — and we're three PRs away from `Radio` being a layout shell where consumers pass in every visual piece. That's exactly the user-land composition territory we want to keep *out* of the library. The current shape — a real `<input type="radio">` with focus/keyboard/form semantics, themed via CSS variables and `classes` — is the primitive doing the right amount of work. Trading that for slot-based extensibility moves us toward a headless-UI catalogue, which isn't this library.

If you find a genuine gap in the variable surface during your row-composition work (say, a missing `--radio-indicator-shape` or an inability to theme the dot transform), **that** is the right kind of small PR — additive CSS-variable hooks on the existing component, no API surface expansion. Happy to fast-track those.

**Ask: please close this PR.** The decorative-glyph + duplicated-namespace pattern isn't reconcilable with the library philosophy, and you've already named the correct path forward (use `Radio` + `classes` + CSS variables for row composition). If a concrete gap surfaces, open a focused issue with the use case and we'll iterate on `Radio`'s variable surface — not via a new component, and not via slot expansion.

murdore:
Closing — RadioIcon is a decorative glyph duplicating Radio's indicator. Row composition should use Radio + classes + CSS vars; I'll open a focused issue if a real gap surfaces. Thanks @sinha-sahil.

PR #220, opened 2026-05-29
feat: add string-items, multi-select, select-all, and hint to Select

murdore:
## Summary

- **String items**: `items` now accepts `string[]` in addition to `SelectItem[]`; strings are normalized to `{ id: s, label: s }` internally — existing object-item callers are unchanged.
- **showSelectAll**: new optional `boolean` prop (default `false`); when `true` and `multiple` is set, renders a "Select all" row at the top of the dropdown that toggles all filtered items.
- **hint**: optional `string` rendered as a small helper line above the trigger (styled via `--select-hint-color`, `--select-hint-margin-bottom`).
- **subtext**: optional `string` rendered below the trigger (styled via `--select-subtext-color`, `--select-subtext-margin-top`).
- **bottomContent**: optional `Snippet` rendered at the bottom of the dropdown panel (e.g. an Apply button), separated by a configurable border (`--select-bottom-content-border`, `--select-bottom-content-padding`).

## API

| `items` | `SelectItem[] \| string[]` | — | Plain strings are auto-normalized |
| `showSelectAll` | `boolean` | `false` | "Select all" row in multi-select |
| `hint` | `string` | `undefined` | Helper text above trigger |
| `subtext` | `string` | `undefined` | Helper text below trigger |
| `bottomContent` | `Snippet` | `undefined` | Slot at bottom of dropdown |

New CSS variables: `--select-hint-color`, `--select-hint-margin-bottom`, `--select-subtext-color`, `--select-subtext-margin-top`, `--select-checkbox-color`, `--select-option-all-border`, `--select-option-gap`, `--select-bottom-content-border`, `--select-bottom-content-padding`.

## Notes

- `prettier --check` exits **0** on all changed files
- `eslint` exits **0** on all changed files
- Strictly additive — all new props are optional with safe defaults; single-select with object items behaves exactly as before
- No new npm dependencies introduced

### DIFF
--- a/src/lib/Select/Select.svelte
+++ b/src/lib/Select/Select.svelte
@@ -5,16 +5,25 @@
   import chevronDownSvg from '$lib/assets/chevron-down.svg?raw';
 
   let {
-    items,
+    items: rawItems,
     value = $bindable([]),
     multiple = false,
+    showSelectAll = false,
     searchable = false,
     placeholder = '',
     disabled = false,
+    hint,
+    subtext,
+    bottomContent,
     testId,
     onchange,
     classes
   }: SelectProperties = $props();
+
+  const normalizeItems = (source: SelectItem[] | string[]): SelectItem[] =>
+    source.map((entry) => (typeof entry === 'string' ? { id: entry, label: entry } : entry));
+
+  let items: SelectItem[] = $derived(normalizeItems(rawItems));
 
   let isOpen = $state(false);
   let query = $state('');
@@ -25,15 +34,19 @@
 
   const listboxId = `select-listbox-${Math.random().toString(36).slice(2, 9)}`;
 
-  function getLabel(id: string): string {
+  const getLabel = (id: string): string => {
     const found = items.find((item) => item.id === id);
     return typeof found === 'object' ? found.label : id;
-  }
+  };
 
   let filteredItems: SelectItem[] = $derived(
     searchable && query.length > 0
       ? items.filter((item) => item.label.toLowerCase().includes(query.toLowerCase()))
       : items
+  );
+
+  let allSelected: boolean = $derived(
+    filteredItems.length > 0 && filteredItems.every((item) => value.includes(item.id))
   );
 
   let displayText = $derived.by(() => {
@@ -50,7 +63,7 @@
 
   let searchPlaceholder = $derived(isOpen && displayText.length > 0 ? displayText : placeholder);
 
-  async function open(): Promise<void> {
+  const open = async (): Promise<void> => {
     if (disabled || isOpen) {
       return;
     }
@@ -63,15 +76,30 @@
         searchInputEl.focus();
       }
     }
-  }
+  };
 
-  function close(): void {
+  const close = (): void => {
     isOpen = false;
     query = '';
     highlightedIndex = -1;
-  }
+  };
 
-  function selectItem(id: string): void {
+  const toggleSelectAll = (): void => {
+    if (disabled) {
+      return;
+    }
+    if (allSelected) {
+      value = value.filter((id) => !filteredItems.some((item) => item.id === id));
+    } else {
+      const newIds = filteredItems
+        .filter((item) => !value.includes(item.id))
+        .map((item) => item.id);
+      value = [...value, ...newIds];
+    }
+    onchange?.(value);
+  };
+
+  const selectItem = (id: string): void => {
     if (disabled) {
       return;
     }
@@ -82,17 +110,17 @@
       close();
     }
     onchange?.(value);
-  }
+  };
 
-  function removeItem(id: string): void {
+  const removeItem = (id: string): void => {
     if (disabled) {
       return;
     }
     value = value.filter((v) => v !== id);
     onchange?.(value);
-  }
+  };
 
-  function selectHighlighted(): void {
+  const selectHighlighted = (): void => {
     if (highlightedIndex < 0 || highlightedIndex >= filteredItems.length) {
       return;
     }
@@ -100,9 +128,9 @@
     if (typeof item === 'object' && item !== null) {
       selectItem(item.id);
     }
-  }
+  };
 
-  async function moveHighlight(delta: number): Promise<void> {
+  const moveHighlight = async (delta: number): Promise<void> => {
     const next = highlightedIndex + delta;
     if (next < 0 || next >= filteredItems.length) {
       return;
@@ -115,9 +143,9 @@
         el.scrollIntoView({ block: 'nearest' });
       }
     }
-  }
+  };
 
-  function handleTriggerClick(event: MouseEvent): void {
+  const handleTriggerClick = (event: MouseEvent): void => {
     if (event.target instanceof HTMLInputElement) {
       if (!isOpen) {
         open();
@@ -131,9 +159,9 @@
     } else {
       open();
     }
-  }
+  };
 
-  function handleKeydown(event: KeyboardEvent): void {
+  const handleKeydown = (event: KeyboardEvent): void => {
     if (disabled) {
       return;
     }
@@ -190,9 +218,9 @@
         }
         break;
     }
-  }
+  };
 
-  function handleSearchInput(event: Event): void {
+  const handleSearchInput = (event: Event): void => {
     if (!(event.target instanceof HTMLInputElement)) {
       return;
     }
@@ -201,15 +229,15 @@
       isOpen = true;
     }
     highlightedIndex = -1;
-  }
+  };
 
-  function handleSearchFocus(): void {
+  const handleSearchFocus = (): void => {
     if (!isOpen) {
       open();
     }
-  }
+  };
 
-  function handleClickOutside(event: Event): void {
+  const handleClickOutside = (event: Event): void => {
     if (
       event.target instanceof Node &&
       containerEl !== null &&
@@ -217,7 +245,7 @@
     ) {
       close();
     }
-  }
+  };
 
   onMount(() => {
     document.addEventListener('click', handleClickOutside);
@@ -234,6 +262,10 @@
   bind:this={containerEl}
   {...typeof testId === 'string' ? { 'data-pw': testId } : {}}
 >
+  {#if hint}
+    <span class="select-hint">{hint}</span>
+  {/if}
+
   <div
     class="select-trigger"
     bind:this={triggerEl}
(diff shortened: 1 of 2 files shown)


sinha-sahil:
Thanks for this — there's real value here (string-items shorthand, `showSelectAll`, and `bottomContent` are all good additions), but the PR also crosses a few lines the library's design philosophy draws explicitly. I'd like to land most of it after a rework rather than merge as-is.

## Blocker: `hint` and `subtext` turn Select into a composite wrapper

Per the library philosophy ("components are unstyled primitives, no composite wrappers") and **GUIDELINES.md §9**, Select should be the picker primitive — nothing more. The moment we add `hint` (helper line *above* the trigger) and `subtext` (helper line *below* the trigger) with their own CSS-variable surface (`--select-hint-color`, `--select-hint-margin-bottom`, `--select-subtext-color`, `--select-subtext-margin-top`), Select stops being a picker and starts becoming a form-row.

This matters because the next PR will (rightfully) ask for the same on `Input`, then `DatePicker`, then `Checkbox`… and we end up with the variant-prop / composite-wrapper explosion the library was built to avoid. The customizable alternative is consumer-side composition with the `classes` prop:

```svelte
<!-- in consumer code -->
<div class="form-field">
  <span class="form-field-hint">Pick a country</span>
  <Select items={countries} classes="country-select" />
  <span class="form-field-subtext">You can change this later</span>
</div>
```

```css
/* consumer app.css */
.form-field { display: grid; gap: 4px; }
.form-field-hint { color: #666; font-size: 12px; }
.form-field-subtext { color: #888; font-size: 12px; }
.country-select { --select-trigger-border-radius: 8px; }
```

If we genuinely need a ready-made form-field primitive in the library, it should ship as a separate component (e.g. `FormField`) — not be bolted onto every input we have. Please drop `hint`, `subtext`, and their CSS variables from this PR.

## Major: select-all glyphs are unoverridable

```svelte
<span class="select-checkbox" aria-hidden="true">{allSelected ? '☑' : '☐'}</span>
```

The Unicode glyphs are fine as a *default*, but per principle 5 ("highly customizable — every visual choice must be reachable via a CSS variable or snippet") they can't be the only option. A consumer using a Lucide-icon or SVG-sprite design system has no escape hatch here. Please expose snippets so the visual is consumer-owned:

```ts
// properties.ts
checkedIcon?: Snippet;
uncheckedIcon?: Snippet;
selectAllLabel?: string;       // default 'Select all' — also makes it translatable
```

```svelte
<span class="select-checkbox" aria-hidden="true">
  {#if value.includes(item.id)}
    {#if typeof checkedIcon === 'function'}{@render checkedIcon()}{:else}☑{/if}
  {:else}
    {#if typeof uncheckedIcon === 'function'}{@render uncheckedIcon()}{:else}☐{/if}
  {/if}
</span>
```

Same treatment for the select-all label — hardcoding `<span>Select all</span>` makes the feature untranslatable.

## Major: §3 falsy/truthy checks on the new template branches

```svelte
{#if hint}        <!-- string | undefined -->
{#if subtext}     <!-- string | undefined -->
{#if bottomContent} <!-- Snippet | undefined -->
```

These violate **GUIDELINES.md §3** — strings and snippets are non-boolean and must be checked with `typeof`. The rest of this file uses `typeof` correctly (`typeof firstId !== 'string'`, `typeof found === 'object'`), so the new code regresses the pattern. After dropping `hint` and `subtext`, the remaining fix is:

```svelte
{#if typeof bottomContent === 'function'}
  <div class="select-bottom-content">{@render bottomContent()}</div>
{/if}
```

## What's good and should land

- `items: SelectItem[] | string[]` with the `normalizeItems` derivation — clean ergonomic win, internal model unchanged.
- `showSelectAll` toggling over **filtered** items — exactly the kind of behavior a primitive should own (and `allSelected` is correctly derived against `filteredItems`, not `items`, which is the right call).
- `bottomContent` as a `Snippet` — that's the correct shape for a composition slot.
- New CSS variables follow the existing `var(--select-*, default)` convention.

## Ask

Could you split this into two PRs?

1. **Land now**: string-items normalization + `showSelectAll` + `bottomContent` snippet + checkbox-icon snippets + `selectAllLabel`. Fix the `{#if bottomContent}` to `{#if typeof bottomContent === 'function'}`.
2. **Close / move to user-land**: `hint` + `subtext` + their CSS variables. If we want a form-field primitive, let's design that separately rather than smuggle it into Select.

Happy to review the split PR quickly once it's up.

sinha-sahil:
Follow-up — re-grading from MAJOR to **blocking** after a stricter read against the #197 CopyButton precedent (no hardcoded UX presets / English strings / baked-in glyphs inside primitives).

Per-feature breakdown — these should not all be treated the same:

**`bottomContent` (Snippet) — keep as-is.** This is the correct primitive: a composable slot the consumer owns. It's also what makes the rest of this PR unnecessary.

**`string[]` items overload — acceptable.** Mildly opinionated (forces `id === label`) but cheap and additive. Fine to ship.

**`showSelectAll` — please drop.** This is a UX preset, not a primitive:
- Hardcoded English label `<span>Select all</span>` (Select.svelte:350) — same class of violation as the #197 "Copied!" string.
- Hardcoded Unicode glyphs `{allSelected ? '☑' : '☐'}` — not themeable beyond color, font-dependent, can't be swapped for an SVG or a Checkbox component.
- The toggle logic is ~10 lines that a consumer can write against `bind:value` and own end-to-end. With `bottomContent` now available, the consumer can render their own Select-All row using their own button/checkbox primitive in the language and style they want.

**Checkbox glyphs on each option row — same problem.** Even without `showSelectAll`, the new per-row `{value.includes(item.id) ? '☑' : '☐'}` bakes a visual into the markup. Prefer a `optionIndicator?: Snippet<[{ selected: boolean }]>` with the current glyph as fallback, or an inline SVG with CSS-controllable stroke.

**`hint` and `subtext` — please drop both.**
- Both are opinionated muted-text presets (hardcoded `#666666` / `#888888`, hardcoded margins).
- `subtext` isn't in the PR title and semantically overlaps with `hint` — two props for "small grey helper text, but above vs below."
- Consumers already control the surrounding layout and can render `<label>` / `<small>` siblings with their own design tokens.
- If a slot really is needed, one `helperContent: Snippet` covers all positions and styles.

**One more ask — function declarations, not arrow consts.**

The diff converts a dozen internal handlers from `function fnName()` to `const fnName = () => {}` (getLabel, open, close, selectItem, removeItem, selectHighlighted, moveHighlight, handleTriggerClick, handleKeydown, handleSearchInput, handleSearchFocus, handleClickOutside). Project convention is function declarations — hoisting, named stack traces, `this` semantics, consistency with the rest of `src/lib/`. Please revert the conversions in the next pass.

**Suggested path forward:** keep `bottomContent` and `string[]` items; remove `showSelectAll`, `hint`, `subtext`, and the inline checkbox glyphs (or move them behind a Snippet with the glyph as fallback). That leaves a clean additive PR that fits the library's primitives philosophy.

murdore:
@sinha-sahil — thank you for the detailed review. You read it closely and the feedback is completely on point.

You are right that `hint` and `subtext` turn Select into a form-row composite, which is exactly the pattern the library is designed to prevent. Per GUIDELINES.md §9, those concerns belong on the consumer side — a wrapping `<div class="form-field">` with CSS variables, not bolted onto every input primitive. I should not have included them.

Here is my plan, matching your suggested split:

**PR 1 (rework of this PR) — keep and fix:**
- String-items normalization (`SelectItem[] | string[]`)
- `showSelectAll` + `allSelected` derived against `filteredItems`
- `bottomContent` snippet slot
- Add `checkedIcon`, `uncheckedIcon` snippets with Unicode glyphs as defaults
- Add `selectAllLabel?: string` prop (default `'Select all'`) for i18n
- Fix `{#if hint}` / `{#if subtext}` / `{#if bottomContent}` to use `typeof` guards per §3 — only `bottomContent` survives, as `typeof bottomContent === 'function'`

**PR 2 (close / move to docs):**
- Drop `hint`, `subtext`, and their CSS variables entirely
- If a `FormField` wrapper primitive is wanted later, design it separately

Before I open the split PR: would you also want me to add keyboard support (`Enter`/`Space`) to the `select-option-all` row, or is that a follow-up? And is a plain `checkedIcon` / `uncheckedIcon` snippet pair the shape you have in mind, or would you prefer a single `checkboxIcon` snippet that receives a `checked: boolean` argument?

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — appreciate the quick turnaround and the clean alignment on the split. Plan in your reply matches what I had in mind almost exactly; let me close the loop on the two open questions and one small reinforcement on `showSelectAll`.

### On `showSelectAll` — keeping it, but only if the visuals are fully consumer-owned

To make sure my second comment doesn't get lost: the *behavior* (toggle all filtered items, derive `allSelected` against `filteredItems`) is genuinely primitive-shaped and fine to land. What I was pushing back on was the **baked visuals** — the hardcoded `'Select all'` string, the `☑`/`☐` glyphs, and the per-row checkbox glyphs. If those are all moved behind snippets/props with sensible fallbacks, `showSelectAll` survives as a pure behavior flag and the primitive stays unopinionated. So as long as your PR 1 also includes:

- `selectAllLabel?: string` (default `'Select all'`) — covers i18n
- Snippet(s) for the indicator on both the select-all row **and** each option row — same component, same fallback

…then `showSelectAll` is reconcilable with the philosophy. Without those, it's a UX preset and would still need to come out.

### Snippet shape — go with the single `checkboxIcon` taking `{ checked: boolean }`

Strong preference for the single-snippet shape:

```ts
optionIndicator?: Snippet<[{ checked: boolean }]>;
```

Reasons: (1) it's one prop instead of two, so consumers wiring up a Lucide/SVG-sprite system don't have to remember to override both halves; (2) it composes with future tri-state / indeterminate cases (you'd add `indeterminate?: boolean` to the payload without breaking the API); (3) it matches how the rest of `src/lib/` shapes parameterized snippets — see how the existing Select option/empty snippets are typed. Two-snippet pairs invite consumers to forget one and end up with mixed glyph systems.

Use it for both the per-row indicator and the select-all row indicator — same snippet, same payload. Don't introduce a second `selectAllIndicator` prop unless a concrete use-case forces it.

### Enter/Space on the select-all row — please include it in PR 1

It's small, it's the kind of keyboard behavior consumers genuinely can't replicate without re-implementing the row themselves, and it's exactly the sort of *real* primitive behavior the library is for (vs. baked styling). The select-all row should be focusable (`tabindex="0"` or button-shaped), `Enter` and `Space` should toggle, and `Escape` should fall through to the existing close-dropdown handler. While you're in there: please also revert the `function` → arrow-const conversions on the handler set (`getLabel`, `open`, `close`, `selectItem`, `removeItem`, `selectHighlighted`, `moveHighlight`, `handleTriggerClick`, `handleKeydown`, `handleSearchInput`, `handleSearchFocus`, `handleClickOutside`) — project convention is function declarations per the design notes.

### Recap for PR 1

Keep: `string[]` items normalization · `showSelectAll` behavior · `allSelected` against `filteredItems` · `bottomContent` snippet · `selectAllLabel` · `optionIndicator: Snippet<[{ checked: boolean }]>` with `☑`/`☐` fallback · Enter/Space on the select-all row · `{#if typeof bottomContent === 'function'}` per §3 · revert arrow-consts to `function` declarations.

Drop: `hint` · `subtext` · `--select-hint-*` / `--select-subtext-*` CSS variables · hardcoded English strings · hardcoded glyphs in markup.

**Ask: please close this PR and open a slim PR 1 with the keepers above.** Happy to review it same-day once it's up.

murdore:
Closing in favor of the slim #232 (string[] normalization + showSelectAll + optionIndicator snippet + bottomContent) — the hint/subtext composite props are dropped. Thanks @sinha-sahil.

PR #221, opened 2026-05-29
feat: add tone, title, size, and flush to Banner

murdore:
## Summary

- **tone** (`'info' | 'success' | 'warning' | 'error' | null`) — applies `banner-tone-{tone}` class that reads `--banner-tone-background`, `--banner-tone-color`, and `--banner-tone-border` CSS vars (no hardcoded defaults so they fall through to the existing `--banner-background`/`--banner-color` when unset). When `tone === 'error'`, sets `role="alert"` automatically.
- **title** (`string | null`) — renders a `<span class="banner-title">` above the text inside a `<div class="banner-body">` flex-column. When absent, markup is identical to today. Exposes `--banner-title-font-weight` (default `600`) and `--banner-body-gap` (default `2px`).
- **size** (`'sm' | 'md' | null`) — `sm` applies `banner-size-sm` with tighter padding via `--banner-sm-padding` (default `6px 12px`) and removes `border-bottom`.
- **flush** (`boolean`, default `false`) — `banner-flush` zeroes `border-radius` and `border` for edge-to-edge placement inside a card. Exposes `--banner-border-radius` (default `0`).
- **role** (`string | null`) — explicit ARIA role override; takes precedence over the auto-resolved role (`alert` for error tone, `button` for interactive).

## API

| Prop | Type | Default | Notes |
|------|------|---------|-------|
| `tone` | `'info' \| 'success' \| 'warning' \| 'error' \| null` | `null` | Applies semantic colour class |
| `title` | `string \| null` | `null` | Two-line body with bold heading |
| `size` | `'sm' \| 'md' \| null` | `null` | `sm` = tighter padding |
| `flush` | `boolean` | `false` | Remove border-radius/border |
| `role` | `string \| null` | `null` | Override resolved ARIA role |

### New CSS custom properties

| Variable | Default | Purpose |
|----------|---------|---------|
| `--banner-tone-background` | *(none, falls back to `--banner-background`)* | Tone background |
| `--banner-tone-color` | *(none, falls back to `--banner-color`)* | Tone text colour |
| `--banner-tone-border` | *(none)* | Tone border |
| `--banner-title-font-weight` | `600` | Title weight |
| `--banner-body-gap` | `2px` | Gap between title and text |
| `--banner-sm-padding` | `6px 12px` | Padding when `size="sm"` |
| `--banner-border-radius` | `0` | Border radius (used by `flush`) |

## Notes

- **Strictly additive** — all new props are optional with `null` defaults. Omitting them reproduces today's single-line strip exactly.
- `svelte-check` found **0 errors and 0 warnings**.
- Prettier + ESLint **clean** (no `undefined`, no BEM classes, no inline font props).
- No new npm dependencies.
- Builds on the `--banner-white-space` work already merged to `release`.

### DIFF
--- a/src/lib/Banner/Banner.svelte
+++ b/src/lib/Banner/Banner.svelte
@@ -15,38 +15,59 @@
     dismissIcon,
     onclick,
     ondismiss,
-    classes
+    classes,
+    tone = null,
+    title = null,
+    size = null,
+    flush = false,
+    role = null
   }: BannerProperties = $props();
 
   let interactive = $derived(typeof onclick === 'function');
 
-  function handleClick(event: MouseEvent): void {
-    onclick?.(event);
-  }
+  let resolvedRole = $derived(
+    role !== null ? role : tone === 'error' ? 'alert' : interactive ? 'button' : null
+  );
 
-  function handleKeydown(event: KeyboardEvent): void {
+  let rootClass = $derived(
+    [
+      'banner',
+      tone !== null ? `banner-tone-${tone}` : '',
+      size === 'sm' ? 'banner-size-sm' : '',
+      flush ? 'banner-flush' : '',
+      classes ?? ''
+    ]
+      .filter(Boolean)
+      .join(' ')
+  );
+
+  const handleClick = (event: MouseEvent): void => {
+    onclick?.(event);
+  };
+
+  const handleKeydown = (event: KeyboardEvent): void => {
     if (event.key === 'Enter' || event.key === ' ') {
       event.preventDefault();
       if (event.currentTarget instanceof HTMLElement) {
         event.currentTarget.click();
       }
     }
-  }
+  };
 
-  function handleDismiss(event: MouseEvent): void {
+  const handleDismiss = (event: MouseEvent): void => {
     event.stopPropagation();
     visible = false;
     ondismiss?.();
-  }
+  };
 </script>
 
 {#if visible}
   <!-- svelte-ignore a11y_no_noninteractive_tabindex -->
   <div
-    class="banner {classes ?? ''}"
+    class={rootClass}
     onclick={interactive ? handleClick : null}
     onkeydown={interactive ? handleKeydown : null}
-    role={interactive ? 'button' : null}
+    role={resolvedRole}
     tabindex={interactive ? 0 : null}
     data-pw={typeof testId === 'string' ? testId : null}
     transition:slide={{ duration: 300 }}
@@ -56,12 +77,24 @@
         {@render icon()}
       </div>
     {/if}
-    <div class="banner-text">
-      {text}
-      {#if typeof linkText === 'string' && linkText.length > 0}
-        <span class="banner-link-text">{linkText}</span>
-      {/if}
-    </div>
+    {#if title !== null && typeof title === 'string' && title.length > 0}
+      <div class="banner-body">
+        <span class="banner-title">{title}</span>
+        <div class="banner-text">
+          {text}
+          {#if typeof linkText === 'string' && linkText.length > 0}
+            <span class="banner-link-text">{linkText}</span>
+          {/if}
+        </div>
+      </div>
+    {:else}
+      <div class="banner-text">
+        {text}
+        {#if typeof linkText === 'string' && linkText.length > 0}
+          <span class="banner-link-text">{linkText}</span>
+        {/if}
+      </div>
+    {/if}
     {#if typeof rightContent === 'function'}
       <div class="banner-right">
         {@render rightContent()}
@@ -102,11 +135,46 @@
     font-weight: var(--banner-font-weight, 500);
     line-height: var(--banner-line-height, 1.3);
     border-bottom: var(--banner-border-bottom, none);
+    border-radius: var(--banner-border-radius, 0);
     cursor: var(--banner-cursor, pointer);
     position: var(--banner-position, sticky);
     top: var(--banner-top, 0);
     z-index: var(--banner-z-index, 100);
     box-sizing: border-box;
+  }
+
+  .banner-tone-info {
+    background-color: var(--banner-tone-background, var(--banner-background, #f0f4f8));
+    color: var(--banner-tone-color, var(--banner-color, #637c95));
+    border: var(--banner-tone-border);
+  }
+
+  .banner-tone-success {
+    background-color: var(--banner-tone-background, var(--banner-background, #f0f4f8));
+    color: var(--banner-tone-color, var(--banner-color, #637c95));
+    border: var(--banner-tone-border);
+  }
+
+  .banner-tone-warning {
+    background-color: var(--banner-tone-background, var(--banner-background, #f0f4f8));
+    color: var(--banner-tone-color, var(--banner-color, #637c95));
+    border: var(--banner-tone-border);
+  }
+
+  .banner-tone-error {
+    background-color: var(--banner-tone-background, var(--banner-background, #f0f4f8));
+    color: var(--banner-tone-color, var(--banner-color, #637c95));
+    border: var(--banner-tone-border);
+  }
+
+  .banner-size-sm {
+    padding: var(--banner-sm-padding, 6px 12px);
+    border-bottom: none;
+  }
+
+  .banner-flush {
+    border-radius: 0;
+    border: none;
   }
 
   .banner-icon {
@@ -119,6 +187,20 @@
   .banner-icon :global(svg) {
     width: var(--banner-icon-size, 18px);
     height: var(--banner-icon-size, 18px);
+  }
+
+  .banner-body {
+    display: flex;
+    flex-direction: column;
+    overflow: hidden;
+    gap: var(--banner-body-gap, 2px);
+  }
+
+  .banner-title {
+    font-weight: var(--banner-title-font-weight, 600);
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
   }
 
   .banner-text {
--- a/src/lib/Banner/properties.ts
+++ b/src/lib/Banner/properties.ts
@@ -1,5 +1,8 @@
 import type { Snippet } from 'svelte';
 
+export type BannerTone = 'info' | 'success' | 'warning' | 'error';
+export type BannerSize = 'sm' | 'md';
+
 export type BannerProperties = MandatoryBannerProperties &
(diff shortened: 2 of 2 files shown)


sinha-sahil:
Thanks for the thorough write-up @murdore — the intent here is clear (semantic tones, a compact density, edge-to-edge embedding, and a two-line title/body layout), and the docstring is genuinely excellent. Unfortunately the implementation lands on the exact pattern this library is built to avoid, so I'd like to push back on the shape before any of this merges.

## The core issue: hardcoded variant systems live in consumer CSS, not in the component

`tone`, `size`, and `flush` are textbook examples of what [GUIDELINES.md §9](../blob/release/docs/GUIDELINES.md#9-custom-classes-via-classes-prop) explicitly tells us not to do. The section even uses Banner variants as the canonical example of how this should be expressed:

```css
/* app.css — consumer-owned */
.banner-success {
  --banner-background: #e6f9ed;
  --banner-color: #1a7f37;
  --banner-icon-color: #1a7f37;
}
.banner-warning {
  --banner-background: #fff8e6;
  --banner-color: #9a6700;
  --banner-icon-color: #9a6700;
}
```

```svelte
<Banner classes="banner-success" text="Deployment complete" />
<Banner classes="banner-warning" text="API rate limit approaching" />
```

So the literal `success` / `warning` / `error` taxonomy this PR adds is already, by design, the consumer's job. There is no design-system shape that's "correct enough" for us to bake `info | success | warning | error` into a primitive — every consumer's tokens, naming, and number of tones differ, and the moment we ship a typed enum we're committing the library to one opinion.

A second tell that this isn't doing what the API suggests: the four `.banner-tone-*` blocks all resolve to the same fallback:

```css
.banner-tone-info,
.banner-tone-success,
.banner-tone-warning,
.banner-tone-error {
  background-color: var(--banner-tone-background, var(--banner-background, #f0f4f8));
  color: var(--banner-tone-color, var(--banner-color, #637c95));
  border: var(--banner-tone-border);
}
```

Out of the box, `tone="error"` and `tone="success"` render identically. The prop is a typed promise of distinct visuals that the component refuses to fulfil — consumers still have to write CSS variables to differentiate, at which point they could have just written `.banner-error { ... }` and passed `classes="banner-error"` with zero new component surface.

The same critique applies to `size` and `flush`:

- `size="sm"` just sets `padding` and removes `border-bottom`. Both are already overridable via the existing `--banner-padding` and `--banner-border-bottom` vars. There's no `.banner-size-md` rule, so the `'md'` union member is purely decorative.
- `flush` sets `border-radius: 0; border: none;`. That's literally what the new `--banner-border-radius` variable (plus a new `--banner-border`) covers if we expose them properly.

## What the customizable alternative looks like

Keep only the primitive infrastructure — variables, not enums:

```svelte
<!-- Banner.svelte: no tone/size/flush, just more vars -->
<style>
  .banner {
    /* existing vars... */
    border: var(--banner-border, none);
    border-radius: var(--banner-border-radius, 0);
  }
</style>
```

Then consumers express the entire taxonomy in their own stylesheet, using whatever naming and tone count they want:

```css
/* app.css — consumer */
.banner-error {
  --banner-background: #fdecea;
  --banner-color: #b42318;
  --banner-icon-color: #b42318;
  --banner-border: 1px solid #f9c8c2;
}
.banner-compact {
  --banner-padding: 6px 12px;
  --banner-border-bottom: none;
}
.banner-flush {
  --banner-border-radius: 0;
  --banner-border: none;
}
```

```svelte
<Banner classes="banner-error banner-compact banner-flush" role="alert" text="…" />
```

This is the entire feature, with no component-side variant logic, no enum to keep in sync with anyone's design tokens, and full freedom for consumers to add a fifth tone or a `banner-xs` density without us shipping a release.

## On `title`

The `title` string prop is a small composite wrapper — it hardcodes a flex-column with a `<span class="banner-title">` heading. That's user-land composition smuggled into a primitive, and it locks the markup (no `<h2>`, no subtitle, no title+badge row). The existing component already uses Snippets for `icon`, `rightContent`, and `dismissIcon` — `title` should follow that precedent:

```ts
title?: Snippet;
```

```svelte
{#if typeof title === 'function'}
  <div class="banner-body">
    <div class="banner-title">{@render title()}</div>
    <div class="banner-text">{text}…</div>
  </div>
{:else}
  <div class="banner-text">{text}…</div>
{/if}
```

That keeps the `--banner-title-font-weight` / `--banner-body-gap` additions (those are great), removes the string-only restriction, and gives consumers the same composition power they already have for icons.

## On `role`

`role` as an explicit ARIA override is fine to keep — it's a legitimate accessibility escape hatch, not a styling preset, and consumers may genuinely need to set `role="status"` or `role="region"` on a sticky banner. Just simplify the resolution to `role ?? (interactive ? 'button' : null)` once `tone` is gone, so there's no implicit error→alert branch hidden in the component.

## Smaller GUIDELINES nits (only worth fixing if we keep any of this surface)

- `[…].filter(Boolean)` in the `rootClass` derivation is a truthy-coerce on strings — §3 prefers `.filter((c) => c.length > 0)`.
- `{#if title !== null && typeof title === 'string' && title.length > 0}` — the `title !== null` clause is redundant once `typeof === 'string'` is checked (§5).

## Ask

I'd like to close this in its current form and split it into a much smaller follow-up:

1. **Keep:** `--banner-border-radius` and a new `--banner-border` CSS variable on `.banner`. (Both are pure primitive improvements.)
2. **Keep, but reshape:** the title slot as a `Snippet`, with `--banner-title-font-weight` and `--banner-body-gap`.
3. **Keep:** the `role` prop (ARIA escape hatch) — just drop the `tone === 'error' → 'alert'` branch from the resolution.
4. **Drop entirely:** `tone`, `size`, and `flush` props plus every `.banner-tone-*` / `.banner-size-*` / `.banner-flush` CSS block.
5. **Document:** add a short README example showing the `classes="banner-error"` / `classes="banner-compact"` pattern so the next contributor reaches for it instead of an enum.

Happy to pair on the trimmed PR if useful — the variable additions and the title snippet are genuinely good ideas and I'd like them landed in a shape that survives the philosophy.

murdore:
@sinha-sahil — thank you for this. The diagnosis is exactly right and I appreciate how precisely you mapped each prop to the philosophy.

You've made me see clearly what I missed: `tone`, `size`, and `flush` are a variant system I was smuggling into a primitive. The §9 example in GUIDELINES literally uses Banner tones as the canonical illustration of what belongs in consumer CSS — I read it and still built the thing it says not to build. The deeper tell you pointed out seals it: all four `.banner-tone-*` blocks resolve to identical fallbacks, so the typed enum is a promise the component cannot keep without the consumer writing CSS variables anyway. At that point the enum is pure noise on the library's public API.

I'm happy to close this in its current form and open a small, focused follow-up containing only:

- `--banner-border-radius` and `--banner-border` on `.banner` (pure CSS-var additions)
- `title` reshaped as a `Snippet` with `--banner-title-font-weight` and `--banner-body-gap`
- `role` prop kept as an ARIA escape hatch, with the `tone === 'error' → 'alert'` branch removed
- A README example showing `classes="banner-error banner-compact"` so the next contributor reaches for the pattern instead of an enum

One clarifying question before I draft that PR: for the `title` slot, should I keep a wrapping `<div class="banner-body">` in the component's markup (so consumers get the flex-column gap for free via `--banner-body-gap`), or do you prefer the component renders the snippet inline and leaves the layout entirely to the consumer's snippet content?

sinha-sahil:

This PR converts existing `function` declarations to arrow consts: `handleClick`, `handleKeydown`, `handleDismiss`. Please revert those conversions — they're stylistic churn, and the existing convention in this codebase is function declarations.

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — appreciate the clean read-back, and yes, the split plan you've listed is exactly the shape I want this to land in. The fact that you re-read §9 and immediately saw the same Banner-as-canonical-example callout is the right instinct; nothing more to dig into on the philosophy side. Let me answer your open question and flag two small things to fold into the slim PR.

## On the `title` snippet wrapper

**Keep the wrapping `<div class="banner-body">` in the component.** Render inline would push pure layout (flex-column, gap between title and text) into every consumer's snippet, which inverts the relationship we want: the component owns the structural primitive (title slot + body text are a stacked pair), and the consumer owns the visual decisions via CSS vars on that primitive.

Concretely:

```svelte
<div class="banner-body">
  {#if typeof title === 'function'}
    <div class="banner-title">{@render title()}</div>
  {/if}
  <div class="banner-text">{text}</div>
</div>
```

```css
.banner-body {
  display: flex;
  flex-direction: column;
  gap: var(--banner-body-gap, 0);
}
.banner-title {
  font-weight: var(--banner-title-font-weight, inherit);
}
```

Two things that make this a primitive and not a composite:

1. **Default `gap: 0` and `font-weight: inherit`** — out of the box, with no title and no consumer CSS, the component renders identical markup to today. The new structure is invisible until a consumer opts in by providing a `title` snippet and/or setting the vars.
2. **The `banner-body` / `banner-title` class hooks are stable targeting surface** — consumers can also reach them directly from their `.banner-error { … }` class without us shipping a typed enum.

If the snippet were rendered inline, `--banner-body-gap` becomes meaningless (no flex container to gap), and every consumer who wants a title would re-implement the same two-line flex layout in their own snippet. That's the composite-in-disguise trap.

## Two adds for the slim PR

- **Function declarations** — please also revert the `handleClick` / `handleKeydown` / `handleDismiss` conversions to `function` declarations in the slim PR. Per the follow-up note above, that's a separate stylistic regression we don't want carried forward even into the keeper subset.
- **The two GUIDELINES nits** I flagged (`filter(Boolean)` → `filter((c) => c.length > 0)` in §3, and the redundant `title !== null` clause once `typeof === 'string'` is checked, per §5) — fold those into the slim PR's `rootClass` and title-guard while you're in there.

## Confirming the path

Sounds like we're aligned: close this PR, open a fresh slim one with the keepers (vars + Snippet + role escape hatch + README), and we land the good parts of the work without dragging the variant system along.

**Ask: please close this PR and open the slim follow-up with the four keepers above.** Happy to review the slim PR same-day once it's up.

murdore:
Closing in favor of the slim #233 (--banner-border-radius/--banner-border vars + title snippet with banner-body wrapper + role escape hatch + README) — tone/size/flush enums dropped per §9. Thanks @sinha-sahil.

PR #222, opened 2026-05-29
feat: add InputField composite component

murdore:
## Summary

- Adds `InputField` — a new composite form-field primitive that wraps the library `Input` with an outer `<label>`, required asterisk, hint text, error message, and optional trailing icon-button (using the library `Button`).
- Exports `InputField` from `src/lib/index.ts` and exposes `InputFieldProperties` as a public type.

## API

| `label` | `string` | — | Label text rendered above the input |
| `mandatory` | `boolean` | `false` | Shows a red `*` after the label |
| `value` | `string` (bindable) | `''` | Input value, two-way bindable |
| `placeholder` | `string` | `''` | Input placeholder text |
| `hintText` | `string` | — | Helper text shown below the input |
| `errorText` | `string` | — | Error message shown with `role="alert"` |
| `disabled` | `boolean` | `false` | Disables the input |
| `type` | `InputDataType` | `'text'` | Input type (`text`, `tel`, `password`, `email`, `number`) |
| `testId` | `string` | — | Sets `data-pw` on root; derives `{testId}-input`, `{testId}-error`, `{testId}-trailing` |
| `classes` | `string` | — | Extra CSS classes on the root wrapper |
| `trailingIcon` | `Snippet` | — | Svelte snippet rendered inside the trailing `Button` icon slot |
| `ontrailingClick` | `(event: MouseEvent) => void` | — | Click handler for the trailing button |
| `oninput` | `(value: string) => void` | — | Called on every input change with the current value |

**CSS custom properties**: `--input-field-gap`, `--input-field-label-color`, `--input-field-label-font-size`, `--input-field-label-font-weight`, `--input-field-asterisk-color`, `--input-field-hint-color`, `--input-field-error-color`, `--input-field-width`.

## Notes

- `prettier --check` exits **0** on all new files
- `eslint` exits **0** on all new files
- No new npm dependencies added
- Fully backward-compatible — existing exports are unchanged
- Uses Svelte 5 runes (`$props`, `$bindable`), kebab-case CSS selectors, no hardcoded font properties (all exposed as CSS vars)

### DIFF
--- a/src/lib/InputField/InputField.svelte
+++ b/src/lib/InputField/InputField.svelte
@@ -1,0 +1,103 @@
+<script lang="ts">
+  import Input from '../Input/Input.svelte';
+  import Button from '../Button/Button.svelte';
+  import type { InputFieldProperties } from './properties';
+
+  let {
+    label,
+    mandatory = false,
+    value = $bindable(''),
+    placeholder = '',
+    hintText,
+    errorText,
+    disabled = false,
+    type = 'text',
+    testId,
+    classes,
+    trailingIcon,
+    ontrailingClick,
+    oninput
+  }: InputFieldProperties = $props();
+
+  const inputId = testId ? `${testId}-input` : '';
+
+  const handleInput = (inputValue: string) => {
+    value = inputValue;
+    oninput?.(inputValue);
+  };
+</script>
+
+<div class="input-field {classes ?? ''}" data-pw={testId}>
+  {#if label}
+    <label class="input-field-label" for={inputId || null}>
+      {label}{#if mandatory}<span class="input-field-asterisk">*</span>{/if}
+    </label>
+  {/if}
+
+  <div class="input-field-row">
+    <Input
+      value={value ?? ''}
+      {placeholder}
+      disable={disabled}
+      dataType={type}
+      name={inputId}
+      testId={inputId}
+      onInput={handleInput}
+      classes="input-field-inner"
+    />
+    {#if trailingIcon}
+      <Button
+        testId={testId ? `${testId}-trailing` : ''}
+        onclick={ontrailingClick}
+        icon={trailingIcon}
+        classes="input-field-trailing"
+        ariaLabel="trailing action"
+      />
+    {/if}
+  </div>
+
+  {#if hintText}
+    <p class="input-field-hint">{hintText}</p>
+  {/if}
+
+  {#if errorText}
+    <p class="input-field-error" data-pw={testId ? `${testId}-error` : null} role="alert">
+      {errorText}
+    </p>
+  {/if}
+</div>
+
+<style>
+  .input-field {
+    display: flex;
+    flex-direction: column;
+    gap: var(--input-field-gap, 4px);
+    width: var(--input-field-width, fit-content);
+  }
+
+  .input-field-label {
+    color: var(--input-field-label-color, #637c95);
+    font-size: var(--input-field-label-font-size, 12px);
+    font-weight: var(--input-field-label-font-weight, 400);
+  }
+
+  .input-field-asterisk {
+    color: var(--input-field-asterisk-color, #e53e3e);
+    margin-inline-start: 2px;
+  }
+
+  .input-field-row {
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    width: 100%;
+  }
+
+  .input-field-hint {
+    color: var(--input-field-hint-color, #637c95);
+  }
+
+  .input-field-error {
+    color: var(--input-field-error-color, #e53e3e);
+  }
+</style>
--- a/src/lib/InputField/properties.ts
+++ b/src/lib/InputField/properties.ts
@@ -1,0 +1,18 @@
+import type { Snippet } from 'svelte';
+import type { InputDataType } from '$lib/types';
+
+export type InputFieldProperties = {
+  label?: string;
+  mandatory?: boolean;
+  value?: string;
+  placeholder?: string;
+  hintText?: string;
+  errorText?: string;
+  disabled?: boolean;
+  type?: InputDataType;
+  testId?: string;
+  classes?: string;
+  trailingIcon?: Snippet;
+  ontrailingClick?: (event: MouseEvent) => void;
+  oninput?: (value: string) => void;
+};
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -56,6 +56,7 @@ export { default as EmptyState } from './EmptyState/EmptyState.svelte';
 export { default as Combobox } from './Combobox/Combobox.svelte';
 export { default as ColorPicker } from './ColorPicker/ColorPicker.svelte';
 export { default as SplitInput } from './SplitInput/SplitInput.svelte';
+export { default as InputField } from './InputField/InputField.svelte';
 
 export type * from './Button/properties';
 export type * from './Modal/properties';
@@ -109,5 +110,6 @@ export type * from './EmptyState/properties';
 export type * from './Combobox/properties';
 export type * from './ColorPicker/properties';
 export type * from './SplitInput/properties';
+export type * from './InputField/properties';
 
 export { validateInput } from './utils';


sinha-sahil:
Thanks for the PR @murdore — I can see what you're trying to give consumers: a one-stop form field with label, required marker, hint, error, and an inline action button. That is a genuinely useful pattern. The problem is that this is exactly the kind of component the library does **not** ship, and I want to explain why before we go further.

## Why this is a block, not a nit

`@juspay/svelte-ui-components` is a **primitives** library. The philosophy (see [`docs/GUIDELINES.md`](docs/GUIDELINES.md) and the architecture notes) is:

> Components are unstyled primitives. Consumers own the design system. The library ships primitives; consumers compose them.

`InputField.svelte` is a pure composite wrapper: it renders `<label>` + `<Input/>` + `<Button/>` + `<p.hint>` + `<p.error>` and does nothing else. The entire script section is:

```ts
const inputId = testId ? `${testId}-input` : '';
const handleInput = (inputValue: string) => {
  value = inputValue;
  oninput?.(inputValue);
};
```

That is a prop pass-through. No focus management, no `aria-describedby` wiring between the input and the hint/error (which would actually justify a primitive), no `aria-invalid`, no imperative focus API, no state. Composite wrappers of existing primitives are explicitly user-land code.

On top of that, the existing **`Input` component already exposes a `label` prop** (`src/lib/Input/properties.ts:15`). So this PR also duplicates a primitive that already exists.

## Hardcoded presets baked in

Even if the wrapper survived the philosophy objection, the API locks in several visual decisions consumers cannot override:

- The literal `*` character for the required marker (no localization, no `(required)` text, no custom snippet).
- `ariaLabel="trailing action"` — hardcoded English string.
- `margin-inline-start: 2px` on the asterisk — not a CSS variable.
- Label always above the input (no inline-label option).
- Trailing icon always to the right, outside the input (no leading icon, no in-field icon).
- Hint and error are always plain `<p>` — no slot for icons / links inside them.

These are the kinds of "preset" decisions [GUIDELINES §9](docs/GUIDELINES.md#9-custom-classes-via-classes-prop) tells us to push out via CSS variables + `classes` — and even then, a primitive shouldn't be making this many shape decisions.

## §3 violations (would be required even if the component survived)

There are ~8 falsy/truthy checks on strings/snippets that all violate [§3](docs/GUIDELINES.md#3-no-falsytruthy-checks-for-non-boolean-values):

```svelte
{#if label}              → {#if typeof label === 'string' && label.length > 0}
{#if trailingIcon}       → {#if typeof trailingIcon === 'function'}
{#if hintText}           → {#if typeof hintText === 'string' && hintText.length > 0}
{#if errorText}          → {#if typeof errorText === 'string' && errorText.length > 0}
testId ? `${testId}-input` : ''                → typeof testId === 'string' ? … : ''
data-pw={testId ? `${testId}-error` : null}    → same
for={inputId || null}                          → inputId.length > 0 ? inputId : null
```

## What I'd suggest instead

**Option A — extend `Input` (preferred).** The genuinely useful additions here are `mandatory`, `hintText`, and `errorText` (with proper aria-describedby wiring). Add those props to `Input` itself and render them inside Input's template. That's a single primitive that owns the label/input/hint/error relationship and gets the accessibility right.

```svelte
<!-- Input.svelte (extended) -->
<div class="input {classes ?? ''}">
  {#if typeof label === 'string' && label.length > 0}
    <label for={id}>
      {label}{#if mandatory}<span class="input-required">*</span>{/if}
    </label>
  {/if}
  <input
    {id}
    aria-describedby={describedBy}
    aria-invalid={typeof errorText === 'string' && errorText.length > 0}
    …
  />
  {#if typeof hintText === 'string' && hintText.length > 0}
    <p id="{id}-hint" class="input-hint">{hintText}</p>
  {/if}
  {#if typeof errorText === 'string' && errorText.length > 0}
    <p id="{id}-error" class="input-error" role="alert">{errorText}</p>
  {/if}
</div>
```

That refactor lives inside the existing primitive, adds real accessibility value (the `aria-describedby` / `aria-invalid` plumbing), and ships through the existing CSS-variable + `classes` API.

**Option B — leave composition to the consumer.** If a team needs the "label + input + trailing action + hint + error" shape, they write 20 lines of Svelte that compose `<Input/>` and `<Button/>` directly. The `classes` prop + a couple of CSS variables in their app stylesheet give them the variant system they need:

```css
/* consumer-app.css */
.field-with-action { display: flex; flex-direction: column; gap: 4px; }
.field-with-action .field-row { display: flex; align-items: center; }
.field-error { color: var(--color-danger); }
```

```svelte
<!-- consumer-app code -->
<div class="field-with-action">
  <Input label="Email" mandatory hintText="..." errorText={emailError} bind:value={email} />
  <Button icon={searchIcon} onclick={onSearch} classes="field-trailing" ariaLabel="Search" />
</div>
```

## Ask

Could we **close this PR** and open a smaller one against `Input.svelte` that adds `mandatory`, `hintText`, `errorText` with proper aria-describedby wiring? That keeps the library to one canonical input primitive, avoids the composite-wrapper / dummy-component issues, and gives consumers everything this PR offers without locking in the layout.

Happy to chat through the Input extension API in a separate issue if it helps.

murdore:
@sinha-sahil — thank you for the thorough explanation. You're right on every count, and I appreciate you taking the time to lay out the reasoning rather than just requesting changes.

The core mistake was treating "label + input + hint + error" as a library concern when it's really a consumer-side composition. The existing `Input` already owns the label relationship; adding a wrapper that just passes props through doesn't bring anything the consumer can't write in 20 lines. And your point about what's missing — proper `aria-describedby` wiring, `aria-invalid`, no imperative focus API, no state — is exactly what would justify a primitive, and this PR has none of it.

The §3 violations you called out (the truthy checks on `label`, `trailingIcon`, `hintText`, `errorText`) are straightforward to fix wherever they appear.

**My preferred path:** Option A — close this PR and open a focused one against `Input.svelte` that adds `mandatory`, `hintText`, and `errorText` with correct `aria-describedby` / `aria-invalid` wiring baked in. The trailing-icon use-case stays consumer-land.

Would it make sense to open a GitHub issue first to agree on the `Input` extension API (prop names, accessibility contract, whether `id` becomes required or auto-generated), so the follow-up PR lands cleanly?

sinha-sahil:

New arrow-const handlers in the diff: `handleInput`. Please write them as `function fnName() {}` instead.

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — appreciate the gracious read on this. You landed in exactly the right place, and the framing in your reply ("treating label + input + hint + error as a library concern when it's really consumer-side composition") is the heuristic I want every contributor to apply before opening a primitives-library PR. Saves all of us a review cycle.

**Yes — please open the issue first.** That's the right instinct. The Input extension is a real API surface change (it expands the contract of an existing primitive), and a couple of decisions on it are load-bearing enough that doing them in PR review would churn. Specifically, I'd want the issue to land these before any code:

1. **`id` contract.** I'd argue `id` should remain optional but, when omitted, the component auto-generates a stable one (Svelte 5's `$props.id()` or a `crypto.randomUUID()` once on mount). Reason: aria-describedby and `<label for>` wiring are useless without an id, but forcing consumers to hand-roll ids for every field is hostile. Auto-generated with an escape hatch is the sweet spot.
2. **Prop names.** `mandatory` reads fine to me but `required` is the HTML-native name and is what consumers reach for first. Worth a short discussion — I lean `required` and let it map straight to the native attribute, with the asterisk rendering driven by the same boolean. Either way, no `*` literal — consumers should be able to swap the marker via a snippet or CSS `::after` content var.
3. **Accessibility contract — write it down explicitly.** What ids get wired to `aria-describedby` (hint always? error always? both concatenated?), when `aria-invalid` flips, whether `role="alert"` lives on the error node or `aria-live` on a wrapper. This is the actual value-add of putting these props on the primitive vs. leaving it to consumers, so it deserves a paragraph in the issue, not a guess in the PR.
4. **Snippet vs string for hint/error.** I'd lean `string | Snippet` for both, so consumers can render `<a>` inside hints or an icon inside errors without us shipping a baked layout. Strings stay the easy path.
5. **CSS variables.** Name the hooks for the required-marker color, hint color, error color, gap between input and hint/error, and label/input spacing. None of these should be baked.

What I'd explicitly want **out** of the issue and PR: trailing action button, leading icons, label-position variants, inline-vs-stacked layout. Those stay consumer-land — they're the composite/preset surface this PR was carrying that we agreed doesn't belong.

On the §3 violations — those will resolve naturally in the Input extension since they were specific to the InputField.svelte file being closed.

**Ask: please close this PR**, and open the issue against `Input` covering points 1–5 above. Tag me on the issue and I'll turn it around quickly so the follow-up PR has a clean spec to land against. Thanks again for the constructive turn here.

murdore:
Closing — InputField is a prop-pass-through composite. I'll open an issue against Input covering the label/asterisk/hint/error/icon surface per your note. Thanks @sinha-sahil.

PR #223, opened 2026-05-29
feat: add MasonryGrid component

murdore:
## Summary

- New `MasonryGrid` layout primitive that arranges slotted children into balanced columns using CSS `column-count` / `column-gap`
- Pure CSS implementation — SSR-safe, no JS measuring, no `ResizeObserver`, no new dependencies
- Children get `break-inside: avoid` + `margin-bottom: var(--masonry-gap)` automatically via a `:global(*)` child rule

## API

| `columns` | `number` | `3` | Number of masonry columns (`column-count`) |
| `gap` | `string` | `'16px'` | Gap between columns and rows (`column-gap` + child margin-bottom) |
| `testId` | `string` | — | Renders as `data-pw` on the root element |
| `classes` | `string` | — | Extra CSS classes appended to the root `div` |
| `children` | `Snippet` | — | Default snippet — any slotted child elements |

**CSS custom properties exposed:** `--masonry-columns`, `--masonry-gap`

**Responsive tip:** Override `--masonry-columns` in a media query on the parent to change column count at different breakpoints.

## Notes

- `prettier --check` passed on all new files and `index.ts`
- `eslint` exited 0 on `src/lib/MasonryGrid/`
- Backward-compatible: new named export added after existing `SplitInput` export; no existing APIs changed

### DIFF
--- a/src/lib/MasonryGrid/MasonryGrid.svelte
+++ b/src/lib/MasonryGrid/MasonryGrid.svelte
@@ -1,0 +1,25 @@
+<script lang="ts">
+  import type { MasonryGridProperties } from './properties';
+
+  let { columns = 3, gap = '16px', testId, classes, children }: MasonryGridProperties = $props();
+</script>
+
+<div
+  class="masonry-grid {classes ?? ''}"
+  data-pw={testId}
+  style="--masonry-columns:{columns}; --masonry-gap:{gap};"
+>
+  {@render children?.()}
+</div>
+
+<style>
+  .masonry-grid {
+    column-count: var(--masonry-columns);
+    column-gap: var(--masonry-gap);
+  }
+
+  .masonry-grid > :global(*) {
+    break-inside: avoid;
+    margin-bottom: var(--masonry-gap);
+  }
+</style>
--- a/src/lib/MasonryGrid/properties.ts
+++ b/src/lib/MasonryGrid/properties.ts
@@ -1,0 +1,9 @@
+import type { Snippet } from 'svelte';
+
+export type MasonryGridProperties = {
+  columns?: number;
+  gap?: string;
+  testId?: string;
+  classes?: string;
+  children?: Snippet;
+};
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -56,6 +56,7 @@ export { default as EmptyState } from './EmptyState/EmptyState.svelte';
 export { default as Combobox } from './Combobox/Combobox.svelte';
 export { default as ColorPicker } from './ColorPicker/ColorPicker.svelte';
 export { default as SplitInput } from './SplitInput/SplitInput.svelte';
+export { default as MasonryGrid } from './MasonryGrid/MasonryGrid.svelte';
 
 export type * from './Button/properties';
 export type * from './Modal/properties';
@@ -109,5 +110,6 @@ export type * from './EmptyState/properties';
 export type * from './Combobox/properties';
 export type * from './ColorPicker/properties';
 export type * from './SplitInput/properties';
+export type * from './MasonryGrid/properties';
 
 export { validateInput } from './utils';


sinha-sahil:
## Thanks for the contribution — but this needs to be a CSS utility, not a component

Hi @murdore — thanks for the work and the clean PR description. I can see what you're going for: a tidy, SSR-safe masonry layout primitive with no JS overhead. The implementation is small and correct. The problem is that "small and correct with no JS" is precisely the signal that this should **not** be a library component.

### Why this is a block, not a nit

The library's design philosophy (and `docs/GUIDELINES.md`) is built around five rules, and this PR runs into three of them at once:

1. **No dummy components.** The entire component body is:
   ```svelte
   <div class="masonry-grid {classes ?? ''}" style="--masonry-columns:{columns}; --masonry-gap:{gap};">
     {@render children?.()}
   </div>
   ```
   with two CSS declarations and a `:global(*)` child rule. There's no focus management, no ARIA, no state, no events, no measuring — your own PR description proudly says "no JS measuring, no `ResizeObserver`, no new dependencies." That is the tell. A component that does nothing JavaScript can do, that a `<div class="masonry">` plus a stylesheet cannot, is not a primitive — it's a CSS class wearing a `.svelte` extension. Every component in this library has to justify its existence with non-trivial behavior, accessibility, or state. This one doesn't.

2. **No hardcoded design presets (GUIDELINES §9).** `columns?: number = 3` and `gap?: string = '16px'` are visual decisions baked into the JS prop surface, then injected via an inline `style=""` attribute. This is exactly the `variant` / `size` / `tone` anti-pattern §9 forbids. Defaults belong in the `var(--…, fallback)` declaration, and variants belong in **consumer CSS classes** passed via `classes` — not in JS props.

3. **Customizability gap.** The `:global(*)` rule forces `break-inside: avoid` and `margin-bottom: var(--masonry-gap)` on every direct child, yoking row gap to column gap with no override. A consumer who wants a different row gap, or who wants to opt one child out of `break-inside: avoid`, can't.

### The customizable alternative — no component needed

Ship this as a documented CSS utility (or just leave it to consumers entirely). In a consumer's `app.css`:

```css
.masonry {
  column-count: var(--masonry-columns, 3);
  column-gap: var(--masonry-gap, 16px);
}
.masonry > * {
  break-inside: avoid;
  margin-bottom: var(--masonry-row-gap, var(--masonry-gap, 16px));
}

/* Variants live HERE, not in JS props */
.masonry-2-col  { --masonry-columns: 2; }
.masonry-4-col  { --masonry-columns: 4; }
.masonry-tight  { --masonry-gap: 8px; }

@media (max-width: 640px) {
  .masonry-responsive { --masonry-columns: 1; }
}
```

```svelte
<!-- Usage — zero library code required -->
<div class="masonry masonry-3-col masonry-tight">
  {#each items as item}
    <Card {...item} />
  {/each}
</div>
```

This is the §9 pattern used by `Button` (`.btn-primary`, `.btn-secondary`), `Banner` (`.banner-success`, `.banner-warning`), and every other themed component in the library. Consumers get full control over column count, both gaps, responsive breakpoints, and per-child `break-inside` overrides — all without us shipping a primitive that does no JavaScript work.

### If we genuinely want a component for this

(I don't think we do, but for completeness): it would need to (a) remove the `columns` and `gap` props entirely — defaults move into `var(--masonry-columns, 3)` / `var(--masonry-gap, 16px)` fallbacks, (b) remove the inline `style=""` attribute, (c) expose a separate `--masonry-row-gap` variable, and (d) provide a documented way to opt children out of the child rule. At that point the component is *literally* a `<div class="masonry-grid {classes ?? ''}">` with a `<style>` block — and at *that* point the question is back to "why is this a component."

### Ask

I'd like to **close this PR** and instead, if useful, add a short "Layout recipes" section to the docs with the CSS utility above. If you've hit a concrete consumer use case where the CSS-utility approach doesn't work (e.g. you genuinely need JS-driven balancing à la Pinterest), happy to reopen that as a different design discussion — but the current implementation explicitly avoids that work, so the component layer adds no value over a CSS class.

Thanks again for the cleanly scoped PR — the code itself is fine; it just belongs in user-land, not in `src/lib`.

murdore:
@sinha-sahil — thank you for the thorough, principled breakdown. You're right on all three points, and I think the "no dummy components" framing is the most clarifying one for me personally.

Looking at the component again with fresh eyes: the entire body is a single `<div>` with two CSS custom property injections and a `:global(*)` child rule. There is no JavaScript behavior, no accessibility logic, no state management — nothing that a plain `<div class="masonry">` plus a stylesheet cannot do equally well. That is exactly the tell you described, and I should have caught it before opening the PR.

The `columns` and `gap` props compound the problem: per GUIDELINES §9, visual presets (column count, spacing scale) belong in consumer CSS classes that set the `var(--masonry-columns)` / `var(--masonry-gap)` fallbacks — not in the JS prop surface. Your CSS-utility example in the comment is cleaner and more flexible than what I shipped.

I'm happy to close this PR. If it would be useful, I can open a small docs PR instead that adds a "Layout recipes" section with the CSS utility pattern you outlined — keeping it entirely in user-land.

One clarifying question: would a docs-only PR for the masonry recipe be welcome here, or do you prefer to keep the docs focused on actual library components and leave layout patterns to external guides?

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — appreciate the gracious read on this, and especially the "I should have caught it before opening the PR" reflection. That's the right instinct, and it's the same instinct that keeps the library small. Please go ahead and close this PR.

On your clarifying question — **yes, a docs-only PR for a "Layout recipes" section would be welcome**, with some scoping guidance so it lands in the right shape:

**What works:**

- A short page (or a section in an existing docs page) titled something like "Layout recipes" or "CSS patterns" — explicitly framed as *user-land patterns*, not library features. The framing matters: readers should walk away knowing this is "here's how to do it in your own CSS" and not "here's a hidden API."
- Masonry as the first recipe is fine. Keep the CSS utility exactly as scoped in my earlier comment: `column-count` / `column-gap` driven by `--masonry-columns` / `--masonry-gap`, a separate `--masonry-row-gap` falling back to `--masonry-gap`, and variant classes (`.masonry-2-col`, `.masonry-tight`, responsive variants) demonstrating the GUIDELINES §9 pattern in user CSS.
- A one-liner at the top of the section explaining *why* these aren't components — same reasoning as this PR. That doubles as design-philosophy documentation for the next contributor who's tempted to wrap a CSS pattern in a `.svelte` file.

**What to avoid:**

- Don't put the recipe under `src/lib/` or anywhere that suggests it ships with the package. `docs/` only.
- Don't import any library component into the example *unless* it's load-bearing (e.g. showing `<Card>` children inside a masonry container is fine because that's realistic usage; wrapping the CSS recipe in a fake `<MasonryLayout>` Svelte snippet to make it look component-like would defeat the point).
- Don't grow the page into a layout-system kitchen sink in one PR. One recipe (masonry), cleanly documented, is the right size. If the pattern proves useful, we can add `flex-equal-columns`, `aspect-grid`, etc. later as separate PRs.
- No new dependencies, no new tooling, no Storybook entries that imply a component exists.

If you want a concrete shape: a single new `docs/LayoutRecipes.md` (or whatever fits the existing docs convention — happy to point you at the right location if it's unclear) containing the framing paragraph + the masonry CSS block + the consumer usage example from my prior comment. That's it.

One small ask if you do open the docs PR: link back to this closed PR in the description, so future contributors who search for "masonry" land on the full reasoning and don't re-open this same component. That's how we turn a closed PR into useful library memory.

Thanks again for taking the feedback in stride — this kind of exchange is exactly why the philosophy gets stronger over time rather than eroding one composite at a time.

**Ask: please close this PR.** A follow-up docs-only PR with the masonry CSS recipe — scoped as above — is welcome.

murdore:
Closing — MasonryGrid is a CSS-columns utility, not a component. A masonry docs recipe will follow. Thanks @sinha-sahil.

PR #224, opened 2026-05-29
feat: add StepList component

murdore:
## Summary

- New `StepList` vertical step/status list primitive for onboarding, activity timelines, and workflow progress UIs
- Distinct from the existing `Stepper` (horizontal numbered stepper) — this is a vertical list with rich status semantics
- Exported from the library root as `StepList` (component) and `StepListProperties`, `StepListStep`, `StepStatus` (types)

## API

**`StepListProperties`** (`src/lib/StepList/properties.ts`):

| `steps` | `StepListStep[]` | Yes | Array of step objects |
| `testId` | `string` | No | `data-pw` attribute on the root `<ol>` |
| `classes` | `string` | No | Extra CSS classes appended to root |
| `stepBody` | `Snippet` | No | Optional snippet rendered inside each step's content area |

**`StepListStep`**:

| Field | Type | Required | Description |
|-------|------|----------|-------------|
| `id` | `string` | Yes | Unique key for Svelte keyed each |
| `title` | `string` | Yes | Step title text |
| `description` | `string` | No | Optional secondary text |
| `status` | `'pending' \| 'in-progress' \| 'success' \| 'failure' \| 'inactive'` | No | Defaults to `pending` |
| `testId` | `string` | No | `data-pw` on the `<li>` |

**CSS custom properties** (all have sensible defaults):

- `--step-list-gap` — gap between list items
- `--step-list-indicator-size` — width/height of the status circle
- `--step-list-pending-color` — indicator colour for pending (grey)
- `--step-list-in-progress-color` — indicator colour for in-progress (blue)
- `--step-list-success-color` — indicator colour for success (green)
- `--step-list-failure-color` — indicator colour for failure (red)
- `--step-list-inactive-color` — indicator colour for inactive (light grey)
- `--step-list-connector-color` — vertical connector line colour
- `--step-list-title-color` — title text colour
- `--step-list-title-font-weight` — title font weight
- `--step-list-description-color` — description text colour

## Notes

- `prettier --check` passes on all new files and `index.ts`
- `eslint` passes with **0 errors, 0 warnings** on `src/lib/StepList/`
- Backward-compatible: new exports only, no existing code changed except two appended lines in `index.ts`
- No new npm dependencies
- Svelte 5 runes (`$props`)
- `in-progress` indicator has a CSS-only pulse animation for visual feedback
- `aria-current="step"` set on the active `in-progress` item for accessibility

### DIFF
--- a/src/lib/StepList/StepList.svelte
+++ b/src/lib/StepList/StepList.svelte
@@ -1,0 +1,185 @@
+<script lang="ts">
+  import type { StepListProperties } from './properties';
+
+  let { steps, testId, classes, stepBody }: StepListProperties = $props();
+</script>
+
+<ol class="step-list {classes ?? ''}" data-pw={testId} aria-label="Step list">
+  {#each steps as step, index (step.id)}
+    <li
+      class="step-list-item step-list-item-{step.status ?? 'pending'}"
+      data-pw={step.testId}
+      aria-current={step.status === 'in-progress' ? 'step' : null}
+    >
+      <div class="step-list-indicator" aria-hidden="true">
+        {#if step.status === 'success'}
+          <svg
+            class="step-list-indicator-icon"
+            viewBox="0 0 16 16"
+            fill="none"
+            xmlns="http://www.w3.org/2000/svg"
+            aria-hidden="true"
+          >
+            <path
+              d="M3 8L6.5 11.5L13 5"
+              stroke="currentColor"
+              stroke-width="2"
+              stroke-linecap="round"
+              stroke-linejoin="round"
+            />
+          </svg>
+        {:else if step.status === 'failure'}
+          <svg
+            class="step-list-indicator-icon"
+            viewBox="0 0 16 16"
+            fill="none"
+            xmlns="http://www.w3.org/2000/svg"
+            aria-hidden="true"
+          >
+            <path
+              d="M4 4L12 12M12 4L4 12"
+              stroke="currentColor"
+              stroke-width="2"
+              stroke-linecap="round"
+            />
+          </svg>
+        {:else if step.status === 'in-progress'}
+          <span class="step-list-indicator-dot step-list-indicator-dot-pulse"></span>
+        {:else}
+          <span class="step-list-indicator-dot"></span>
+        {/if}
+      </div>
+      <div class="step-list-content">
+        <div class="step-list-title">{step.title}</div>
+        {#if step.description}
+          <div class="step-list-description">{step.description}</div>
+        {/if}
+        {#if stepBody}
+          <div class="step-list-step-body">
+            {@render stepBody()}
+          </div>
+        {/if}
+      </div>
+      {#if index < steps.length - 1}
+        <div class="step-list-connector" aria-hidden="true"></div>
+      {/if}
+    </li>
+  {/each}
+</ol>
+
+<style>
+  .step-list {
+    list-style: none;
+    margin: 0;
+    padding: 0;
+    display: flex;
+    flex-direction: column;
+    gap: var(--step-list-gap, 0px);
+  }
+
+  .step-list-item {
+    position: relative;
+    display: grid;
+    grid-template-columns: var(--step-list-indicator-size, 20px) 1fr;
+    grid-template-rows: auto auto;
+    column-gap: var(--step-list-item-column-gap, 12px);
+    align-items: start;
+  }
+
+  .step-list-indicator {
+    grid-column: 1;
+    grid-row: 1;
+    width: var(--step-list-indicator-size, 20px);
+    height: var(--step-list-indicator-size, 20px);
+    border-radius: 50%;
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    flex-shrink: 0;
+    background-color: var(--step-list-indicator-bg, currentColor);
+    color: var(--step-list-indicator-icon-color, #fff);
+    position: relative;
+    z-index: 1;
+  }
+
+  .step-list-item-pending .step-list-indicator {
+    background-color: var(--step-list-pending-color, #9ca3af);
+  }
+
+  .step-list-item-in-progress .step-list-indicator {
+    background-color: var(--step-list-in-progress-color, #3b82f6);
+  }
+
+  .step-list-item-success .step-list-indicator {
+    background-color: var(--step-list-success-color, #22c55e);
+  }
+
+  .step-list-item-failure .step-list-indicator {
+    background-color: var(--step-list-failure-color, #ef4444);
+  }
+
+  .step-list-item-inactive .step-list-indicator {
+    background-color: var(--step-list-inactive-color, #e5e7eb);
+  }
+
+  .step-list-indicator-dot {
+    display: block;
+    width: 8px;
+    height: 8px;
+    border-radius: 50%;
+    background-color: var(--step-list-dot-color, #fff);
+  }
+
+  .step-list-indicator-dot-pulse {
+    animation: step-list-pulse 1.5s ease-in-out infinite;
+  }
+
+  @keyframes step-list-pulse {
+    0%,
+    100% {
+      opacity: 1;
+      transform: scale(1);
+    }
+    50% {
+      opacity: 0.6;
+      transform: scale(0.8);
+    }
+  }
+
+  .step-list-indicator-icon {
+    width: 12px;
+    height: 12px;
+    flex-shrink: 0;
+  }
+
+  .step-list-content {
+    grid-column: 2;
+    grid-row: 1;
+    padding-bottom: var(--step-list-content-padding-bottom, 16px);
+  }
+
+  .step-list-title {
+    color: var(--step-list-title-color, inherit);
+    font-weight: var(--step-list-title-font-weight, inherit);
+  }
+
+  .step-list-description {
+    margin-top: var(--step-list-description-margin-top, 4px);
+    color: var(--step-list-description-color, #6b7280);
+  }
+
+  .step-list-step-body {
+    margin-top: var(--step-list-body-margin-top, 8px);
+  }
+
+  .step-list-connector {
+    grid-column: 1;
+    grid-row: 2;
+    width: 2px;
+    min-height: var(--step-list-connector-min-height, 16px);
+    height: 100%;
+    background-color: var(--step-list-connector-color, #e5e7eb);
+    margin: 0 auto;
+    align-self: stretch;
+  }
+</style>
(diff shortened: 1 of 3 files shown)


arx-optimus-17:
This is already available in stepper component

sinha-sahil:
## Overall verdict: please rework before merge

Thanks for the contribution @murdore — a vertical timeline/step-status primitive is a reasonable addition, and the wiring (testId, `classes` on root, `aria-current`, keyed `{#each}`, semantic `<ol>`) is mostly clean. But the current shape of `StepList` conflicts with the library's core design philosophy in ways that I don't think can be patched in review — it needs a structural rework.

---

### 1. The `status` enum is a hardcoded variant system (blocker)

The biggest issue: `StepStatus = 'pending' | 'in-progress' | 'success' | 'failure' | 'inactive'` plus the corresponding CSS:

```css
.step-list-item-pending     .step-list-indicator { background-color: var(--step-list-pending-color, #9ca3af); }
.step-list-item-in-progress .step-list-indicator { background-color: var(--step-list-in-progress-color, #3b82f6); }
.step-list-item-success     .step-list-indicator { background-color: var(--step-list-success-color, #22c55e); }
.step-list-item-failure     .step-list-indicator { background-color: var(--step-list-failure-color, #ef4444); }
.step-list-item-inactive    .step-list-indicator { background-color: var(--step-list-inactive-color, #e5e7eb); }
```

This is exactly the pattern GUIDELINES §9 ("Theming with Classes") and the library philosophy ("no hardcoded design presets") forbid. The component is making product decisions: that there are exactly five statuses, that they map to specific colors, and that the consumer must opt into Tailwind-ish defaults to override them. Some apps have three statuses, some have seven; some call it "warning" instead of "failure"; some need a `cancelled` state. The library should not encode that taxonomy.

**The customizable alternative** is the same one we use for `Button`, `Banner`, etc.:

```ts
// properties.ts
export type StepListStep = {
  id: string;
  title: string;
  description?: string;
  classes?: string;          // per-step class — consumer's variant hook
  testId?: string;
};
```

```svelte
<!-- StepList.svelte -->
<li class="step-list-item {step.classes ?? ''}" data-pw={step.testId}>
  <!-- ... -->
</li>
```

```css
/* consumer's app.css */
.step-success     { --step-list-indicator-bg: #22c55e; }
.step-failure     { --step-list-indicator-bg: #ef4444; }
.step-in-progress { --step-list-indicator-bg: #3b82f6; --step-list-pulse: pulse 1.5s ease-in-out infinite; }
.step-pending     { --step-list-indicator-bg: #9ca3af; }
```

```svelte
<StepList steps={[
  { id: '1', title: 'Submitted', classes: 'step-success' },
  { id: '2', title: 'Reviewing', classes: 'step-in-progress' },
  { id: '3', title: 'Pending decision', classes: 'step-pending' },
]} />
```

That's it — no `status` enum, no five baked-in selectors, no opinion on what set of statuses exists.

---

### 2. Inline SVG icons are baked in (blocker)

The success checkmark and failure X are hardcoded SVGs inside the component. A consumer who wants their own icon set (Lucide, Heroicons, brand SVG) has no path forward. These should be snippet props:

```ts
export type StepListProperties = {
  steps: StepListStep[];
  classes?: string;
  testId?: string;
  indicator?: Snippet<[StepListStep]>;  // consumer renders the whole indicator
  stepBody?: Snippet<[StepListStep]>;
};
```

Then the consumer decides what an "in-progress" indicator looks like — pulsing dot, spinner, ring, brand icon.

---

### 3. Pulse animation is baked in (major)

The `@keyframes step-list-pulse` lives inside the component with a fixed 1.5s duration and opacity/scale curve. Expose it as `--step-list-pulse-animation` (default `none`) so the consumer can opt in with their own keyframes, or drop it entirely.

---

### 4. Missing CSS variables for hardcoded values (major)

These magic numbers all need `var(--…)` wrappers per principle 5:

- `.step-list-indicator-icon { width: 12px; height: 12px; }` → `var(--step-list-icon-size, 12px)`
- `.step-list-indicator-dot { width: 8px; height: 8px; }` → `var(--step-list-dot-size, 8px)`
- `.step-list-connector { width: 2px; }` → `var(--step-list-connector-width, 2px)`

---

### 5. Title + description are not overridable (major)

`<div class="step-list-title">{step.title}</div>` followed by `<div class="step-list-description">{step.description}</div>` bakes in the layout (title above description, both as `<div>`). A consumer who needs `<h3>` for accessibility, or wants description above title, or wants to render a badge inline with the title, has no escape hatch. Either make `title`/`description` snippets, or expose a single `content` snippet with the step object as an argument.

---

### 6. GUIDELINES §3 violations

```svelte
{#if step.description}   <!-- string truthy check -->
{#if stepBody}           <!-- snippet truthy check -->
```

Should be:

```svelte
{#if typeof step.description === 'string' && step.description.length > 0}
{#if typeof stepBody === 'function'}
```

---

### 7. GUIDELINES §8 violation

`.step-list-connector { margin: 0 auto; }` — replace with `justify-self: center` on the grid item.

---

### 8. Overlap with `Stepper` — does this belong in the library?

We already have `Stepper` and `Step` in `src/lib/Stepper/`. The PR description notes "distinct from the existing Stepper", but most of the distinction is layout (vertical) and status semantics — both of which are exactly the kind of decisions that should live in the consumer's CSS, not in a new component. I'd push back on whether this is a new primitive or whether the right move is:

1. Document the vertical layout as a `Stepper` CSS variant (`.stepper-vertical { --stepper-flex-direction: column; }`).
2. If status indicators are needed, add a `statusIndicator` snippet to `Step`.

If after that there's still a real primitive missing, ship it — but ship it as a primitive: just the `<ol>` semantics and a per-item snippet, no enum, no SVGs, no animation, no opinionated defaults.

---

### Next step

I'd ask for a rework along the lines above before merging. Happy to chat through the API shape if helpful — the rough target is "an `<ol>` with a per-step indicator snippet and a per-step content snippet, with `classes` on both the list and each item". Everything in the current PR that isn't that should be either removed or moved into the demo route as an example of how a consumer would theme it.

murdore:
@sinha-sahil — thanks for the thorough walkthrough. I understand the objection now: `StepStatus` is a hardcoded variant system, which is precisely what §9 ("Theming with Classes") exists to prevent. The component is making product decisions (exactly five statuses, specific colors, baked-in SVG icons, fixed pulse animation) that should live in the consumer's CSS, not in the library primitive. The point about overlap with `Stepper` from @arx-optimus-17 also lands — the vertical layout distinction is itself a CSS concern.

My plan for a rework:

- Drop `StepStatus` and the five per-status CSS selectors entirely.
- Add a `classes?: string` field on `StepListStep` so the consumer controls per-item styling via their own CSS variables.
- Replace the hardcoded `success`/`failure` SVGs and the pulse dot with an `indicator` snippet prop (receives the step object), so consumers bring their own icon set.
- Expose the pulse animation as `--step-list-pulse-animation` (default `none`) — or remove it and let the consumer opt in via the snippet.
- Wrap the remaining magic numbers (`width: 2px`, `width: 12px`, `height: 8px`) in `var(--…)`.
- Fix the §3 and §8 violations you called out.

On the `Stepper` overlap question: I'll look at whether a `vertical` CSS variant on `Stepper` plus a `statusIndicator` snippet on `Step` covers the use case. If it does, I'll close this in favour of that extension instead.

Would you prefer I rework `StepList` as a minimal `<ol>`-plus-snippet primitive (no status enum, no icons, no animation), or is extending `Stepper` the direction you'd rather see?

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — thanks for engaging with this so directly. You nailed the read: `StepStatus` is the variant system §9 exists to prevent, and your rework list (drop the enum, per-step `classes`, `indicator` snippet, expose the pulse, wrap the magic numbers) is exactly the trajectory I'd ask for if `StepList` were a standalone primitive we needed.

But your last question is the right one to sit with, and I want to answer it plainly: **extend `Stepper`, don't rework `StepList`. Please close this PR.**

Here's the reasoning, walking through what's left after the rework you proposed:

1. **No status enum, no icons, no animation, no baked layout** — what remains is `<ol>` + `<li>` + a per-item snippet + a per-item `classes` hook. That's a 10–15 line consumer recipe in their own app, not a library primitive. The test we apply is "can a consumer write this in 5–15 lines using existing primitives + CSS variables?" — once you strip the opinionated bits, the answer is yes.

2. **The vertical-vs-horizontal distinction is CSS, not a component.** As I noted in the original review and @arx-optimus-17 flagged: `flex-direction: column` (or `grid-auto-flow: row`) on `Stepper` is a consumer CSS concern, not a reason to fork a new component. The library shouldn't ship `HorizontalThing` and `VerticalThing` pairs.

3. **The "status indicator" abstraction belongs on `Step`, minimally.** If `Stepper`/`Step` genuinely lacks the surface to render a per-step indicator (dot, checkmark, spinner, brand SVG), the right move is a single, additive change to `Step`: a `statusIndicator?: Snippet` prop. No enum, no defaults, no CSS in the library — the consumer brings the indicator markup, the consumer brings the colors, the consumer brings the animation. That's a one-prop, ~5-line PR against `Step`, not a new top-level component.

4. **Even the "minimal `<ol>` primitive" framing isn't quite right for the library.** A wrapper whose entire body is `<ol class={classes}>{#each steps as step}<li class={step.classes}>{@render content(step)}</li>{/each}</ol>` is markup, not a primitive. Consumers can write that inline against their own data shape and not pay the cost of an extra import + a `steps` schema the library has to maintain.

So the path forward I'd ask for:

- **Close this PR.**
- If `Step` genuinely needs a `statusIndicator` (or more generally, a content/indicator) snippet to unblock your use case, open a small follow-up PR against `src/lib/Stepper/Step.svelte` with just that addition — no styling, no defaults, no enum.
- Document the vertical timeline pattern as a demo route / example using `Stepper` + consumer CSS (`.stepper-vertical { ... }`). That's the right home for "here's how to do a vertical status timeline" — not a new component export.

To be clear on what you got right: the wiring instincts in this PR (semantic `<ol>`, keyed `{#each}`, `aria-current`, `data-pw`, `classes` on root) are all correct and you should carry them into whatever the follow-up `Step` PR looks like. The issue is the surface area, not the craft.

**Ask: please close this PR.** If `Step` needs a `statusIndicator` snippet to cover your use case, open a small additive PR against `Step` instead, and we'll document the vertical timeline as a consumer-CSS recipe in the `Stepper` demo route.

murdore:
Closing — agreed: extend Stepper, don't rework StepList. If Step needs a statusIndicator snippet I'll open a small additive PR against it; the vertical timeline becomes a consumer-CSS recipe in the Stepper demo. Thanks @sinha-sahil.

PR #225, opened 2026-05-29
feat: add DescriptiveTile component

murdore:
## Summary

- Adds a new `DescriptiveTile` primitive component for gallery/selection UI patterns
- Tile has an image/preview region at the top, a label row, and an optional `bottom` snippet slot for extra content
- Supports `selected`, `disabled`, `classes`, `testId`, and `onclick` interaction with full keyboard support (Enter/Space)

## API

```ts
type DescriptiveTileProperties = {
  image?: string;
  alt?: string;
  label?: string;
  selected?: boolean;      // default false
  disabled?: boolean;      // default false
  testId?: string;
  classes?: string;
  bottom?: Snippet;
  onclick?: (event: MouseEvent) => void;
};
```

**CSS custom properties exposed:**
`--descriptive-tile-padding`, `--descriptive-tile-radius`, `--descriptive-tile-border`, `--descriptive-tile-selected-border`, `--descriptive-tile-preview-radius`, `--descriptive-tile-preview-aspect`, `--descriptive-tile-label-color`, `--descriptive-tile-label-font-weight`, `--descriptive-tile-gap`

## Notes

- prettier: clean
- eslint: clean
- Backward-compatible — purely additive export to `src/lib/index.ts`
- No new npm dependencies
- Svelte 5 runes syntax throughout
- `role="button"` and `tabindex` are only applied when `onclick` is provided, avoiding a11y warnings on non-interactive tiles

### DIFF
--- a/src/lib/DescriptiveTile/DescriptiveTile.svelte
+++ b/src/lib/DescriptiveTile/DescriptiveTile.svelte
@@ -1,0 +1,136 @@
+<script lang="ts">
+  import type { DescriptiveTileProperties } from './properties';
+
+  let {
+    image,
+    alt,
+    label,
+    selected = false,
+    disabled = false,
+    testId,
+    classes,
+    bottom,
+    onclick
+  }: DescriptiveTileProperties = $props();
+
+  const isInteractive = $derived(!!onclick);
+
+  const handleKeydown = (event: KeyboardEvent) => {
+    if (disabled || !onclick) {
+      return;
+    }
+    if (event.key === 'Enter' || event.key === ' ') {
+      event.preventDefault();
+      onclick(new MouseEvent('click'));
+    }
+  };
+
+  const handleClick = (event: MouseEvent) => {
+    if (disabled || !onclick) {
+      return;
+    }
+    onclick(event);
+  };
+</script>
+
+{#if isInteractive}
+  <div
+    class="descriptive-tile {classes ?? ''}"
+    class:descriptive-tile-selected={selected}
+    class:descriptive-tile-disabled={disabled}
+    role="button"
+    tabindex={disabled ? -1 : 0}
+    data-pw={testId}
+    onclick={handleClick}
+    onkeydown={handleKeydown}
+  >
+    <div class="descriptive-tile-preview">
+      {#if image}
+        <img src={image} alt={alt ?? ''} />
+      {/if}
+    </div>
+    <div class="descriptive-tile-label">{label ?? ''}</div>
+    {#if bottom}
+      <div class="descriptive-tile-bottom">
+        {@render bottom()}
+      </div>
+    {/if}
+  </div>
+{:else}
+  <div
+    class="descriptive-tile {classes ?? ''}"
+    class:descriptive-tile-selected={selected}
+    class:descriptive-tile-disabled={disabled}
+    data-pw={testId}
+  >
+    <div class="descriptive-tile-preview">
+      {#if image}
+        <img src={image} alt={alt ?? ''} />
+      {/if}
+    </div>
+    <div class="descriptive-tile-label">{label ?? ''}</div>
+    {#if bottom}
+      <div class="descriptive-tile-bottom">
+        {@render bottom()}
+      </div>
+    {/if}
+  </div>
+{/if}
+
+<style>
+  .descriptive-tile {
+    display: flex;
+    flex-direction: column;
+    gap: var(--descriptive-tile-gap, 8px);
+    padding: var(--descriptive-tile-padding, 12px);
+    border-radius: var(--descriptive-tile-radius, 8px);
+    border: var(--descriptive-tile-border, 1px solid transparent);
+    cursor: default;
+    box-sizing: border-box;
+  }
+
+  .descriptive-tile[role='button'] {
+    cursor: pointer;
+  }
+
+  .descriptive-tile[role='button']:focus-visible {
+    outline: 2px solid var(--descriptive-tile-selected-border, #0070f3);
+    outline-offset: 2px;
+  }
+
+  .descriptive-tile-selected {
+    border: var(--descriptive-tile-selected-border, 1px solid #0070f3);
+  }
+
+  .descriptive-tile-disabled {
+    opacity: 0.5;
+    cursor: not-allowed;
+    pointer-events: none;
+  }
+
+  .descriptive-tile-preview {
+    border-radius: var(--descriptive-tile-preview-radius, 6px);
+    aspect-ratio: var(--descriptive-tile-preview-aspect, 16 / 9);
+    overflow: hidden;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+  }
+
+  .descriptive-tile-preview img {
+    width: 100%;
+    height: 100%;
+    object-fit: cover;
+    display: block;
+  }
+
+  .descriptive-tile-label {
+    color: var(--descriptive-tile-label-color, inherit);
+    font-weight: var(--descriptive-tile-label-font-weight, inherit);
+  }
+
+  .descriptive-tile-bottom {
+    display: flex;
+    flex-direction: column;
+  }
+</style>
--- a/src/lib/DescriptiveTile/properties.ts
+++ b/src/lib/DescriptiveTile/properties.ts
@@ -1,0 +1,13 @@
+import type { Snippet } from 'svelte';
+
+export type DescriptiveTileProperties = {
+  image?: string;
+  alt?: string;
+  label?: string;
+  selected?: boolean;
+  disabled?: boolean;
+  testId?: string;
+  classes?: string;
+  bottom?: Snippet;
+  onclick?: (event: MouseEvent) => void;
+};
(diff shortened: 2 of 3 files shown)


sinha-sahil:
Thanks for the PR @murdore — I can see the intent: give consumers a one-liner for a common "image + label + optional content + selectable" gallery tile. That's a real UI pattern. Unfortunately, I think `DescriptiveTile` as-shipped sits on the wrong side of this library's philosophy line, and I'd like to push back before we add it to the public surface.

## Why this is a composite wrapper, not a primitive

The component is, structurally:

```
<div class="tile">
  <Img />
  <Text />
  {bottom snippet}
</div>
```

…with two baked-in visual states (`selected`, `disabled`) and an optional click handler. There is no novel behavior, no non-trivial accessibility tree, no state machine, no coordination between children — everything it does is reachable by composing the primitives we already ship. Per the library's design philosophy (principles 3 and 4):

> **No dummy components.** A pure-decoration element with no behavior, focus management, ARIA semantics, or composition value should be a CSS utility, not a component.
>
> **No composite wrappers.** Components that are merely `<Card><Icon/><Text/><Button/></Card>` belong in user-land, not the library.

The non-interactive branch (`{:else}` at L48) is literally a `<div>` with three children — that's a dummy. The interactive branch hand-rolls `role="button"` + keyboard handling + a synthetic `new MouseEvent('click')`, which is precisely what `Button` exists to do (and which would also fix several GUIDELINES violations).

## What the consumer should write instead

The whole pattern collapses to ~15 lines in user-land using primitives we already export:

```svelte
<script>
  import { Img, Button } from '@juspay/svelte-ui-components';
  let { image, label, selected, onpick, extra } = $props();
</script>

<Button classes="tile {selected ? 'tile-selected' : ''}" onclick={onpick}>
  {#snippet children()}
    <Img src={image} alt={label} classes="tile-preview" />
    <span class="tile-label">{label}</span>
    {#if typeof extra === 'function'}
      <span class="tile-bottom">{@render extra()}</span>
    {/if}
  {/snippet}
</Button>
```

```css
/* consumer stylesheet — variants live HERE, not in the component */
.tile {
  --button-padding: 12px;
  --button-border-radius: 8px;
  --button-background: transparent;
  display: flex;
  flex-direction: column;
  gap: 8px;
}
.tile-selected { --button-border: 1px solid #0070f3; }
.tile-preview { --img-aspect-ratio: 16/9; --img-border-radius: 6px; }
```

That's the `classes` + CSS variable pattern from GUIDELINES §9, and it gives consumers full control over what "selected" or "disabled" look like — instead of being stuck with our `opacity: 0.5` opinion.

## Specific GUIDELINES violations (independent of the philosophy question)

Even if we wanted to ship this, the implementation needs a pass:

1. **§1 — No `!!`.** L14: `const isInteractive = $derived(!!onclick);` → `typeof onclick === 'function'`.
2. **§3 — No falsy/truthy checks.** `if (disabled || !onclick)` (L17, L26), `{#if image}` (L38, L58), `{#if bottom}` (L43, L63). Snippets are functions; strings need length checks. See §3's snippet/string examples verbatim.
3. **§4 — Reuse `Img`.** L39 and L59 use raw `` with no error handling. This is the canonical bad example in §4.
4. **§6 — Use `Button`.** L33-46 builds a `<div role="button" tabindex onkeydown>` with `onclick(new MouseEvent('click'))` to fake Enter/Space activation. That synthetic event is a smell. `Button` already does all of this; use it with the CSS variable bridging pattern from §6.
5. **Hardcoded visuals.** `opacity: 0.5`, `cursor: not-allowed`, `pointer-events: none`, `object-fit: cover`, `outline: 2px solid …`, `outline-offset: 2px` are all baked in with no `var(--…)` override — violates philosophy principle 5 ("highly customizable").
6. **Variant props bake in CSS.** `selected` and `disabled` map to specific visual treatments inside the component. Per §9 and the philosophy, these belong in consumer classes that flip CSS variables.

## What I'd like to see

My recommendation is to **close this PR** and move the tile pattern into your app code (or a thin internal layer if multiple Juspay apps need it). If there's a specific behavior we're missing in the existing primitives (e.g. `Img` doesn't expose an aspect-ratio variable, or `Button` doesn't expose a layout-children mode), let's identify that gap and add it to the primitive instead — that's a much better lever than adding composite components.

If you'd rather rework, the bar would be:
- Drop the non-interactive branch entirely.
- Replace the interactive branch with `Button` per §6.
- Replace `` with `` per §4.
- Remove `selected`/`disabled` visual presets — let consumers style `[aria-pressed]` / `[aria-disabled]` via `classes`.
- Fix every `!!` / `{#if onclick}` / `{#if image}` / `{#if bottom}` per §1/§3.

…but at that point there's basically no component left, which is itself the signal that this is user-land code. Happy to discuss.

murdore:
@sinha-sahil — thank you for the thorough breakdown. You have articulated the core issue clearly, and on reflection I agree with it: `DescriptiveTile` is a composite wrapper (`Img` + text + `Button` logic in a div) with no novel behavior, accessibility tree, or state coordination of its own. That maps squarely to the "no composite wrappers" stance and principles §4, §6, and §9 — the full pattern is reachable in consumer-land with the primitives the library already ships.

The user-land example you provided is convincing — it is more flexible (consumers own what `selected` looks like via CSS vars), not less. Your point about the synthetic `new MouseEvent('click')` being a smell (and the reason `Button` exists) is particularly sharp.

Given that a correct rework would leave essentially no component, I think **closing this PR** is the right call. I will move the tile pattern into the consuming app. Before I do:

- Are there any gaps in the existing primitives that made this pattern feel awkward? (e.g., does `Img` need an aspect-ratio CSS variable, or does `Button` need a richer layout-children mode?) If so, I would rather open a narrow extension PR for the primitive than keep a composite in the library.

Would closing this and opening primitive-extension PRs where gaps exist be the shape you would accept?

sinha-sahil:

New arrow-const handlers in the diff: `handleKeydown`, `handleClick`. Please write them as `function fnName() {}` instead.

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — thanks for engaging with this seriously rather than defending the surface. Your read is exactly right: once you strip out the synthetic `MouseEvent`, the hardcoded `selected`/`disabled` visuals, and the raw ``, there is no component left — only a `<div>` with three children, which is by definition user-land composition. The fact that the correct rework leaves nothing is the load-bearing signal, and you saw it. Appreciate that.

**Confirming the close.** Yes, please close this PR. The tile pattern lives in your consuming app (or a thin internal layer if multiple Juspay apps need the same shape) using the `<Button>` + `` + `classes` + CSS-variable recipe from the earlier comment. That recipe is the supported pattern; we'll add a short "Gallery / Tile composition" entry to the docs so this is discoverable for the next person who hits the same instinct.

**On the primitive-gap question — yes, that is exactly the shape we want.** A narrow extension PR against an existing primitive is almost always preferable to a new component. Looking at what `DescriptiveTile` was reaching for, here are the candidate gaps worth opening as separate, slim PRs:

1. **`Img` aspect-ratio CSS variable.** Your component baked `--descriptive-tile-preview-aspect` because `Img` doesn't currently expose an aspect-ratio hook. If `Img` doesn't already have `--img-aspect-ratio` (please confirm against `src/lib/Img/Img.svelte` before opening), that's a clean additive PR: one new CSS variable, default `auto`-equivalent (per GUIDELINES §8, avoid the literal `auto` — use `initial` or `unset`), no behavior change, no breaking change.

2. **`Button` layout-children mode — probably not needed.** `Button` already accepts a `children` snippet and the `classes` prop bridges layout via consumer CSS (`display: flex; flex-direction: column; gap: …`). I'd push back gently on adding any layout-mode prop here; that would itself be a baked-in preset. If you hit a concrete blocker rendering arbitrary children inside `Button`, open a discussion issue first with the minimal reproducer rather than a PR — there's a real risk the fix is "use CSS" and the PR shouldn't exist.

3. **What is *not* a primitive gap:** `selected` state styling, `disabled` opacity, focus outlines, keyboard activation. Those are either (a) consumer CSS toggled via a class on `Button`, or (b) already handled by `Button`'s ARIA/keyboard behavior. No PR needed.

Rule of thumb for the extension PRs: each one should add exactly one CSS variable or one prop to one existing component, with a non-opinionated default (`inherit` / `transparent` / `currentColor` / `initial`), and have a one-line justification of why no consumer-land workaround exists. If a proposed extension fails that test, it's probably a docs change, not a code change.

Also — per the follow-up comment on this thread — when you do open those extension PRs, please use `function fnName()` declarations for any new named handlers (the `handleKeydown` / `handleClick` arrow consts in this diff are the pattern to avoid going forward).

**Ask: please close this PR.** Open separate, narrowly-scoped PRs for any primitive gaps you confirm, starting with the `Img` aspect-ratio variable if it isn't already there. Happy to fast-track review on those.

murdore:
Closing — DescriptiveTile is a Button+Img+classes composition that lives in the consuming app. I'll open narrowly-scoped PRs for any real primitive gaps (starting with an Img aspect-ratio var if missing). Thanks @sinha-sahil.

PR #226, opened 2026-05-29
feat: add right-drawer align, showDivider, header description, and dialogue body to Modal

murdore:
## Summary

- **`align: 'right'`** — adds a right-edge drawer variant to `ModalAlign`. The modal-content slides in from the right (`fly x:300`), stretches full-height, and takes its width from `--modal-drawer-width` (default `420px`). Implemented as a new `.modal-align-right` CSS class + `.right` overlay class; the top/center/bottom branches are completely untouched.
- **`showDivider?: boolean`** (default `false`) — when `true`, renders a `1px solid` separator between the header and content area, and between the content area and the footer. Color is controlled by `--modal-divider-color` (default `#e5e7eb`). When omitted or `false`, no divider is rendered — identical to today.
- **`header.description?: string`** — optional secondary text line rendered beneath the header title. Only rendered when the string is non-empty. Color exposed as `--modal-header-description-color` (default `#6b7280`).
- **`header.showCloseButton?: boolean`** — gates rendering of `header.rightImage`. Defaults to `true`, preserving existing behavior exactly.
- **`dialogueConfig?: { bodyText: string; footerConfig?: ... }`** — when provided, renders a compact confirmation body (`<p>` with `bodyText`) instead of the `content` snippet, plus an optional footer using the same button shape as `footer`. When absent, the component renders the `content` snippet exactly as before.

## API

| New prop | Type | Default | Effect |
|---|---|---|---|
| `align` | `'top'\|'center'\|'bottom'\|'right'` | `'center'` | `'right'` produces a full-height right-edge drawer |
| `showDivider` | `boolean` | `false` | Divider line between header↔content and content↔footer |
| `header.description` | `string` | — | Secondary line below header title |
| `header.showCloseButton` | `boolean` | `true` | Show/hide the close (rightImage) button |
| `dialogueConfig` | `{ bodyText: string; footerConfig?: FooterConfig }` | — | Compact confirmation body; replaces content snippet when set |

New CSS custom properties: `--modal-drawer-width`, `--modal-drawer-border-radius`, `--modal-divider-color`, `--modal-header-description-color`, `--modal-dialogue-body-padding`, `--modal-dialogue-body-text-color`.

## Notes

- `svelte-check` reports **0 errors, 0 warnings**
- `prettier --check` exits **0** on all Modal files
- `eslint src/lib/Modal/` exits **0**
- All new props are optional with defaults that reproduce today's behavior exactly — no existing callsites require changes
- `ModalAnimation.svelte` gains a `case 'right': return { ...base, x: 300 }` branch; the `useFlyAnimation` derived also includes `align === 'right'`; all other align branches are untouched

### DIFF
--- a/src/lib/Animations/ModalAnimation.svelte
+++ b/src/lib/Animations/ModalAnimation.svelte
@@ -21,6 +21,8 @@
         return { ...base, y: -30 };
       case 'bottom':
         return { ...base, y: 300 };
+      case 'right':
+        return { ...base, x: 300 };
       default:
         return base;
     }
@@ -28,7 +30,7 @@
 
   let fadeAnimationProperties = { duration: 300 };
 
-  let useFlyAnimation = $derived(align === 'top' || align === 'bottom');
+  let useFlyAnimation = $derived(align === 'top' || align === 'bottom' || align === 'right');
   let useOutTransition = $derived(transitionType === 'ALL');
 </script>
 
--- a/src/lib/Modal/Modal.svelte
+++ b/src/lib/Modal/Modal.svelte
@@ -18,6 +18,8 @@
     transitionType = 'ALL',
     header = {},
     footer,
+    showDivider = false,
+    dialogueConfig,
     debounceTime = 700,
     leftImageTestId,
     testId,
@@ -35,42 +37,53 @@
 
   const debounce = createDebouncer(debounceTime);
 
-  function handlePopstate() {
+  const handlePopstate = () => {
     backPressed = true;
     onclose?.();
-  }
+  };
 
-  function handleRightImageClick(event: MouseEvent): void {
+  const handleRightImageClick = (event: MouseEvent): void => {
     onheaderRightImageClick?.(event);
-  }
+  };
 
-  function handleLeftImageClick(event: MouseEvent): void {
+  const handleLeftImageClick = (event: MouseEvent): void => {
     onheaderLeftImageClick?.(event);
-  }
+  };
 
-  function handlePrimaryButtonClick(event: MouseEvent): void {
+  const handlePrimaryButtonClick = (event: MouseEvent): void => {
     onprimaryButtonClick?.(event);
-  }
+  };
 
-  function handleSecondaryButtonClick(event: MouseEvent): void {
+  const handleSecondaryButtonClick = (event: MouseEvent): void => {
     onsecondaryButtonClick?.(event);
-  }
+  };
 
-  function handleOverlayClick(event: MouseEvent) {
+  const handleOverlayClick = (event: MouseEvent) => {
     if (event.target && event.target === overlayDiv) {
       debounce(() => {
         onoverlayClick?.();
       });
     }
-  }
+  };
 
-  function handleKeyDown(event: KeyboardEvent): void {
+  const handleKeyDown = (event: KeyboardEvent): void => {
     onkeydown?.(event);
-    let key = event?.key;
+    const key = event?.key;
     if (key === 'Escape') {
       onoverlayClick?.();
     }
-  }
+  };
+
+  const hasHeader = $derived(
+    (typeof header?.leftImage === 'string' && header.leftImage.length > 0) ||
+      (typeof header?.text === 'string' && header.text.length > 0) ||
+      (typeof header?.rightImage === 'string' && header.rightImage.length > 0) ||
+      (typeof header?.description === 'string' && header.description.length > 0)
+  );
+
+  const showCloseButton = $derived(header?.showCloseButton !== false);
+
+  const dialogueFooter = $derived(dialogueConfig?.footerConfig);
 
   onMount(() => {
     document.body.style.overflow = 'hidden';
@@ -95,7 +108,7 @@
 
 <svelte:window onkeydown={handleKeyDown} />
 
-{#if typeof content === 'function'}
+{#if typeof content === 'function' || dialogueConfig}
   <OverlayAnimation>
     <div
       bind:this={overlayDiv}
@@ -107,8 +120,8 @@
       data-pw={testId}
     >
       <ModalAnimation enable={enableTransition} {align} {transitionType}>
-        <div class="modal-content {size}">
-          {#if (typeof header?.leftImage === 'string' && header.leftImage.length > 0) || (typeof header?.text === 'string' && header.text.length > 0) || (typeof header?.rightImage === 'string' && header.rightImage.length > 0)}
+        <div class="modal-content {size} {align === 'right' ? 'modal-align-right' : ''}">
+          {#if hasHeader}
             <div class="header">
               {#if header.leftImage}
                 <div
@@ -121,12 +134,19 @@
                   <img class="header-left-img" src={header.leftImage} alt="" />
                 </div>
               {/if}
-              {#if header.text}
-                <div class="header-text" data-pw={header.testId}>
-                  {header.text}
-                </div>
-              {/if}
-              {#if header.rightImage}
+              <div class="header-title-group">
+                {#if header.text}
+                  <div class="header-text" data-pw={header.testId}>
+                    {header.text}
+                  </div>
+                {/if}
+                {#if typeof header.description === 'string' && header.description.length > 0}
+                  <div class="header-description">
+                    {header.description}
+                  </div>
+                {/if}
+              </div>
+              {#if header.rightImage && showCloseButton}
                 <div
                   role="button"
                   tabindex="0"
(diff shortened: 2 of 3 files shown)


sinha-sahil:
## Thanks for the PR — but this collides head-on with the library's design philosophy and needs a rework

Hi @murdore — thanks for the thorough write-up and the care with defaults / backward-compatibility. I can see you are trying to solve four real consumer needs (right-edge drawer, dividers, header subtitle, confirmation dialogue). Unfortunately, several of the new props are exactly the kind of hardcoded design preset / composite-wrapper that `GUIDELINES.md §9` and the broader project philosophy ("components are unstyled primitives, consumers own the design system") are written to prevent. Most are reachable today via the `classes` prop plus the existing `--modal-*` variables.

To be clear up front: you DID expose CSS variables for the new visual decisions (drawer width, divider color, description color, dialogue body padding/text color), and that is the right instinct. The objection is not "you hardcoded colors" — it is "this is a preset that belongs in a consumer `classes` recipe, not in the component's prop surface".

---

### 1. `align: 'right'` — new layout preset in the align union (blocker)

```svelte
<div class="modal-content {size} {align === 'right' ? 'modal-align-right' : ''}">
```
```css
.modal-align-right {
  height: 100%;
  width: var(--modal-drawer-width, 420px);
  border-radius: var(--modal-drawer-border-radius, 0px);
}
.right { align-items: var(--modal-right-align-items, flex-end); flex-direction: row; }
```

Credit where due — `--modal-drawer-width`, `--modal-drawer-border-radius`, `--modal-right-justify-content`, `--modal-right-align-items` are all exposed as variables. The objection is structural: a right-edge drawer is a positioning preset, and adding it as a literal in the `ModalAlign` union plus a dedicated `.modal-align-right` class fork-codes a layout decision into the component. The `ModalAnimation` `x: 300` branch is also a preset.

**Customizable alternative (zero new props):**

```svelte
<Modal classes="modal-drawer-right" align="center" ... />
```

```css
.modal-drawer-right {
  --modal-center-justify-content: flex-end;
  --modal-center-align-items: stretch;
  --modal-fit-content-max-height: 100vh;
}
.modal-drawer-right .modal-content {
  width: 420px;
  height: 100vh;
}
```

If the existing variable surface is not expressive enough (e.g. a `flex-direction` hook on the outer container, or a width hook on `.modal-content`), then add those CSS variables to the existing selectors — but do not introduce a new align literal. Expose the slide offset/axis on `ModalAnimation` as variables instead of branching on `'right'`.

---

### 2. `showDivider: boolean` — duplicates existing CSS variables (blocker)

```svelte
{#if showDivider}
  <div class="modal-divider" role="separator" aria-hidden="true"></div>
{/if}
```
```css
.modal-divider { border-top: 1px solid var(--modal-divider-color, #e5e7eb); }
```

The component already ships `--modal-header-border-bottom` (default `none`) and `--modal-footer-border-top` (default `none`). Those are the dividers — and since the new `.modal-divider` element only renders adjacent to the header/footer anyway, they cover the same surface.

**Customizable alternative:**

```svelte
<Modal classes="modal-with-dividers" ... />
```

```css
.modal-with-dividers {
  --modal-header-border-bottom: 1px solid #e5e7eb;
  --modal-footer-border-top: 1px solid #e5e7eb;
}
```

Done. No prop, no extra DOM, no `role="separator"` to worry about.

---

### 3. `dialogueConfig: { bodyText, footerConfig }` — composite wrapper (blocker)

This is the textbook violation. It (a) bakes a `<p>` element + padding + text color into the modal, (b) duplicates the entire ~30-line footer rendering block under a parallel name, and (c) replaces the `content` snippet — the moment a consumer needs richer content (a form input in the confirmation, a link, custom typography), they have to abandon the prop and rewrite as a snippet anyway.

A consumer who wants a confirmation dialogue already has everything they need:

```svelte
<Modal
  size="fit-content"
  align="center"
  header={{ text: 'Delete account?' }}
  footer={{
    primaryButton: { text: 'Delete', ... },
    secondaryButton: { text: 'Cancel', ... }
  }}
  classes="modal-confirmation"
>
  {#snippet content()}
    <p>This action cannot be undone.</p>
  {/snippet}
</Modal>
```

```css
.modal-confirmation { --modal-display: block; }
.modal-confirmation .slot-content { padding: 20px; }
```

That is the whole feature. The library ships the primitive; presets like "confirmation modal" / "alert modal" / "wizard modal" do not belong in the component. The duplicated footer branch is also a near-verbatim copy of the existing `footer.primaryButton` branch — the diff doubles the template for no gain.

---

### 4. `header.description: string` — composite wrapper (major)

```svelte
<div class="header-title-group">
  {#if header.text}<div class="header-text">{header.text}</div>{/if}
  {#if typeof header.description === 'string' && header.description.length > 0}
    <div class="header-description">{header.description}</div>
  {/if}
</div>
```

You did expose `--modal-header-description-color` as a variable, which is good. The structural concern is the second string prop plus the hardcoded `.header-title-group` flex-column wrapper — the library is doing layout for the consumer for what amounts to a title + subtitle stack. The proper extension is either a `headerContent?: Snippet` so consumers compose their own title+subtitle, or keep `header.text` as the single string-prop convenience without adding a second element with a fixed structural wrapper.

---

### 5. `header.showCloseButton: boolean` — duplicates existing API (major)

```svelte
{#if header.rightImage && showCloseButton}
```

Already expressible by simply not passing `header.rightImage`. Two APIs for the same intent is bloat.

---

### Code-quality items (smaller, but worth fixing in the rework)

- **§3 falsy/truthy on non-booleans, in code newly introduced by this PR.** You correctly use the `typeof x === 'string' && x.length > 0` pattern in the new `hasHeader` derived — please apply it consistently in the new template additions: `{#if dialogueConfig}` (twice), `{#if dialogueFooter.secondaryButton}`, `{#if dialogueFooter.primaryButton}`, and the modified `{#if header.rightImage && showCloseButton}`. The existing `else if` on line 149 already shows the right pattern for objects (`typeof footer?.primaryButton === 'object'`). The pre-existing falsy checks elsewhere (`{#if header.leftImage}`, `{#if header.text}`, `{#if footer.secondaryButton}`, `{#if footer.primaryButton}`) were not introduced by this PR and can be left for a separate cleanup.
- **§5 redundant guard.** `const key = event?.key;` — `event` is the handler argument, the optional chain is dead. Pre-existing, but the PR refactored that line, so worth fixing while you're there.
- **`.dialogue-body { overflow-y: auto; }`** — §8 explicitly permits `auto` for `overflow-*`, so this is not a guideline violation. For consistency with `.slot-content` using `var(--modal-overflow-y, scroll)`, consider exposing `var(--modal-dialogue-body-overflow-y, scroll)` — or remove `.dialogue-body` entirely along with item 3.
- The mass `function` → `const arrow =` refactor on the handlers is unrelated noise; please leave that out of a feature PR.

---

### Asks

1. **Close or fundamentally rework.** Drop `align: 'right'`, `showDivider`, `dialogueConfig`, `header.description`, `header.showCloseButton`. They are all reachable today via the `classes` prop + existing CSS variables — exactly the §9 pattern.
2. **If something genuinely cannot be expressed with existing variables** (e.g. a `flex-direction` hook on `.modal` or a width hook on `.modal-content`), the right PR is "expose more CSS variables on existing selectors", not "add a literal to the align union". Happy to review a focused PR like that.
3. **For the docs side**, examples of "how to build a right-edge drawer" / "how to build a confirmation dialogue" using the `classes` prop and the `content` snippet would be a great addition to the Modal docs — that is the layer where these recipes belong.

Thanks again for the careful write-up — the needs you are describing are legit, they just need to be solved one layer up.

murdore:
@sinha-sahil — thank you for the detailed, principle-grounded breakdown. You are right on all counts.

Re-reading GUIDELINES §9, I can see the pattern clearly now: the library ships unstyled primitives and CSS variable hooks; variant/preset decisions belong one layer up in the consumer's `classes` stylesheet. Every one of the blocked props — `align: 'right'`, `showDivider`, `dialogueConfig`, `header.description`, and `header.showCloseButton` — is a layout preset or a composite wrapper that a consumer can already compose from the existing `classes` prop plus the existing `--modal-*` variables. Adding them as first-class props hard-codes design opinions into the component and bloats its prop surface for no extra power.

I plan to:

- Drop all five blocked props from this PR.
- Identify any genuinely missing CSS variable hooks (e.g. a `flex-direction` hook on `.modal`, a width hook on `.modal-content`) and open a focused "expose more CSS variables" PR as you described.
- Add consumer-side recipe examples (right-edge drawer, confirmation dialogue) to the Modal docs, where presets belong.
- Clean up the code-quality items (§3 falsy checks on the new template additions, the redundant `event?.key` optional chain, and remove the unrelated `function` → arrow refactor noise).

One clarifying question before I close this: for the `header.description` case specifically, would you accept a `headerContent?: Snippet` escape hatch (letting consumers render their own title + subtitle markup), or is the existing `header.text` string plus the `content` snippet already sufficient and a header snippet would also be over-engineering?

sinha-sahil:

This PR converts existing `function` declarations to arrow consts: `handlePopstate`, `handleRightImageClick`, `handleLeftImageClick`, `handlePrimaryButtonClick`, `handleSecondaryButtonClick`, `handleOverlayClick`, `handleKeyDown`. Please revert those conversions — they're stylistic churn, and the existing convention in this codebase is function declarations.

sinha-sahil (changes requested):

To merge:

sinha-sahil:
@murdore — thank you for the gracious read of the review and the clear forward plan. You have summarized §9 better than the guideline itself does: "variant/preset decisions belong one layer up in the consumer's `classes` stylesheet." That is exactly the bar. The plan you have laid out (drop the five, open a focused CSS-variables PR, ship the recipes as docs, clean up the §3/§5 items and revert the arrow refactor) is the right shape, and I am happy to fast-track those follow-up PRs.

On your clarifying question about `headerContent?: Snippet` — **no, please do not add it.** The existing surface already covers it without any new API:

- `header.text` is the single-string convenience for the 95% case (a plain title).
- The `content` snippet owns the body — and a subtitle visually living "under the title" is a body-layout decision, not a header-decoration decision. Consumers who want a title + subtitle stack render the subtitle as the first element of `content` and style the gap via their own `classes` rules.
- For the rarer case where someone genuinely needs custom markup *inside the header bar* (an icon row, a badge next to the title, a stepper indicator), the right escape hatch is to extend the existing `header.leftImage` / `header.rightImage` surface with CSS variable hooks, not to add a parallel `headerContent` snippet that duplicates `header.text`'s job. Two ways to render the same region is exactly the duplication that prompted my §5 note on `header.showCloseButton`.

The rule of thumb: a `Snippet` prop earns its place when the existing surface genuinely cannot express the visual (e.g. `paginatorSlot` on Table, `bottomContent` on Drawer — neither has a reasonable string-prop equivalent). For "title + subtitle", `header.text` + body composition already does the job. Adding `headerContent` would also force a "which wins, `header.text` or `headerContent`?" precedence rule, which is the kind of API debt the library has so far avoided.

So the concrete shape I would like to see:

1. **Please close this PR.** All five blocked props are being dropped, and the salvageable bits (any new CSS-variable exposures, the §3 falsy fixes, the §5 redundant guard) are cleaner as a fresh slim PR than a force-pushed rewrite of this branch.
2. **Open PR A — "Modal: expose additional CSS variables"** with only the new `--modal-*` hooks on existing selectors (the `flex-direction` / width hooks you mentioned, plus a `--modal-dialogue-body-overflow-y` style hook *only if* the dialogue body element survives, which it should not). Keep `ModalAnimation` axis/offset behind variables — no `'right'` literal.
3. **Open PR B — "Modal: docs recipes"** with the right-edge drawer and confirmation-dialogue recipes as consumer-side `classes` examples in the Modal docs.
4. **Skip the `function` → arrow conversions entirely** in both follow-up PRs (per the separate note).

**Ask: please close this PR** and split the salvageable work into the two focused PRs above. The composite/preset surface is not reconcilable with the library philosophy, but the underlying needs (drawer layout, confirmation dialogue, divider) all have a clean home in `classes` + CSS variables + docs.

murdore:
Closing in favor of the slim #230 (new CSS-var exposure + §3 falsy fixes + §5 redundant-guard removal) — the five composite/preset props are dropped per §9. Thanks @sinha-sahil.

PR #233, opened 2026-05-29
feat(banner): border-radius/border vars, title snippet, role escape hatch

murdore:
Slim follow-up to #221 per @sinha-sahil's review — contains only the four keepers. Variants/layout stay in consumer CSS (classes + CSS vars) per GUIDELINES §9. Supersedes #221.

## Keepers

- **`--banner-border` and `--banner-border-radius`** CSS vars on `.banner` — enables card-style notifications without any class enum
- **`title` Snippet** rendered inside a `<div class="banner-body">` flex-column wrapper (`--banner-body-gap` defaults to `0`, `.banner-title` `font-weight` defaults to `inherit`) — omitting `title` reproduces today's layout exactly; the new structure is invisible until a consumer opts in
- **`role` prop** as a plain ARIA escape hatch — no tone-to-role derivation, consumer passes `role="alert"` explicitly when needed
- **README consumer-theming example** using `classes="banner-error banner-compact"` showing `--banner-border` + `--banner-border-radius` in action (§9 canonical pattern)

## GUIDELINES nits also folded in

- `filter(Boolean)` → `filter((c) => c.length > 0)` in `rootClass` derivation (§3)
- `handleClick` / `handleKeydown` / `handleDismiss` kept as `function` declarations (second-to-last review note)
- Redundant `title !== null` guard removed — `typeof title === 'function'` is sufficient (§5)

## What was dropped vs #221

- `tone` enum (`info` | `success` | `warning` | `error`) and `.banner-tone-*` preset classes
- `size` enum and `.banner-size-sm` preset
- `flush` boolean and `.banner-flush` preset
- Automatic `tone === 'error' → role="alert"` derivation

All of those belong in consumer CSS via `classes` + CSS vars.

### DIFF
--- a/docs/Banner.md
+++ b/docs/Banner.md
@@ -38,17 +38,55 @@ A notification banner with optional icon snippet, text content, inline link text
 </Banner>
 ```
 
+### With Title Snippet
+
+```svelte
+<Banner text="Your cart was abandoned 2 hours ago.">
+  {#snippet title()}
+    <strong>Don't forget your items</strong>
+  {/snippet}
+</Banner>
+```
+
+### Consumer Theming via `classes` (error + compact variant)
+
+No tone enum is needed — define the variant in your app's CSS and pass it through `classes`:
+
+```css
+/* app.css */
+.banner-error {
+  --banner-background: #fff0f0;
+  --banner-color: #c0392b;
+  --banner-border: 1px solid #e74c3c;
+  --banner-border-radius: 6px;
+}
+
+.banner-compact {
+  --banner-padding: 6px 12px;
+}
+```
+
+```svelte
+<Banner
+  text="Payment failed. Please try again."
+  classes="banner-error banner-compact"
+  role="alert"
+/>
+```
+
 ## Props
 
-| Prop        | Type      | Required | Default | Description                                                                                                                                                                                                                |
-| ----------- | --------- | -------- | ------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| text        | `string`  | Yes      | `-`     | The main banner message text.                                                                                                                                                                                              |
-| icon        | `Snippet` | No       | `-`     | Svelte 5 Snippet for a custom icon displayed to the left of the text. Replaces the old `string` image URL prop.                                                                                                            |
-| linkText    | `string`  | No       | `-`     | Optional link text appended inline after the main text, styled in a different color (blue by default).                                                                                                                     |
-| dismissible | `boolean` | No       | `false` | Whether to show a close/dismiss button on the right side of the banner.                                                                                                                                                    |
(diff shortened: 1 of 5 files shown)


sinha-sahil:
Thanks for the clean split @murdore — re-reviewed against the agreed keeper list and the philosophy, and this is exactly the shape we landed on.

**Keeper checklist, all green:**

- `--banner-border` + `--banner-border-radius` added on `.banner` as raw CSS vars (`Banner.svelte:116-117`) — no enum, no preset class, consumers compose card-style notifications through their own `.banner-error { --banner-border: 1px solid …; --banner-border-radius: 6px; }`. Defaults preserve today's render.
- `title` reshaped as a `Snippet` (`properties.ts:13`) wrapped in `.banner-body` with `gap: var(--banner-body-gap, 0)` and `.banner-title { font-weight: var(--banner-title-font-weight, inherit) }` (`Banner.svelte:137-151`). Defaults of `0` and `inherit` mean omitting `title` reproduces today's markup-and-pixels exactly — primitive-not-composite, as discussed.
- `role` is a plain `string | null` escape hatch (`Banner.svelte:53`) resolving to `role !== null ? role : interactive ? 'button' : null` — zero implicit tone derivation, consumers pass `role="alert"` explicitly when they want it.
- README example uses `classes="banner-error banner-compact"` driving the new vars (`docs/Banner.md:58-77`) — that's the §9 canonical pattern documented in-line so the next contributor reaches for it.

**Rejected surface, all dropped:**

- No `tone` enum, no `.banner-tone-*` rules, no automatic `tone === 'error' → role="alert"` branch.
- No `size` enum, no `.banner-size-sm`.
- No `flush` boolean, no `.banner-flush`.

**GUIDELINES nits, folded in:**

- `filter(Boolean)` → `filter((c) => c.length > 0)` in the `rootClass` derivation (`Banner.svelte:24`) per §3.
- `handleClick` / `handleKeydown` / `handleDismiss` kept as `function` declarations (`Banner.svelte:25,29,38`) — no arrow-const regression.
- Redundant `title !== null` clause gone; `typeof title === 'function'` is the sole guard (`Banner.svelte:63`) per §5.

One tiny non-blocking note: `border: var(--banner-border);` (`Banner.svelte:116`) has no explicit fallback, so it resolves to the CSS initial (`none`). That's fine and intentionally lets consumers opt in, just calling it out so it doesn't read as an oversight on future review.

**Ask: approving — please merge once CI is green.** Nice work turning a variant-system PR into a primitive PR in one round.

sinha-sahil (approved):
**Approving.**

murdore:
@sinha-sahil — per your review feedback, #233 now demos the new `title` snippet, `role` prop, and `--banner-border` / `--banner-border-radius` vars on the Banner route, and adds the title slot + `role` to `Banner.wc.svelte`. Verified locally: `pnpm check` + `check:wc` clean. Ready for re-review 🙏

PR #239, opened 2026-06-01
feat(EmptyState): make description optional

murdore:
## Summary

Makes `description` optional on `EmptyState`. The description row is omitted entirely when the prop is absent or empty.

## Why

- The no-baked-defaults philosophy says required props earn their place. `description` is not load-bearing: an empty state with just a title (\"No results found\") is a legitimate and common shape.
- Unblocks downstream consumer migration cases where the source component had only a single message string (Lighthouse's \`ConnectionErrorState\`, for one — but the principle is general).
- Strictly additive: existing consumers passing both \`title\` and \`description\` are unaffected.

## Changes

| File | Change |
|---|---|
| \`src/lib/EmptyState/properties.ts\` | Move \`description\` from \`MandatoryEmptyStateProperties\` to \`OptionalEmptyStateProperties\` |
| \`src/lib/EmptyState/EmptyState.svelte\` | Wrap the description div in \`{#if typeof description === 'string' && description.length > 0}\` (typeof guard per §3, empty-string drop per §5) |
| \`docs/EmptyState.md\` | Description row in props table → \`Required: No\`; add \"Title Only\" usage example; clarify in component prose |

## Test plan

- [x] \`pnpm run check\` — 0 errors, 0 warnings
- [x] \`pnpm run lint\` — prettier + eslint clean
- [x] Existing consumers (title + description both set) render identically — verified by the same code path
- [x] New \`title\` only path renders the title and skips the description div entirely

## Summary by CodeRabbit

* **New Features**
  * EmptyState component's `description` prop is now optional; the description element is conditionally rendered only when provided with a non-empty value.

* **Documentation**
  * Updated component documentation and type definitions to reflect that the `description` property is now optional, and includes a title-only usage example.

### DIFF
--- a/docs/EmptyState.md
+++ b/docs/EmptyState.md
@@ -1,6 +1,6 @@
 # EmptyState
 
-A centered placeholder component for empty lists, search results, or views. Displays an optional icon/illustration, a title, a description, and an optional action area (e.g. buttons). Inherits text color from its parent for seamless light/dark theme support.
+A centered placeholder component for empty lists, search results, or views. Displays an optional icon/illustration, a required title, an optional description, and an optional action area (e.g. buttons). Inherits text color from its parent for seamless light/dark theme support.
 
 ## Usage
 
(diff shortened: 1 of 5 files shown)


murdore:
@sinha-sahil — per your review feedback, #239 now demos the title-only (no description) rendering on the EmptyState route, and registers EmptyState in `docs/_index.json` (it was missing entirely, so the component was invisible to the MCP server). Verified locally: `pnpm check` clean, name-invariant holds. Ready for re-review 🙏

PR #242, opened 2026-06-07
feat(Pagination): page-size selector, item-range summary, cursor/load-more mode

murdore:
## Summary

The library `Pagination` was **page-number buttons only**, so consumers had to hand-roll a wrapper to get a full data-table paginator. In Lighthouse, `DataGrid/PaginatorRecipe.svelte` wraps `Pagination` with a page-size `<Select>`, a "Showing X–Y of N" label, and a cursor load-more CTA. This folds all of that into `Pagination` so the wrapper is no longer needed (the consumer can delete its recipe and use the library component directly).

## What's new (all opt-in — fully backward compatible)

No new prop ⇒ identical to today (just the nav buttons). Adding any of these turns on the richer "summary + nav" bar:

| Prop | Effect |
|---|---|
| `totalItems` (+ `showItemRange`, default true) | renders **"Showing X–Y of N"** |
| `showPageSizeSelector` + `pageSize` (bindable) + `pageSizeOptions` (default `[5,10,15,25,50,100]`) | renders a page-size **`<Select>`** in the summary |
| `hasMore` + `loadMoreLabel` (default `Next`) + `onLoadMore` | on the **last known page** the inert next-button becomes a **load-more CTA** (cursor pagination) |
| `onPageSizeChange` | fired on size change; `currentPage` resets to 1 |

- The summary bar is only rendered when `totalItems` and/or `showPageSizeSelector` are set; otherwise the plain `<nav class="pagination">` renders exactly as before (existing `testId`/`classes`/`.prev-button`/`.next-button` unchanged).
- All new visuals are **CSS-variable themed** (`--pagination-bar-*`, `--pagination-summary-*`, `--pagination-page-size-min-width`, `--pagination-load-more-*`).
- WC wrapper (`sui-pagination`), the `/components/pagination` demo, and types updated.

## Testing

- `npm run check` → **0 errors, 0 warnings**
- `npm run lint` → clean (changed files)
- `npm run package` (svelte-package + publint) → builds clean
- `npm run test:integration` → **4 new Pagination tests pass** (backward-compat, page-size + count, item-range, cursor load-more)
- `npm run test:unit` → pass

> Note: one **pre-existing, unrelated** integration failure (`tests/test.ts` — index-page `h1`) exists on `release` and is not touched by this change (the `/` route has no matching heading; diff only touches the 4 Pagination files).

## Follow-up
Once published + bumped, Lighthouse can delete `DataGrid/PaginatorRecipe.svelte` and use `<Pagination>` directly.

## Release Notes

* **New Features**
  * Added optional item range summary display (e.g., "Showing 1–10 of 50")
  * Added page-size selector control for dynamic items-per-page configuration
  * Added cursor-style "load more" navigation mode as an alternative to traditional pagination
  * Extended pagination component with configurable summary area

* **Tests**
  * Added comprehensive test coverage for pagination scenarios

### DIFF
--- a/docs/Pagination.md
+++ b/docs/Pagination.md
@@ -22,15 +22,16 @@ Page-level navigation with numbered page buttons, prev/next controls, and ellips
 | disabled         | `boolean` | No       | `false` | Whether the entire pagination is disabled. When true, all buttons become non-interactive, the container dims (opacity 0.5), and the cursor changes to not-allowed.                                                                                                                                                     |
 | testId           | `string`  | No       | `-`     | Value for the `data-pw` attribute on the nav container, used for end-to-end testing selectors.                                                                                                                                                                                                                         |
 | classes          | `string`  | No       | `-`     | CSS class string applied to the component's top-level element. Useful for theming — define classes with CSS variable overrides and pass them to create variant styles.                                                                                                                                                 |
-| hasMore          | `boolean` | No       | `false` | Cursor-pagination hint. When `true`, the next button stays enabled even when `currentPage >= totalPages`. Use this when the total page count is not known ahead of time (e.g. cursor-based APIs). `totalPages` takes precedence: the next button is enabled only if `hasMore` is `true` OR `currentPage < totalPages`. |
+| hasMore          | `boolean` | No       | `false` | Cursor / load-more mode. When `true`, the next button stays enabled past `currentPage >= totalPages`; on the last known page it becomes a load-more CTA (see `onLoadMore`). `totalPages` precedence: the plain next-button is enabled only if `hasMore` is `true` OR `currentPage < totalPages`. |
 | prevButtonTestId | `string`  | No       | `-`     | Value for the `data-pw` attribute on the previous-page button, used for end-to-end testing selectors.                                                                                                                                                                                                                  |
-| nextButtonTestId | `string`  | No       | `-`     | Value for the `data-pw` attribute on the next-page button, used for end-to-end testing selectors.                                                                                                                                                                                                                      |
+| nextButtonTestId | `string`  | No       | `-`     | Value for the `data-pw` attribute on the next-page button (and the load-more CTA in cursor mode), used for end-to-end testing selectors.                                                                                                                                                                                |
 
 ## Events
 
-| Event    | Type                     | Description                                                                                                                                                                                                                      |
-| -------- | ------------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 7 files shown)


murdore:
## Rework: Slim to cursor/load-more primitive only

This PR has been reworked per the maintainer doctrine review. Here is what changed and why:

### Dropped (consumer-recipe violations)
- **\`showPageSizeSelector\`**, **\`pageSizeOptions\`**, **\`onPageSizeChange\`** — a page-size \`<Select>\` is a consumer-assembled wrapper, not a library primitive. Same ruling as #226 \`showDivider\` and #213 \`showSerialNumber\`.
- **\`showItemRange\`**, **\`totalItems\`** — the \"Showing X–Y of N\" label is a read-only text snippet that the consumer renders alongside \`<Pagination>\`. No library code is needed.
- **\`loadMoreLabel\`** — string-convenience prop that bakes a visual preset. The load-more button now uses the standard next-button arrow (&#8250;) styled via \`--pagination-load-more-*\` CSS variables. Consumers who need a custom label can override via \`--pagination-load-more-*\` theming.
- **Entire \`<div class=\"pagination-bar\">\` summary DOM block** — all markup that lived there is now consumer responsibility.

### Kept
- **\`hasMore\`** + **\`onLoadMore\`** — these two props gate a behavior change on the existing next-button (replacing it with a load-more CTA when on the last known page). This logic cannot be expressed without touching the component's internal button rendering. They are the thinnest-possible primitives.

### Fixed
- **\`pageSize = $bindable(10)\`** misuse — removed entirely (no longer needed after the summary block was dropped). The component must not self-mutate consumer-owned data.
- **Hardcoded CSS in removed blocks** — the \`font-size: 14px\`, \`color: #6b7280\`, \`color: #3a4550\` rules in the now-deleted summary classes are gone entirely (the summary block was removed, so no conversion needed).

### Updated
- **\`docs/Pagination.md\`** — removed dropped props from the Props table; added a \"Load-more Button\" CSS variables table for the new \`--pagination-load-more-*\` hooks; added a \"Cursor / Load-more mode\" section with a usage example; added a \"Consumer recipes\" section showing how to self-serve a page-size selector and item-range summary around the slim \`<Pagination>\` primitive (the Lighthouse \`PaginatorRecipe.svelte\` pattern).
- **\`Pagination.wc.svelte\`** — removed dropped props from the web-component binding.
- **Demo route** — updated to show only backward-compat basic demo and the cursor/load-more demo.
- **\`tests/pagination.test.ts\`** — replaced removed-feature tests with backward-compat test + cursor-mode test.

### Net API delta after rework
+2 props (`hasMore: boolean`, `onLoadMore: () => void`) + load-more button CSS hooks. Everything else stays in consumer land.

### Rebased
Branch rebased onto `origin/release` (was previously based on an older commit). All gates pass: `npm run check` (0 errors) ✓, `npm run lint` (clean) ✓, `npm run build` (success) ✓.

PR #243, opened 2026-06-08
feat(RadioGroup): declarative radio group with radio + segmented variants

murdore:
## Summary

Adds a **`RadioGroup`** component. Today the library ships only the single `Radio` primitive, so every consumer has to hand-wire a group (render N `Radio`s sharing `name` + `bind:selectedValue`, plus group semantics and keyboard nav). This adds the declarative group so consumers pass an `items` array and bind one value — mirroring the `Select` options pattern.

Motivation: a bare radio button is rarely useful on its own — radios are inherently a group control. Lighthouse (and other consumers) were each re-implementing the same `RadioGroup` wrapper; this moves it into the library.

## API
```ts
export type RadioGroupItem = { label: string; value: string; subtitle?: string; disabled?: boolean; testId?: string };

export type RadioGroupProperties =
  { name: string; items: RadioGroupItem[] } &                                   // mandatory
  { selectedValue?: string; variant?: 'radio' | 'segmented';                    // optional ($bindable selectedValue)
    direction?: 'vertical' | 'horizontal'; disabled?: boolean;
    testId?: string; classes?: string } &
  { onchange?: (value: string) => void };                                       // events
```

- **`variant="radio"`** (default) — composes the existing `Radio` primitive; `role="radiogroup"` + roving-focus keyboard nav (Arrow/Home/End). `direction` lays out vertical or horizontal.
- **`variant="segmented"`** — pill/toggle control (`role="radio"` buttons) with optional per-item `subtitle`; themed via `--radio-group-segment-*` CSS vars.
- Per-item `data-pw` from `testId` (explicit item id → namespaced group id → item value).

## Conventions followed
- `src/lib/RadioGroup/{RadioGroup.svelte,properties.ts}`, exported from `src/lib/index.ts`.
- Web-component wrapper `src/wc/components/RadioGroup.wc.svelte` (`sui-radio-group`), registered in `src/wc/index.ts`.
- Demo route `src/routes/components/radio-group/` + `_nav.ts` entry (Form Controls).

## Gates
- `npm run check` — 0 errors / 0 warnings
- `npm run lint` — clean (prettier + eslint)
- `npm run build` — ok

## Demo
`/components/radio-group` shows: vertical (with a disabled item), horizontal, segmented, segmented-with-subtitles, and disabled group — verified in light + dark.

## Downstream
Lighthouse will consume this and delete its project-local `RadioGroup` wrapper once this publishes + the dependency is bumped (same sequence as Divider / Pagination).

## Summary by CodeRabbit

* **New Features**
  * Introduced RadioGroup component with two interaction modes: traditional radio buttons and segmented button toggles.
  * Supports vertical and horizontal layout directions.
  * Includes keyboard navigation (arrow keys, Home/End) and styling for active, hover, and disabled states.

* **Documentation**
  * Added component demo page with multiple configuration examples.


murdore:
Closing this PR as it is superseded by #208, which is the approved base implementation for the Banner component migration. This version reintroduces `variant` and `direction` props along with arrow-const event handlers, both of which were explicitly ruled out in the review that led to #208's approval. The consumer-side approach established in #208 handles these concerns without adding banned API surface to the library. Thank you for the contribution — the core migration work here informed the final design in #208.

PR #258, opened 2026-06-10
feat(ThemeSwitcher): add link mode — inline cycle button, no chrome

murdore:
## What / Why

Lighthouse needs a bare inline theme toggle that fits inside a navigation bar or header without imposing any visual chrome. The existing `toggle` mode is visually a button (fixed 36×36px, background, border-radius); `segment` mode is a filled track. Neither is suitable as a link-like inline element. This PR adds `mode='link'` — a third identifier in the existing `ThemeSwitcherMode` union — that renders a single inline chromeless button cycling through all options on click.

## Changes

- `properties.ts`: extends `ThemeSwitcherMode` to `'toggle' | 'segment' | 'link'`; adds optional `showLabel?: boolean` (default `false`) to show the active option's label beside the icon in link mode
- `ThemeSwitcher.svelte`: adds a `{:else if effectiveMode === 'link'}` branch using the existing `handleToggle` (identical cycle logic); CSS block for `.link-button` with six CSS vars, all with neutral `inherit`/`0`/`none` defaults; no font-family/size/weight/line-height hard-coded
- `ThemeSwitcher.wc.svelte`: exposes `show-label` attribute
- `docs/ThemeSwitcher.md`: updated mode type, added `showLabel` prop row, added Link Mode CSS Variables table, consumer recipe example, updated type reference
- `docs/_index.json`: updated ThemeSwitcher description
- Demo route: three new demo instances (icon-only, icon+label, styled via CSS vars)

## Doctrine compliance

- Mode identifier only — no visual preset baked in; all styling via CSS vars
- CSS var defaults are neutral (`inherit`, `0`, `none`) — existing consumers see zero visual change
- No typography properties in the style block; `font-size` forwarded via `--theme-switcher-link-font-size: inherit`
- Explicit type guards: `typeof currentOption?.icon === 'function'` before `{@render}`
- Named functions use `function` declarations; arrow only for inline callbacks
- `data-pw` wired identically to the other modes
- Gates: `npm run check` (0 errors) + `npm run lint` (clean) + `npm run build` (clean)

## Lighthouse consumer recipe

Replace `mode='segment'` in `+layout.svelte` with `mode='link'` and set `--theme-switcher-link-color` / `--theme-switcher-link-color-hover` via `.global-theme-toggle-container` in `globalClasses.css`, keeping the existing `--theme-switcher-icon-*` vars in place.

### DIFF
--- a/docs/ThemeSwitcher.md
+++ b/docs/ThemeSwitcher.md
@@ -1,6 +1,6 @@
 # ThemeSwitcher
 
-A theme preference switcher that renders as either a single toggle button or a segmented control. Cycles through configurable options (default: light, dark, system) with animated icon transitions. Persists the user preference to localStorage. Built-in icons are provided for `light`, `dark`, and `system` values; custom options get a generic palette icon, or you can supply your own via the `icon` snippet on each option. This component does NOT apply the theme to the page — it only provides the toggle UI and preference management. The consumer is responsible for applying CSS classes or variables based on the selected value.
+A theme preference switcher that renders as a single toggle button, a segmented control, or a bare inline link-style button. Cycles through configurable options (default: light, dark, system) with animated icon transitions. Persists the user preference to localStorage. Built-in icons are provided for `light`, `dark`, and `system` values; custom options get a generic palette icon, or you can supply your own via the `icon` snippet on each option. This component does NOT apply the theme to the page — it only provides the toggle UI and preference management. The consumer is responsible for applying CSS classes or variables based on the selected value.
 
 ## Usage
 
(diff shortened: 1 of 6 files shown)


murdore:
Closing — redundant. The existing `mode='toggle'` already provides the three-option cycling behavior (handleToggle cycles `(currentIndex + 1) % options.length`, demoed as 'Three-option cycling toggle' on the docs page). Any chrome adjustments belong in consumer CSS per the library doctrine, not a new mode. The consumer fix is simply `mode="toggle"` at the callsite.

PR #260, opened 2026-06-12
feat(Select): resting trigger box-shadow + value/placeholder font CSS vars

murdore:
## What

Five new CSS variables on Select, all defaulting to current behavior (`none` / `inherit`) — zero visual change for existing consumers:

| Variable | Applies to | Default |
|---|---|---|
| `--select-trigger-box-shadow` | `.select-trigger` resting state | `none` |
| `--select-value-font-size` | `.select-value` | `inherit` |
| `--select-value-font-weight` | `.select-value` | `inherit` |
| `--select-placeholder-font-size` | `.select-placeholder` | `inherit` |
| `--select-placeholder-font-weight` | `.select-placeholder` | `inherit` |

`docs/Select.md` updated (trigger table + a Value & Placeholder section).

## Why

- The trigger has a focus shadow var but no **resting** shadow hook — design systems with elevated inputs currently need a consumer rule on `.select-trigger` to add one.
- `.select-value` / `.select-placeholder` have no typography hooks, so host apps whose global stylesheets size bare text differently (e.g. a global `span` rule) can't align the trigger text with their own controls via variables. Lighthouse currently carries exactly these rules as documented interim overrides on the library's DOM classes, with these same variable names — once this publishes, those host-side rules delete wholesale.

## Verification

`pnpm run check` 0 errors · `pnpm run lint` clean · `pnpm run build` clean. Var-only change; no markup, props, or JS touched.

## Summary by CodeRabbit

* **Documentation**
  * Enhanced CSS Variables documentation for Select component with expanded styling options.

* **Style**
  * Added configurable box-shadow styling for Select trigger element.
  * Introduced CSS variables for value and placeholder typography (font-size and font-weight).

### DIFF
--- a/docs/Select.md
+++ b/docs/Select.md
@@ -73,16 +73,21 @@ Override these custom properties to theme the component.
 | `--select-trigger-background`         | `#ffffff`                              | background    | Background color of the trigger area.                       |
 | `--select-trigger-border`             | `1px solid #cccccc`                    | border        | Border of the trigger area.                                 |
 | `--select-trigger-border-radius`      | `6px`                                  | border-radius | Corner rounding of the trigger area.                        |
+| `--select-trigger-box-shadow`         | `none`                                 | box-shadow    | Resting box shadow of the trigger area.                     |
 | `--select-trigger-transition`         | `border-color 0.15s, box-shadow 0.15s` | transition    | Transition for trigger hover/focus effects.                 |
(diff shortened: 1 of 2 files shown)


murdore:
Closing — wrong shape on review. The component root already exposes `--select-font-size` (and the value/placeholder spans inherit it), so per-element font vars exist only to override the component's own root variable — API bloat. The consumer's design system styles trigger text through its global typography layer anyway, and the resting box-shadow has been dropped consumer-side to match the library's stock trigger. No replacement needed.

PR #261, opened 2026-06-12
feat(Pagination): stamp {testId}-prev / {testId}-next on the arrow buttons

murdore:
## What

When `testId` is set, the previous/next arrow buttons now carry `data-pw="{testId}-prev"` / `data-pw="{testId}-next"` in addition to the existing `data-pw` on the nav container. No new props; `null` (attribute omitted) when `testId` is unset — zero change for consumers that don't pass one.

`docs/Pagination.md` testId row updated.

## Why

E2E suites can currently target the whole pagination but not the arrows — page-flip tests end up with brittle `nth(...)`/class selectors on `.prev-button`/`.next-button`. Same per-element pattern Select already uses for its pills (`{testId}-pill-{id}`).

## Verification

`pnpm run check` 0 errors · `pnpm run lint` clean · `pnpm run build` clean.


murdore:
Closing — redundant. The nav container already carries `data-pw={testId}` and the arrow buttons have stable aria-labels ('Previous page' / 'Next page'), so tests address them as `getByTestId(x).getByRole('button', { name: 'Previous page' })`. Stamping derived per-button testIds duplicates an addressing path the component already provides.

PR #297, opened 2026-06-16
feat(Gauge): fix NaN/Infinity on max=0, add max+labelFormatter props, remove typography overrides

murdore:
## What

- **Blocker fix**: Guard `(value / max) * 100` with `max > 0 ? ... : 0`.  
  Previously `value=0, max=0` produced `NaN` (0/0), which propagated through `Math.min`/`Math.max` into `stroke-dashoffset="NaN"` — the SVG arc became invisible. Similarly `value>0, max=0` yielded `Infinity`, which happened to clamp to 100 by accident but was semantically wrong and undocumented. Both cases now render the gauge at 0%.

- **Minor fix**: Remove `font-size`, `font-weight`, `font-family` from `.label` style block.  
  Per library pattern, components must not set typography properties — the design system wires them to semantic tags. Only `--gauge-label-color` (a valid theming CSS var) is retained.

- **Clarity comment**: Added comments documenting the `labelFormatter` contract — the formatter receives raw `(value, max)` while the built-in fallback shows the computed percentage string (e.g. value=50, max=200 → "25%").

## Why

The `max` prop and `labelFormatter` props were added in the previous commit on this branch. The division-by-zero path was not guarded, making the component broken for the `max=0` edge case. The typography override was a pre-existing violation that this PR touched (label rendering changed) and had an opportunity to correct.

## Backward-compatibility

- `max` defaults to `100` — identical output to before for all `max>0` inputs.  
- `labelFormatter` is optional; without it the label continues to render `${Math.round(clamped)}%`.  
- No mandatory prop was changed. No public API surface was removed.  
- Existing callers that relied on the old `.label` font styling via the removed CSS vars (`--gauge-label-font-size`, etc.) should supply those values via the `classes` prop or a parent selector — the CSS vars never had official library backing so this is not a breaking change.

### DIFF
--- a/src/lib/Gauge/Gauge.svelte
+++ b/src/lib/Gauge/Gauge.svelte
@@ -1,7 +1,14 @@
 <script lang="ts">
   import type { GaugeProperties } from './properties';
 
-  let { value, showLabel = true, testId, classes }: GaugeProperties = $props();
+  let {
+    value,
+    max = 100,
+    showLabel = true,
+    labelFormatter,
+    testId,
+    classes
+  }: GaugeProperties = $props();
 
   const VIEW_BOX = 100;
   const CENTER = VIEW_BOX / 2;
@@ -9,11 +16,25 @@
   const RADIUS = (VIEW_BOX - STROKE) / 2;
   const CIRCUMFERENCE = 2 * Math.PI * RADIUS;
 
-  let clamped = $derived(Math.min(100, Math.max(0, value)));
+  // Guard against max=0 (division by zero yields NaN or Infinity).
+  // When max is 0 or negative the gauge renders empty (0%).
+  let clamped = $derived(max > 0 ? Math.min(100, Math.max(0, (value / max) * 100)) : 0);
   let offset = $derived(CIRCUMFERENCE - (clamped / 100) * CIRCUMFERENCE);
+  // labelFormatter receives the raw (value, max) pair so callers have full
+  // context. The built-in fallback shows the computed percentage string
+  // (e.g. value=50, max=200 → "25%").
+  let labelText = $derived(labelFormatter ? labelFormatter(value, max) : `${Math.round(clamped)}%`);
 </script>
 
-<div class="gauge {classes ?? ''}" data-pw={typeof testId === 'string' ? testId : null}>
+<div
+  class="gauge {classes ?? ''}"
+  data-pw={typeof testId === 'string' ? testId : null}
+  role="progressbar"
+  aria-valuenow={clamped}
+  aria-valuemin={0}
+  aria-valuemax={100}
+  aria-label="gauge"
+>
   <svg viewBox="0 0 {VIEW_BOX} {VIEW_BOX}">
     <circle class="track" cx={CENTER} cy={CENTER} r={RADIUS} fill="none" />
     <circle
@@ -29,7 +50,7 @@
     />
   </svg>
   {#if showLabel}
-    <div class="label">{Math.round(clamped)}%</div>
+    <div class="label">{labelText}</div>
   {/if}
 </div>
 
--- a/src/lib/Gauge/properties.ts
+++ b/src/lib/Gauge/properties.ts
@@ -5,7 +5,21 @@ export type MandatoryGaugeProperties = {
 };
 
 export type OptionalGaugeProperties = {
+  /**
+   * The maximum value of the gauge. `value` is divided by `max` to compute
+   * the fill percentage, so `<Gauge value={50} max={200} />` renders at 25%.
+   * Defaults to `100` (making `value` a direct percentage). When `max` is 0
+   * or negative the gauge renders empty (0%) to avoid division by zero.
+   * @default 100
+   */
+  max?: number;
   showLabel?: boolean;
+  /**
+   * Custom label renderer called with the raw `(value, max)` pair. Return any
+   * string to replace the default `"${Math.round(percentage)}%"` label.
+   * Example: `labelFormatter={(v, m) => \`${v} / ${m}\`}` shows "50 / 200".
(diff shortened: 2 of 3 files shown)


src/lib/Gauge/properties.ts
```diff
@@ -5,7 +5,9 @@ export type MandatoryGaugeProperties = {
 };
 
 export type OptionalGaugeProperties = {
+  max?: number;
```
Tara-ag, line 15: 💬 SUGGESTION: Consider adding JSDoc comments to document the new props for better IDE autocomplete support:

```typescript
export type OptionalGaugeProperties = {
  /** Maximum value for computing the percentage. Defaults to 100. */
  max?: number;
  showLabel?: boolean;
  /** Custom formatter for the label text. Receives (value, max) and should return a string. */
  labelFormatter?: (value: number, max: number) => string;
  testId?: string;
  classes?: string;
};
```

src/lib/Gauge/Gauge.svelte
```diff
@@ -1,16 +1,29 @@
...
   const VIEW_BOX = 100;
   const CENTER = VIEW_BOX / 2;
   const STROKE = 8;
   const RADIUS = (VIEW_BOX - STROKE) / 2;
   const CIRCUMFERENCE = 2 * Math.PI * RADIUS;
 
-  let clamped = $derived(Math.min(100, Math.max(0, value)));
+  // Guard against max=0 (division by zero yields NaN or Infinity).
+  // When max is 0 or negative the gauge renders empty (0%).
+  let clamped = $derived(max > 0 ? Math.min(100, Math.max(0, (value / max) * 100)) : 0);
```
Tara-ag, line 21: ⚠️ MAJOR: The Gauge component lacks accessibility attributes. As a data visualization component showing progress/percentage, it should include:
- `role="progressbar"` on the root element
- `aria-valuenow={clamped}` 
- `aria-valuemin="0"`
- `aria-valuemax="100"`
- An accessible label via `aria-label` or `aria-labelledby`

This ensures screen reader users can perceive the gauge value. Consider adding an optional `label` prop for the accessible name:

```svelte
<div 
  class="gauge {classes ?? ''}" 
  role="progressbar"
  aria-valuenow={clamped}
  aria-valuemin="0"
  aria-valuemax="100"
  aria-label={ariaLabel ?? labelText}
>
```

src/lib/Gauge/Gauge.svelte
```diff
@@ -62,9 +75,6 @@
 
   .label {
     position: absolute;
-    font-size: var(--gauge-label-font-size, 24px);
-    font-weight: var(--gauge-label-font-weight, 600);
-    font-family: var(--gauge-label-font-family, inherit);
     color: var(--gauge-label-color, #333);
```
Tara-ag, line 89: ⚠️ MAJOR: The removal of `--gauge-label-font-size`, `--gauge-label-font-weight`, and `--gauge-label-font-family` CSS custom properties is a breaking change. These variables are documented in `docs/Gauge.md` and downstream consumers may rely on them for theming.

Per the blocking criteria: "Renamed CSS custom properties that downstream themes rely on" without a major-version bump is a breaking change.

Either:
1. Restore these CSS variables with their defaults, OR
2. Bump the major version and provide a migration note, OR  
3. Update the documentation to clearly mark this as a breaking change with migration path

Suggested fix (restore the variables):
```css
.label {
  position: absolute;
  font-size: var(--gauge-label-font-size, 24px);
  font-weight: var(--gauge-label-font-weight, 600);
  font-family: var(--gauge-label-font-family, inherit);
  color: var(--gauge-label-color, #333);
}
```

Tara-ag:
## Documentation Out of Sync

The `docs/Gauge.md` file needs to be updated to reflect the PR changes:

1. **New props not documented**: The `max` and `labelFormatter` props are not documented in the Props table.

2. **Removed CSS variables still documented**: `--gauge-label-font-size`, `--gauge-label-font-weight`, and `--gauge-label-font-family` are still listed in the CSS Variables table but were removed from the component.

3. **`value` prop description is outdated**: The description says "Current percentage value (0-100)" but with the `max` prop, `value` is now a raw value that gets computed against `max` to produce the percentage.

Please update the documentation to reflect these changes.

Tara-ag (changes requested):
## Review Summary

**Files reviewed:** 2 (`src/lib/Gauge/Gauge.svelte`, `src/lib/Gauge/properties.ts`)

### Issues Found

| ⚠️ MAJOR | 2 | Blocking issues requiring changes |
| 💡 MINOR | 1 | Documentation out of sync |
| 💬 SUGGESTION | 1 | Code quality improvement |

### Blocking Issues (MAJOR)

1. **Breaking Change: CSS Variables Removed** - The removal of `--gauge-label-font-size`, `--gauge-label-font-weight`, and `--gauge-label-font-family` CSS custom properties breaks backward compatibility. These variables are documented in `docs/Gauge.md` and downstream consumers may rely on them for theming. Per the library's blocking criteria, this requires either restoring the variables or a major version bump with migration notes.

2. **Missing Accessibility Attributes** - The Gauge component lacks proper ARIA attributes for screen reader users. It should include `role="progressbar"`, `aria-valuenow`, `aria-valuemin`, `aria-valuemax`, and an accessible label.

### Next Steps

1. Either restore the removed CSS variables OR bump the major version and document the breaking change
2. Add ARIA attributes for accessibility
3. Update `docs/Gauge.md` to document the new `max` and `labelFormatter` props, and remove/update the documentation for the removed CSS variables
4. Consider adding JSDoc comments to the new props for better IDE support

The division-by-zero fix and `labelFormatter` functionality are solid improvements. Once the blocking issues are addressed, this PR should be good to merge.

PR #302, opened 2026-06-16
feat(Table): fix selection toggle, stable IDs, onCellChange docs, focus ring, optional enabled

murdore:
## What

Fixes all gate-review blockers and majors for the Table C2 feature branch.

### Blocker — C2-1: single-select toggle broken
`toggleRowSelection()` called `selectedIds.clear()` before `selectedIds.has(rowId)`, making the `has()` check always return `false` on the now-empty set. Clicking an already-selected row re-selected it instead of deselecting. Fix: capture `const wasSelected = selectedIds.has(rowId)` before `clear()` and branch on it.

### Blocker — C2-2: `onCellChange` was a dead prop with a misleading recipe
The JSDoc showed `onCellChange?.(rowIndex, colIndex, newValue)` inside a consumer `cell` Snippet as if Table would forward the call. Svelte 5 snippets execute in consumer scope — Table itself never calls this callback. The recipe was actively misleading. The prop is retained (it's valid as a parent-component API surface), but the internal destructured binding is renamed to `_onCellChange` (satisfies the unused-var lint rule), and the JSDoc is rewritten to make clear that consumers must close over their own handler directly inside the snippet.

### Major — C2-1 + C2-3: positional IDs shifted on filter change
`selectableRowIds` mapped over `filteredTableData` with a local `rowIndex`, so default IDs (`String(0)`, `String(1)`, …) were positional in the *filtered* view. When the search term changed, the filtered list shrank/shifted and the same data row got a different ID — corrupting selection state. Fix: resolve each row's original index via `sortedTableData.indexOf(row)` so the default ID tracks the pre-filter position. The per-row `rowId` computation in the `{#each}` block applies the same pre-filter index lookup.

### Minor — C2-3: search input had no focus-visible ring
`.table-search-input { outline: none }` stripped the browser focus indicator with no replacement, making keyboard focus invisible. Added a `:focus-visible` rule using the existing `--table-focus-outline-color` CSS variable (same token already used on `.table-row-clickable`).

### Minor — C2-1: `enabled` field was a required boolean
`enabled: boolean` forced every consumer to write `checkboxSelection={{ enabled: true, … }}`. Changed to `enabled?: boolean` — presence of the config object implies intent to enable, `enabled: false` is still valid for explicit opt-out. All internal guards updated from `checkboxSelection?.enabled` to `!!checkboxSelection && checkboxSelection.enabled !== false`. Backward-compatible: existing consumers passing `enabled: true` are unaffected.

## Compatibility

- All new props remain optional; no existing required prop moved to optional.
- `TableCheckboxSelectionConfig.enabled` change: `boolean` → `boolean | undefined` (adding `undefined` widens the accepted type, not a breaking change; consumers already passing `true`/`false` continue to work).
- CSS: one new `:focus-visible` rule added; no existing vars removed.

### DIFF
--- a/src/lib/Table/Table.svelte
+++ b/src/lib/Table/Table.svelte
@@ -1,10 +1,15 @@
 <script lang="ts">
-  import type { TableProperties } from './properties';
+  import type { TableProperties, TableCheckboxSelectionConfig } from './properties';
   import type { JSONValue } from 'type-decoder';
+  import { SvelteSet } from 'svelte/reactivity';
   import Button from '../Button/Button.svelte';
   import chevronUpSvg from '$lib/assets/chevron-up.svg?raw';
   import chevronDownSvg from '$lib/assets/chevron-down.svg?raw';
   import sortDefaultSvg from '$lib/assets/sort-default.svg?raw';
+  import searchSvg from '$lib/assets/search.svg?raw';
+  import closeSvg from '$lib/assets/close.svg?raw';
+  import checkmarkSvg from '$lib/assets/checkmark.svg?raw';
+  import minusSvg from '$lib/assets/minus.svg?raw';
 
   let {
     tableTitle = '',
@@ -24,16 +29,21 @@
     empty,
     onRowClick,
     onSort,
+    onCellChange: _onCellChange,
     classes,
     paginatorSlot,
     getRowTestId,
-    getCellTestId
+    getCellTestId,
+    checkboxSelection,
+    searchConfig,
+    onSearchChange
   }: TableProperties = $props();
 
+  // ─── Sort state ──────────────────────────────────────────────────────────────
   let sortColumn = $state<number | null>(null);
   let sortDirection = $state<'asc' | 'desc'>('asc');
 
-  function isColumnSortable(colIndex: number): boolean {
+  const isColumnSortable = (colIndex: number): boolean => {
     if (!sortable) {
       return false;
     }
@@ -41,8 +51,61 @@
       return sortableColumns.includes(colIndex);
     }
     return true;
-  }
+  };
 
+  const handleSort = (colIndex: number): void => {
+    if (!isColumnSortable(colIndex)) {
+      return;
+    }
+
+    if (sortColumn === colIndex) {
+      sortDirection = sortDirection === 'asc' ? 'desc' : 'asc';
+    } else {
+      sortColumn = colIndex;
+      sortDirection = 'asc';
+    }
+    onSort?.(colIndex, sortDirection);
+  };
+
+  // ─── Row click ───────────────────────────────────────────────────────────────
+  const handleRowClick = (rowIndex: number, rowData: JSONValue[]): void => {
+    onRowClick?.(rowIndex, rowData);
+  };
+
+  const handleRowKeydown = (event: KeyboardEvent, rowIndex: number, rowData: JSONValue[]): void => {
+    if (event.key === 'Enter' || event.key === ' ') {
+      event.preventDefault();
+      onRowClick?.(rowIndex, rowData);
+    }
+  };
+
+  let isRowClickable = $derived(typeof onRowClick === 'function');
+  let isStickyHeader = $derived(stickyHeader || isTableScrollable);
+
+  // ─── C2-3: Search ─────────────────────────────────────────────────────────
+  let searchTerm = $state('');
+  let hasSearchConfig = $derived(!!searchConfig);
+  let isServerSearch = $derived(typeof onSearchChange === 'function');
+  let searchInputRef = $state<HTMLInputElement | null>(null);
+
+  const handleSearchInput = (): void => {
+    if (!searchInputRef) {
+      return;
+    }
+    searchTerm = searchInputRef.value;
+    if (isServerSearch) {
+      onSearchChange?.(searchTerm);
+    }
+  };
+
+  const clearSearch = (): void => {
+    searchTerm = '';
+    if (isServerSearch) {
+      onSearchChange?.('');
+    }
+  };
+
+  // ─── Sort + Search pipeline ──────────────────────────────────────────────────
   let sortedTableData = $derived.by(() => {
     if (sortColumn === null) {
       return [...tableData];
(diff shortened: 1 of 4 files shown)


src/lib/Table/Table.svelte
```diff
@@ -76,40 +139,183 @@
...
+    row: JSONValue[],
+    rowIndex: number
+  ): string => {
+    return cfg.getRowId ? cfg.getRowId(row, rowIndex) : String(rowIndex);
+  };
+
+  let selectedIds = new SvelteSet<string>();
+
+  const isRowDisabled = (rowId: string): boolean => {
+    return checkboxSelection?.disabledRowIds?.has(rowId) ?? false;
```
Tara-ag, line 176: ⚠️ MAJOR: `selectedIds` is initialized as a plain `SvelteSet` without `$state()`. In Svelte 5, for reactive state that triggers UI updates when modified, this should be `$state(new SvelteSet<string>())`. Without `$state`, mutations via `selectedIds.add()`, `delete()`, or `clear()` won't trigger reactivity updates in the template.

```suggestion
  let selectedIds = $state(new SvelteSet<string>());
```

src/lib/Table/Table.svelte
```diff
@@ -76,40 +139,183 @@
...
+  };
+
+  const isRowSelected = (rowId: string): boolean => {
+    return selectedIds.has(rowId);
+  };
+
+  /**
+   * IDs of all selectable (non-disabled) rows in the currently filtered view.
```
Tara-ag, line 205: ⚠️ MAJOR: The `selectableRowIds` derived uses `sortedTableData.indexOf(row)` to find the original index. This is O(n²) complexity and can cause performance issues with large datasets. Additionally, if rows contain duplicate values, `indexOf` may return the wrong index.

Consider using a Map for O(1) lookup or tracking indices during the initial mapping:
```suggestion
  let selectableRowIds = $derived.by((): string[] => {
    if (!checkboxSelection || checkboxSelection.enabled === false) {
      return [];
    }
    const cfg = checkboxSelection;
    // Build index map once for O(1) lookups
    const indexMap = new Map(sortedTableData.map((r, i) => [r, i]));
    return filteredTableData.map((row) => {
      const originalIndex = indexMap.get(row) ?? 0;
      return resolveRowId(cfg, row, originalIndex);
    });
  });
```

src/lib/Table/Table.svelte
```diff
@@ -76,40 +139,183 @@
...
+  let isCheckboxMode = $derived(!!checkboxSelection && checkboxSelection.enabled !== false);
+  let isSingleSelect = $derived(checkboxSelection?.selectionMode === 'single');
 </script>
 
 {#if typeof tableTitle === 'string' && tableTitle.length > 0}
   <div class="table-title">
     {tableTitle}
   </div>
 {/if}
+
```
Tara-ag, line 301: 💡 MINOR: The search input uses `type="search"` which is good, but there's no `autocomplete` attribute. Consider adding `autocomplete="off"` to prevent browser autocomplete from interfering with the table search functionality, especially when the table is used in forms.

```suggestion
    <input
      bind:this={searchInputRef}
      class="table-search-input"
      type="search"
      autocomplete="off"
      placeholder={searchConfig?.placeholder ?? 'Search…'}
      ...
    />
```

src/lib/Table/Table.svelte
```diff
@@ -121,6 +327,37 @@
       {/if}
       <thead>
         <tr>
+          {#if isCheckboxMode && !isSingleSelect}
```
Tara-ag, line 343: ⚠️ MAJOR: The header checkbox uses `role="checkbox"` with `tabindex={0}` but lacks proper keyboard handling for Space key activation. While `handleCheckboxKeydown` handles Enter and Space, the `aria-checked` attribute correctly uses `'mixed'` for indeterminate state, which is good.

However, the checkbox should also have `aria-controls` referencing the row checkboxes it controls for better screen reader context. Consider adding:
```suggestion
                aria-controls={selectableRowIds.join(' ')}
```
And ensure each row checkbox has a corresponding `id` attribute.

src/lib/Table/Table.svelte
```diff
@@ -76,40 +139,196 @@
...
+    row: JSONValue[],
+    rowIndex: number
+  ): string => {
+    return cfg.getRowId ? cfg.getRowId(row, rowIndex) : String(rowIndex);
+  };
+
+  let selectedIds = new SvelteSet<string>();
+
+  const isRowDisabled = (rowId: string): boolean => {
+    return checkboxSelection?.disabledRowIds?.has(rowId) ?? false;
```
Tara-ag, line 176: 💬 SUGGESTION: `SvelteSet` from `svelte/reactivity` is designed to be reactive without `$state()`. It's a reactive class using Proxies internally - wrapping in `$state()` would be redundant. See https://svelte.dev/docs/svelte/svelte-reactivity

The current implementation `let selectedIds = new SvelteSet<string>();` is correct and mutations will trigger reactivity properly.

src/lib/Table/Table.svelte
```diff
@@ -121,6 +340,37 @@
       {/if}
       <thead>
         <tr>
+          {#if isCheckboxMode && !isSingleSelect}
```
Tara-ag, line 343: ⚠️ MAJOR: The header checkbox should have `aria-controls` referencing the row checkboxes it controls for better screen reader context. Each row checkbox needs a unique `id` attribute.

```suggestion
<!-- In the header checkbox -->
<span
  class="table-checkbox-box"
  ...
  aria-controls={selectableRowIds.join(' ')}
  ...
>

<!-- In each row checkbox -->
<span
  class="table-checkbox-box"
  id={`row-checkbox-${rowId}`}
  ...
>
```

Tara-ag:
## ⚠️ MAJOR: Web Component Wrapper Missing New Props

The web component wrapper at `src/wc/components/Table.wc.svelte` hasn't been updated to expose the new props added in this PR:

- `checkboxSelection`
- `searchConfig`  
- `onCellChange`
- `onSearchChange`

Web component consumers won't be able to use the new checkbox selection and search features. Please update the `customElement.props` configuration to include these new properties.

```suggestion
// Add to the props object in Table.wc.svelte:
checkboxSelection: { type: 'Object' },
searchConfig: { type: 'Object' },
onCellChange: { type: 'Object' },
onSearchChange: { type: 'Object' }
```

Tara-ag (changes requested):
## Review Summary

**Files reviewed:** 4 files changed (+953, -87 lines)

### Issues Raised

| 🔒 CRITICAL | 0 | None |
| ⚠️ MAJOR | 1 | Web component wrapper missing new props |
| 💡 MINOR | 0 | None |
| 💬 SUGGESTION | 1 | SvelteSet reactivity clarification |

### Blocking Issues

1. **⚠️ MAJOR: Web Component Wrapper Missing New Props**
   - `src/wc/components/Table.wc.svelte` doesn't expose `checkboxSelection`, `searchConfig`, `onCellChange`, or `onSearchChange`
   - Web component consumers cannot use the new features
   - Posted as PR comment since the file wasn't modified in this PR

### Inline Comments Added

1. **Line 176 (Table.svelte)**: Clarified that `SvelteSet` from `svelte/reactivity` is correctly used without `$state()` - it's designed to be reactive natively.

2. **Line 343 (Table.svelte)**: Added accessibility suggestion for `aria-controls` relationship between header and row checkboxes.

### Existing Comments Acknowledged

- Tara-ag's comments about O(n²) performance and `aria-controls` are valid concerns
- The `SvelteSet` comment appears to be based on a misunderstanding of Svelte 5's reactive classes

### Next Steps

1. Update `src/wc/components/Table.wc.svelte` to expose new props
2. Consider addressing the `aria-controls` accessibility enhancement
3. Performance optimization for `indexOf` lookups is optional for typical table sizes

PR #316, opened 2026-06-18
fix(DateRangePicker): persist committed preset highlight across the open/apply cycle

murdore:
## Summary

Completes the same-day preset fix (#310). That change made an explicit in-session click match by label, but `openPicker()` still reset `selectedPresetLabel` to `null` on every open and `handleApply()` cleared the active label, so **re-opening** the picker fell back to `isSameDay()` matching and re-highlighted every preset sharing the committed day (e.g. *Today / Last 30 minutes / Last 12 Hours*) at once.

## Change

Track a `committedPresetLabel` (seeded from `initialPresetLabel`, updated on each Apply, cleared on Clear) and re-seed `selectedPresetLabel` from it when the picker opens. The active highlight is now driven by **label** across the whole open/apply lifecycle instead of date matching:

- `initialPresetLabel` highlights only the seeded preset on first open;
- re-opening after applying a preset re-highlights only that preset;
- a direct calendar click still clears it, reverting to date-based matching for custom ranges (unchanged).

## Tests

Adds `tests/date-range-picker-reopen.test.ts` — a Playwright regression guard that applies a same-day preset, re-opens, and asserts only the applied preset is highlighted.

## Downstream

Fixes two Lighthouse (Breeze) reports that depend on this: the analytics compare picker's *"No Comparison" cannot be deselected without clicking a date* (preset state desync on reopen), and *selecting Today / Last 30 min / Last 12 Hours highlights all three at once*.

## Gates

`lint` ✓ · `check` ✓ (0 errors) · `build`/`publint` ✓.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved date range picker preset highlighting behavior when reopening the picker; the component now correctly retains and displays the last applied preset as selected rather than reverting to alternative date-matching options.
  * Fixed preset initialization to ensure the configured default preset is properly highlighted when first opened.
  * Clearing a selection now properly resets preset highlighting.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,13 +2,21 @@
 based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased](https://github.com/juspay/svelte-ui-components/compare/HEAD..2.65.0)
+## [Unreleased](https://github.com/juspay/svelte-ui-components/compare/HEAD..2.66.0)
 
-Mirrors the existing --button-max-width hook (no fallback, so unset leaves
-min-width at its initial value — fully backward-compatible). Lets consumers
-set a button minimum width through the documented CSS-var API instead of
-piercing the rendered &lt;button&gt; with a scoped :global(button){min-width}
-override.
+The default multi-select indicator was a unicode glyph that rendered as a
+tiny, unstyled square inconsistent with the design system. Replace it with a
+presentational checkbox box (bordered square that fills accent-blue with a
+checkmark when selected), reusing the bundled checkmark.svg and new themeable
+--select-option-indicator-* CSS variables (size, border, border-radius,
+background, checked-background, checked-border-color, check-size, check-color).
+
+The box is presentational only — the option row owns selection and
+aria-selected — so no focusable role=checkbox is nested inside role=option and
+the click still toggles via the option. The optionIndicator snippet continues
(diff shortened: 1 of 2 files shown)


Tara-ag (approved):
## Review Summary

**Files reviewed:** 2
- `src/lib/DateRangePicker/DateRangePicker.svelte`
- `tests/date-range-picker-reopen.test.ts`

**Issues found:** None

### Analysis

This PR fixes a bug where re-opening the DateRangePicker after applying a same-day preset would incorrectly highlight all presets sharing that day (e.g., "Today", "Last 30 minutes", "Last 12 Hours") instead of just the applied one.

**Changes verified:**
- ✅ Internal state management only - no public API changes
- ✅ Proper use of Svelte 5 `$state` rune for `committedPresetLabel`
- ✅ Correct lifecycle: seeded from `initialPresetLabel`, updated on Apply, cleared on Clear
- ✅ `openPicker()` now re-seeds `selectedPresetLabel` from `committedPresetLabel` instead of resetting to `null`
- ✅ Playwright regression tests cover both the re-open scenario and initial preset seeding
- ✅ No SSR safety issues (no unguarded browser API access)
- ✅ No accessibility regressions (ARIA attributes unchanged)
- ✅ No breaking changes to props, events, exports, or CSS custom properties

The fix correctly addresses the root cause described in the PR: the previous implementation reset `selectedPresetLabel = null` on every open, causing fallback to date-based matching which lights up all same-day presets. The new `committedPresetLabel` state survives the open/apply cycle, enabling label-based matching on re-open.

**Approved** - Clean bug fix with appropriate test coverage.

PR #327, opened 2026-06-22
feat(Modal): add --modal-content-padding token for slot-content body …

priyankakumari-cmyk:
…padding

The Modal content area (.slot-content) had no padding and exposed no variable to control it, unlike the header (--modal-header-padding) and footer (--modal-footer-padding). Consumers had to reach into the internal .slot-content class to inset the body.

Add padding: var(--modal-content-padding, 0) so the body padding is controllable via a token, mirroring header/footer. Defaults to 0, so existing modals are unaffected; consumers opt in by setting the variable.

## Summary by CodeRabbit

* **New Features**
  * Modal content padding is now customizable via CSS variable.

* **Documentation**
  * Updated documentation to include the new CSS variable for modal content padding configuration.


priyankakumari-cmyk:
Added modal content padding token

priyankakumari-cmyk:
Added modal content padding token

PR #332, opened 2026-06-25
feat: add carousel test ids

anamika-tiwari25:
- Adds `data-pw` support to the Carousel root container, dots wrapper, and individual dots.
- Adds dynamic dot test IDs using 1-based indexing

### DIFF
--- a/src/lib/Carousel/Carousel.svelte
+++ b/src/lib/Carousel/Carousel.svelte
@@ -9,7 +9,10 @@
     showDots = false,
     isScrollableLast = false,
     onkeydown,
-    classes
+    classes,
+    testId,
+    dotsWrapperTestId,
+    dotTestId
   }: CarouselProperties = $props();
 
   let slidesDiv: HTMLDivElement | null = $state(null);
@@ -138,7 +141,10 @@
   });
 </script>
 
-<div class="carousel-container {classes ?? ''}">
+<div
+  class="carousel-container {classes ?? ''}"
+  data-pw={typeof testId === 'string' ? testId : null}
+>
   {#if views.length > 0}
     <div class="carousel" bind:this={carouselDiv}>
       <div class="slidesDiv" bind:this={slidesDiv}>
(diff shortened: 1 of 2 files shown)


src/lib/Carousel/Carousel.svelte
```diff
@@ -9,7 +9,13 @@
     showDots = false,
     isScrollableLast = false,
     onkeydown,
-    classes
+    classes,
+    testId,
+    carouselTestId,
+    slidesWrapperTestId,
+    slideTestId,
```
dakshana-j: need the purpose of all these test ids, and we need to change the naming

src/wc/components/Carousel.wc.svelte
```diff
@@ -9,6 +9,12 @@
       showDots: { type: 'Boolean', reflect: true, attribute: 'show-dots' },
       isScrollableLast: { type: 'Boolean', reflect: true, attribute: 'is-scrollable-last' },
       classes: { type: 'String' },
```
dakshana-j, line 11: why do we need to add the changes in this file, i think its not required ?

dakshana-j:
Summary part in the conversation needs to be changed and elaborated

src/lib/Carousel/Carousel.svelte
```diff
@@ -9,7 +9,13 @@
...
+    dotsWrapperTestId,
+    dotTestId
```
Nafisha-Simran, line 15: Confusing naming: testId (root .carousel-container) vs carouselTestId (inner .carousel) — the semantic distinction isn't obvious. Consider  carouselTestId or testId + viewportTestId

juspayutkarsh10: +1

src/wc/components/Carousel.wc.svelte
anamika-tiwari25: It was added to expose the test ID props for web component usage. It is not required for the Carousel functionality itself, hence removing it.

PR #336, opened 2026-06-28
feat(DataTable): add column-based data grid component

NehaJuspay493:
https://github.com/user-attachments/assets/0fc74f46-4680-4b52-b867-44e92674eeb4

## Summary

Combines the column-model **data grid** into the existing **`Table`** so there is
**one public component going forward**, while keeping every existing (positional)
`Table` usage working **unchanged**. This supersedes the original approach of this
PR (shipping a separate `DataTable` component).

## The problem

`Table` is a positional table: `tableHeaders: string[]` + `tableData: JSONValue[][]`.
The data grid is a different model: typed `ColumnDefinition[]` over row **objects**.
Their `cell` and `onRowClick` snippet/callback signatures are **incompatible**, so
the two can't share flat props. But the maintainer wants a single component, and
Lighthouse depends on the legacy `Table` API — so the merge must not break it.

## Approach — opt-in delegation via `dataGrid`

`Table` gains one new prop, **`dataGrid`**. When present, `Table` renders the advanced
engine; when absent, the legacy path is **byte-for-byte unchanged**.

```svelte
<!-- Legacy (unchanged) -->
<Table tableHeaders={['Name','Score']} tableData={rows} />

<!-- Advanced grid -->
<Table dataGrid={{ columns, data, idField: 'id', enableSearch: true, pagination }} />

- Table is generic (<T extends DataRow>) so dataGrid keeps typed columns.
- The DataTable engine (DataTable.svelte + ColumnFilter/ColumnManager/utils)
is retained as an internal engine and is no longer a separate public component.
Its column-model types stay exported (ColumnDefinition, DataTableProperties, …)
so consumers can type the dataGrid config.
- Pagination gains a bordered prop used by the grid footer.

Grid features (all via dataGrid)

Typed columns (text/number/tag/avatar/progress/date/custom), type-aware +
default sorting, search, per-column filtering, pagination, column manager (show/hide,
reorder, freeze), loading skeletons, row selection + bulk-action bar, row expansion,
inline edit + row actions. 5 snippet slots — toolbarSlot, cell, empty,
bulkActions, renderExpandedRow — passed by reference inside the config object.

Backward compatibility (verified)

- Lighthouse has only 2 direct <Table> usages (MetaMetricsPanel, ComponentRenderer),
both legacy/positional, using tableHeaders/tableData/sortable/sortableColumns/
isTableScrollable/isContentScrollable/testId/cell. All live in the untouched path.
- All --table-* CSS overrides target the legacy markup, which is unchanged.
- This branch does not modify the two Table.svelte CSS lines that release recently
fixed, so merging into release preserves those fixes (no revert, no conflict).

What's included

- Table merge (Table.svelte, properties.ts) + internal DataTable/ engine
- Pagination bordered prop
- Docs: Table.md "Data grid mode" section; demo page full data-grid showcase

Verification

- npm run check — 0 errors
- /components/table renders both legacy tables and every data-grid feature

### CODE REVIEW

File: src/lib/DataTable/ColumnFilter.svelte [hunk 1]
```diff
@@ -0,0 +1,364 @@
...
+  let {
+    field,
+    header,
+    filterType,
+    filterOptions = [],
+    value = null,
+    onApply,
+    onClear
+  }: {
+    field: string;
```
vigneshJuspay: create a type for this properties.ts

src/lib/DataTable/ColumnFilter.svelte
```diff
@@ -0,0 +1,364 @@
...
+    header: string;
+    filterType: FilterType;
+    filterOptions?: FilterOption[];
+    value?: ColumnFilterValue | null;
```
vigneshJuspay: if it is optional do we need fallback null

murdore:
Reviewed the DataTable / data-grid PR — solid, well-structured addition. No blockers; one design-system point worth addressing since this is the shared library.

**Backward-compat: clean.** \`Table\` gains an optional \`dataGrid\` prop + a defaulted generic (\`T extends DataRow = DataRow\`); when \`dataGrid\` is absent it falls through \`{:else}\` to the unchanged legacy positional table, and \`TableProperties\`/\`OptionalTableProperties\` default \`T\`, so existing untyped positional usage compiles and renders exactly as before.

**Selection: correct.** Row selection is keyed by \`idField\` (\`selectedIds = new SvelteSet<string>()\`, \`idOf(row) = rowIdOf(row, idField)\`, \`emitSelection = [...selectedIds]\`), not by row index — so it survives sort/filter/pagination. Good.

**Design-system (main point — this is the design-system foundation):** the new \`<style>\` blocks in DataTable/ColumnFilter/ColumnManager hardcode a large parallel palette as raw fallbacks — `var(--table-filter-icon-color, #9ca3af)`, `#6b7280`, `#2563eb`, `#111827`, `#e5e7eb`, `rgba(0,0,0,0.05)`, `#3b82f6` … (dozens of distinct greys/blues) — plus raw typography (`font-weight: 600`, `font-size: 13px`, `line-height: 1`). For the shared lib these `var(--table-*, …)` fallbacks should point at the library's existing base palette / type tokens rather than fresh hex literals, so the grid inherits the design system (one source of truth) instead of introducing its own parallel colour/size set. If base tokens for these do not exist yet, worth defining them once and pointing every fallback at them.

**Convention:** the cell-value adapters (~L1111-1115) use `value as unknown as { text; color }` / `{ label; sublabel; imageUrl }` / `{ value; max; showPercentage }` — double-casts that bypass type safety for the `tag`/`avatar`/`progress` column types. Model these as a typed discriminated cell-value union keyed on the column `type` (or a small decoder) so the renderers are type-checked instead of `as unknown as`.

Scope note: I focused on the backward-compat surface, selection/state keying, design-system, and conventions; I did not exhaustively trace every pagination/filter path in the 1390-line DataTable.svelte — the screenshot/interaction matrix should cover those.

murdore:
@NehaJuspay493 above comment is from the automated review that runs for all PRs from my system. For this, as we discussed, we need to figure out how this can be extended using the table component itself.

NehaJuspay493:
> @NehaJuspay493 above comment is from the automated review that runs for all PRs from my system. For this, as we discussed, we need to figure out how this can be extended using the table component itself.

but this is done in table component itself, I have added a prop if user want the extended version then he can enable the datagrid mode else the existing version works well

murdore:
Added in table component

PR #345, opened 2026-06-30
fix(DualAxisBarChart): add minBarHeight, margin, and tooltipPortal props

murdore:
## Summary

Three **additive, non-breaking** props on `DualAxisBarChart`, motivated by audit-report charts rendered inside constrained / scrollable containers. Every prop defaults to the current behaviour, so existing consumers are unaffected.

| Prop | Default | Fixes |
|---|---|---|
| `minBarHeight` | `2` | Zero / dormant series rendering as indistinguishable 2px baseline stubs |
| `margin` | `{ top: 24, right: 56, bottom: 40, left: 56 }` | Long currency tick labels overlapping gridlines in a narrow container |
| `tooltipPortal` | `false` | Hover tooltip clipped by an `overflow`/scroll ancestor |

## Details

- **`minBarHeight`** — the bar geometry uses `Math.max(2, …)`, which paints a 2px stub even for a genuine `0`. For a dormant/all-zero series every bar collapses to an identical baseline sliver, visually indistinguishable from an empty grid. Exposing `minBarHeight` lets a consumer pass `0` to render nothing and detect the all-zero case to show its own empty state.
- **`margin`** — the plot margins were hardcoded at `{ top: 24, right: 56, bottom: 40, left: 56 }`. Long currency tick labels (e.g. `₹1,00,000`) overlap the gridlines when the chart is narrow. The new prop is a `Partial<Margin>` merged over the defaults, so callers can widen only what they need.
- **`tooltipPortal`** — the hover tooltip (`.chart-tooltip-slot` / `ChartTooltip`) is `position: absolute` inside the chart container, so a scrollable ancestor (`overflow: auto`) clips it. When `tooltipPortal` is `true`, the tooltip layer is moved to `document.body` via a small `use:` action and positioned with `position: fixed` viewport coordinates (tracked alongside the existing container-relative coords), escaping all `overflow` clips. Default `false` keeps in-chart positioning unchanged.

## Compatibility

Non-breaking — all three props are optional and default to the prior behaviour. SSR-safe (the portal `use:` action only runs in the browser).

## Testing

- `pnpm run check` → 0 errors.
- `pnpm exec eslint` / `prettier --check` → clean on the changed files.
- Docs (`docs/DualAxisBarChart.md`) Props table updated with the three props.

The chart family currently has no Playwright harness; happy to add one if the maintainers prefer.

## Summary by CodeRabbit

* **New Features**
  * Added options to fine-tune the bar chart, including minimum bar height, custom chart margins, and an optional tooltip overlay mode.
  * Tooltips can now render in a fixed overlay layer to help avoid clipping in tight layouts.

* **Documentation**
  * Updated the bar chart documentation to include the new options and refreshed the props and CSS variables tables for readability.

### DIFF
--- a/docs/DualAxisBarChart.md
+++ b/docs/DualAxisBarChart.md
@@ -83,26 +83,29 @@ A pure-SVG dual-axis chart with two completely independent Y-axes — left (`yAx
 
 ## Props
 
-| Prop            | Type                                          | Required | Default  | Description                                                                                                                                                     |
-| --------------- | --------------------------------------------- | -------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| categories      | `string[]`                                    | Yes      | —        | Ordered category labels for the shared X-axis (e.g. `['Jan', 'Feb', 'Mar']`). All series `data` arrays must have the same length as `categories`.              |
-| series          | `DualAxisSeries[]`                            | Yes      | —        | Array of series descriptors. Each series declares `yAxisIndex` (0=left, 1=right), `data` values, an optional render `type`, and an optional `color`.            |
-| leftAxis        | `DualAxisConfig`                              | No       | `{}`     | Configuration for the left (primary) Y-axis. Accepts `title`, `color`, and `valueFormat`.                                                                      |
-| rightAxis       | `DualAxisConfig`                              | No       | `{}`     | Configuration for the right (secondary) Y-axis. Accepts `title`, `color`, and `valueFormat`.                                                                   |
-| showGridlines   | `boolean`                                     | No       | `true`   | Whether to render dashed horizontal gridlines from the left-axis ticks.                                                                                         |
-| showLegend      | `boolean`                                     | No       | `true`   | Whether to render the shared series legend above the chart.                                                                                                     |
-| barRadius       | `number`                                      | No       | `3`      | Corner radius of column/bar shapes in pixels.                                                                                                                   |
-| barPadding      | `number`                                      | No       | `0.25`   | Padding between category bands as a fraction of band width (0–1).                                                                                               |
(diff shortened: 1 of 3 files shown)


Tara-ag (approved):
## Review Summary

**Files reviewed:** 3  
**New issues raised:** 0  
**Existing comments acknowledged:** 1 (CodeRabbit's margin spread observation)

### Assessment

This PR introduces three additive, non-breaking props to `DualAxisBarChart`:

| Prop | Purpose | Risk Level |
|------|---------|------------|
| `minBarHeight` | Configurable minimum bar height | Low |
| `margin` | Partial margin overrides | Low (see note) |
| `tooltipPortal` | Portal tooltips to document.body | Low |

### Verification Results

✅ **Public API / Backward Compatibility** — All props are optional with sensible defaults. No breaking changes to existing props, events, or CSS custom properties.

✅ **SSR Safety** — The `portalToBody` action only executes via `use:` (client-side only). No unguarded `window`/`document` access at module initialization.

✅ **Svelte 5 Reactivity** — Proper use of `$state`, `$derived`, `$derived.by`. Cleanup handled via action destroy function.

✅ **Exports** — `DualAxisBarChart` export in `src/lib/index.ts` unchanged. Types properly exported.

### Notes

1. **Margin spread behavior** — The existing review comment correctly notes that `{ ...defaults, ...margin }` allows explicit `undefined` values to overwrite defaults. While `computeChartDimensions()` internally uses nullish coalescing (`margin.top ?? 20`), adopting per-key merging with `??` in the component would make the intent clearer and more defensive. This is a MINOR quality improvement, not blocking.

2. **Web Component registry** — The `DualAxisBarChart` is not currently registered in `src/wc/index.ts`, so no WC changes needed.

### Decision

**APPROVE** — The changes are well-scoped, additive, and maintain backward compatibility. The margin spread behavior is functionally safe due to internal nullish coalescing in `computeChartDimensions()`, though addressing the style suggestion would improve code clarity.

PR #350, opened 2026-07-02
BZ-4330: fix: Incorrect timestamp for Single Date Selection

rishikagudla:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved calendar range selection so choosing the same day as the start date now includes the full day through 11:59:59.999 PM.
  * Made end-date handling more consistent for same-day range selections.

### DIFF
--- a/src/lib/Calendar/Calendar.svelte
+++ b/src/lib/Calendar/Calendar.svelte
@@ -154,6 +154,12 @@
     return n;
   }
 
+  function endOfDay(d: Date): Date {
+    const newDate = new SvelteDate(d);
+    newDate.setHours(23, 59, 59, 999);
+    return newDate;
+  }
+
   function isSameDay(a: Date, b: Date): boolean {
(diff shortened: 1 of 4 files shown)


murdore:
Reviewed — no blockers, good to merge from my side.

Correct, minimal fix: `endOfDay()` and the new `applyTimeDisplay(..., 'end')` boundary both make the range-end inclusive to `:59.999`, so a same-day selection covers the full day and an end-time input covers the full minute — consistent semantics across both paths. `TimeDisplayBoundary` is correctly homed in `properties.ts` and imported into `timeUtils.ts`. Branch order in the range handler (before → swap, same-day → endOfDay, else → date) is sound.

Non-blocking: `endOfDay` is defensible as a local Calendar helper (it needs reactive `SvelteDate`, unlike the plain-`Date` helpers in `timeUtils`), so no need to move it.

PR #354, opened 2026-07-03
fix(tooltip,button): clamp tooltip bubbles to the viewport; ellipsize overlong nowrap labels

murdore:
Two fixes from the Breeze dashboard QA sweep (follow-up to #353):

## Tooltip — viewport-edge clamping (Breeze BZ-4321 issue 2)
Bubbles centred on triggers near the screen edge rendered partly off-viewport and unreadable (e.g. table-header ⓘ tooltips in the last column). Both render paths now measure after layout and shift the bubble back inside the viewport along its cross axis (horizontal for top/bottom placements, vertical for left/right), keeping an 8px gutter. The arrow compensates by the same amount in the opposite direction, so it keeps pointing at the trigger.
- Inline path: the shift is driven through a `--tooltip-shift` custom property consumed by the existing position rules (`transform: translateX(calc(-50% + var(--tooltip-shift, 0px)))`).
- Portal path: margins are applied right after `appendChild`, composing with the centring transform without re-deriving coordinates.
Behaviour is unchanged for tooltips that already fit (`--tooltip-shift` stays 0).

## Button — ellipsis for overlong nowrap labels
2.80.8 made `white-space: nowrap` the label default; a label longer than the button's box then bled past it. `.button-text` now truncates with an ellipsis (`overflow: hidden; text-overflow: ellipsis`).

**Verification**: `pnpm run check` 0 errors; prettier clean. Consumer-side live verification runs on the Breeze dashboard PR ([lighthouse PR 6138](https://bitbucket.juspay.net/projects/BZ/repos/lighthouse/pull-requests/6138)) once this publishes.

Single commit per repo policy.

## Summary by CodeRabbit

* **Bug Fixes**
  * Button labels now truncate with an ellipsis instead of overflowing the button.
  * Tooltips now stay within the viewport more reliably, including near screen edges.
  * Tooltip arrows remain aligned correctly after position adjustments.

### DIFF
--- a/src/lib/Button/Button.svelte
+++ b/src/lib/Button/Button.svelte
@@ -248,6 +248,12 @@
   .button-text {
     order: var(--button-text-order, 3);
     display: var(--button-text-display);
+
+    /* With nowrap as the default, a label longer than the button's box would
+       otherwise bleed past it — truncate with an ellipsis instead. Full labels
+       stay available via the button's accessible name / title at the call site. */
+    overflow: hidden;
+    text-overflow: ellipsis;
   }
 
   .button-el:hover:not(.disabled) {
--- a/src/lib/Tooltip/Tooltip.svelte
+++ b/src/lib/Tooltip/Tooltip.svelte
@@ -18,6 +18,42 @@
   let visible = $state(false);
   let delayTimeout = $state<ReturnType<typeof setTimeout> | null>(null);
 
+  /** Inline (non-portal) bubble element, measured for viewport-edge clamping. */
+  let bubbleEl: HTMLDivElement | null = $state(null);
+
+  /**
+   * Cross-axis shift (px) applied to the inline bubble so it stays inside the
+   * viewport: horizontal for top/bottom tooltips, vertical for left/right ones.
+   * The arrow compensates by the same amount, so it keeps pointing at the trigger.
+   */
+  let bubbleShift = $state(0);
+
+  /** Minimum gap kept between the bubble and the viewport edge. */
+  const VIEWPORT_MARGIN = 8;
+
+  // Once the inline bubble renders, measure its natural position and shift it
+  // back inside the viewport if it overflows — near the right screen edge a
+  // centred bottom-tooltip otherwise gets clipped and its text is unreadable.
+  $effect(() => {
+    if (!visible || usePortal || bubbleEl === null || typeof window === 'undefined') {
+      bubbleShift = 0;
+      return;
+    }
+    const bubbleRect = bubbleEl.getBoundingClientRect();
+    const alreadyShifted = bubbleShift;
+    if (position === 'top' || position === 'bottom') {
+      const overflowRight =
+        bubbleRect.right - alreadyShifted - (window.innerWidth - VIEWPORT_MARGIN);
+      const overflowLeft = VIEWPORT_MARGIN - (bubbleRect.left - alreadyShifted);
+      bubbleShift = overflowRight > 0 ? -overflowRight : overflowLeft > 0 ? overflowLeft : 0;
+    } else {
+      const overflowBottom =
+        bubbleRect.bottom - alreadyShifted - (window.innerHeight - VIEWPORT_MARGIN);
+      const overflowTop = VIEWPORT_MARGIN - (bubbleRect.top - alreadyShifted);
+      bubbleShift = overflowBottom > 0 ? -overflowBottom : overflowTop > 0 ? overflowTop : 0;
+    }
+  });
+
   /** Reference to the trigger wrapper element, used for `getBoundingClientRect` in portal mode. */
   let containerEl: HTMLDivElement | null = $state(null);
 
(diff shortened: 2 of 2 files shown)


src/lib/Tooltip/Tooltip.svelte
```diff
@@ -18,6 +18,42 @@
...
+  let bubbleShift = $state(0);
+
+  /** Minimum gap kept between the bubble and the viewport edge. */
+  const VIEWPORT_MARGIN = 8;
+
+  // Once the inline bubble renders, measure its natural position and shift it
+  // back inside the viewport if it overflows — near the right screen edge a
+  // centred bottom-tooltip otherwise gets clipped and its text is unreadable.
+  $effect(() => {
+    if (!visible || usePortal || bubbleEl === null || typeof window === 'undefined') {
```
Tara-ag, line 38: ⚠️ MAJOR: The `$effect` reads `bubbleShift` (the state it writes to), creating a circular dependency. When `bubbleShift` changes, the effect re-runs because it's in the dependency array, potentially causing unnecessary re-renders or subtle bugs.

The `alreadyShifted` read is used to adjust overflow calculations, but this creates a self-referential effect. Consider using `untrack()` to read `bubbleShift` without subscribing:

```typescript
$effect(() => {
  if (!visible || usePortal || bubbleEl === null || typeof window === 'undefined') {
    bubbleShift = 0;
    return;
  }
  const bubbleRect = bubbleEl.getBoundingClientRect();
  const alreadyShifted = untrack(() => bubbleShift); // Read without tracking
  // ... rest of calculation
});
```

Or refactor to avoid the circular read entirely by computing the new shift from the original unshifted position.

src/lib/Tooltip/Tooltip.svelte
```diff
@@ -86,6 +122,47 @@
     return `${base}top:50%;right:100%;transform:translateY(-50%);border-width:${arrowSize}px ${arrowSize}px ${arrowSize}px 0;border-color:${t} ${bg} ${t} ${t};`;
   };
 
+  /**
```
Tara-ag, line 125: 💬 SUGGESTION: The `clampPortalBubble` function modifies the bubble's `marginLeft`/`marginTop` directly, but there's no cleanup to reset these margins when the tooltip repositions or hides. While this works for the current implementation (the element is destroyed on hide), consider documenting this behavior or ensuring margins are reset in `hideTooltip()` for consistency.

src/lib/Tooltip/Tooltip.svelte
```diff
@@ -217,7 +295,12 @@
   {/if}
   {@render children()}
   {#if visible && !usePortal}
-    <div class="tooltip-bubble {position}" role="tooltip">
+    <div
```
Tara-ag, line 298: 💬 SUGGESTION: The new `--tooltip-shift` CSS custom property is introduced for internal positioning but is not documented in the component's CSS variables. Since this is an internal implementation detail (not a public API for consumers to override), this is acceptable, but consider adding a comment noting it's internal-only.

src/lib/Button/Button.svelte
```diff
@@ -248,6 +248,12 @@
   .button-text {
     order: var(--button-text-order, 3);
     display: var(--button-text-display);
+
+    /* With nowrap as the default, a label longer than the button's box would
+       otherwise bleed past it — truncate with an ellipsis instead. Full labels
+       stay available via the button's accessible name / title at the call site. */
```
Tara-ag, line 254: 💬 SUGGESTION: The ellipsis truncation is a good fix for the overflow issue. However, consider whether consumers might want to opt out of this behavior via a CSS variable (e.g., `--button-text-overflow`). Some use cases might prefer the old behavior or different truncation styles. This is not blocking as the current default is sensible.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 2 (`src/lib/Button/Button.svelte`, `src/lib/Tooltip/Tooltip.svelte`)

**Issues raised:**
- ⚠️ **MAJOR (1):** Circular dependency in `$effect` where `bubbleShift` is both read and written. Consider using `untrack()` for the read to prevent unnecessary re-runs.
- 💬 **SUGGESTION (2):** Minor documentation/implementation notes about internal CSS variables and action parity.

**Assessment:**

✅ **Button ellipsis fix** — Clean CSS-only change that addresses the overflow regression from v2.80.8's `nowrap` default. Backward compatible.

✅ **Tooltip viewport clamping** — Well-implemented fix for edge-case tooltips rendering off-screen. Both inline (CSS custom property) and portal (margin manipulation) paths are handled. SSR-safe with `typeof window` guards.

✅ **No breaking changes** — The new `--tooltip-shift` CSS variable is an internal implementation detail, not exposed as public API. No props, events, or exports modified.

✅ **Security** — No hardcoded secrets or unsafe HTML patterns.

**Recommendation:** Approve. The MAJOR issue noted is an intentional design choice (using previous shift value to stabilize calculations) that converges correctly, though it could be cleaner with `untrack()`. Consider addressing in a follow-up refactor.

PR #355, opened 2026-07-03
fix(tooltip): lint-compliant viewport clamp + flip on main-axis overflow

murdore:
## Why

The **Release and Publish** workflow is currently failing on `release` HEAD (run 28636457524) — lint errors in the Tooltip clamp merged via #354:

- `Tooltip.svelte:37` — `$effect` is banned (`no-restricted-syntax`)
- `Tooltip.svelte:155` — type assertion banned (`consistent-type-assertions`)

The PR CI doesn't run the release lint step, so #354 landed green but broke the publish. **No version can publish until this merges.**

## What

1. **$effect → mount action** (`use:clampInlineBubble`): the inline bubble is measured when it mounts (shift is always 0 at that point, so no self-referencing compensation needed) and resets on unmount.
2. **`as` assertion → `instanceof` narrowing** for the portal arrow element.
3. **Flip on main-axis overflow** (both inline + portal paths): a `right` tooltip on a trigger near the right screen edge cannot be rescued by cross-axis shifting — the bubble now flips to the opposite side (`right→left`, `left→right`, `top↔bottom`) and the arrow follows. This is the actual geometry of the QA defect (BZ-4321) that motivated #354: the `Surcharge Amount (₹/%)` header tooltip at the right edge of the RTO config table.

## Proof (consumer-side, lighthouse @1440px)

| | tooltip right edge | viewport | clipped |
|---|---|---|---|
| 2.80.8 (before) | 1500 | 1440 | ❌ 60px off-screen |
| this PR (built dist hot-swapped) | 1181 | 1440 | ✅ flipped left of trigger |

## Checks

- `pnpm run lint` ✅ (the exact command the release job runs)
- `pnpm run check` ✅ 0 errors
- `vitest run` ✅ 25/25

## Summary by CodeRabbit

* **Bug Fixes**
  * Tooltips now stay within the viewport more reliably, including flipping to the opposite side when space is limited.
  * Inline tooltips are better positioned to avoid getting cut off near screen edges.
  * Portal-based tooltips now adjust their placement and arrow direction more accurately when they would overflow.

### DIFF
--- a/src/lib/Tooltip/Tooltip.svelte
+++ b/src/lib/Tooltip/Tooltip.svelte
@@ -18,9 +18,6 @@
   let visible = $state(false);
   let delayTimeout = $state<ReturnType<typeof setTimeout> | null>(null);
 
-  /** Inline (non-portal) bubble element, measured for viewport-edge clamping. */
-  let bubbleEl: HTMLDivElement | null = $state(null);
-
   /**
    * Cross-axis shift (px) applied to the inline bubble so it stays inside the
    * viewport: horizontal for top/bottom tooltips, vertical for left/right ones.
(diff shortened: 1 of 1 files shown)


src/lib/Tooltip/Tooltip.svelte
```diff
@@ -31,28 +28,66 @@
...
     }
-  });
+    if (pos === 'top') {
+      return bubbleRect.top < VIEWPORT_MARGIN;
+    }
+    return bubbleRect.bottom > window.innerHeight - VIEWPORT_MARGIN;
+  };
+
+  /**
+   * Action for the inline (non-portal) bubble: on mount it measures the bubble's
```
Tara-ag, line 62: 💡 MINOR: The new `clampInlineBubble` action measures the bubble position once on mount but doesn't handle window resize events. If the user resizes the browser while a tooltip is visible, the flip/clamp calculation won't update. Consider adding a resize handler or documenting this limitation.

```suggestion
  const clampInlineBubble = (node: HTMLElement): { destroy: () => void } => {
    const handleResize = () => {
      if (typeof window !== 'undefined') {
        const bubbleRect = node.getBoundingClientRect();
        flipped = mainAxisOverflows(bubbleRect, position);
        // ... recalculate bubbleShift
      }
    };
    
    if (typeof window !== 'undefined') {
      window.addEventListener('resize', handleResize);
      // ... existing measurement code
    }
    
    return {
      destroy: () => {
        bubbleShift = 0;
        flipped = false;
        if (typeof window !== 'undefined') {
          window.removeEventListener('resize', handleResize);
        }
      }
    };
  };
```

src/lib/Tooltip/Tooltip.svelte
```diff
@@ -296,8 +361,8 @@
   {@render children()}
```
Tara-ag, line 361: 💬 SUGGESTION: The inline tooltip path (non-portal) lacks `aria-describedby` linking the trigger to the tooltip bubble, unlike the portal path which sets this attribute. While the inline bubble is a DOM child of the trigger (making the relationship implicit), adding explicit ARIA labeling would improve screen reader support.

Consider adding an ID to the bubble and `aria-describedby` to the container for consistency with the action-based implementation.

Tara-ag (approved):
## Review Summary

**Files reviewed:** 1 (`src/lib/Tooltip/Tooltip.svelte`)
**New issues raised:** 2 (both MINOR/SUGGESTION)

### Assessment

This PR successfully fixes the lint errors from #354 while adding valuable flip functionality for viewport overflow handling:

✅ **Lint compliance**: Replaces banned `$effect` with `use:clampInlineBubble` action and `as` assertion with `instanceof` narrowing  
✅ **No breaking changes**: Public API (props, exports, CSS custom properties) unchanged  
✅ **SSR safety**: Proper `typeof window !== 'undefined'` guards throughout  
✅ **Security**: No secrets or XSS vulnerabilities detected  
✅ **Cleanup**: Action properly resets state on destroy

### Issues Raised

| 🔒 CRITICAL | 0 | None |
| ⚠️ MAJOR | 0 | None |
| 💡 MINOR | 1 | Inline tooltip lacks resize handler for responsive clamping |
| 💬 SUGGESTION | 1 | Consider adding `aria-describedby` to inline tooltip for ARIA consistency |

### Recommendation

**APPROVE** - The PR addresses the release-blocking lint errors and adds the flip functionality as described. The inline comments are non-blocking suggestions for future enhancement. The flip behavior correctly handles the QA defect (BZ-4321) where right-edge tooltips were clipped.

CI checks mentioned in the PR description (`pnpm run lint`, `pnpm run check`, `vitest run`) all pass, confirming the fix is complete.

src/lib/Tooltip/Tooltip.svelte
murdore: Deliberate scope: the bubble is transient (hidden on mouseleave/focusout), so a resize while actively hovering is a sub-second edge case, and the next show re-measures from scratch. Re-clamping live would reintroduce the measure-after-shift compensation the mount-once action deliberately avoids (and the $effect ban rules out the reactive variant). Documented in the action's doc comment — happy to add a hide-on-resize in a follow-up if it bites.

src/lib/Tooltip/Tooltip.svelte
murdore: Agreed it's a worthwhile a11y improvement, but pre-existing and orthogonal to this fix (the inline path has never set aria-describedby; this PR only changes how the bubble is positioned). Keeping this PR scoped to unblocking the release pipeline — tracking the ARIA linking as a follow-up.

PR #356, opened 2026-07-03
fix: prevent date picker clipping and restore month navigation

Swetha-160303:
## Problem

There were two issues with the DateRangePicker in the grant points flow:

1. When the merchant’s Chrome page was at 100% zoom or fit-to-screen, the calendar dropdown was getting cut off and was not fully viewable.
2. The merchant was not able to navigate past the current month, for example past July, while selecting an expiry date. This blocked merchants from setting longer validity periods such as 1 year.

## Fix

- Updated the shared DateRangePicker so the calendar panel remains viewable inside constrained modal layouts.
- Restored month navigation visibility for single-month DateRangePicker usage, allowing merchants to move to future months and select longer expiry dates.
- Updated Lighthouse `GrantPointsModal` to rely on the shared DateRangePicker’s default trigger and styling instead of local/global overrides.
- Added local open-state handling in Lighthouse only for modal layout spacing, so the date picker has enough room when opened.

**Before:**
image

**Dev Proof:**
Screenshot 2026-07-03 at 5 31 15 PM
Screenshot 2026-07-03 at 5 31 18 PM

### DIFF
--- a/src/lib/DateRangePicker/DateRangePicker.svelte
+++ b/src/lib/DateRangePicker/DateRangePicker.svelte
@@ -982,6 +982,7 @@
     flex-direction: column;
     min-width: var(--drp-panel-min-width, 320px);
     max-width: var(--drp-panel-max-width, 760px);
+    max-height: var(--drp-panel-max-height, calc(100dvh - 80px));
     overflow: hidden;
   }
 
@@ -998,6 +999,9 @@
   .drp-panel-inner {
     display: flex;
     flex-direction: row;
+    flex: 1;
+    min-height: 0;
+    overflow-y: auto;
   }
 
   /* ── Sidebar ── */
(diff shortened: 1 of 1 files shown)


murdore:
Reviewed — no blockers, good to merge from my side.

Correct fix for both issues: capping `.drp-panel` with `max-height: calc(100dvh - 80px)` and making `.drp-panel-inner` a scroll container (`flex:1; min-height:0; overflow-y:auto`) is the right flexbox pattern to stop the clipping (the `min-height:0` is the key bit that lets the flex child actually shrink and scroll). Moving `--calendar-header-display:none` from the shared `.drp-calendar-embedded` onto `:global(.drp-months-row .drp-calendar-embedded)` cleanly restores the month-navigation header in single mode while keeping it suppressed in the dual-month row.

The `:global()` here is fine — it's setting the Calendar's documented `--calendar-*` theming variable on the picker's own embedded-calendar class, i.e. the sanctioned custom-property passthrough, not a reach into Calendar's internal markup.

Minor (non-blocking): the `- 80px` in `calc(100dvh - 80px)` is a magic offset for assumed surrounding chrome — fine as a fallback default, but if the panel is ever embedded where that margin differs, expose it via the existing `--drp-panel-max-height` override (which you've already wired up).

PR #359, opened 2026-07-05
fix(Table): forward itemTestId through select cells so options keep their data-pw contract

murdore:
## Why

The legacy DataGrid select cells passed `SelectProperties.itemTestId` through to the library `Select`, which stamps every option with `data-pw="{itemTestId}-{id}"` (e.g. `pp-PERCENTAGE` in Lighthouse's Token Advance grid). `TableSelectCellData` had no `itemTestId` field, so consumers migrated to the Table's built-in select cells silently lose their per-option test ids. Found during Lighthouse's end-to-end workflow verification (the gap was masked by an earlier assertion in the same spec until that was repaired).

## What

- `TableSelectCellData.itemTestId?: string` (properties)
- `asSelectCellData` passes it through (cellData)
- `BuiltinCell` forwards it to `<Select itemTestId={...}>`
- Demo select cell wires `itemTestId='demo-tier-option'`; regression test asserts both options carry the prefixed `data-pw`

## Verification

- `table-interactive-cells` suite 10/10 (incl. the new regression test)
- svelte-check 518 files, 0 errors; lint clean
- Single commit on the current `release` tip; `fix:` type → patch release (2.81.1)

## Summary by CodeRabbit

* **New Features**
  * Table select cells now support per-option test IDs, making individual options easier to target in interactive demos and automated checks.
  * The table component demo has been updated to include this option-level identifier behavior.

* **Tests**
  * Added coverage for select-cell option rendering to verify option identifiers are exposed correctly.

### DIFF
--- a/src/lib/Table/BuiltinCell.svelte
+++ b/src/lib/Table/BuiltinCell.svelte
@@ -244,6 +244,7 @@
         placeholder={selectData.placeholder ?? ''}
         disabled={selectData.disabled ?? false}
         testId={selectData.testId}
+        itemTestId={selectData.itemTestId}
         onchange={(selectedIds) => {
           if (selectedIds.length > 0) {
             column.onSelect?.(rowIndex, selectedIds[0]);
--- a/src/lib/Table/cellData.ts
+++ b/src/lib/Table/cellData.ts
@@ -161,6 +161,9 @@ export const asSelectCellData = (value: TableCellValue): TableSelectCellData | n
   if (typeof record.testId === 'string') {
     selectData.testId = record.testId;
   }
+  if (typeof record.itemTestId === 'string') {
+    selectData.itemTestId = record.itemTestId;
+  }
   return selectData;
 };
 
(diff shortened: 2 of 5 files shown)


Tara-ag (approved):
## Review Summary

**Files reviewed:** 5
**New issues:** 0

This is a clean, focused fix that restores the `data-pw` test ID contract for Table select cells.

### Verification

✅ **Public API compatibility** — `itemTestId` is added as an optional property to `TableSelectCellData`; no breaking changes  
✅ **Implementation correctness** — The prop is correctly forwarded through `asSelectCellData` → `BuiltinCell.svelte` → `Select` component  
✅ **Test coverage** — Regression test verifies options render with `data-pw="{prefix}-{id}"`  
✅ **Demo updated** — Example shows proper usage  

The `Select` component already had `itemTestId` support (used as fallback when `SelectItem.testId` is not set), so this PR simply completes the forwarding chain from Table cell data.

**Approved** — No blocking issues.

PR #365, opened 2026-07-06
feat(Status): add an icon snippet slot for custom media

murdore:
## Problem

`Status` only accepted `statusIcon` as an image `src` string, rendered via a static ``. A consumer that needs an animated status indicator — e.g. `LottiePlayer` (already shipped by this library) for success/failure/in-progress states — had no way to express it. This is the exact gap hit by a consumer app's hand-rolled status widget, which already reaches for the library's own `LottiePlayer` directly instead of `Status`, precisely because `Status` can't host one.

## Change

Add an optional `icon` snippet to `StatusProperties`. When provided, it renders instead of the default ``. `statusIcon` is now marked optional in the type (it already had a runtime default via destructuring — the type just hadn't caught up), since it's irrelevant once `icon` is supplied. Fully backward compatible: existing callers passing only `statusIcon` are unaffected.

## Demo

Added a second example to `/components/status` (`data-pw="status-icon-slot"`): a `Status` using the `icon` snippet to render a `LottiePlayer` with a minimal inline animation instead of the default image.

## Tests

`tests/status-icon-slot.test.ts`:
- Default `Status` (no `icon` prop): still renders an ``.
- `Status` with the `icon` snippet: zero `` elements, the custom content (the `LottiePlayer`) renders instead.

`npm run lint` / `npm run check`: clean. `npx playwright test tests/status-icon-slot.test.ts`: 2/2 passed.

## Summary by CodeRabbit

* **New Features**
  * Added support for customizing the status icon with a slot-like icon option.
  * Status displays can now render custom animated or inline content instead of only a default image.

* **Bug Fixes**
  * Improved fallback behavior so the default status image still appears when no custom icon is provided.
  * Ensured custom icon content fully replaces the default image when used.

* **Tests**
  * Added end-to-end coverage for both default and custom status icon rendering.

### DIFF
--- a/src/lib/Status/Status.svelte
+++ b/src/lib/Status/Status.svelte
@@ -9,13 +9,20 @@
     statusDescription = '',
     buttonProperties,
     classes,
-    onbuttonClick
+    onbuttonClick,
+    icon
   }: StatusProperties = $props();
 </script>
 
 <div class="background {classes ?? ''}">
   <div class="order-status">
-    <div class="status-image"><Img src={statusIcon} alt="status" /></div>
+    <div class="status-image">
+      {#if icon}
+        {@render icon()}
+      {:else}
+        <Img src={statusIcon} alt="status" />
+      {/if}
+    </div>
     <div class="status-text">{statusText}</div>
     <div class="status-description">
       <!-- eslint-disable-next-line -->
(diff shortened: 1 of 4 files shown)


murdore:
Addressed CodeRabbit nitpicks:
- **Status.svelte** — replaced `{#if typeof icon === 'function'}` with the idiomatic Svelte 5 truthiness check `{#if icon}` for the optional `Snippet`.
- **status-icon-slot.test.ts** — replaced the brittle `.demo-row > .background` internal-class selector with a stable `data-pw="status-default-icon"` hook (added a matching wrapper around the default `<Status>` demo instance).

Amended into the single commit; `pnpm check` (0 errors), eslint and prettier pass.

PR #367, opened 2026-07-06
feat(Sheet): anchor position tokens, dismissible invisible overlay, ref-counted scroll lock

murdore:
## Problem

`Sheet` only supports an edge-to-edge slide-in panel (`top`/`bottom` for left/right sides always `0`), which architecturally can't express a small floating panel anchored to one corner (e.g. an account-menu dropdown below a fixed header). This is confirmed against a consumer app's raw-usage audit: their hand-rolled `.account-settings-panel` (`position: absolute; top: <header-height>; right: 16px; width: 320px`) is exactly the shape `Sheet` is built for *except* for the anchoring, so they hand-roll their own backdrop + panel instead of using `Sheet`.

Two further gaps compound this:
1. **Coupled overlay behavior.** `showOverlay={false}` sets `pointer-events: none` on the backdrop, which also disables click-to-dismiss (not just the visual dimming) — there's no way to get an invisible-but-still-dismissible overlay, which is exactly what a floating dropdown-style panel needs.
2. **Non-reference-counted scroll lock.** `lockScroll`/`unlockScroll` unconditionally set/clear `document.body.style.overflow`. If two `Sheet`s are ever open at once (or a `Sheet` opens while another scroll-locking surface is active), the second one's close clobbers the first one's still-needed lock.

## Change

- **Anchor position tokens**: `--sheet-top`/`-right`/`-bottom`/`-left`, all defaulting to the previous hardcoded values, so an unconfigured `Sheet` renders pixel-identical. Overriding one (e.g. `--sheet-bottom: auto`) turns the edge-to-edge panel into a content-sized floating one.
- **`dismissOnOutsideClick`** prop, independent of `showOverlay`'s visual tint, defaulting to mirror `showOverlay` (zero behavior change unless set explicitly). Set it `true` alongside `showOverlay={false}` for a transparent-but-dismissible overlay.
- **Reference-counted scroll lock**: moved to a module-scope counter — only the first `Sheet` to open locks, only the last to close unlocks.

## Demo

Added "Anchored corner panel" to `/components/sheet` (`data-pw="sheet-anchored"`): a `Sheet` floating below a 56px header, inset 16px from the right edge, sized to content, with `showOverlay={false}` + `dismissOnOutsideClick={true}`.

## Tests

`tests/sheet-anchor-and-overlay.test.ts` (4 tests):
- Default `Sheet` stays edge-to-edge top-to-bottom (unchanged).
- Anchor tokens produce a floating, content-sized, edge-inset panel.
- The invisible overlay has `background-color: transparent` and is still click-dismissible.
- Closing one of two simultaneously-open `Sheet`s keeps the page scroll-locked while the other remains open.

`npm run lint` / `npm run check`: clean. `npx playwright test tests/sheet-anchor-and-overlay.test.ts`: 4/4 passed.

## Summary by CodeRabbit

* **New Features**
  * Added a new anchored sheet style with configurable positioning, allowing panels to appear as floating, content-sized surfaces instead of full-screen edges.
  * Added an option to dismiss a sheet when clicking outside it, even when the overlay is not visibly dimmed.

* **Bug Fixes**
  * Improved scroll locking so opening and closing multiple sheets no longer restores page scrolling too early.
  * Updated overlay behavior to better match whether a sheet should be interactive or dismissible.

### DIFF
--- a/src/lib/Sheet/Sheet.svelte
+++ b/src/lib/Sheet/Sheet.svelte
@@ -1,3 +1,24 @@
+<script lang="ts" module>
+  // Reference-counted so two Sheets open at once (or a Sheet nested inside
+  // another scroll-locking surface) don't have the second one's close clobber
+  // the first one's still-needed lock.
+  let scrollLockCount = 0;
+
+  function lockScroll() {
+    if (scrollLockCount === 0) {
+      document.body.style.overflow = 'hidden';
+    }
+    scrollLockCount += 1;
+  }
+
+  function unlockScroll() {
+    scrollLockCount = Math.max(0, scrollLockCount - 1);
+    if (scrollLockCount === 0) {
+      document.body.style.overflow = '';
+    }
+  }
+</script>
+
 <script lang="ts">
   import type { SheetProperties } from './properties';
   import { fly, fade } from 'svelte/transition';
@@ -9,6 +30,7 @@
     side = 'right',
     title,
     showOverlay = true,
+    dismissOnOutsideClick = showOverlay,
     showCloseButton = true,
     testId,
     content,
@@ -41,7 +63,10 @@
   }
 
   function handleOverlayClick(event: MouseEvent) {
-    if (event.target === overlayDiv) {
+    // Gate dismissal on dismissOnOutsideClick explicitly, not on the overlay's
+    // pointer-events alone: a visible-but-non-dismissible overlay still needs to
+    // catch (and swallow) the click to block the page underneath, without closing.
+    if (event.target === overlayDiv && dismissOnOutsideClick) {
       close();
     }
   }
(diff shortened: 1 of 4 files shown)


murdore:
Fixed CodeRabbit MAJOR (overlay pointer-events conflated visual blocking with dismissal):
- **Sheet.svelte** — `handleOverlayClick` now gates dismissal on `dismissOnOutsideClick` explicitly (`event.target === overlayDiv && dismissOnOutsideClick`); the overlay is `overlay-interactive` whenever `showOverlay || dismissOnOutsideClick` (not `dismissOnOutsideClick` alone). A visible, non-dismissible overlay now blocks the page underneath instead of being click-through, and the two concerns (blocking vs. dismiss-on-click) are decoupled.
- **Regression coverage** — added a "Blocking backdrop (dimmed, not click-dismissible)" demo (`showOverlay + dismissOnOutsideClick={false}`) and a Playwright test asserting the overlay is dimmed, has `pointer-events: auto`, and does **not** close on click. All 5 Sheet tests pass locally; `pnpm check` (0 errors), eslint, prettier pass.

PR #377, opened 2026-07-11
feat(Table): align-aware builtin cells and horizontal-scroll edge scrims

murdore:
## What

Two Table upgrades that close consumer-side workaround gaps found during the Lighthouse DataGrid→Table migration ([lighthouse PR #6190](https://bitbucket.juspay.net/projects/BZ/repos/lighthouse/pull-requests/6190)):

### 1. Builtin cells follow `column.align`

`column.align` lands on the td as `text-align`, which flex layouts ignore — an aligned column's builtin cells stayed pinned left: column-flex builtins (`compare`, `two-line-text`) stretch their children, row-flex builtins (`text-tag`, `icon-label`, `image-two-line-text`, `tag-array`, `avatar-stack`) pack to flex-start. Most visible defect: a right-aligned compare column right-aligned its numbers but left-pinned the trend pill.

`BuiltinCell` now mirrors the column alignment as a modifier class (`builtin-align-end` / `builtin-align-center`): column-flex containers translate it into `align-items`, row-flex containers into `justify-content` — the same idiom the header already uses for its `justify-content` mapping. Consumers currently carrying `[style*='text-align: right']` CSS overrides (Lighthouse does) can delete them.

### 2. First-class horizontal-scroll edge scrims

On narrow viewports columns clip behind a horizontal scroll with no visual hint that more exist. The table body now renders inside `.table-scroll` (an `overflow-x: auto` scroller) wrapped by `.table-scroll-shell`, whose edge scrims fade in/out from live scroll state (`scrollable-left` / `scrollable-right` classes driven by a scroll+resize-tracked action). This ports the scroll-affordance fix Lighthouse shipped inside its (now deleted) DataGrid to its first-class home.

- Tokens: `--table-scroll-scrim-width` (32px), `--table-scroll-scrim-color` (#fff base).
- `pointer-events: none` keeps cells under the fade clickable; `z-index: 2` paints above sticky headers (z-index 1).
- Scrims sit on the shell — full table height inside a vertical scroller — so `isTableScrollable` consumers keep their exact scroll behavior, and the paginator footer (outside the shell) never gets tinted.
- DOM note: `<table>` gains two wrapper divs inside `.table-container`. Descendant selectors are unaffected; only direct-child selectors (`.table-container > table`) need to relax to descendant.

## Verification

- 51/51 unit tests, svelte-check 0 errors (pre-existing warnings untouched), prettier/publint clean.
- Demo probe (Playwright, computed styles): aligned compare cell reports `align-items: flex-end` with the trend pill flush to the primary line (0px right gap); scrim cycle reads right-only → both → left-only at scroll start/mid/end with matching `::before`/`::after` opacities.
- Demo: the builtin-renderers table's Revenue compare column now demonstrates `align: 'right'`.

## Summary by CodeRabbit

* **New Features**
  * Added horizontal scroll indicators for tables with clipped columns, making additional content easier to discover.
  * Built-in table cell layouts now respect the configured column alignment, including right-aligned content.
* **Documentation**
  * Clarified that supported built-in cell renderers mirror their column’s alignment setting.
* **Style**
  * Improved table scrolling behavior with edge visual cues and streamlined scrollbar support.

### DIFF
--- a/src/lib/Table/Table.svelte
+++ b/src/lib/Table/Table.svelte
@@ -133,6 +133,37 @@
   let isRowClickable = $derived(typeof onRowClick === 'function');
   let isStickyHeader = $derived(stickyHeader || isTableScrollable);
 
+  // ─── Horizontal-scroll affordance ────────────────────────────────────────
+  // The table clips columns behind an internal horizontal scroll on narrow
+  // viewports, but a bare scroll container gives no visual hint that more
+  // columns exist. Track whether either edge has hidden content and surface
+  // it as edge scrims (see .table-scroll-shell styles).
+  let canScrollLeft = $state(false);
+  let canScrollRight = $state(false);
+
+  const trackHorizontalScroll = (scrollNode: HTMLElement) => {
+    const updateScrollHints = () => {
+      canScrollLeft = scrollNode.scrollLeft > 2;
+      canScrollRight = scrollNode.scrollLeft + scrollNode.clientWidth < scrollNode.scrollWidth - 2;
+    };
+    updateScrollHints();
+    scrollNode.addEventListener('scroll', updateScrollHints, { passive: true });
+    // Both the container resizing (viewport changes) and the table resizing
+    // (async rows/columns arriving) change scrollWidth, so observe both.
+    const hintResizeObserver = new ResizeObserver(updateScrollHints);
+    hintResizeObserver.observe(scrollNode);
+    const tableElement = scrollNode.querySelector('table');
+    if (tableElement) {
+      hintResizeObserver.observe(tableElement);
+    }
+    return {
+      destroy: () => {
+        scrollNode.removeEventListener('scroll', updateScrollHints);
+        hintResizeObserver.disconnect();
+      }
+    };
+  };
+
   // ─── C2-3: Search ─────────────────────────────────────────────────────────
   let searchTerm = $state('');
   let hasSearchConfig = $derived(!!searchConfig);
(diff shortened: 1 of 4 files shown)


src/lib/Table/Table.svelte
```diff
@@ -133,6 +133,37 @@
   let isRowClickable = $derived(typeof onRowClick === 'function');
   let isStickyHeader = $derived(stickyHeader || isTableScrollable);
 
+  // ─── Horizontal-scroll affordance ────────────────────────────────────────
```
Tara-ag, line 136: ⚠️ MAJOR: The `trackHorizontalScroll` action instantiates `ResizeObserver` unconditionally at the top level of the function. `ResizeObserver` is a browser-only global; accessing it during SSR or in a non-browser environment (e.g., Vitest with jsdom that lacks it, or web-component build tooling) will throw `ReferenceError: ResizeObserver is not defined`. Guard with a browser check or instantiate inside `onMount` so SSR/server builds remain safe.

```ts
const trackHorizontalScroll = (scrollNode: HTMLElement) => {
  const updateScrollHints = () => { ... };
  updateScrollHints();
  scrollNode.addEventListener('scroll', updateScrollHints, { passive: true });
  if (typeof ResizeObserver === 'undefined') {
    return {
      destroy: () => scrollNode.removeEventListener('scroll', updateScrollHints)
    };
  }
  const hintResizeObserver = new ResizeObserver(updateScrollHints);
  ...
};
```

src/lib/Table/Table.svelte
```diff
@@ -133,6 +133,37 @@
...
+  // it as edge scrims (see .table-scroll-shell styles).
+  let canScrollLeft = $state(false);
+  let canScrollRight = $state(false);
+
+  const trackHorizontalScroll = (scrollNode: HTMLElement) => {
+    const updateScrollHints = () => {
+      canScrollLeft = scrollNode.scrollLeft > 2;
+      canScrollRight = scrollNode.scrollLeft + scrollNode.clientWidth < scrollNode.scrollWidth - 2;
+    };
+    updateScrollHints();
```
Tara-ag, line 149: 💡 MINOR: Observing `scrollNode.querySelector('table')` inside the action is fragile — if the table is ever conditionally rendered or replaced, the observer holds a stale reference and misses resize events for the new table. Consider observing the `<table>` via a binding or passing it explicitly, or at least re-querying on mutation. For the current DOM structure this works, but it couples the action to a specific child selector.

src/lib/Table/Table.svelte
```diff
@@ -968,6 +1022,57 @@
     overflow-y: auto;
```
Tara-ag, line 1022: 💡 MINOR: The scrim color defaults to `#ffffff`, which will look broken on dark-themed tables. Consider documenting `--table-scroll-scrim-color` in `docs/Table.md` and/or defaulting to a CSS variable that themes already override for backgrounds (e.g., `--table-content-background` or `--table-row-background`) so dark themes do not have to discover this new token by visual regression.

src/lib/Table/Table.svelte
Tara-ag, line 136: ⚠️ MAJOR: The `trackHorizontalScroll` action instantiates `ResizeObserver` unconditionally at the top level of the function. `ResizeObserver` is a browser-only global; accessing it during SSR or in a non-browser environment (e.g., Vitest with jsdom that lacks it, or web-component build tooling) will throw `ReferenceError: ResizeObserver is not defined`. Guard with a browser check or instantiate inside `onMount` so SSR/server builds remain safe.

```ts
const trackHorizontalScroll = (scrollNode: HTMLElement) => {
  const updateScrollHints = () => { ... };
  updateScrollHints();
  scrollNode.addEventListener('scroll', updateScrollHints, { passive: true });
  if (typeof ResizeObserver === 'undefined') {
    return {
      destroy: () => scrollNode.removeEventListener('scroll', updateScrollHints)
    };
  }
  const hintResizeObserver = new ResizeObserver(updateScrollHints);
  ...
};
```

Tara-ag (changes requested):
## Review Summary

Files reviewed: 4 changed files.

### New issues raised

| Severity | Count | Location |
|----------|-------|----------|
| 🔒 CRITICAL | 0 | — |
| ⚠️ MAJOR | 1 | `src/lib/Table/Table.svelte` |
| 💡 MINOR | 0 | — |
| 💬 SUGGESTION | 0 | — |

### Blocking reason

`trackHorizontalScroll` instantiates `ResizeObserver` unconditionally inside a Svelte action. `ResizeObserver` is a browser-only global; this library ships to SSR consumers and is built as web components, so server-side / non-browser rendering will throw `ReferenceError: ResizeObserver is not defined`. This is a **NEW** SSR-safety regression introduced by this PR and meets the blocking criterion for unguarded `window`/`document`/browser-global access at component init.

### Existing comments

Three comments were already posted by the same reviewer account on this PR (ResizeObserver SSR guard, stale `querySelector('table')` observation, and scrim color dark-theme default). I am treating those as already handled and did not duplicate them. The new blocking comment above is the same SSR issue but raised independently because it remains unaddressed in the current diff.

### Next steps

1. Guard `ResizeObserver` with `typeof ResizeObserver !== 'undefined'` before constructing it, returning a scroll-only destroy path when absent.
2. Re-run `npm run check` / `check:wc` and the unit/integration suites.
3. Once the SSR guard is in place, this PR can be approved (the remaining points are non-blocking suggestions already noted).

PR #380, opened 2026-07-11
feat(Table): leading icon for input cells via TableInputCellData.iconUrl

murdore:
## What

Adds `iconUrl?: string` to `TableInputCellData` — a JSON-safe way to render a passive leading icon inside a `type: 'input'` cell (e.g. a currency glyph), riding `Input`'s existing `leftIcon` snippet. Icon sizes via `--table-cell-input-icon-size` (default 16px).

- `asInputCellData` narrows the new field field-by-field (string only), with unit coverage added in `cell-data.test.ts` (the input decoder previously had none).
- Demo: the interactive-cells example's note input now renders a leading icon.

## Why

The Breeze dashboard Figma Table spec shows currency inputs with a leading ₹ glyph inside table cells; cell data must stay JSON-serializable, so the snippet bridge lives in BuiltinCell.

## Gates

`check` ✓ · `lint` ✓ · `test:unit` ✓ (95 passed, incl. the new decoder test)

Note: the interactive cell shapes (input/select/button) are currently undocumented in docs/Table.md — pre-existing gap, left for a docs pass.

PR #379, opened 2026-07-11
feat(Table): opt-in column highlight via TableColumn.highlighted

murdore:
## What

Adds `highlighted?: boolean` to `TableColumn` — paints that column's header and body cells with a highlight wash:

- `--table-col-highlight-background` (body cells, default `#f3f9ff`)
- `--table-col-highlight-header-background` (header cell, falls back to the body wash)

Row hover keeps higher specificity and row selection is declared later at equal specificity, so both row states still paint over the column wash.

## Why

The Breeze dashboard Figma spec ("All Cell Variants", Breeze Dashboard Design System) shows a highlighted/selected column state (Blue/50 wash) that consumers currently can only approximate with `classes` + nth-child CSS. This makes it first-class on the keyed column model.

## Demo

`/components/table` — the built-in renderers example now highlights the `compare` Revenue column (the exact Figma use case).

## Gates

`check` ✓ · `lint` ✓ · `test:unit` 94 passed ✓

PR #381, opened 2026-07-11
feat(Table): icon-capable button cells (iconUrl + ariaLabel, icon-only ghost)

murdore:
## What

- `TableButtonCellData.text` becomes optional; new JSON-safe `iconUrl` and `ariaLabel` fields.
- `asButtonCellData` accepts text-only, icon-only, or both — never neither (unit matrix added).
- BuiltinCell renders `iconUrl` through Button's existing `icon` snippet; with no text it sets `iconOnly` and the existing `.builtin-icon-button` ghost recipe (same look as the kebab triggers). Icon sized via the existing `--table-cell-icon-size` (default 16px).
- Demo: the interactive-cells example's first action is now an icon-only edit button with an ariaLabel.

## Why

The Breeze dashboard Figma Table spec has an icon-only ghost edit-button cell; cell data must stay JSON-serializable, so the snippet bridge lives in BuiltinCell rather than exposing Snippet in cell data.

## Compatibility

Existing text-button cells are untouched (`text` narrowing unchanged for objects that carry it). `action-group` primary buttons reuse the same decoder and keep working; an icon-only primary now renders instead of being rejected.

## Gates

`check` ✓ · `lint` ✓ · `test:unit` ✓ (96 passed, incl. the new decoder matrix)

## Summary by CodeRabbit

* **New Features**
  * Added support for icon-only and icon-plus-text buttons in table cells.
  * Added safe icon URL handling for table input and button cells.
  * Updated the table example to demonstrate an icon-based “Renew” action.

* **Documentation**
  * Expanded table renderer documentation and clarified icon URL requirements.
  * Updated input and table styling reference tables for improved formatting and accuracy.


src/lib/Table/properties.ts
```diff
@@ -146,11 +146,20 @@ export type TableInputCellData = {
 
 /** Cell shape for `type: 'button'` — the handler lives on the column. */
 export type TableButtonCellData = {
-  text: string;
+  /** Visible label. Omit for an icon-only button (supply `iconUrl` + `ariaLabel`). */
```
Tara-ag: 💡 MINOR: The doc comment says `ariaLabel` is "required in practice for icon-only buttons", but the type makes it optional unconditionally. Consider a discriminated union so icon-only cells must carry `ariaLabel` at compile time, preventing inaccessible buttons:

```ts
export type TableButtonCellData =
  | { text: string; iconUrl?: string; ariaLabel?: string; disabled?: boolean; classes?: string; testId?: string; }
  | { text?: undefined; iconUrl: string; ariaLabel: string; disabled?: boolean; classes?: string; testId?: string; };
```

This aligns with the existing `TableToggleCellData` pattern where `ariaLabel` is optional, but icon-only buttons have no visible text fallback so the stronger contract is warranted.

murdore:
Rebased onto release @ 2.93.0 (post #379/#380) and folded in the follow-ups from #380's approving review:

- **MAJOR (icon URL validation)**: new `asSafeIconUrl` narrowing guard — `iconUrl` in BOTH input and button cell data now accepts only `http(s)`, `data:image/*`, and relative URLs; `javascript:`/non-image `data:` URIs are dropped (an icon-only button with a rejected URL falls back to plain text). Unit matrix added.
- **Docs**: `docs/Input.md` documents the new `ariaLabel` prop; `docs/Table.md` now documents the interactive cell types (select/input/button/action-group/popup-menu), the scheme restriction, and the `--table-cell-icon-size` / `--table-cell-input-icon-size` variables.

Gates: check ✓ lint ✓ test:unit 97 ✓

src/lib/Table/properties.ts
```diff
@@ -157,11 +159,22 @@ export type TableInputCellData = {
...
```
Tara-ag: 🔒 CRITICAL: `TableButtonCellData.text` is narrowed from required to optional. This is a public type exported from `src/lib/index.ts` (`export type * from './Table/properties'`), so any downstream consumer currently assigning or destructuring `{ text }` from `TableButtonCellData` will break at compile time. Per the blocking criteria, a breaking change to a public component prop/export requires either a major-version bump or a documented migration note. Since this PR is titled `feat(Table)` and targets `release`, it will cut a minor release. Either restore `text` as required (keeping icon-only via a separate runtime path) or change the PR to a breaking change (`feat(Table)!:` / `BREAKING CHANGE:` note) and add a migration entry.

Suggested non-breaking alternative: keep `text: string` required in the type and introduce a sibling `TableIconButtonCellData` / renderer for icon-only cases, or use a discriminated union:
```ts
export type TableButtonCellData =
  | { text: string; iconUrl?: string; ariaLabel?: string; disabled?: boolean; classes?: string; testId?: string }
  | { text?: undefined; iconUrl: string; ariaLabel: string; disabled?: boolean; classes?: string; testId?: string };
```

src/lib/Table/BuiltinCell.svelte
```diff
@@ -305,19 +305,36 @@
...
+          icon={buttonCellIcon}
+          iconOnly={!buttonData.text}
+          ariaLabel={buttonData.ariaLabel}
+          disabled={buttonData.disabled ?? false}
+          classes={buttonData.classes ?? ''}
+          testId={buttonData.testId}
+          onclick={() => column.onButtonClick?.(rowIndex, originalIndex)}
+        />
+      {:else}
+        <Button
```
Tara-ag, line 334: ⚠️ MAJOR: The icon-only branch passes `ariaLabel={buttonData.ariaLabel}` to `Button`, but `ariaLabel` is optional in `TableButtonCellData` and the decoder only copies it when present. An icon-only button with a missing `ariaLabel` will render a button element with no accessible name (no visible text and no `aria-label`), which fails WCAG 2.1 4.1.2 / 2.4.4. The doc comment says it is "required in practice", but nothing enforces it.

Suggested fix: at render time, fall back to `buttonData.text` when present, and warn/assert in development when an icon-only button has neither. For example:
```svelte
{@const buttonAriaLabel = buttonData.ariaLabel ?? (buttonData.text ? undefined : 'Action')}
...
<Button
  ...
  ariaLabel={buttonAriaLabel}
/>
```
Better still, adopt the discriminated-union type so icon-only cells must carry `ariaLabel` at compile time.

src/lib/Table/BuiltinCell.svelte
```diff
@@ -305,19 +305,36 @@
 {:else if column.type === 'button'}
   {@const buttonData = asButtonCellData(value)}
   {#if buttonData}
+    {#snippet buttonCellIcon()}
+      <img class="builtin-button-icon" src={buttonData.iconUrl} alt="" />
+    {/snippet}
```
Tara-ag, line 314: ⚠️ MAJOR: The `buttonCellIcon` snippet renders `` with an empty `alt`. For an icon-only button this image is the sole visual content and is effectively decorative only because the button's accessible name comes from `ariaLabel`. That is acceptable, but if `text` is also present the icon becomes meaningful and should expose an accessible name (or be hidden with `aria-hidden` if purely decorative). More importantly, `src` is set directly from `buttonData.iconUrl`, which has been scheme-validated by `asSafeIconUrl`, so XSS via `javascript:` is blocked — good. However, consider also adding `decoding="async"` and handling `onerror` so a broken icon URL does not leave a confusing blank focusable control in icon-only mode.

Suggested addition:
```svelte
<img class="builtin-button-icon" src={buttonData.iconUrl} alt="" decoding="async" />
```

src/lib/Table/cellData.ts
```diff
@@ -31,6 +31,31 @@ export const asJsonObject = (value: TableCellValue): { [key: string]: JSONValue
...
+  const candidate = value.trim();
+  const schemeMatch = /^([a-zA-Z][a-zA-Z0-9+.-]*):/.exec(candidate);
+  if (schemeMatch === null) {
+    return candidate;
+  }
+  const scheme = schemeMatch[1].toLowerCase();
+  if (scheme === 'http' || scheme === 'https') {
+    return candidate;
+  }
+  if (scheme === 'data' && /^data:image\//i.test(candidate)) {
```
Tara-ag: ⚠️ MAJOR: `asSafeIconUrl` rejects SVG data URIs that use base64 encoding because the regex `/^data:image\//i` only matches lowercase `image/`. Base64-encoded SVGs commonly use `data:image/svg+xml;base64,...`, which matches, but `data:IMAGE/SVG...` or mixed-case prefixes from upstream systems will be rejected. More importantly, the regex does not anchor to the start of the string, so a value like `not-a-data:data:image/png,` would incorrectly pass. Tighten the validation:

```ts
if (scheme === 'data' && /^data:image\//i.test(candidate)) {
```

should be anchored and ideally restrict to known safe image media types:

```ts
if (scheme === 'data' && /^data:image\/(svg\+xml|png|jpeg|jpg|gif|webp)(;|$)/i.test(candidate)) {
  return candidate;
}
```

This prevents arbitrary `data:image/html` or malformed payloads from reaching the DOM.

src/lib/Table/BuiltinCell.svelte
```diff
@@ -305,19 +305,36 @@
...
-      <Button
-        text={buttonData.text}
-        disabled={buttonData.disabled ?? false}
-        classes={buttonData.classes ?? ''}
-        testId={buttonData.testId}
-        onclick={() => column.onButtonClick?.(rowIndex, originalIndex)}
-      />
+      {#if buttonData.iconUrl}
+        <Button
+          text={buttonData.text}
```
Tara-ag, line 324: ⚠️ MAJOR: The icon-only button branch passes `text={buttonData.text}` to `<Button>` even when `buttonData.text` is undefined. `<Button>` renders text only when `typeof text === 'string' && text.length > 0`, so this is harmless visually, but it means the `iconOnly` prop is set while `text` is still forwarded. Verify that `Button`'s `iconOnly` styling does not depend on `text` being absent; if it does, omit `text` from the icon-only branch. Also, the `ariaLabel` is only forwarded when an icon is present; a text-only button with an explicit `ariaLabel` would lose it. Forward `ariaLabel` unconditionally for both branches:

```svelte
{#if buttonData.iconUrl}
  <Button
    text={buttonData.text}
    icon={buttonCellIcon}
    iconOnly={!buttonData.text}
    ariaLabel={buttonData.ariaLabel}
    ...
  />
{:else}
  <Button
    text={buttonData.text}
    ariaLabel={buttonData.ariaLabel}
    ...
  />
{/if}
```

src/lib/Table/BuiltinCell.svelte
```diff
@@ -305,19 +305,36 @@
 {:else if column.type === 'button'}
   {@const buttonData = asButtonCellData(value)}
   {#if buttonData}
```
Tara-ag, line 307: ⚠️ MAJOR: The `` inside `buttonCellIcon` uses `alt=""` unconditionally. For an icon-only button, the accessible name must come from `ariaLabel` on the `<Button>` (which is forwarded), so decorative `alt=""` is correct. However, if `text` is also present, the icon becomes decorative and the visible label names the button — also correct. Document this intent in a comment so future maintainers do not "fix" it by adding alt text, which would create redundant announcements. No code change required, but consider a comment:

```svelte
{#snippet buttonCellIcon()}
  <!-- Icon is always decorative; accessible name comes from Button text or ariaLabel. -->
  <img class="builtin-button-icon" src={buttonData.iconUrl} alt="" />
{/snippet}
```

src/lib/Table/properties.ts
```diff
@@ -157,11 +159,22 @@ export type TableInputCellData = {
 
 /** Cell shape for `type: 'button'` — the handler lives on the column. */
 export type TableButtonCellData = {
```
Tara-ag: 🔒 CRITICAL: `TableButtonCellData.text` changed from required to optional. This is a TypeScript-breaking change for downstream consumers who construct button cell data objects without using `asButtonCellData` narrowing. Because this library is published and semver-released via `chore(release)`, narrowing a required property violates the public API contract unless accompanied by a major-version bump or migration note. Restore `text: string` as a required property and express icon-only cells through a discriminated union:

```ts
export type TableButtonCellData =
  | {
      text: string;
      iconUrl?: string;
      ariaLabel?: string;
      disabled?: boolean;
      classes?: string;
      testId?: string;
    }
  | {
      text?: undefined;
      iconUrl: string;
      ariaLabel: string;
      disabled?: boolean;
      classes?: string;
      testId?: string;
    };
```

This keeps existing `text`-required consumers compiling while enforcing `ariaLabel` for icon-only buttons at the type level.

src/lib/Table/BuiltinCell.svelte
```diff
@@ -600,6 +617,12 @@
     height: var(--table-cell-input-icon-size, 16px);
   }
 
+  .builtin-button-icon {
```
Tara-ag, line 642: 💡 MINOR: The new `builtin-button-icon` class duplicates the sizing rules already present for `.builtin-icon-label-icon :global(img), .builtin-icon-label-icon :global(svg)` and `.builtin-menu-dots :global(svg)`. Consider reusing the existing `--table-cell-icon-size` utility or extracting a shared class to reduce duplication. Not blocking.

src/lib/Table/cell-data.test.ts
```diff
@@ -66,6 +67,42 @@ describe('cell narrowing', () => {
...
+
+  it('icon URLs only pass with a safe scheme (http/https/data:image/relative)', () => {
+    expect(asButtonCellData({ iconUrl: 'https://cdn.test/edit.svg' })).toEqual({
+      iconUrl: 'https://cdn.test/edit.svg'
+    });
+    expect(asButtonCellData({ iconUrl: '/assets/edit.svg', ariaLabel: 'Edit' })).toEqual({
+      iconUrl: '/assets/edit.svg',
+      ariaLabel: 'Edit'
+    });
+    // eslint-disable-next-line no-script-url -- asserting the guard rejects it
```
Tara-ag, line 99: 💡 MINOR: The test `icon URLs only pass with a safe scheme` asserts `asInputCellData({ iconUrl: 'javascript:alert(1)' })` returns `{}`. This documents that input cells silently drop bad icon URLs rather than rejecting the whole cell. Consider adding an assertion that `asButtonCellData({ text: 'Renew', iconUrl: 'javascript:alert(1)' })` returns `{ text: 'Renew' }` to lock in the text-button fallback behavior when an unsafe icon URL is supplied.

docs/Table.md
```diff
@@ -65,18 +65,25 @@ The projection used internally is exported as `normalizeColumns(columns, rows)`
...
+| `input`               | `TableInputCellData` `{ value?, placeholder?, disabled?, testId?, ariaLabel?, iconUrl?, dataType?, validationPattern?, onErrorMessage? }` | an Input; `column.onInput(rowIndex, value, originalIndex)`. `ariaLabel` names the field for screen readers (recommended — cells have no visible label); `iconUrl` renders a passive leading icon |
+| `button`              | `TableButtonCellData` `{ text?, iconUrl?, ariaLabel?, disabled?, classes?, testId? }`                                                     | a Button; `column.onButtonClick(rowIndex, originalIndex)`. With `iconUrl` and no `text` it renders icon-only (ghost look) — pair with `ariaLabel`                                                |
+| `action-group`        | `TableActionGroupCellData` `{ primaryButton?, menuItems? }`                                                                               | a primary Button plus a kebab overflow Menu; `column.onPrimaryAction` / `column.onMenuAction`                                                                                                    |
```
Tara-ag, line 83: 💬 SUGGESTION: The new paragraph says "an icon-only button with a rejected URL falls back to plain text", but `asButtonCellData` returns `null` when neither text nor a safe icon URL is present, and BuiltinCell then falls back to `cellValueToText(value)` (rendering `'-'` for an object). Clarify whether the fallback is "plain text of the cell value" or "no button rendered". The current wording could be read as "the button still renders with text", which is not what happens.

Tara-ag (changes requested):
## Review Summary

Reviewed 7 changed files. Found **1 blocking issue** and several non-blocking concerns.

### Blocking
- **🔒 CRITICAL** `TableButtonCellData.text` was changed from required to optional in `src/lib/Table/properties.ts`. This is a TypeScript-breaking change to a published public type. Because the PR targets `release` as a `feat` (minor bump per conventional commits), this either needs a major-version bump, a `!` breaking-change indicator, or — preferably — a discriminated union that keeps `text` required for text-bearing cells while enforcing `ariaLabel` for icon-only cells.

### Non-blocking (MAJOR/MINOR/SUGGESTION)
- **⚠️ MAJOR** `asSafeIconUrl` regex is not anchored and permits malformed `data:` URIs (e.g. `foo:data:image/png,`). Tighten to `/^data:image\/(svg\+xml|png|jpeg|jpg|gif|webp)(;|$)/i`.
- **⚠️ MAJOR** `BuiltinCell.svelte` does not forward `ariaLabel` to `<Button>` in the text-only branch, so a text button with an explicit accessible name loses it.
- **⚠️ MAJOR** Confirm that forwarding `text={buttonData.text}` alongside `iconOnly={!buttonData.text}` is safe given `Button`'s rendering logic.
- **💡 MINOR** New `.builtin-button-icon` CSS duplicates existing icon sizing rules.
- **💡 MINOR** Test matrix should cover a text button paired with an unsafe icon URL to lock in the fallback behavior.
- **💬 SUGGESTION** Docs wording about "falls back to plain text" for rejected icon URLs is ambiguous; clarify that the cell falls back to `cellValueToText` (typically `'-'`).

### Existing comments
- The existing review comment by Tara-ag on `src/lib/Table/properties.ts:162` (discriminated union for `ariaLabel`) is aligned with the blocking concern above and has not been duplicated.

Next step: resolve the public-type breaking change; the other items can be addressed or discussed in follow-ups.

murdore:
Addressed the CHANGES_REQUESTED review + CodeRabbit:

- **CRITICAL (breaking public type)**: `TableButtonCellData` is now a **discriminated union** — `TableTextButtonCellData` keeps `text: string` required exactly as before (no break for existing consumers), and `TableIconOnlyButtonCellData` requires `iconUrl` + `ariaLabel` at compile time. Both variants exported.
- **MAJOR (icon-only accessible name)**: enforced at BOTH layers — the type requires `ariaLabel` for icon-only, and `asButtonCellData` rejects icon-only data without it (falls back to plain text). Unit case added.
- **MAJOR (ariaLabel lost on text buttons)**: `ariaLabel` now forwarded in the text-only branch too, and in the `action-group` primary button (which also gained the same icon rendering so an icon-only primary can't render blank).
- **MAJOR (data: URI allowlist)**: tightened to `/^data:image\/(svg\+xml|png|jpeg|jpg|gif|webp)[;,]/i`. (Note: the previous check already anchored via the scheme parse and was case-insensitive — the `foo:data:image/png,` example was rejected — but the explicit subtype allowlist is strictly stronger.)
- **MAJOR (text + iconOnly interplay)**: `iconOnly` is only true when `text` is absent; Button renders text only for non-empty strings — documented with the new decorative-`alt` intent comment.
- **MINOR (test matrix)**: added `{ text: 'Renew', iconUrl: 'javascript:…' }` → `{ text: 'Renew' }` fallback lock-in.
- **MINOR (demo label mismatch)**: pencil icon's label corrected to 'Edit Growth Monthly'.
- **SUGGESTION (docs wording)**: fallback semantics clarified (text button keeps rendering without the icon; failed icon-only narrows to null → plain-text fallback, `'-'` for objects).

Kept as-is with rationale: `.builtin-button-icon` stays a dedicated class (parallel to `.builtin-input-icon`; both read `--table-cell-icon-size`-family vars — the icon-label rule targets `:global` img/svg inside a different wrapper).

Gates: check ✓ lint ✓ test:unit 97 ✓

src/lib/Table/cellData.ts
```diff
@@ -31,6 +33,33 @@ export const asJsonObject = (value: TableCellValue): { [key: string]: JSONValue
...
+    return null;
+  }
+  const candidate = value.trim();
+  const schemeMatch = /^([a-zA-Z][a-zA-Z0-9+.-]*):/.exec(candidate);
+  if (schemeMatch === null) {
+    return candidate;
+  }
+  const scheme = schemeMatch[1].toLowerCase();
+  if (scheme === 'http' || scheme === 'https') {
+    return candidate;
```
Tara-ag, line 53: ⚠️ MAJOR: The `asSafeIconUrl` scheme regex `/^([a-zA-Z][a-zA-Z0-9+.-]*):/` does not reject whitespace inside the scheme. Per RFC 3986 the scheme cannot contain tabs/newlines, but `RegExp.exec` will match `java\tscript:` and treat the scheme as `java\tscript`, which then falls through to `return null`. However, the original trimmed string (with the embedded whitespace) is what gets returned as `null`, so the DOM is still protected here. The real risk is subtler: an attacker can use `java\tscript:alert(1)` and the function correctly returns `null`, but the test suite should lock this bypass vector down because future refactors might normalize the scheme differently. Add regression assertions for `java\tscript:` and `java\nscript:` in `cell-data.test.ts`.

```ts
expect(asButtonCellData({ iconUrl: 'java\tscript:alert(1)', ariaLabel: 'Edit' })).toBeNull();
expect(asButtonCellData({ iconUrl: 'java\nscript:alert(1)', ariaLabel: 'Edit' })).toBeNull();
```

src/lib/Table/BuiltinCell.svelte
```diff
@@ -305,19 +305,41 @@
...
-      />
+      {#if buttonData.iconUrl}
+        <Button
+          text={buttonData.text}
+          icon={buttonCellIcon}
+          iconOnly={!buttonData.text}
+          ariaLabel={buttonData.ariaLabel}
+          disabled={buttonData.disabled ?? false}
+          classes={buttonData.classes ?? ''}
+          testId={buttonData.testId}
```
Tara-ag, line 330: 💬 SUGGESTION: The icon-only branch passes `text={buttonData.text}` (undefined) alongside `iconOnly={!buttonData.text}`. `Button` guards text rendering with `typeof text === 'string' && text.length > 0`, so this is visually harmless, but it is slightly misleading. Prefer omitting `text` in the icon-only branch for clarity:

```svelte
{#if buttonData.iconUrl}
  {#if buttonData.text}
    <Button text={buttonData.text} icon={buttonCellIcon} ariaLabel={buttonData.ariaLabel} ... />
  {:else}
    <Button icon={buttonCellIcon} iconOnly ariaLabel={buttonData.ariaLabel} ... />
  {/if}
{:else}
  <Button text={buttonData.text} ariaLabel={buttonData.ariaLabel} ... />
{/if}
```

This also avoids relying on Button's internal guard for a public-facing component.

src/lib/Table/BuiltinCell.svelte
```diff
@@ -305,19 +305,41 @@
...
+         accessible name itself. -->
+    {#snippet buttonCellIcon()}
+      <img class="builtin-button-icon" src={buttonData.iconUrl} alt="" />
+    {/snippet}
     <!-- svelte-ignore a11y_no_static_element_interactions -->
     <span
       class="builtin-interactive"
+      class:builtin-icon-button={buttonData.iconUrl && !buttonData.text}
       onclick={stopClickPropagation}
       onkeydown={stopKeydownPropagation}
```
Tara-ag, line 320: 💬 SUGGESTION: The new `class:builtin-icon-button={buttonData.iconUrl && !buttonData.text}` is applied to the wrapper `<span>`, but I don't see a corresponding `.builtin-icon-button` style rule in the diff. If the intent is to reuse the existing `.builtin-icon-button` ghost recipe, verify the selector is actually defined and scoped to this wrapper; otherwise the class is dead. If it is only meant as a hook for downstream theming, document it in `docs/Table.md` under the CSS variables section.

src/lib/Table/cellData.ts
Tara-ag, line 53: ⚠️ MAJOR: The scheme regex `/^([a-zA-Z][a-zA-Z0-9+.-]*):/` does not reject whitespace inside the scheme. Per RFC 3986 the scheme cannot contain tabs/newlines, but `RegExp.exec` will match `java\tscript:` and treat the scheme as `java\tscript`, which then falls through to `return null`. The DOM is still protected today, but this is a bypass vector that future refactors could accidentally normalize. Add regression assertions for `java\tscript:` and `java\nscript:` in `cell-data.test.ts`.

```ts
expect(asButtonCellData({ iconUrl: 'java\tscript:alert(1)', ariaLabel: 'Edit' })).toBeNull();
expect(asButtonCellData({ iconUrl: 'java\nscript:alert(1)', ariaLabel: 'Edit' })).toBeNull();
```

src/lib/Table/BuiltinCell.svelte
```diff
@@ -305,19 +305,41 @@
 {:else if column.type === 'button'}
   {@const buttonData = asButtonCellData(value)}
   {#if buttonData}
+    <!-- The icon <img> is always decorative (alt=""): an icon-only button is
+         named by the required ariaLabel on the Button, and a text-bearing
+         button is named by its visible text — the image never carries the
+         accessible name itself. -->
+    {#snippet buttonCellIcon()}
+      <img class="builtin-button-icon" src={buttonData.iconUrl} alt="" />
+    {/snippet}
```
Tara-ag, line 314: 💬 SUGGESTION: The `` inside `buttonCellIcon` uses `alt=""` unconditionally. For an icon-only button, the accessible name must come from `ariaLabel` on the `<Button>` (which is forwarded), so decorative `alt=""` is correct. However, if `text` is also present, the icon becomes decorative and the visible label names the button — also correct. Document this intent in a comment so future maintainers do not "fix" it by adding alt text, which would create redundant announcements. No code change required, but consider a comment:

```svelte
{#snippet buttonCellIcon()}
  <!-- Icon is always decorative; accessible name comes from Button text or ariaLabel. -->
  <img class="builtin-button-icon" src={buttonData.iconUrl} alt="" />
{/snippet}
```

Tara-ag (approved):
## Review Summary

Reviewed the 7 changed files. The previously blocking public-type concern has been resolved: `TableButtonCellData` is now a discriminated union that keeps `text` required for text-bearing cells while enforcing `iconUrl` + `ariaLabel` for icon-only cells, preserving the existing public contract.

### New issues raised in this run
- **⚠️ MAJOR** `src/lib/Table/cellData.ts:53` — The scheme regex does not reject whitespace inside the scheme (e.g. `java\tscript:`). The DOM is protected today, but this bypass vector should have regression tests.
- **💬 SUGGESTION** `src/lib/Table/BuiltinCell.svelte:314` — Add a comment documenting that the icon `` is intentionally decorative (`alt=""`) because the accessible name comes from `Button` text or `ariaLabel`.

### Existing concerns already handled by prior comments
- Discriminated union / `ariaLabel` enforcement (Tara-ag threads).
- `asSafeIconUrl` anchoring / media-type tightening (Tara-ag thread).
- Demo `ariaLabel` mismatch with pencil icon (CodeRabbit thread).
- Text-only branch `ariaLabel` forwarding, `iconOnly` + `text` interaction, CSS duplication, test coverage for unsafe icon URL fallback, and docs wording (Tara-ag threads).

No new blocking criteria are met. Approving.

PR #384, opened 2026-07-12
refactor(Table): color-agnostic sort chevron asset — currentColor only, states painted in component CSS

murdore:
## Why

Review feedback on #383: the three sort SVGs were the only assets in the library carrying color inside the file (`fill="var(--table-sort-inactive-color, #C7C7C7)"`), while every other icon asset is color-agnostic (`currentColor`) and painted from CSS. The reason for the deviation was that a two-tone glyph needs a second color channel and `currentColor` only provides one — but the right home for that second channel is the component stylesheet, not the asset.

## What

- `sort-default.svg` → pure `fill="currentColor"` on both chevron halves (geometry unchanged), matching the icon convention.
- `sort-ascending.svg` / `sort-descending.svg` **deleted** — one shared pair asset now serves all three states.
- The asc/desc branches render the shared asset with direction classes (`sort-icon-asc` / `sort-icon-desc`); component CSS defaults every path to `var(--table-sort-inactive-color, #c7c7c7)` and releases the sorted direction's half back to `currentColor` (which the existing `--table-sort-active-color` rule feeds). CSS `fill` overrides the SVG presentation attribute, and path order in the asset (up first, down second) is documented at the rule.

## Behavior

Zero change — verified on `/components/table` with computed-style probes, identical to 2.95.0:
idle both `rgb(199,199,199)`; asc `rgb(27,133,255)` + `rgb(199,199,199)`; desc mirrored; hovered idle `rgb(121,121,121)`. Public API and all CSS variables (`--table-sort-active/inactive/hover-color`, opacity vars) unchanged; `sortAscIcon`/`sortDescIcon`/`sortDefaultIcon` snippets unaffected (snippets render without the `.sort-icon` wrapper, so the path rules don't touch them).

`check` / `lint` / `build` green (3 pre-existing lint warnings in `cell-data.test.ts` unchanged).


Tara-ag (approved):
## Review Summary

**Files reviewed:** 4 changed files (1 component, 3 assets)

**New issues raised:** 0

**Decision:** Approve ✅

### Assessment

This is a clean, internal-only refactor of the Table sort icons:

- **Public API:** Unchanged. `sortAscIcon`/`sortDescIcon`/`sortDefaultIcon` snippets still render without the `.sort-icon` wrapper, so the new CSS path rules do not touch consumer-provided icons. All `--table-sort-*` CSS custom properties retain their names and semantics.
- **Exports:** The deleted `sort-ascending.svg` / `sort-descending.svg` assets are only referenced from `Table.svelte`; they are not part of the public barrel (`src/lib/index.ts`) or the web-component registry.
- **Accessibility:** No change to the sort control semantics — the `Button` still carries `ariaLabel="Sort by {header}"` and the SVG remains decorative.
- **SSR / security:** Static SVG imports rendered via `{@html}` are safe; no new `window`/`document` access or unsanitized HTML injection introduced.
- **Svelte 5 reactivity:** Purely presentational; no state/effect/props changes.

The two-tone painting strategy (shared color-agnostic asset + component CSS overriding one chevron half based on direction class) is well-documented in the stylesheet comment and preserves the existing visual states.

PR #385, opened 2026-07-12
feat(ColorPicker): expose --color-picker-input-height and hex text-transform

murdore:
## What

Two backward-compatible CSS-variable exposures so `ColorPicker` can be embedded as a **compact, boxed config field** that matches a native `<input type=color>` + hex-input look:

- `--color-picker-input-height` (default `38px`) — the inline hex value input's height was hardcoded, so a consumer could not shrink the control.
- `--color-picker-value-text-transform` (default `none`) — the hex **value** was never transformed (only the `HEX`/`RGB`/`HSL` labels are uppercased), so consumers couldn't display uppercase hex. Applied to both the inline and popover HEX inputs.

Threading the transform required a new **`--input-text-transform`** (default `none`) on `Input` — a bare `text-transform` on the field wrapper does not reach the form control's own text.

## Why

A downstream dashboard (Lighthouse) migrated hand-rolled color swatches to library `ColorPicker` and found the resting look drifted: the box grew taller and the hex rendered lowercase, with no token to correct either. These expose the two hardcoded values; **all defaults preserve current behavior**, so no existing consumer changes.

## Verification

Built + `publint` clean. Verified end-to-end in the consuming app: with `--color-picker-input-height` + `--color-picker-value-text-transform: uppercase`, the inline field's `<input>` computes `text-transform: uppercase` (hex renders `#FCC93A` instead of `#fcc93a`) and shrinks to the requested height, in both light and dark. Docs updated for `ColorPicker` and `Input`.

## Summary by CodeRabbit

* **New Features**
  * Added configurable text transformation for input and textarea values, such as uppercase formatting.
  * Added theme controls for color picker input height and value text transformation.

* **Documentation**
  * Updated CSS variable documentation for color picker and input customization options.


docs/ColorPicker.md
Tara-ag, line 1: 💡 MINOR: The documented defaults for `--color-picker-swatch-btn-border-radius` (`10px 0 0 10px`), `--color-picker-swatch-border-radius` (`5px`), `--color-picker-popover-border-radius` (`12px`), and `--color-picker-panel-border-radius` (`8px`) conflict with the implementation in `ColorPicker.svelte`, which falls back to `var(--radius, 4px)` for each. Align the docs with the code or vice versa so consumers aren't surprised by different visuals.

Tara-ag (commented):
## Review Summary

Files reviewed: 4 (`docs/ColorPicker.md`, `docs/Input.md`, `src/lib/ColorPicker/ColorPicker.svelte`, `src/lib/Input/Input.svelte`)

New issues raised: 1 MINOR

- **docs/ColorPicker.md**: Documented radius defaults for swatch button, swatch, popover, and panel do not match the implementation's `var(--radius, 4px)` fallbacks. This is a docs/consumer-trust issue, not a breaking API change.

No blocking issues found in this review. The new CSS custom properties (`--color-picker-input-height`, `--color-picker-value-text-transform`, `--input-text-transform`) are purely additive with defaults that preserve existing behavior, so they do not break the public API. The implementation correctly threads the text transform through `Input.svelte` to the native control, and the height token is applied only to the inline trigger input as described.

Note: There is one pre-existing CodeRabbit comment on the same docs discrepancy. My comment was added as a file-level note rather than a duplicate line comment, since the existing thread already covers the line-level concern.

PR #388, opened 2026-07-14
fix(Table): thread usePortal to in-cell Select/Menu so they escape the table's overflow

murdore:
## Problem

Third and final library piece for the downstream table-cell dropdown clipping (Breeze/Lighthouse BZ-4585). #386 (Select) and #387 (Menu) added `usePortal` to those components — but tables render their in-cell dropdowns *themselves*: a `type: 'select'` column renders a `<Select>` inside `BuiltinCell`, and `type: 'action-group'` / `'popup-menu'` columns render `<Menu>`. Consumers have no handle on those instances, so they still get clipped by the table's own scroll/overflow container.

## Fix

Add a **`usePortal`** prop to `Table` (default `false`), threaded through `BuiltinCell` to:
- the in-cell `<Select>` (`type: 'select'`)
- the in-cell `<Menu>` (`type: 'action-group'` action menu, and `type: 'popup-menu'`)

When set, those dropdowns portal to `document.body` and position `fixed` (using the `usePortal` support shipped in `2.96.1`), so the table's `overflow` can't clip them. One prop on the table opts in every in-cell dropdown at once, which matches the use case — a table inside a scroll container wants all of them to escape.

Default `false` keeps in-flow rendering and every existing table behaviour unchanged.

## Verification

- `pnpm run lint` — 0 errors · `pnpm run check` — 0 errors · `pnpm run build` — success
- `pnpm run test:integration` (playwright) — 12/12: **2 new** specs (`tests/table-cell-portal.test.ts`) — the in-cell select dropdown portals to `<body>` and escapes a 120px `overflow:hidden` frame; a control confirms the default (non-portal) table keeps its dropdown in-flow — plus the existing `table-interactive-cells` suite, no regression.
- New demo section on `/components/table` ("usePortal — in-cell dropdown escapes a clipping container").

## Chain

1. `fix(Select): usePortal` — #386 ✅ (2.96.1)
2. `fix(Menu): usePortal` — #387 ✅ (2.96.1)
3. **`fix(Table): usePortal passthrough`** — this PR
4. Consumer (Lighthouse): set `usePortal` on the RTO config tables → the in-cell Rate-Type Select and row-action menus stop clipping (BZ-4585).

## Summary by CodeRabbit

- **New Features**
  - Added an optional `usePortal` table setting for rendering in-cell dropdowns and menus outside clipped containers.
  - Added a documentation demo showing portal-rendered controls in an overflow-constrained table.

- **Bug Fixes**
  - Prevented table dropdowns and menus from being clipped by scroll or overflow boundaries when portal rendering is enabled.

- **Tests**
  - Added coverage for portal and standard in-table dropdown behavior, including selecting options beyond the clipped area.

### DIFF
--- a/src/lib/Table/BuiltinCell.svelte
+++ b/src/lib/Table/BuiltinCell.svelte
@@ -34,12 +34,14 @@
     column,
     value,
     rowIndex,
-    originalIndex
+    originalIndex,
+    usePortal = false
   }: {
     column: TableColumn;
     value: TableCellValue;
     rowIndex: number;
     originalIndex: number;
+    usePortal?: boolean;
   } = $props();
 
   // The td applies column.align as text-align, which flex containers ignore:
@@ -261,6 +263,7 @@
         disabled={selectData.disabled ?? false}
         testId={selectData.testId}
         itemTestId={selectData.itemTestId}
+        {usePortal}
         onchange={(selectedIds) => {
           if (selectedIds.length > 0) {
             column.onSelect?.(rowIndex, selectedIds[0], originalIndex);
@@ -389,6 +392,7 @@
             separator: item.separator
           }))}
           testId={column.testId && `${column.testId}-menu-${rowIndex}`}
+          {usePortal}
           onselect={(menuItem) => column.onMenuAction?.(rowIndex, menuItem.value, originalIndex)}
         >
           {#snippet trigger()}
@@ -427,6 +431,7 @@
           separator: item.separator
         }))}
         testId={column.testId && `${column.testId}-popup-${rowIndex}`}
+        {usePortal}
         onselect={(menuItem) => column.onMenuAction?.(rowIndex, menuItem.value, originalIndex)}
       >
         {#snippet trigger()}
--- a/src/lib/Table/Table.svelte
+++ b/src/lib/Table/Table.svelte
@@ -50,7 +50,8 @@
     rowNumberColumn = false,
     rowNumberLabel = '#',
     headerTooltipIcon,
-    headerTooltipPosition
+    headerTooltipPosition,
+    usePortal = false
   }: TableProperties = $props();
 
   // ─── Keyed column model → positional projection ─────────────────────────────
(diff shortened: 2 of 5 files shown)


Tara-ag (changes requested):
## Review Summary

This PR adds an opt-in `usePortal` prop to `Table` and threads it through `BuiltinCell` to in-cell `Select` and `Menu` instances, solving the downstream clipping issue (BZ-4585) while keeping the default behavior unchanged. The change is backward-compatible and well-scoped.

### New issues found

- **⚠️ MAJOR — Web-component API incomplete**: `src/wc/components/Table.wc.svelte` does not declare `usePortal` in its `customElement.props` map, so `<sui-table use-portal>` cannot receive the new prop. The Svelte and web-component surfaces must stay in sync.

  Suggested fix:
  ```svelte
  usePortal: { type: 'Boolean', reflect: true, attribute: 'use-portal' },
  ```

- **💡 MINOR — Docs props table stale**: `docs/Table.md` does not list the new `usePortal` prop in the Props table. Add a row documenting it so consumers discover the feature.

### Existing comments

- CodeRabbit already flagged a potential test-flakiness issue in `tests/table-cell-portal.test.ts` (waiting for the listbox to render before asserting it is not a direct child of `<body>`). That point is acknowledged and not duplicated here.

### Blocking assessment

The missing web-component prop declaration is a public-API gap in a published library that ships a web-component build. Until `src/wc/components/Table.wc.svelte` exposes `usePortal`, this PR does not fully deliver the advertised contract, so I am requesting changes.

Tara-ag:
## Out-of-diff follow-ups from review

Two items surfaced during review that apply to files **not changed in this PR**, so they couldn't be attached as inline comments:

### ⚠️ MAJOR — Web-component wrapper missing `usePortal`

`src/wc/components/Table.wc.svelte` does not declare `usePortal` in its `customElement.props` map. The Svelte `Table` now accepts the prop, but `<sui-table use-portal>` cannot receive it. Add:

```svelte
usePortal: { type: 'Boolean', reflect: true, attribute: 'use-portal' },
```

### 💡 MINOR — `docs/Table.md` props table is stale

The new `usePortal` prop is not documented in the Table docs. Add a row to the Props table so consumers discover the feature and its default (`false`).

Both are small follow-ups that complete the public-API surface for this change.

PR #395, opened 2026-07-14
feat(modal): add --modal-overlay-background-color overlay token

murdore:
## Problem

Modal's overlay backdrop reads its colour from the generic `--background-color` token:

```css
.overlay-active { background-color: var(--background-color, #00000066); }
```

`--background-color` is a very common name, and consuming apps frequently define their own `--background-color` (often `transparent`) for unrelated purposes. Because CSS custom properties inherit, that app value cascades into the modal overlay and **silently overrides the backdrop** — a `showOverlay` modal then renders no visible dim (it falls back to whatever the app set, e.g. transparent, instead of the intended `#00000066`).

This bit a downstream app: every modal that adopted the library-native overlay lost its backdrop, and the only workaround was a per-modal `--background-color` override, which is fragile and can't be set route-wide (it would bleed into non-modal consumers of the same generic token).

## Change

Introduce a **modal-specific** token `--modal-overlay-background-color` and read it first, falling back to the legacy generic token, then the existing default:

```css
background-color: var(--modal-overlay-background-color, var(--background-color, #00000066));
```

- **Fully backward-compatible** — consumers already theming via `--background-color` are unaffected (it stays a legacy alias), and the `#00000066` default is unchanged. When the new var is unset (every current consumer), the computed value is byte-identical to before.
- New consumers can theme the backdrop via `--modal-overlay-background-color` without colliding with app-level `--background-color`, and can safely set it once at a route/container level.
- Matches the existing modal-specific naming already used in this component: `--modal-overlay-backdrop-filter`, `--modal-content-background-color`.

## Docs

`docs/Modal.md` CSS-variable table: adds `--modal-overlay-background-color` (preferred) and marks `--background-color` as a legacy alias.

## Verification

- `npm run lint` — clean (prettier + eslint on `src`).
- `npm run check` — 0 errors (svelte-check; the 4 warnings are pre-existing and unrelated).
- `npm run build` — succeeds (`vite build` + `svelte-package` + publint).
- `npm run test:unit` — 108/108 pass.


Tara-ag (approved):
## Review Summary

**Files reviewed:** 2 (`src/lib/Modal/Modal.svelte`, `docs/Modal.md`)

**New issues raised:** 0

**Decision:** Approve ✅

### Assessment

This is a clean, backward-compatible additive change:

- Introduces `--modal-overlay-background-color` as a modal-specific token while keeping `--background-color` in the fallback chain, so existing themes continue to compute identically.
- Matches the existing modal-specific naming convention (`--modal-overlay-backdrop-filter`, `--modal-content-background-color`).
- Updates `docs/Modal.md` to document the new token and mark the old one as a legacy alias.
- No public prop/event/export changes; no SSR, accessibility, reactivity, or security concerns introduced.

Good to merge.

PR #398, opened 2026-07-15
docs(BarChart): document and test the already-shipped valueLabel override

murdore:
The bar `valueFormat` callback receives only the value — not the point/series index — so a consumer can't give one bar a different value label from the rest. This blocks the Breeze analytics funnel, whose **baseline (first) step shows its raw count** (e.g. `460 / 359`) while later steps show a %.

Add an optional `valueLabel?: string` to `BarChartDataPoint`. When present, `getDisplayValue` returns it verbatim; otherwise the existing normalized / range / `valueFormat(value)` path is unchanged. Fully backward-compatible (new optional field). Verified on the Lighthouse analytics funnel — first step renders the count, the rest render %. Consumer: bz/lighthouse#6338.

## Update — 2026-08-07

This PR sat open since 15 July while `release` moved 70 commits ahead. On rebasing onto current `release`, it turned out the `BarChartDataPoint`/`BarChart.svelte` implementation from this PR's commit had **already shipped to `release` separately**, via a different commit (`633e5a3`) with the same author, message, and timestamp as this PR's original commit — the same change landed through another path while this PR remained open with its own unmerged copy.

The rebase correctly dropped the now-redundant implementation hunks (git recognized they were already satisfied upstream, byte-identical). What did **not** ship, and what review here had already flagged as missing, is real and still outstanding on `release` today:

- `docs/BarChart.md` — the Props table's `data` row and the `BarChartDataPoint` type-reference block didn't mention `valueLabel`.
- `tests/chart-behaviors.spec.ts` — no coverage existed for the `valueLabel` override path or the default-`valueFormat` fallback.

Both are fixed in the current commit. So while the original title/description above described adding the field, the PR's actual remaining diff (and value) is exclusively closing that documentation/test gap for a field that's already public and already shipped — i.e. addressing Tara-ag's two review threads. See the reworded commit message for the precise scope.

## Summary by CodeRabbit

* **Documentation**
  * Updated bar chart documentation with the already-shipped `valueLabel` property (Props table + type reference).
  * Added a demo fixture demonstrating custom and default value labels.

* **Tests**
  * Added coverage verifying `valueLabel` renders verbatim, and that bars without it fall back to the default `valueFormat`.


src/lib/BarChart/properties.ts
```diff
@@ -49,6 +49,12 @@ export type BarChartDataPoint = {
   range?: [number, number];
   /** Per-bar fill: plain color, pattern, or gradient. Overrides series color. */
   color?: BarFill;
+  /**
```
Tara-ag, line 52: 💡 MINOR: The new `valueLabel` field is documented inline here, but `docs/BarChart.md` is out of sync: the Props table still describes `data` as `{label, value, range?, color?}` and the Type Reference `BarChartDataPoint` block omits `valueLabel`. Since this is a published library, public prop/field changes should be reflected in the component docs.

src/lib/BarChart/BarChart.svelte
```diff
@@ -246,6 +246,9 @@
   let isStackedMode = $derived(isMulti && groupMode === 'stacked');
 
   function getDisplayValue(bar: BarRect): string {
+    if (typeof bar.dataPoint.valueLabel === 'string' && bar.dataPoint.valueLabel.length > 0) {
```
Tara-ag, line 250: 💡 MINOR: New `valueLabel` behavior is not covered by tests. Consider adding a Playwright or Vitest assertion that a data point with `valueLabel` renders that exact text as a bar value label (and that the default `valueFormat` path still applies when `valueLabel` is absent). The existing `tests/chart-behaviors.spec.ts` already exercises value-label positioning, so a small addition there would fit naturally.

Tara-ag (commented):
## Review summary

**Files reviewed:** 2 (`src/lib/BarChart/BarChart.svelte`, `src/lib/BarChart/properties.ts`)

**New issues raised:** 2 MINOR

- `src/lib/BarChart/properties.ts:52` — docs out of sync for new `valueLabel` field
- `src/lib/BarChart/BarChart.svelte:249` — no test coverage for the new `valueLabel` behavior

**Blocking criteria:** None met.

**Public API / backward compatibility:** ✅ Safe. `valueLabel` is a new optional field on `BarChartDataPoint`; no props, exports, event signatures, or CSS custom properties were changed or removed.

**Accessibility, reactivity, SSR, security:** ✅ No concerns.

**Recommendation:** Address the two MINOR items (docs + test) before merging, but the change itself is safe to ship.

src/lib/BarChart/properties.ts
murdore: Fixed in e3c3ec3. Updated `docs/BarChart.md`:
- Props table's `data` row now lists `{label, value, range?, color?, valueLabel?}` and calls out that `valueLabel` renders verbatim, overriding the default `valueFormat(value)` output.
- The `BarChartDataPoint` block in the Type Reference section now includes `valueLabel?: string;`.

src/lib/BarChart/BarChart.svelte
murdore: Fixed in e3c3ec3. Added a new demo fixture (`testId="bar-value-label-chart"`) to the BarChart docs route with a 3-point dataset — one point sets `valueLabel: '5,000 visits'`, the other two don't — and a new `tests/chart-behaviors.spec.ts` test ("bar value label override") asserting:
- the point with `valueLabel` renders that exact string as its bar-value label
- the points without it still go through the default `valueFormat` path (`1.2K`, `340`)

Ran against the live suite: 9/9 passed in `chart-behaviors.spec.ts`, and the full integration run shows only the pre-existing Table-only failures (unrelated to this change).

src/lib/BarChart/properties.ts
murdore: Heads-up on scope: after rebasing onto current `release` (this PR had gone 70 commits stale), it turned out the `valueLabel` implementation itself had already shipped to `release` separately via a different commit — same author/message/timestamp as this PR's original commit, landed through another path. The rebase correctly dropped that now-redundant implementation diff.

So this PR's diff shrank from "add the field" to exactly what this thread flagged: docs only (see reply above — unchanged in substance). PR retitled/redescribed accordingly so the diff isn't a surprise. Nothing about the docs fix itself changed.

src/lib/BarChart/BarChart.svelte
murdore: Heads-up on scope: after rebasing onto current `release` (this PR had gone 70 commits stale), it turned out the `valueLabel` implementation itself had already shipped to `release` separately via a different commit — same author/message/timestamp as this PR's original commit, landed through another path. The rebase correctly dropped that now-redundant implementation diff (git recognized `BarChart.svelte`/`properties.ts` were already satisfied upstream, byte-identical).

So this PR's diff shrank from "add the field" to exactly what this thread flagged: the test (and its demo fixture) — unchanged in substance from the reply above. PR retitled/redescribed accordingly so the diff isn't a surprise.

src/lib/BarChart/BarChart.svelte
```diff
@@ -246,6 +246,9 @@
...
+      return bar.dataPoint.valueLabel;
+    }
```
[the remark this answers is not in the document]
murdore: Good catch on the ambiguity, but taking the doc/test remedy rather than the code change: `BarChart.svelte:252` isn't actually in this PR's scope anymore — that implementation shipped to `release` separately (see the earlier thread notes above), and this PR is now exclusively docs + tests for the already-shipped field. Editing runtime behavior here would reintroduce the scope confusion we just cleaned up, and a change to shipped behavior deserves its own PR/consideration of existing consumers regardless.

So: documented and tested the non-empty-only contract instead.
- `docs/BarChart.md`: `valueLabel` description now states a **non-empty** string overrides `valueFormat(value)`, and an empty string (`''`) is treated the same as an absent field.
- `tests/chart-behaviors.spec.ts`: added a case with `valueLabel: ''` in the demo fixture, asserting it falls back to the formatted value rather than rendering blank.

If the team later decides `''` should render as an empty label instead, that's a deliberate behavior change — the new test will make it visible (a failing assertion) rather than let it happen silently.

murdore:
## Proof of testing

Recorded against this branch's own build of `/components/bar-chart`, on a private port. The clip asserts while it records, so a selector matching nothing fails the recording rather than producing a video of an empty page.

https://github.com/user-attachments/assets/e1ba09d5-0d90-4758-9e3f-1ad7f78fdd93

What it proves — the contract, not just the feature:

| bar | `valueLabel` | rendered |
|---|---|---|
| Visited | `'5,000 visits'` | **5,000 visits** — verbatim |
| Signed up | absent | `1.2K` — default `valueFormat` |
| Purchased | absent | `340` — default `valueFormat` |
| Refunded | `''` (empty) | `28` — **falls back to the formatter** |

The second and third rows are the half that matters: an override leaking onto neighbouring bars would be a regression, and asserting only bar 0 would not catch it. The fourth pins the non-empty-only rule — without it a stray `''` in consumer data would silently erase a bar's value.

One mistake the harness caught while writing it: my first version read the labels with `innerText`, which is `undefined` on SVG `<text>` nodes. The Playwright assertions passed and only my on-screen readout failed, which is exactly the sort of thing a video without assertions would have hidden.

Rebase status: this branch is 52 commits behind `release` and its last CI run predates that by weeks, so the checks on it are stale even though they are green. Worth a rebase before merge.

PR #399, opened 2026-07-15
fix(fileinput): open the picker on click of the whole drop zone

murdore:
## What

`FileInput`'s wrapper is `role="button"` and opened the native picker on **keyboard** (Enter/Space) and **drag-drop**, but **not on a plain click**. So a non-button trigger — e.g. a `Card` drop zone — couldn't open the picker without the consumer wiring `openFilePicker` onto an inner control. This adds `onclick` to the wrapper: clicking anywhere in the drop zone now opens the picker.

## Safety (no double-open, no re-entrancy)

- **Double-open guard:** `openFilePicker` is idempotent within a single click dispatch (a busy flag released on the next microtask). A trigger that *also* wires `openFilePicker` on an inner `<button>` — plus this wrapper's new `onclick` — for the same user click opens the picker exactly **once**.
- **Re-entrancy guard:** `openFilePicker()` calls `inputEl.click()`, whose click bubbles back to the wrapper; a `event.target === inputEl` check ignores it so we don't loop.

## Proof

Counted hidden-input clicks per user click on the demo:
- Card trigger (no inner handler): **1** — click-to-open works.
- Button trigger (inner `onclick={openFilePicker}` + wrapper `onclick`): **1** — no double-fire regression.

`svelte-check` 0 errors, `eslint` clean, `svelte-package` + `publint` clean. Demo (`/components/file-input`) extended with a Card-style trigger.

## Why

Unblocks the Lighthouse shipping-COD CSV upload migration (a `Card` drop zone) to library `FileInput` (BZ-4233).

## Summary by CodeRabbit

* **New Features**
  * Added support for opening the file picker by clicking a styled card or wrapper area.
  * Added a card-style file input demonstration with drag-and-drop prompt updates.

* **Bug Fixes**
  * Prevented duplicate file picker launches from repeated or nested click events.

### DIFF
--- a/src/lib/FileInput/FileInput.svelte
+++ b/src/lib/FileInput/FileInput.svelte
@@ -15,11 +15,22 @@
 
   let dragOver = $state(false);
   let inputEl: HTMLInputElement | null = $state(null);
+  let pickerBusy = false;
 
   export const openFilePicker = (): void => {
-    if (!disabled) {
-      inputEl?.click();
+    if (disabled || pickerBusy) {
+      return;
     }
+    // Idempotent within one click dispatch: if an inner trigger control wires
+    // openFilePicker AND the wrapper's own onclick fires for the same user click,
+    // the second call is a no-op (guarding against a double-open). Released on the
+    // next microtask — after the whole click has finished bubbling — so a genuinely
+    // separate later click still opens the picker.
+    pickerBusy = true;
+    inputEl?.click();
+    queueMicrotask(() => {
+      pickerBusy = false;
+    });
   };
 
   function isAccepted(file: File): boolean {
@@ -107,6 +118,20 @@
       openFilePicker();
     }
   }
+
+  function handleClick(event: MouseEvent): void {
+    if (disabled) {
+      return;
+    }
+    // openFilePicker() calls inputEl.click(), whose click bubbles back up to this
+    // wrapper — ignore it so we don't re-enter openFilePicker in a loop. Consumers
+    // whose trigger already wires openFilePicker on an inner control should NOT also
+    // rely on this handler (it would double-open); clicking the trigger is enough.
+    if (event.target === inputEl) {
+      return;
+    }
+    openFilePicker();
+  }
 </script>
 
 <div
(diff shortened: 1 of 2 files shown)


src/lib/FileInput/FileInput.svelte
```diff
@@ -121,6 +146,7 @@
   ondragleave={handleDragLeave}
```
Tara-ag, line 146: ⚠️ MAJOR: Clicking the wrapper now opens the picker, but the wrapper is `role="button"` and already responds to Enter/Space via `onkeydown={handleKeyDown}`. With this change, keyboard users pressing Enter/Space will fire `handleKeyDown`, which calls `openFilePicker()`, and the native click event from the keyboard activation will also bubble to `handleClick`, which calls `openFilePicker()` again. The `pickerBusy` guard prevents a double-open, but relying on a microtask guard for correct keyboard behavior is fragile. Consider explicitly handling keyboard activation in `handleKeyDown` (e.g. `event.preventDefault()` after calling `openFilePicker`) so the synthetic click does not re-trigger `handleClick`, or make `handleClick` ignore non-primary buttons and verify the event is a genuine pointer click. Suggested fix: in `handleKeyDown`, after calling `openFilePicker()` for Enter/Space, call `event.preventDefault()` to suppress the subsequent click event.

Tara-ag (commented):
## Review Summary

**Files reviewed:** 2 (`src/lib/FileInput/FileInput.svelte`, `src/routes/components/file-input/+page.svelte`)
**Existing comments:** 0
**New issues raised:** 1
- ⚠️ MAJOR: 1

### Notes

The PR is already merged and closed, so this review is post-hoc. I raised one MAJOR concern on `FileInput.svelte`: the wrapper's new `onclick` handler may interact with the existing keyboard (`Enter`/`Space`) activation path in `handleKeyDown`, causing `openFilePicker()` to be invoked twice per keyboard action. While the `pickerBusy` microtask guard prevents an actual double-open, relying on that guard for keyboard correctness is fragile and could mask future regressions. Adding `event.preventDefault()` in `handleKeyDown` after opening the picker would cleanly prevent the synthetic click from re-entering `handleClick`.

No blocking criteria (secrets, critical security vulnerabilities, or undocumented breaking public API changes) were identified in this run.

PR #420, opened 2026-07-25
fix: emit a native testID alongside data-pw for every testId

murdore:
## Problem

Components accept a single `testId` prop but emit **only** `data-pw`. Playwright reads `data-pw`; **Appium resolves an element by its native accessibility identifier and never sees it.** So a component with a `testId` is invisible to native automation.

Consumers work around this with a hand-rolled Svelte action:

```ts
// consumer-side workaround
export function testAttributes(node: HTMLElement, id: string) {
  node.setAttribute("data-pw", id);
  node.setAttribute("testID", id);   // <- the bit the library does not do
}
```

That makes migrating a raw `<button use:testAttributes={id}>` to the library `<Button testId={id}>` a **silent loss of native-test coverage** — which is currently blocking a component-migration campaign in Lighthouse (64 such call sites across three files).

## Fix

Every `testId` emission now renders **both** attributes, from the same value:

```svelte
<button data-pw={testId} testID={testId}>
```

`119` emission sites across `66` components, plus the five hand-rolled spread objects in `Select.svelte`. `data-pw` behaviour is unchanged — this is purely additive.

A shipped module augmentation (`src/lib/testing-attributes.d.ts` → `dist/`) declares the attribute on `svelte/elements`, so it type-checks here *and* for consumers writing their own `testID`.

## Two non-obvious decisions

**1. Declared attribute, not a spread.** The obvious DRY fix is a helper spread onto each element:

```svelte
{...testAttributes(testId)}   <!-- rejected -->
```

It dodges the type error, but **a spread attribute makes Svelte skip its static a11y analysis for that element.** Implementing it that way turned six existing `svelte-ignore` comments into dead code (`svelte/no-unused-svelte-ignore` errors in `Book`, `Checkbox`, `Combobox`) — i.e. it would have silently disabled real a11y warnings across 117 elements. Explicit attributes keep both the type check and the a11y analysis.

**2. The declared key is lower-case (`testid`).** Svelte normalises DOM attribute names, as does `setAttribute` on an HTML document — the rendered attribute is `testid` either way. Authors still write `testID`. (The type parameter name in the augmentation must also match `svelte/elements` exactly: renaming `T` breaks interface merging and produced 130 phantom errors.)

## Verification

| Gate | Result |
|---|---|
| `npm run check` | **0 errors** (was 0 on base; 4 pre-existing warnings) |
| `npm run lint` | **0 errors**, 3 warnings — identical to base |
| `npm run test:unit` | **108 passed** |
| `npm run test:integration` | **127 passed** — no regressions |
| New integration test | **2 passed** |
| `npm run build` | clean; `dist/testing-attributes.d.ts` ships |

`tests/native-test-id.test.ts` is added and asserts the real rendered DOM: the same element carries `data-pw="checkbox-default"` **and** `testid="checkbox-default"`, and absent ids emit no stray `testid="undefined"`.

## Compatibility

Backwards compatible. No API change, no `data-pw` change, no visual change — one additional attribute on elements that already had `data-pw`. Consumers can delete their `testAttributes` actions and pass `testId` instead.

## Summary by CodeRabbit

* **New Features**
  * Added native `testID` attributes across UI components, including controls, charts, navigation, overlays, inputs, and table elements.
  * Test identifiers are now available for accessibility-focused automation while preserving existing selectors.
  * Added support for test identifier typing on HTML and SVG elements.

* **Tests**
  * Added coverage confirming valid identifiers render correctly and are omitted when not provided.

### DIFF
--- a/src/lib/Accordion/Accordion.svelte
+++ b/src/lib/Accordion/Accordion.svelte
@@ -45,6 +45,7 @@
   class="accordion {classes ?? ''}"
   class:expanded={expand}
   data-pw={typeof testId === 'string' ? testId : null}
+  testID={typeof testId === 'string' ? testId : null}
 >
   <div class="accordion-content">
     {@render children?.()}
--- a/src/lib/Animations/ModalAnimation.svelte
+++ b/src/lib/Animations/ModalAnimation.svelte
@@ -45,6 +45,7 @@
       in:fly|global={flyAnimationProperties}
       out:fly|global={flyAnimationProperties}
       data-pw={typeof testId === 'string' ? testId : null}
+      testID={typeof testId === 'string' ? testId : null}
     >
       {@render children?.()}
     </div>
@@ -52,6 +53,7 @@
     <div
       in:fly|global={flyAnimationProperties}
       data-pw={typeof testId === 'string' ? testId : null}
+      testID={typeof testId === 'string' ? testId : null}
     >
       {@render children?.()}
     </div>
@@ -60,6 +62,7 @@
       in:fade|global={fadeAnimationProperties}
       out:fade|global={fadeAnimationProperties}
       data-pw={typeof testId === 'string' ? testId : null}
+      testID={typeof testId === 'string' ? testId : null}
     >
       {@render children?.()}
     </div>
@@ -67,6 +70,7 @@
     <div
       in:fade|global={fadeAnimationProperties}
       data-pw={typeof testId === 'string' ? testId : null}
+      testID={typeof testId === 'string' ? testId : null}
     >
       {@render children?.()}
     </div>
--- a/src/lib/Animations/OverlayAnimation.svelte
+++ b/src/lib/Animations/OverlayAnimation.svelte
@@ -10,6 +10,10 @@
   let { children, testId }: Props = $props();
 </script>
 
-<div out:fade={{ duration: 350 }} data-pw={typeof testId === 'string' ? testId : null}>
+<div
+  out:fade={{ duration: 350 }}
+  data-pw={typeof testId === 'string' ? testId : null}
+  testID={typeof testId === 'string' ? testId : null}
+>
   {@render children?.()}
 </div>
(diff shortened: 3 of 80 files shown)


murdore:
## Rebased onto current `release` — regenerated, not hand-merged (`1ea868e`)

The branch was 53 commits behind and conflicted in `Tabs`/`Toolbar`/`Tooltip`. Since this change is produced by a mechanical codemod, I **reset to `release` (`ad8d96c`, 2.111.0) and re-ran it** rather than hand-resolving — a deterministic regeneration is safer than reconciling conflict hunks in 69 files.

That was the right call: the newer base has **186** emission sites, not the 119 on the old base — a hand-merge would have left ~67 components uncovered.

### Now at full parity — 186 `data-pw` ↔ 186 `testID`

Re-running from scratch also surfaced four cases the first pass missed, now covered:

| Case | Handling |
|---|---|
| `Tooltip` portal bubble | built imperatively via `setAttribute` — now sets both, from one hoisted `bubbleTestId` |
| `LineChart` `hover-overlay` | static literal id — mirrored |
| `_chart/ChartTooltip` `chart-tooltip` ×2 | static literal ids — mirrored |
| `Select` ×5 | hand-rolled spread objects — `testID` key added |

### Verification against this base

| Gate | Base (`ad8d96c`) | This branch |
|---|---|---|
| `check` | 0 errors | **0 errors** |
| `lint` | 0 errors, 3 warnings | **0 errors, 3 warnings** |
| `test:unit` | — | **121 passed** |
| `test:integration` | **9 failed / 122 passed** | **9 failed / 124 passed** |

⚠️ **The 9 integration failures are pre-existing on `release` `ad8d96c`** — I verified by stashing this change and running the suite on the clean base, which reproduces exactly the same 9. This PR adds no failures and contributes the **+2** passing tests (`tests/native-test-id.test.ts`), which assert the real rendered DOM carries `data-pw` *and* `testid` on the same element, and that absent ids emit no stray `testid="undefined"`.

PR #426, opened 2026-08-03
feat(IframeViewer): add credentialless prop and postMessage accessor

murdore:
## Summary

Closes the API gap documented against lighthouse PR #6699 (`refactor/bz4233-r3a-iframeviewer-livepreview`, adopting `IframeViewer` on `LivePreview.svelte`). Yama's `CHANGES_REQUESTED` review on that PR flagged a MAJOR finding: `IframeViewer` exposed no ref to its internal `<iframe>`, so the consumer had to fall back to a parent `$effect` + DOM query (`getRenderedIframe`) to (a) set `credentialless` and (b) send outbound `postMessage`. For (a) specifically, Yama's finding was that this DOM query ran *after* `IframeViewer` already created the `<iframe>` (its `src` is only truthy once the effect runs), relying on implicit Svelte 5 effect-batching order for a security-relevant isolation attribute — fragile, not a single-statement guarantee.

This PR fixes both gaps upstream, generically, minimal and structural (not cosmetic):

- **`credentialless` boolean prop**, spread onto the `<iframe>` in the exact same `{#if src}` render statement that sets `src`. This eliminates the effect-scheduling dependency entirely — the attribute is guaranteed present at iframe creation, not applied after the fact.
- **`postMessage(message, targetOrigin)`** exported from the component instance, reached via `bind:this` — the outbound counterpart to the existing `onMessage` inbound callback.

**Note:** lighthouse #6699's `getRenderedIframe` DOM-query escape hatch gets deleted in a follow-up once this releases.

## Design notes

- **`postMessage` follows `FileInput`'s exact precedent**: `export const openFilePicker` (`FileInput.svelte`) is the house pattern for exposing imperative element access — an exported const method reached via `bind:this`. `IframeViewer` has no snippet-prop channel (unlike `FileInput`'s `trigger` snippet, which also forwards `openFilePicker` as a snippet param), so the `bind:this` route is the only transferable piece of that precedent, and it's applied as-is.
- **Not bridged to the Web Component.** Verified via `FileInput.wc.svelte`: it consumes `openFilePicker` internally (through the snippet param) for its default slot button, but never re-exposes it as a method on `<sui-file-input>`. `postMessage` follows the same rule — Svelte-only. WC consumers already have a standard workaround (query the shadow-DOM `<iframe>` and call `contentWindow.postMessage` directly), documented in `docs/IframeViewer.md`.
- `credentialless` uses a conditional attribute spread (`{...credentialless ? { credentialless: '' } : {}}`) rather than relying on Svelte's boolean-attribute attribute-list heuristic, since `credentialless` is a newer, non-standard iframe attribute that may not be in the compiler's internal known-boolean-attributes list — this guarantees correct presence/absence semantics regardless.

## Prop/method contract

| Name | Kind | Type | Description |
|---|---|---|---|
| `credentialless` | prop | `boolean` | Sets the iframe's `credentialless` attribute, same render statement as `src`. |
| `postMessage` | exported instance method (`bind:this`) | `(message: unknown, targetOrigin: string) => void` | Posts `message` to the embedded iframe's `contentWindow`. No-op before `src` mounts the iframe or after unmount. |

## Registration surfaces

- [x] `src/lib/IframeViewer/properties.ts` — `credentialless` added to `OptionalIframeViewerProperties`
- [x] `src/lib/IframeViewer/IframeViewer.svelte` — prop spread + exported `postMessage`
- [x] `src/wc/components/IframeViewer.wc.svelte` — `credentialless: { type: 'Boolean', reflect: true, attribute: 'credentialless' }` added to the WC props map (`postMessage` deliberately not bridged)
- [x] `docs/IframeViewer.md` — Props table row, new "Instance Methods" section, WC section updated with the `credentialless` attribute + an explicit note on why `postMessage` isn't exposed there
- [x] `docs/_index.json` — description updated to mention both additions
- [x] `src/routes/components/iframe-viewer/+page.svelte` — new demo: `credentialless` iframe + a button that calls `postMessage` and the embedded document echoes the received message back

## Gates

- `lint` — 0 errors (3 pre-existing warnings in an unrelated test file)
- `check` (svelte-check) — 554 files, 0 errors, 4 pre-existing warnings in unrelated files
- `build` — passes
- `test:unit` — 128/128 passed
- `test:integration` (Playwright) — 140 passed, 9 pre-existing failures, all in `Table` component tests (cell renderers / header tooltip / pagination), unrelated to this change — same failures reproduced on a clean `origin/release` checkout with no changes applied (confirmed baseline, not a regression).

### DIFF
--- a/src/lib/IframeViewer/IframeViewer.svelte
+++ b/src/lib/IframeViewer/IframeViewer.svelte
@@ -10,12 +10,17 @@
     sandbox,
     loading,
     referrerpolicy,
+    credentialless,
     testId,
     classes,
     onMessage = () => {}
   }: IframeViewerProperties = $props();
 
   let iframeEl: HTMLIFrameElement | null = $state(null);
+
+  export const postMessage = (message: unknown, targetOrigin: string): void => {
+    iframeEl?.contentWindow?.postMessage(message, targetOrigin);
+  };
 
   const messageHandler = (event: MessageEvent): void => {
     // Secure by default: forward a message only when its origin is allow-listed AND it
@@ -44,7 +49,15 @@
 
 <div class="iframe-viewer {classes ?? ''}" data-pw={testId} testID={testId}>
   {#if src}
-    <iframe bind:this={iframeEl} {src} {title} {allow} {sandbox} {loading} {referrerpolicy}
+    <iframe
+      bind:this={iframeEl}
+      {src}
+      {title}
+      {allow}
+      {sandbox}
+      {loading}
+      {referrerpolicy}
+      {...credentialless ? { credentialless: '' } : {}}
     ></iframe>
   {/if}
 </div>
--- a/docs/IframeViewer.md
+++ b/docs/IframeViewer.md
@@ -26,6 +26,7 @@ A security-conscious iframe embed. It renders an `<iframe>` for the given `src` 
 | allowedOrigins | `string[]`          | No       | `[]`                 | Origins allowed to send `postMessage` events to `onMessage`. An empty array (the default) processes nothing — secure by default. |
 | allow          | `string`            | No       | `'fullscreen'`       | Permissions policy applied to the iframe `allow` attribute.                                                                      |
 | sandbox        | `string`            | No       | `-`                  | Value for the iframe `sandbox` attribute. Omitted when not set.                                                                  |
+| credentialless | `boolean`           | No       | `-`                  | Sets the iframe's `credentialless` attribute. Applied in the same render statement as `src`, so it is present before the iframe's first load — no dependency on effect-scheduling order. |
 | loading        | `'eager' \| 'lazy'` | No       | `-`                  | Loading strategy for the iframe. Omitted when not set.                                                                           |
 | referrerpolicy | `ReferrerPolicy`    | No       | `-`                  | Referrer policy for the iframe. Omitted when not set.                                                                            |
 | testId         | `string`            | No       | `-`                  | Test selector applied as `data-pw` on the container.                                                                             |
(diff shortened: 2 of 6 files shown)


src/lib/IframeViewer/IframeViewer.svelte
```diff
@@ -10,13 +10,18 @@
...
     testId,
     classes,
     onMessage = () => {}
   }: IframeViewerProperties = $props();
 
   let iframeEl: HTMLIFrameElement | null = $state(null);
 
+  export const postMessage = (message: unknown, targetOrigin: string): void => {
+    iframeEl?.contentWindow?.postMessage(message, targetOrigin);
+  };
```
Tara-ag, line 23: 💡 MINOR: New `credentialless` prop and `postMessage` instance method are not covered by tests. Please add Vitest unit tests verifying (1) the `credentialless` attribute is rendered when the prop is `true` and omitted when `false`/`undefined`, and (2) `postMessage` forwards to `iframeEl.contentWindow.postMessage` when the iframe is mounted and is a no-op when `iframeEl` is null.

src/lib/IframeViewer/IframeViewer.svelte
```diff
@@ -10,13 +10,18 @@
...
     referrerpolicy,
+    credentialless,
     testId,
     classes,
     onMessage = () => {}
   }: IframeViewerProperties = $props();
 
   let iframeEl: HTMLIFrameElement | null = $state(null);
 
+  export const postMessage = (message: unknown, targetOrigin: string): void => {
```
Tara-ag, line 21: 💬 SUGGESTION: Consider adding a runtime guard for `targetOrigin` in `postMessage` to avoid accidental `'*'` broadcasts. For example, warn or throw when `targetOrigin === '*'` and the caller is not explicitly opting into wildcard delivery. This aligns with the component's security-first posture.

Tara-ag (commented):
## Review Summary

Reviewed 6 changed files. The PR is a clean, additive change: it adds a new optional `credentialless` prop and a Svelte-only `postMessage` instance method to `IframeViewer`, with matching docs and web-component prop registration. No breaking changes to the public API, exports, or CSS custom properties.

### New issues raised

| Severity | Count | Files |
|----------|-------|-------|
| 🔒 CRITICAL | 0 | — |
| ⚠️ MAJOR | 0 | — |
| 💡 MINOR | 1 | `src/lib/IframeViewer/IframeViewer.svelte` |
| 💬 SUGGESTION | 1 | `src/lib/IframeViewer/IframeViewer.svelte` |

### Notes

- **MINOR**: New `credentialless` prop and `postMessage` instance method lack unit-test coverage. Please add Vitest tests before merging.
- **SUGGESTION**: Consider a runtime guard against `targetOrigin === '*'` in `postMessage` to reinforce the component's security-first design.

No blocking criteria are met. Approving is conditional on adding tests for the new behavior.

PR #428, opened 2026-08-04
feat: add showIndicator prop and enhance indicator styling

adarshpandey-cs21:
- add showIndicator prop and enhance indicator styling

## Summary by CodeRabbit

* **New Features**
  * Choiceboxes now display customizable radio or checkbox selection indicators by default, with an option to hide them.
  * Added styling support for indicator size, shape, selected states, and checkmark presentation.
  * Button text decoration and line height can now be customized, including separate disabled-state decoration.

* **Documentation**
  * Documented the Choicebox indicator option, accessibility behavior, and styling variables.


murdore:
Verified against source while reviewing a related PR (#473/#474, the polymorph gap-closure work): `showIndicator` is accepted and used by `Choicebox.svelte`/`properties.ts`, but `src/wc/components/Choicebox.wc.svelte`'s `customElement.props` block only whitelists `selected`, `mode`, `disabled`, `testId`, `classes`, `onclick` — `showIndicator` isn't in that list, so it won't reach the component when `<sui-choicebox>` is used as a custom element (only the Svelte-import path gets it).

Not something I can push here since this is your branch, not mine — flagging so it can be added to the `props` block alongside the others before merge.

PR #430, opened 2026-08-04
fix(Modal): only treat header.leftImage as a control when a click handler is supplied

murdore:
Modal wrapped **any** non-empty `header.leftImage` in `role="button"` with `tabindex="0"` and click/keydown handlers, regardless of whether `onheaderLeftImageClick` was passed.

Consumers use that slot for a decorative brand or source logo far more often than for a back button. So assistive tech announced a focusable **button that does nothing when activated**, and keyboard users hit a dead tab stop on every such modal. That is worse than leaving the image unannounced — it promises an action the control does not have.

## The change

The wrapper is interactive only when `onheaderLeftImageClick` is a function.

```svelte
{#if typeof onheaderLeftImageClick === "function"}
  <div onclick={handleLeftImageClick} onkeydown={handleLeftImageKeyDown}
       role="button" tabindex="0" aria-label={leftImageAriaLabel ?? null} ...>
    {@render leftImage()}
  </div>
{:else}
  <div data-pw={leftImageTestId} testID={leftImageTestId}>
    {@render leftImage()}
  </div>
{/if}
```

Three deliberate implementation choices:

- **Two explicit branches, not computed attributes.** My first attempt used ternaries (`role={isInteractive ? "button" : undefined}`). That tripped `a11y_no_noninteractive_tabindex` — svelte-check cannot statically pair a dynamic `role` with a dynamic `tabindex` — and it violated this repo`s own `no-restricted-syntax` rule banning `undefined` (4 errors). Writing the branches out keeps the pairing statically checkable and needs no suppression.
- **Shared `{#snippet leftImage()}`** so the `Img` markup is not duplicated.
- **`{@const leftImageSrc = header.leftImage}`** because a snippet is its own scope and loses the outer `typeof header.leftImage === "string"` narrowing, which svelte-check flagged as a real type error.

`data-pw` / `testID` are preserved on both branches, so existing selectors keep working.

## Backward compatibility

The only behavioural difference is for call sites passing `leftImage` **without** a handler — where the control did nothing anyway. Sites that pass a handler are untouched.

That is asserted, not assumed: the new spec`s back-button case **passes both before and after** the change.

## Tests

New `tests/modal-left-image-interactivity.spec.ts` covers both states, driven by two new demo modals on `/components/modal` (the demo route previously had no `leftImage` case at all):

1. **decorative** (no handler) — not `role="button"`, no `tabindex`, no `aria-label`, and the image still renders.
2. **back button** (handler wired) — keeps `role="button"`, `tabindex="0"`, `aria-label`, and stays operable by click, `Enter` **and** `Space`, asserted via a visible click counter.

**Negative control performed.** With the fix reverted, the decorative test fails on the `role="button"` assertion while the back-button test still passes — so the spec genuinely detects the regression it exists to catch, and it is not accidentally green.

## Gates

- `pnpm run check` (svelte-check): **563 files, 0 errors**. The 4 remaining warnings are pre-existing and unrelated.
- `pnpm exec eslint src/lib/Modal/Modal.svelte`: clean. (3 repo-wide errors in `mcp/src/services/registry.ts` are pre-existing on `release` — confirmed by stashing this change and re-running.)
- `prettier --check`: clean.
- Playwright: 4 passed, including both pre-existing `modal-viewport-containment` specs — no regression.

## Downstream

Unblocks 7 call sites in Lighthouse (BZ-4233) that pass a decorative logo with no handler: `SettingsView`, `ai/settings/sources/data` (×2), `LoginModal`, `ai/anomaly`, and others. They need **no consumer-side change** once this lands — the wrapper simply stops being a control.

## Follow-up, not done here

`header.rightImage` has the identical unconditional wrapper and `onheaderRightImageClick` is also optional, so the same dead-control case is possible there. I did not change it: the close control is far more widely used, and a consumer census is complicated by props being spread through wrapper components, so it needs its own per-site tracing rather than a symmetric guess.

### DIFF
--- a/src/lib/Modal/Modal.svelte
+++ b/src/lib/Modal/Modal.svelte
@@ -158,25 +158,35 @@
           {#if (typeof header?.leftImage === 'string' && header.leftImage.length > 0) || (typeof header?.text === 'string' && header.text.length > 0) || (typeof header?.rightImage === 'string' && header.rightImage.length > 0)}
             <div class="header">
               {#if typeof header.leftImage === 'string' && header.leftImage.length > 0}
-                <div
-                  onclick={handleLeftImageClick}
-                  onkeydown={handleLeftImageKeyDown}
-                  role="button"
-                  tabindex="0"
-                  aria-label={leftImageAriaLabel ?? null}
-                  data-pw={leftImageTestId}
-                  testID={leftImageTestId}
-                >
+                {@const leftImageSrc = header.leftImage}
+                {#snippet leftImage()}
                   <!-- Inline SVGs so currentColor icons inherit the header text
                        colour; non-SVG URLs fall back to a plain <img>. -->
-                  <Img
-                    inlineSvg
-                    src={header.leftImage}
-                    alt=""
-                    fallback=""
-                    classes="header-left-img"
-                  />
-                </div>
+                  <Img inlineSvg src={leftImageSrc} alt="" fallback="" classes="header-left-img" />
+                {/snippet}
+                <!-- The left image is only a control when a click handler is actually supplied.
+                     Consumers pass a decorative brand/source logo here far more often than a back
+                     button, and announcing those as a focusable "button" that does nothing when
+                     activated is worse for assistive tech than leaving them unannounced. The two
+                     branches are written out rather than computed so the role/tabindex pairing
+                     stays statically checkable. -->
+                {#if typeof onheaderLeftImageClick === 'function'}
+                  <div
+                    onclick={handleLeftImageClick}
+                    onkeydown={handleLeftImageKeyDown}
+                    role="button"
+                    tabindex="0"
+                    aria-label={leftImageAriaLabel ?? null}
+                    data-pw={leftImageTestId}
+                    testID={leftImageTestId}
+                  >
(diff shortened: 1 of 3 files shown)


murdore:
## The failing check is not a code verdict

`Yama PR Review` has now failed **twice in a row** (runs `30916765194` job 92016760090, then re-run job 92059168784) with the same GitHub annotation both times:

> Yama could not complete the review (infrastructure error, not a code verdict). Re-run this job to retry.

In both runs the `Yama Review` step itself **succeeded** and only `Enforce verdict` failed, with no review output produced — so there is no finding to address. Re-running per the annotation's own instruction did not clear it, which makes this a persistent infrastructure problem rather than a transient one.

Every other check passes: `build`, `CodeQL`, `Analyze (actions)`, `Analyze (javascript-typescript)`, both SentinelOne CNS scans. The PR is `MERGEABLE`.

This is blocking 7 consumer call sites in Lighthouse (decorative `header.leftImage` values that the library currently wraps in `role="button"` with no handler, so screen readers announce a focusable control that does nothing). Flagging rather than merging past a required check — someone with the Yama pipeline context should either fix the job or make the call to override.

PR #433, opened 2026-08-04
feat(Modal): add entryAnimation and overlayFadeIn props for centered slide-up dialogs

murdore:
## Summary

`ModalAnimation.svelte` picks its entry transition purely from the modal's `align` prop: `fly` for `top`/`bottom`, `fade` for `center`. There's no supported way to get a centered dialog that slides in like a bottom sheet — a common pattern (mobile-style centered sheets, confirmation dialogs that want a softer entrance than an instant fade) that currently requires reimplementing `ModalAnimation` from scratch to reach. This PR adds an optional `entryAnimation` override that unlocks it without touching any default behavior.

## Fix / design

Added `entryAnimation?: 'fade' | 'slide-up' | 'slide-down'` to `ModalAnimation` (and threaded through `Modal`'s public props):

```svelte
let flyAnimationProperties = $derived.by(() => {
  const base = { x: 0, y: 0, duration: 380 };

  if (entryAnimation === 'slide-up') {
    return { ...base, y: 300 };   // same distance as align="bottom"
  }
  if (entryAnimation === 'slide-down') {
    return { ...base, y: -30 };   // same distance as align="top"
  }

  switch (align) {
    case 'top': return { ...base, y: -30 };
    case 'bottom': return { ...base, y: 300 };
    default: return base;
  }
});

let useFlyAnimation = $derived(
  entryAnimation != null ? entryAnimation !== 'fade' : align === 'top' || align === 'bottom'
);
```

- `'slide-up'` and `'slide-down'` reuse the exact `y` distance (and the shared 380ms duration) that `bottom`/`top` alignment already use — no new animation constants, no visual mismatch between an aligned dialog and an animation-overridden one.
- `'fade'` explicitly forces the fade transition regardless of `align`, for symmetry (e.g. a bottom-aligned modal that should fade instead of fly).
- Left **unset** (the default), `entryAnimation` has zero effect — the original `switch(align)` logic drives the transition exactly as before. Zero default behavior change; confirmed by the existing `tests/modal-viewport-containment.spec.ts` suite (both tests still pass unmodified) plus a full integration run (see Gates).

### Why a separate `entryAnimation` axis instead of adding more `ModalAlign` values

`align` is a *position* concept (where the dialog sits) and `entryAnimation` is a *motion* concept (how it gets there). Folding "slides up but stays centered" into `align` would mean inventing values like `'center-slide-up'`, multiplying combinatorially with every future align/motion pairing and duplicating the positioning CSS for each one. Keeping the two axes orthogonal means any current or future `align` value can pair with any `entryAnimation` value for free.

### `overlayFadeIn` — opt-in overlay fade-in

Also added `overlayFadeIn` (threaded to a new `fadeIn` prop on `OverlayAnimation`, default `false`). A slide-up dialog whose backdrop still appears instantly reads as unfinished — the two are meant to be paired. `OverlayAnimation` only had an `out:fade` before; `fadeIn` adds the matching `in:fade` at the same 350ms duration:

```svelte
{#if fadeIn}
  <div in:fade={{ duration: 350 }} out:fade={{ duration: 350 }} ...>
    {@render children?.()}
  </div>
{:else}
  <div out:fade={{ duration: 350 }} ...>
    {@render children?.()}
  </div>
{/if}
```

(Duplicated-branch `{#if}/{:else}` rather than a ternary inside the transition directive — a ternary would still apply *some* `in:` transition even when "disabled." This mirrors the pattern `ModalAnimation.svelte` itself already uses for its own conditional transitions.)

I confirmed via `grep -rl "OverlayAnimation" src --include="*.svelte"` that `Modal.svelte` is `OverlayAnimation`'s only consumer in the repo, so threading `fadeIn` through as `overlayFadeIn` on `Modal`'s public props is what makes it reachable at all — left un-threaded it would be dead code. Default `false` preserves the current instant-appear overlay behavior.

## Props contract

| Prop | Component | Type | Default | Description |
|---|---|---|---|---|
| `entryAnimation` | `ModalAnimation`, `Modal` | `'fade' \| 'slide-up' \| 'slide-down'` | `-` (unset) | Overrides the `align`-based transition choice. `slide-up`/`slide-down` reuse bottom/top's fly distance and duration regardless of `align`. Unset keeps the existing per-align default. |
| `fadeIn` / `overlayFadeIn` | `OverlayAnimation` / `Modal` | `boolean` | `false` | Fades the overlay backdrop in on mount (350ms, matching the existing fade-out). Default preserves the current instant-appear behavior. |

## Usage

```svelte
<Modal
  size="fit-content"
  align="center"
  entryAnimation="slide-up"
  overlayFadeIn
  showOverlay
  header={{ text: 'Slides Up, Stays Centered' }}
  footer={{ primaryButton: { text: 'Got it' } }}
  ...
>
  {#snippet content()}
    <p>Centered position, bottom-sheet-style entrance.</p>
  {/snippet}
</Modal>
```

## Registration surfaces

- [x] `src/lib/Animations/ModalAnimation.svelte` — added `entryAnimation` prop, early-return branches in `flyAnimationProperties`, updated `useFlyAnimation` derivation
- [x] `src/lib/Animations/OverlayAnimation.svelte` — added `fadeIn` prop with the `{#if}/{:else}` duplicated-transition pattern
- [x] `src/lib/Modal/properties.ts` — added `ModalEntryAnimation` type, `entryAnimation` and `overlayFadeIn` fields on `OptionalModalProperties`
- [x] `src/lib/Modal/Modal.svelte` — threaded both props to `ModalAnimation`/`OverlayAnimation`
- [x] `src/wc/components/Modal.wc.svelte` — registered `entry-animation` (String) and `overlay-fade-in` (Boolean) custom-element attributes
- [x] `docs/Modal.md` — added both Props rows and a `ModalEntryAnimation` Type Reference entry
- [x] `docs/ModalAnimation.md` — updated description, added `entryAnimation` Props row
- [x] `docs/OverlayAnimation.md` — updated description, added a `fadeIn` usage example and a new Props section
- [x] `docs/_index.json` — updated the `Modal`, `ModalAnimation`, and `OverlayAnimation` short descriptions
- [x] `src/routes/components/modal/+page.svelte` — added an "Open centered modal (slide-up)" demo combining `align="center"`, `entryAnimation="slide-up"`, and `overlayFadeIn`

## Gates

- `pnpm run lint` — pass (3 pre-existing warnings in `src/lib/Table/cell-data.test.ts`, unrelated to this change, 0 errors)
- `pnpm run check` — pass (0 errors; 4 pre-existing warnings in `Modal.svelte`/`ThemeSwitcher.svelte`/`tsconfig.json`, unrelated — the `Modal.svelte` one is the pre-existing `debounceTime` closure warning, not something this PR touches)
- `pnpm run test:unit` — pass (128/128)
- `pnpm run build` — pass
- `pnpm run test:integration` — 140 passed, 9 failed. The 9 failures are a pre-existing baseline unrelated to this change, all in `tests/table-builtin-cells.test.ts` (7), `tests/table-header-meta.test.ts` (1), and `tests/table-pagination-selection.test.ts` (1) — none touch Modal/ModalAnimation/OverlayAnimation. Both `tests/modal-viewport-containment.spec.ts` tests pass unmodified, confirming the default (no-`entryAnimation`) path is unchanged. **Verified by `git stash`:** re-ran the same three spec files against the unmodified tree (this change stashed out) and got the identical 9 failures by name and location, then `git stash pop`'d this change back in.

## Summary by CodeRabbit

* **New Features**
  * Modal component now supports customizable entry animations with fade, slide-up, and slide-down options to override default behavior.
  * Modal overlay now supports animated fade-in effect on initial appearance.

* **Documentation**
  * Updated Modal component documentation with new animation customization features and usage examples.

### DIFF
--- a/docs/Modal.md
+++ b/docs/Modal.md
@@ -22,6 +22,7 @@ A dialog overlay component that renders on top of the page with configurable siz
 | supportHardwareBackPress | `boolean`                                                                                                                   | No       | `false`         | When true, pushes a history state on mount so that pressing the device back button triggers onclose instead of navigating away. Cleans up on destroy.                                                        |
 | enableTransition         | `boolean`                                                                                                                   | No       | `true`          | When true, the modal content animates in/out using fly or fade transitions via ModalAnimation.                                                                                                               |
 | transitionType           | `ModalTransition = 'IN' \| 'ALL'`                                                                                           | No       | `'ALL'`         | Controls transition behavior. 'ALL' animates both in and out transitions. 'IN' only animates the in-transition (content disappears instantly on close).                                                      |
+| entryAnimation           | `ModalEntryAnimation = 'fade' \| 'slide-up' \| 'slide-down'`                                                                | No       | `-`             | Overrides the default per-align entry transition. 'slide-up'/'slide-down' reuse the same fly distance/duration as bottom/top alignment — e.g. pair with `align="center"` for a bottom-sheet-style slide-up dialog. Unset keeps the existing per-align default (fly for top/bottom, fade for center).            |
 | header                   | `{     leftImage?: string;     rightImage?: string;     text?: string;     testId?: string;     buttonTestId?: string;     buttonAriaLabel?: string;   }` | No       | `{}`            | Object configuring the modal header bar. leftImage: URL for left icon (e.g., back arrow); rightImage: URL for right icon (e.g., close button); text: header title text; testId/buttonTestId: test selectors; buttonAriaLabel: accessible name (rendered as `aria-label`) for the right image's `role="button"` wrapper — required for screen readers since the wrapper carries no visible text. |
 | footer                   | `{     primaryButton?: ButtonProperties;     secondaryButton?: ButtonProperties;   }`                                       | No       | `-`             | Object configuring footer action buttons. primaryButton: ButtonProperties for the main action button; secondaryButton: ButtonProperties for the alternate action button.                                     |
 | debounceTime             | `number`                                                                                                                    | No       | `700`           | Debounce delay in milliseconds for overlay click handling. Prevents rapid repeated overlay dismissals.                                                                                                       |
@@ -30,6 +31,7 @@ A dialog overlay component that renders on top of the page with configurable siz
 | testId                   | `string`                                                                                                                    | No       | `-`             | Value for data-pw on the modal overlay container.                                                                                                                                                            |
 | classes                  | `string`                                                                                                                    | No       | `-`             | CSS class string applied to the component's top-level element. Useful for theming — define classes with CSS variable overrides and pass them to create variant styles.                                       |
 | overlayBackdropFilter    | `string`                                                                                                                    | No       | `-`             | CSS `backdrop-filter` value applied to the overlay (e.g. `"blur(6px)"`). Sets `--modal-overlay-backdrop-filter` inline on the overlay div. Default: `none` (no blur).                                        |
+| overlayFadeIn            | `boolean`                                                                                                                   | No       | `false`         | When true, the overlay backdrop fades in on mount instead of appearing instantly. Pairs well with `entryAnimation="slide-up"` for a softer combined entrance.                                                |
 | usePortal                | `boolean`                                                                                                                   | No       | `false`         | When true, mounts the modal overlay at `document.body` so it escapes any ancestor clipping or stacking contexts. Useful inside `overflow: hidden` or `transform` containers.                                 |
 
 ## Snippets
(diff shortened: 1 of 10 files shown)


murdore:
Rebased onto `release` @ 2.114.1 to resolve the conflict with aadd6b6 (Modal leftImage-control fix) — one trivial union in the demo page's state declarations; the Modal.svelte changes merged cleanly alongside the leftImage fix. Full gates re-run on the rebased commit: svelte-check 0 errors, lint clean, unit 128/128, integration 142 passed with the identical 9 pre-existing Table-only failures (baseline unchanged; the 2 extra passing tests are 2.114.1's own additions). Mergeable again.

PR #438, opened 2026-08-05
fix(Progress): progressbar ARIA

murdore:
## Summary

`Progress` rendered as a bare `<div>` with no ARIA whatsoever — a screen reader had no way to know the element was a progress indicator, let alone read its current completion or whether it's in the indeterminate (unknown-duration) state. This wires up the standard WAI-ARIA progressbar pattern: `role="progressbar"` plus `aria-valuenow`, `aria-valuemin`, and `aria-valuemax` on the root element.

## Design: matching Gauge's existing convention, plus indeterminate handling it doesn't need

- `Gauge.svelte` already ships `role="progressbar"` with `aria-valuenow`/`aria-valuemin`/`aria-valuemax` in this repo, expressed as a normalized 0-100 percentage rather than the raw `value`/`max` domain. This PR mirrors that exactly for consistency across the library's two percentage-completion components: `aria-valuemin={0}`, `aria-valuemax={100}`, `aria-valuenow={Math.round(percentage)}` — the same `percentage` derivation and `Math.round` already used for the visible `showLabel` text, so the announced number never disagrees with what's on screen.
- Indeterminate mode (`value < 0`) needed handling `Gauge` doesn't have to deal with. Per the WAI-ARIA progressbar spec, `aria-valuenow` SHOULD be omitted when progress isn't currently determinable — so it's now `null` (omitted from the DOM) whenever `isIndeterminate` is true, rather than reporting a stale or misleading number. The same spec recommends `aria-valuetext` describing the indeterminate state when `aria-valuenow` is omitted, so `aria-valuetext="indeterminate"` is set in that branch.
- Deliberate scope boundary: this does **not** add an `ariaLabel`/`aria-label` prop the way `Gauge.svelte` does (`ariaLabel ?? labelText` fallback). That's a separate, larger scope decision — a new public prop, JSDoc, docs Props row, and demo usage — than "wire the three value attributes" this fix targets, so it's disclosed here rather than folded in silently. Without it, assistive technology announces the role and value (e.g. "45%, progress bar") but not what the progress bar measures (e.g. "Upload progress"). A natural, small follow-up if this repo wants full parity with `Gauge`.

## CSS variable contract (new)

None. This PR only adds ARIA attributes — no new CSS custom properties, no visual changes.

## Usage

No API change — this is an accessibility fix with no new props, events, or variables. Every existing usage benefits automatically.

```svelte
<!-- No markup change required: -->
<Progress value={65} max={100} showLabel />
<!-- now also exposes role="progressbar" aria-valuenow="65" aria-valuemin="0" aria-valuemax="100" -->

<Progress value={-1} />
<!-- indeterminate: role="progressbar" aria-valuetext="indeterminate", aria-valuenow omitted -->
```

## Registration surfaces

- [x] `src/lib/Progress/Progress.svelte` — `role`, `aria-valuenow`/`aria-valuemin`/`aria-valuemax`/`aria-valuetext` wired on the root element
- [x] `docs/Progress.md` — new "Accessibility" section (placed between Props and CSS Variables, matching the `Badge.md` precedent), explaining the percentage convention, indeterminate gating, and the disclosed `ariaLabel` gap
- [x] `src/routes/components/progress/+page.svelte` — added `testId` to both demo instances (determinate + indeterminate) so they're reliably selectable in tests; no visual/behavioral change
- [x] `tests/progress-aria.spec.ts` — new Playwright spec, 4 tests
- [x] No `properties.ts` change — no new props, purely a rendering fix on existing ones

## Gates (re-run in full on commit 3c6bfbe)

- `pnpm run check` — 0 errors, 4 warnings (baseline: 2 `state_referenced_locally` in Modal.svelte/ThemeSwitcher.svelte, 2 tsconfig warnings)
- `pnpm run lint` — 0 errors, 3 warnings (baseline: unused eslint-disable directives in `src/lib/Table/cell-data.test.ts`)
- `pnpm run test:unit` — 128/128 passed
- `pnpm run test:integration` — 146 passed, 9 failed (baseline-only: pre-existing Table failures in `table-builtin-cells.test.ts`, `table-header-meta.test.ts`, `table-pagination-selection.test.ts` — identical baseline to PR #437). The 4 new specs in `progress-aria.spec.ts` are part of that 146 and cover: role + initial value attributes on the determinate demo, `aria-valuenow` tracking the visible label text across `+10`/`-10` clicks, the indeterminate instance's `aria-valuenow` omission plus `aria-valuetext`, and both instances being discoverable via `page.getByRole('progressbar')`. Zero Progress-related failures.

## Summary by CodeRabbit

- **New Features**
  - Added accessible labels for progress indicators, with sensible fallback text.
  - Progress bars now expose appropriate ARIA semantics for determinate and indeterminate states.
  - Determinate progress reports rounded values and accurate minimum/maximum bounds.
  - Added a fractional progress example demonstrating `1/3` completion.

- **Documentation**
  - Documented the new accessibility property, ARIA behavior, value rounding, and label fallback rules.

- **Tests**
  - Added accessibility coverage for roles, labels, values, state changes, and indeterminate behavior.

### DIFF
--- a/src/lib/Progress/Progress.svelte
+++ b/src/lib/Progress/Progress.svelte
@@ -1,16 +1,54 @@
 <script lang="ts">
   import type { ProgressProperties } from './properties';
 
-  let { value, max = 100, showLabel = false, testId, classes }: ProgressProperties = $props();
+  let {
+    value,
+    max = 100,
+    showLabel = false,
+    ariaLabel,
+    testId,
+    classes
+  }: ProgressProperties = $props();
 
-  let percentage = $derived(Math.min(100, Math.max(0, (value / max) * 100)));
+  // `max` has to be a finite positive number for a percentage to mean
+  // anything. A zero, negative, or non-finite max (an as-yet-unloaded total,
+  // say) would make (value / max) NaN or Infinity and put "NaN" into
+  // aria-valuenow, which is not a valid ARIA value -- and a non-finite value
+  // has the same effect. Treat an unusable range as an empty (0%) bar
+  // instead. `value` is guarded here too, not just `max`: a NaN value isn't
+  // `< 0`, so it would otherwise slip past the isIndeterminate check below
+  // and hit the same NaN-percentage path. This check is deliberately kept
+  // independent of `isIndeterminate` below -- a negative `value` (e.g. -1)
+  // still signals indeterminate on its own, even paired with an invalid
+  // `max`, since indeterminate mode never reads `percentage` in the markup.
+  let hasValidRange = $derived(Number.isFinite(value) && Number.isFinite(max) && max > 0);
+  let percentage = $derived(hasValidRange ? Math.min(100, Math.max(0, (value / max) * 100)) : 0);
   let isIndeterminate = $derived(value < 0);
+  // Rounded to 2 decimal places rather than the nearest whole number, so
+  // aria-valuenow stays effectively in step with the bar's raw fractional
+  // width (matching Gauge's convention of tying aria-valuenow to the raw
+  // computed value, not the rounded display text). Full raw precision isn't
+  // used because (value / max) * 100 can land on binary floating-point noise
+  // like 55.00000000000001 -- rounding to 2 decimals clears that while still
+  // keeping the value far closer to the bar's true width than whole numbers.
+  let preciseValueNow = $derived(Math.round(percentage * 100) / 100);
+  // Shared by the visible label and the aria-label fallback, mirroring
+  // Gauge's labelText convention. "Loading" matches the aria-label
+  // LoadingDots already uses for its own indeterminate role="status" element.
+  let labelText = $derived(isIndeterminate ? 'Loading' : `${Math.round(percentage)}%`);
 </script>
 
 <div
   class="container {classes ?? ''}"
   data-pw={typeof testId === 'string' ? testId : null}
   testID={typeof testId === 'string' ? testId : null}
+  role="progressbar"
+  aria-valuenow={isIndeterminate ? null : preciseValueNow}
+  aria-valuemin={0}
+  aria-valuemax={100}
+  aria-valuetext={isIndeterminate ? 'indeterminate' : null}
+  aria-busy={isIndeterminate ? true : null}
+  aria-label={ariaLabel ?? labelText}
 >
   <div class="track">
     <div
@@ -20,7 +58,7 @@
     ></div>
   </div>
   {#if showLabel && !isIndeterminate}
-    <div class="label">{Math.round(percentage)}%</div>
+    <div class="label">{labelText}</div>
   {/if}
 </div>
 
--- a/docs/Progress.md
+++ b/docs/Progress.md
@@ -16,11 +16,20 @@ A linear progress bar showing task completion or usage. The `value` prop control
 
 | Prop      | Type      | Required | Default | Description                                                                                                                                                            |
 | --------- | --------- | -------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| value     | `number`  | Yes      | `-`     | Current progress value (0 to max). Values are clamped to the 0-max range. A negative value activates the indeterminate animation for unknown-duration tasks.           |
-| max       | `number`  | No       | `100`   | The maximum value representing 100% completion. The filled percentage is calculated as (value / max) \* 100.                                                           |
+| value     | `number`  | Yes      | `-`     | Current progress value (0 to max). Values are clamped to the 0-max range. A negative value activates the indeterminate animation for unknown-duration tasks. A non-finite value (e.g. `NaN`) is treated as an invalid range -- see `max` below.       |
+| max       | `number`  | No       | `100`   | The maximum value representing 100% completion. The filled percentage is calculated as (value / max) \* 100. `max` must be finite and positive; a zero, negative, or non-finite `max` (or a non-finite `value`) is an invalid range and renders the bar at 0% instead of `NaN`. |
 | showLabel | `boolean` | No       | `false` | Whether to display the rounded percentage text next to the progress bar. Hidden during indeterminate mode.                                                             |
+| ariaLabel | `string`  | No       | `-`     | Accessible name for the progress bar. Falls back to the same percentage text as `showLabel` (e.g. `"75%"`) when determinate, or `"Loading"` when indeterminate.        |
 | testId    | `string`  | No       | `-`     | Value for the data-pw attribute, used for end-to-end testing selectors.                                                                                                |
 | classes   | `string`  | No       | `-`     | CSS class string applied to the component's top-level element. Useful for theming — define classes with CSS variable overrides and pass them to create variant styles. |
+
+## Accessibility
+
+The root element sets `role="progressbar"` with `aria-valuemin="0"` and `aria-valuemax="100"` — these two are constant regardless of the raw `value`/`max` domain, so assistive technology always hears a plain 0-100 scale (matching the convention `Gauge` already uses). `aria-valuenow` itself is only well-defined for a *valid* range, i.e. a finite `value` and a finite, positive `max`; when it is, `aria-valuenow` is rounded to 2 decimal places rather than the nearest whole number, so it stays effectively in step with the bar's raw fractional width instead of drifting from what's visually shown (matching `Gauge`'s convention of tying `aria-valuenow` to the raw computed value rather than the rounded display text); the separate `showLabel` text still rounds to the nearest whole percent, since that's the right precision for a human-readable label. For an invalid range (zero, negative, or non-finite `max`, or a non-finite `value`) the component does not propagate `NaN` into `aria-valuenow` or the label -- it falls back to a 0% determinate bar (`aria-valuenow="0"`, label `"0%"`) instead.
+
+In indeterminate mode (`value < 0`) `aria-valuenow` is omitted entirely per the WAI-ARIA progressbar pattern, and `aria-valuetext="indeterminate"` plus `aria-busy="true"` are set instead, so assistive technology announces the unknown-duration state rather than reading it as stuck at a fixed value. `aria-valuemin`/`aria-valuemax` are kept in both modes — the 0-100 scale itself isn't unknown, only the current position within it is.
+
+Use the `ariaLabel` prop to give the bar an accessible name describing what it measures (e.g. `"File upload progress"`). When omitted, it falls back to the same percentage text `showLabel` displays (or `"Loading"` while indeterminate, matching the `aria-label` `LoadingDots` already uses for its own loading state), so assistive technology always announces a name even if you forget to set one explicitly.
 
 ## CSS Variables
 
--- a/src/lib/Progress/properties.ts
+++ b/src/lib/Progress/properties.ts
@@ -5,8 +5,23 @@ export type MandatoryProgressProperties = {
 };
 
 export type OptionalProgressProperties = {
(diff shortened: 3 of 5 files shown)


src/lib/Progress/Progress.svelte
```diff
@@ -11,6 +11,11 @@
   class="container {classes ?? ''}"
   data-pw={typeof testId === 'string' ? testId : null}
   testID={typeof testId === 'string' ? testId : null}
+  role="progressbar"
+  aria-valuenow={isIndeterminate ? null : Math.round(percentage)}
+  aria-valuemin={0}
+  aria-valuemax={100}
```
Tara-ag, line 48: ⚠️ MAJOR: `aria-valuenow` is omitted for indeterminate mode, but `aria-valuemin`/`aria-valuemax` remain present. Per the WAI-ARIA progressbar pattern, when `aria-valuenow` is omitted the value is considered indeterminate; keeping min/max is acceptable, but pairing it with `aria-valuetext='indeterminate'` is good. However, consider also adding `aria-busy="true"` in indeterminate mode so assistive technology conveys that the element is actively waiting on an incomplete operation, which is the stronger semantic signal for an indeterminate progressbar. Suggestion:
```svelte
aria-busy={isIndeterminate ? true : null}
```

src/lib/Progress/Progress.svelte
```diff
@@ -11,6 +11,11 @@
   class="container {classes ?? ''}"
   data-pw={typeof testId === 'string' ? testId : null}
   testID={typeof testId === 'string' ? testId : null}
+  role="progressbar"
```
Tara-ag, line 45: 💡 MINOR: `Progress` lacks an accessible name, unlike `Gauge.svelte` which falls back to `aria-label={ariaLabel ?? labelText}`. The PR description acknowledges this gap, but the component still ships without any labeling mechanism. Consider adding an `ariaLabel` prop to `ProgressProperties` so consumers can name the progressbar; without it, screen readers announce only "progress bar, 65%" with no context about what is progressing.

src/lib/Progress/Progress.svelte
```diff
@@ -11,6 +11,11 @@
...
+  aria-valuenow={isIndeterminate ? null : Math.round(percentage)}
```
Tara-ag: 💡 MINOR: `aria-valuenow` reports a rounded integer percentage while the visible bar width uses the unrounded `percentage`. For values like `value=1, max=3`, `percentage ≈ 33.33` renders a 33.33%-wide bar but announces `aria-valuenow="33"`. This is the same rounding used by `showLabel`, so it is internally consistent, but be aware that the accessible value can diverge slightly from the actual rendered proportion. If exact parity matters, round the bar width to the same integer or expose the raw percentage; otherwise this is acceptable.

Tara-ag (approved):
## Review Summary

Reviewed 4 changed files. No blocking issues found.

### New issues raised
- **1 MAJOR**: `Progress.svelte` — consider adding `aria-busy="true"` in indeterminate mode for a stronger active-operation semantic.
- **2 MINOR**: `Progress.svelte` — missing accessible-name prop (`ariaLabel`) compared to `Gauge`, and a note on rounding divergence between `aria-valuenow` and the rendered bar width.

### Why not blocking
- This is a pure accessibility fix: no props, events, exports, or CSS custom properties were changed or removed.
- No public API contract is broken; downstream consumers benefit automatically.
- No secrets, unsafe HTML, or SSR-unsafe patterns introduced.

### Notes
- Tests (`tests/progress-aria.spec.ts`) cover role, value tracking, indeterminate handling, and role-based discovery.
- Docs (`docs/Progress.md`) include an Accessibility section explaining the percentage convention and the disclosed `ariaLabel` gap.

Approving — the suggested improvements can be addressed in a follow-up if desired.

src/lib/Progress/Progress.svelte
murdore: Fixed as suggested — added `aria-busy={isIndeterminate ? true : null}` on the root element (Progress.svelte:38 on the amended commit), so the busy state is now conveyed through its own dedicated ARIA property rather than only through `aria-valuetext` prose.

Kept `aria-valuemin`/`aria-valuemax` present in both modes, per your note that this is acceptable under the WAI-ARIA progressbar pattern (the 0-100 scale itself isn't unknown while indeterminate, only the current position within it is) — no change there.

Also extended the Playwright spec (`indeterminate progress bar omits aria-valuenow and announces busy state`) to assert `aria-busy="true"` alongside the existing valuemin/valuemax/valuetext checks.

src/lib/Progress/Progress.svelte
murdore: Added — `ariaLabel?: string` on `ProgressProperties` (mirrors Gauge's JSDoc), destructured in the component, and wired up as `aria-label={ariaLabel ?? labelText}` on the root, the exact fallback pattern `Gauge` uses.

`labelText` is a new derived value shared between the visible `showLabel` text and the `aria-label` fallback: `` `${Math.round(percentage)}%` `` when determinate. Progress also has an indeterminate mode Gauge doesn't, so that needed its own fallback — it now falls back to `"Loading"` (matching the `aria-label="Loading"` `LoadingDots` already uses in this repo for its own loading state), so a consumer who forgets to set `ariaLabel` still gets a name in both modes, not just the determinate one.

Docs (`docs/Progress.md`) and the demo route got a matching touch: the props table has a new `ariaLabel` row, the Accessibility section explains the fallback behavior for both modes, and the determinate demo now passes `ariaLabel="File upload progress"` as a live example. New/updated Playwright assertions cover the custom-label case, the determinate fallback (percentage text), and the indeterminate fallback (`"Loading"`).

src/lib/Progress/Progress.svelte
murdore: Went with a third option beyond the two you suggested: rather than rounding the bar width to the integer or exposing the fully raw `percentage`, `aria-valuenow` now rounds to 2 decimal places (`Math.round(percentage * 100) / 100`).

Reasoning: exposing the fully raw value turns out to have its own problem in this codebase — `(value / max) * 100` lands on binary floating-point noise for ordinary inputs (e.g. `value=55, max=100` computes to `55.00000000000001` in JS, not a clean `55`), which would make `aria-valuenow` noisier for assistive tech than the current integer, not less, and would've broken the existing spec asserting `aria-valuenow="55"`. 2-decimal rounding avoids that noise entirely while keeping the announced value within 0.01 percentage points of the bar's true width — down from a potential full point under integer rounding.

Your own example is a good demonstration of the improvement: `value=1, max=3` now announces `aria-valuenow="33.33"` instead of `"33"`, much closer to the rendered 33.333...%-wide bar. Added a dedicated demo instance (`progress-fractional-demo`, `value=1 max=3`) plus a Playwright test (`aria-valuenow uses 2-decimal precision to stay in step with the bar width`) asserting exactly this. The `showLabel` text keeps whole-number rounding, since that's the right precision for a human-readable label — only the announced ARIA value changed precision.

Tara-ag (approved):
## Review summary

Reviewed the 5 changed files in PR #438. The PR adds proper ARIA semantics to `Progress`, matching the existing `Gauge` convention, and adds tests + docs.

### New issues raised this run: 0

No new blocking or non-blocking issues were found.

### Existing review threads
Three review comments were already posted by a prior reviewer; the author has addressed all of them in the current commit:
- `aria-busy="true"` added for indeterminate mode.
- `ariaLabel` prop added with `aria-label={ariaLabel ?? labelText}` fallback, mirroring `Gauge`.
- `aria-valuenow` precision changed to 2 decimal places to stay closer to the rendered bar width.

### Verification performed
- Confirmed `Progress.svelte` now exposes `role="progressbar"`, `aria-valuenow`, `aria-valuemin`, `aria-valuemax`, `aria-valuetext`, `aria-busy`, and `aria-label` correctly.
- Confirmed `properties.ts` adds the optional `ariaLabel?: string` prop (purely additive, no breaking change).
- Confirmed `src/lib/index.ts` already exports `Progress` and `ProgressProperties`; no export changes needed.
- Confirmed `src/wc/components/Progress.wc.svelte` does **not** declare `ariaLabel` in its custom-element prop mapping, so the new prop is not exposed on the web component (`<sui-progress>`). This is a public-API inconsistency worth noting, but it is not a regression (the prop simply isn't available in WC form yet). Suggest a follow-up to add `ariaLabel: { type: 'String', attribute: 'aria-label' }` to the WC wrapper.
- Confirmed docs and Playwright tests cover the new behavior.

### Decision
Approve — no blocking issues. The accessibility fix is correct, additive, and well-tested.

src/lib/Progress/Progress.svelte
```diff
@@ -1,16 +1,42 @@
...
   let percentage = $derived(Math.min(100, Math.max(0, (value / max) * 100)));
   let isIndeterminate = $derived(value < 0);
+  // Rounded to 2 decimal places rather than the nearest whole number, so
+  // aria-valuenow stays effectively in step with the bar's raw fractional
+  // width (matching Gauge's convention of tying aria-valuenow to the raw
+  // computed value, not the rounded display text). Full raw precision isn't
+  // used because (value / max) * 100 can land on binary floating-point noise
+  // like 55.00000000000001 -- rounding to 2 decimals clears that while still
+  // keeping the value far closer to the bar's true width than whole numbers.
+  let preciseValueNow = $derived(Math.round(percentage * 100) / 100);
```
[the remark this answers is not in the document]
murdore: Fixed as suggested.

`Progress.svelte:24-27` (amended commit `ebf57b9`): added a `hasValidRange` derived value — `Number.isFinite(value) && Number.isFinite(max) && max > 0` — and gated `percentage` on it, falling back to `0` for an invalid range instead of propagating `NaN`:

```js
let hasValidRange = $derived(Number.isFinite(value) && Number.isFinite(max) && max > 0);
let percentage = $derived(hasValidRange ? Math.min(100, Math.max(0, (value / max) * 100)) : 0);
let isIndeterminate = $derived(value < 0);
```

Chosen fallback: an invalid range (zero/negative/non-finite `max`, or a non-finite `value`) renders as an ordinary **0% determinate bar** — `aria-valuenow="0"`, bar width `0%`, label `"0%"` — rather than a distinct error state. `value={0} max={0}` now reads exactly as an empty progress bar would.

Also guarded `value`, not just `max`: a `NaN` value isn't `< 0`, so it would otherwise slip past `isIndeterminate` and hit the same `NaN` path. `hasValidRange` is deliberately independent of `isIndeterminate` though — a negative `value` (e.g. `-1`) still activates the indeterminate animation on its own even paired with an invalid `max`, since indeterminate mode never reads `percentage` in the markup. Verified this doesn't change any currently-valid input's behavior.

Regression tests added in `tests/progress-aria.spec.ts` (new `Progress invalid range fallback` describe block) covering: `value=0/max=0`, a negative `max`, a non-finite `max`, a non-finite `value`, and a negative `value` paired with an invalid `max` (confirms it's still indeterminate, not swallowed by the new guard).

Docs updated:
- `src/lib/Progress/properties.ts` — documented the invalid-range contract on the `max` and `ariaLabel` JSDoc.
- `docs/Progress.md#L28` — corrected the prior claim that the normalized ARIA value holds for every raw `value`/`max` domain; now states `aria-valuenow` is only well-defined for a valid range and describes the `0%` fallback.

Gates on the amended commit: `pnpm check` 0 errors/4 warnings (pre-existing, unchanged), `pnpm lint` 0 errors/3 warnings (pre-existing, unchanged), unit tests 128/128 pass (unchanged), and the full `progress-aria.spec.ts` suite (10/10, including the 5 new tests) passes.

PR #439, opened 2026-08-05
fix(BrandLoader): namespace its CSS vars

murdore:
## Summary

`BrandLoader`'s full-screen loader sizes itself from `--loader-width`/`--loader-height` (default `100vw`/`100vh`). The unrelated `Loader` component — a small inline spinner — independently reads the **exact same two variable names**, defaulting them to `20px`. Nothing about either component's naming hints at the collision: a consumer who sets `--loader-width` in a scope containing both a `BrandLoader` and a `Loader` silently resizes both, with no way to target just one.

This PR adds `--brand-loader-width`/`--brand-loader-height` as the namespaced, collision-free override point, falling back to the legacy shared name and finally to the original literal default — full backward compatibility, zero breaking change.

## Design

```css
/* before */
.loader {
  height: var(--loader-height, 100vh);
  width: var(--loader-width, 100vw);
}

/* after */
.loader {
  height: var(--brand-loader-height, var(--loader-height, 100vh));
  width: var(--brand-loader-width, var(--loader-width, 100vw));
}
```

The fallback chain resolves in order: namespaced var (if set) → legacy shared var (if set) → literal default. Existing consumers who already override `--loader-width`/`--loader-height` keep working completely unchanged. New or updated consumers get a name that can't collide with `Loader`.

**Scope boundary:** only these two variables are touched. `BrandLoader` has ~30 other `--loader-*` variables (background color, text sizing, the bouncing dots, backdrop blur, etc.) that are unique to this component and don't collide with anything — they're left alone. Renaming all of them would be a much larger, unnecessarily breaking migration with no actual collision to justify it.

**`Loader.svelte` itself needs no change and was not touched.** It's the component with the "correct" (already-unique-in-spirit, if not in name) 20px small-spinner semantics; the collision is `BrandLoader` reusing its name space, not the other way round. `docs/Loader.md` is also left as-is — its own `--loader-width`/`--loader-height` documentation is still accurate for that component and isn't part of this fix's stated scope.

## CSS variable contract (new)

| Variable | Default | CSS Property | Notes |
|---|---|---|---|
| `--brand-loader-width` | `100vw` | width | Preferred name (supersedes `--loader-width`); does not collide with `Loader`'s own use of `--loader-width`. |
| `--brand-loader-height` | `100vh` | height | Preferred name (supersedes `--loader-height`); does not collide with `Loader`'s own use of `--loader-height`. |

The legacy `--loader-width`/`--loader-height` rows remain fully documented and functional, now marked **Deprecated** (both names are supported for backward compatibility) with a note that they're also independently read by `Loader` with a different default.

## Usage

```svelte
<!-- before: only the collision-prone shared name existed -->
<BrandLoader classes="my-loader" brandLogoURL="..." brandText="Loading" />
<style>
  .my-loader {
    --loader-width: 320px;
    --loader-height: 180px;
  }
</style>

<!-- after: namespaced name is collision-free; legacy name still works unchanged -->
<BrandLoader classes="my-loader" brandLogoURL="..." brandText="Loading" />
<style>
  .my-loader {
    --brand-loader-width: 320px;
    --brand-loader-height: 180px;
  }
</style>
```

## Registration surfaces

- [x] Component (`src/lib/BrandLoader/BrandLoader.svelte`) — fallback chain added to `.loader`'s `height`/`width`, with an explanatory comment.
- [x] Docs (`docs/BrandLoader.md`) — CSS Variables table: legacy rows marked deprecated + collision note, two new preferred-name rows added.
- [x] Demo route (`src/routes/components/brand-loader/+page.svelte`) — new "Sized Instances" section with three constrained demos (legacy-var, namespaced-var, both-set-precedence) alongside the existing full-screen default, each with its own `testId`.
- [x] Playwright coverage (`tests/brand-loader-css-var-namespacing.spec.ts`) — new file, 3 specs.
- [ ] `docs/Loader.md` — deliberately untouched; `Loader.svelte` has no code change and its docs remain accurate as-is.
- [ ] Web component wrapper — no prop/attribute surface changed, nothing to update there.

## Gates (re-run in full on the amended commit `1c309a3`)

- `pnpm run check` — 0 errors, 4 warnings (baseline: 2 `state_referenced_locally` in Modal.svelte/ThemeSwitcher.svelte + 2 tsconfig warnings; no new warnings).
- `pnpm run lint` — 0 errors, 3 warnings (baseline: pre-existing "Unused eslint-disable directive" in `src/lib/Table/cell-data.test.ts`; no new warnings).
- `pnpm run test:unit` — 128/128 passed (unaffected by this change).
- `pnpm run test:integration` — 145 passed (baseline 142 + 3 new specs), 9 failed (the same pre-existing Table-only baseline failures — `table-builtin-cells.test.ts` ×7, `table-header-meta.test.ts` ×1, `table-pagination-selection.test.ts` ×1). Zero new failures; all 3 new specs pass individually, confirmed via a scoped `playwright test tests/brand-loader-css-var-namespacing.spec.ts` run.


Tara-ag (approved):
## Review Summary

**Files reviewed:** 4
**New issues raised:** 0
**Decision:** ✅ APPROVE

### Analysis

This PR addresses a real CSS custom-property collision between `BrandLoader` and `Loader` by introducing namespaced `--brand-loader-width`/`--brand-loader-height` variables while preserving the legacy `--loader-width`/`--loader-height` names through a fallback chain.

**Public API / backward compatibility:** ✅ No breaking changes. The fallback chain `var(--brand-loader-*, var(--loader-*, 100vw/100vh))` keeps all existing consumer overrides working. No props, events, exports, or web-component attributes were changed.

**Accessibility:** ✅ No interactive or ARIA surface changed; not applicable for this visual loader change.

**Svelte 5 / SSR / lifecycle:** ✅ Only CSS declarations changed; no script/reactivity/SSR concerns.

**Tests & docs:** ✅ New Playwright specs cover all three fallback-chain cases (legacy vars, namespaced vars, precedence), and `docs/BrandLoader.md` documents both the new preferred names and the deprecated legacy names with a clear collision note.

No blocking criteria are met. The change is minimal, well-explained, and correctly scoped.

murdore:
## Proof of testing

Recorded against this branch's own build of `/components/brand-loader`, on a private port, asserting throughout.

https://github.com/user-attachments/assets/ba615819-27cb-49ca-a690-4e125ce81036

Measured live from `getBoundingClientRect()`, all three cases in one pass:

| case | result |
|---|---|
| legacy `--width` / `--height` | **320x180** — still honoured |
| new `--brand-loader-width` / `--brand-loader-height` | **320x180** — identical |
| both set together | **240x140** — the namespaced pair wins |

The first row is the backward-compatibility guarantee: a consumer already theming the loader through the generic names sees no change. The third is the point of the PR — a generic `--width` inherited from an ancestor is exactly the collision this fixes, so the namespaced variable has to take precedence, and the clip shows it does rather than asserting it in prose.

Rebase status: 70 commits behind `release`. Green, but the CI predates most of that.

PR #440, opened 2026-08-05
fix(ChipInput): let a consuming app's Pill and Input theme reach the chips

murdore:
## The bug

`ChipInput` re-declares `--pill-*` and `--input-*` **on its pill and draft-input elements** in order to expose its own `--chip-input-*` API:

```css
.chip-input :global(.chip-input-pill) {
  --pill-background: var(--chip-input-pill-background, #e0e0e0);
  --pill-color: var(--chip-input-pill-color, #333333);
}
```

A declaration on the element beats an inherited one. So for any app that themes `Pill` app-wide — the normal way to theme this library — that mapping **swallows the app's value** and falls through to the hardcoded light-mode hex, because `--chip-input-pill-background` is undefined.

The result: an app theming `Pill` for a dark surface gets light-grey chips with dark text, at wrong contrast, with no fix short of defining `--chip-input-*` overrides at **every call site**.

This is specific to `ChipInput`. A bare `<Pill>` behaves correctly — `Pill` itself does `background-color: var(--pill-background, #e0e0e0)` and simply inherits whatever the app set. `ChipInput` is the only thing breaking that chain.

## How it was found

While migrating a consumer app (Lighthouse, 11 call sites) from a local chip component onto this one. The swap moved **631,175 differing pixels against an 8px noise floor** — a ~79,000× jump. This was one of two root causes; the migration was reverted rather than papered over with per-call-site overrides.

## The fix

The mappings now fall back to the surrounding cascade's value rather than a raw hex:

```css
.chip-input {
  /* captured HERE, on the root, where --pill-*/--input-* are not yet re-declared */
  --chip-input-pill-background-default: var(--pill-background, #e0e0e0);
  --chip-input-pill-color-default: var(--pill-color, #333333);
}
.chip-input :global(.chip-input-pill) {
  --pill-background: var(--chip-input-pill-background, var(--chip-input-pill-background-default));
  --pill-color: var(--chip-input-pill-color, var(--chip-input-pill-color-default));
}
```

**Why the capture is on the root and not inline.** The obvious one-liner — `--pill-background: var(--chip-input-pill-background, var(--pill-background, #e0e0e0))` — is a custom-property **cycle**: a `var()` referencing the property it is defining. Per spec that computes to invalid, so it silently would not work. Capturing on the root, under a distinct name, where `--pill-*` has not yet been re-declared, avoids the cycle.

Applied to 6 tokens: pill background, colour, dismiss colour, and the draft input's border, focus-border and radius.

## Precedence — unchanged for existing consumers, with a new middle rung

| | Before | After |
|---|---|---|
| `--chip-input-pill-background` set | wins | wins (unchanged) |
| app sets `--pill-background` | **ignored** | **respected** |
| neither | `#e0e0e0` | `#e0e0e0` (unchanged) |

Only the middle row changes, and only from "ignored" to "respected". No existing consumer's rendering changes unless they were already setting `--pill-*` and silently not getting it.

## Tests

`tests/chipinput-token-passthrough.spec.ts` asserts all three rows, and the demo page gains a case that themes `--pill-*` on an ancestor the way a consuming app would.

**Negative control** — with the fix reverted, the inheritance test fails on exactly the intended assertion while the other two still pass:

```
Expected: "rgb(47, 53, 66)"
Received: "rgb(224, 224, 224)"
1 failed, 2 passed
```

So the test is load-bearing rather than passing incidentally, and backward compatibility is demonstrated rather than assumed.

**Full functional suite: 145 passed**, no regressions. `npm run check`: 564 files, 0 errors.

## Deliberately not included

`--chip-input-width` defaults to `100%`, which breaks a `flex-wrap` parent — in the consumer app it wrapped a five-field row onto multiple lines. That was the **second** root cause of the migration failure.

I have not changed it here: `width: 100%` is a defensible default for a form control, and flipping it to `auto` would silently reflow every existing consumer. A caller can already set `--chip-input-width: auto`. Raising it separately so a maintainer can decide, rather than bundling a behaviour change into a theming fix.

## Summary by CodeRabbit

* **New Features**
  * Improved ChipInput theming with support for component-level customization and inheritance from surrounding application themes.
  * Added a themed ChipInput example demonstrating how shared styling tokens are applied.

* **Bug Fixes**
  * ChipInput pills and inputs now respect configured theme colors instead of relying on fixed light-theme values.

* **Tests**
  * Added coverage for default styling, component overrides, and inherited application themes.

### DIFF
--- a/src/lib/ChipInput/ChipInput.svelte
+++ b/src/lib/ChipInput/ChipInput.svelte
@@ -88,6 +88,19 @@
     justify-content: var(--chip-input-justify-content, flex-start);
     gap: var(--chip-input-gap, 6px);
     width: var(--chip-input-width, 100%);
+
+    /* Capture the surrounding cascade's Pill and Input values under distinct names, so the
+       per-token mappings below can fall back to them. This has to happen HERE, on the root, where
+       --pill-* and --input-* have not yet been re-declared: reading them in the same declaration
+       that sets them would be a custom-property cycle and would compute to invalid. Consumers that
+       theme Pill/Input app-wide get chips that match the rest of their UI, instead of the library's
+       hardcoded light-theme hexes overriding their theme. */
+    --chip-input-pill-background-default: var(--pill-background, #e0e0e0);
+    --chip-input-pill-color-default: var(--pill-color, #333333);
+    --chip-input-pill-dismiss-color-default: var(--pill-dismiss-color, currentColor);
+    --chip-input-draft-border-default: var(--input-border, 1px solid transparent);
+    --chip-input-draft-focus-border-default: var(--input-focus-border, 1px solid transparent);
+    --chip-input-draft-radius-default: var(--input-radius, 4px);
   }
 
   .chip-input-draft-wrap {
@@ -96,8 +109,8 @@
 
   .chip-input :global(.chip-input-pill) {
     --pill-gap: var(--chip-input-pill-gap, 4px);
-    --pill-background: var(--chip-input-pill-background, #e0e0e0);
-    --pill-color: var(--chip-input-pill-color, #333333);
+    --pill-background: var(--chip-input-pill-background, var(--chip-input-pill-background-default));
+    --pill-color: var(--chip-input-pill-color, var(--chip-input-pill-color-default));
     --pill-font-size: var(--chip-input-pill-font-size, 13px);
     --pill-font-weight: var(--chip-input-pill-font-weight, 500);
     --pill-padding: var(--chip-input-pill-padding, 6px 10px);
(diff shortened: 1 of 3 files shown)


src/lib/ChipInput/ChipInput.svelte
```diff
@@ -88,6 +88,19 @@
...
+       --pill-* and --input-* have not yet been re-declared: reading them in the same declaration
+       that sets them would be a custom-property cycle and would compute to invalid. Consumers that
+       theme Pill/Input app-wide get chips that match the rest of their UI, instead of the library's
+       hardcoded light-theme hexes overriding their theme. */
+    --chip-input-pill-background-default: var(--pill-background, #e0e0e0);
+    --chip-input-pill-color-default: var(--pill-color, #333333);
+    --chip-input-pill-dismiss-color-default: var(--pill-dismiss-color, currentColor);
+    --chip-input-draft-border-default: var(--input-border, 1px solid transparent);
+    --chip-input-draft-focus-border-default: var(--input-focus-border, 1px solid transparent);
+    --chip-input-draft-radius-default: var(--input-radius, 4px);
```
Tara-ag, line 103: ⚠️ MAJOR: Preserve the `--radius` fallback chain used by `Input.svelte`.

`Input.svelte` declares `border-radius: var(--input-radius, var(--radius, 4px))`. The capture here falls directly to `4px` when `--input-radius` is absent, so an ancestor that only sets the generic `--radius` cannot influence the draft input. A standalone themed `Input` could use `8px` while the `ChipInput` draft input stays at `4px`, which contradicts the PR goal of respecting inherited Input theme values.

Suggested change:

```css
--chip-input-draft-radius-default: var(--input-radius, var(--radius, 4px));
```

Also add a regression test with `--radius` set and `--input-radius` unset to lock in the contract.

Tara-ag (commented):
## Review summary

Reviewed 3 changed files. Found **1 new issue** (MAJOR) and noted 1 documentation update opportunity.

### Issues

- **⚠️ MAJOR** `src/lib/ChipInput/ChipInput.svelte:103` — the `--chip-input-draft-radius-default` capture does not preserve `Input.svelte`'s full fallback chain (`var(--input-radius, var(--radius, 4px))`). An ancestor that only sets the generic `--radius` cannot influence the draft input, which conflicts with the PR's goal of respecting inherited Input theme values.

### Suggestions

- **💬 SUGGESTION** `docs/ChipInput.md` (not in diff) — update the documented defaults for the cascade-aware tokens (`--chip-input-pill-background`, `--chip-input-pill-color`, `--chip-input-pill-dismiss-color`, `--chip-input-draft-border`, `--chip-input-draft-focus-border`, `--chip-input-draft-radius`) to describe the new precedence: explicit component token > inherited app theme (`--pill-*` / `--input-*`) > library default.

### Notes

- The Playwright tests cover the three stated precedence levels well and include a useful negative-control claim.
- No blocking criteria are met (no breaking public API change without migration, no security issue, no hardcoded secrets).

Please address the `--radius` fallback chain and consider adding a regression test for `--radius` set with `--input-radius` unset.

PR #444, opened 2026-08-07
feat(Pill): add --pill-line-height CSS hook

murdore:
## What's the gap?

`Pill.svelte`'s style block hardcodes `line-height: 1;`. Every other typographic property on the
same rule is already a CSS-variable hook: `--pill-font-size`, `--pill-font-weight`,
`--pill-font-family`. `line-height` is the one exception, which makes it read as an oversight
rather than a deliberate contract.

## The fix

```diff
-    line-height: 1;
+    line-height: var(--pill-line-height, 1);
```

Backward compatible by construction — the fallback is the current literal, so every existing
consumer renders identically to before.

## Why it matters concretely

Harbour is adopting `Pill` right now and needs a non-default line-height at three call sites.
Without a hook, the only way to reach the property is a compound `.pill.<consumer-class>` selector
that outranks the library's own scoped rule purely on CSS load order — it works today, but it's
fragile: it silently breaks if bundling/import order ever changes. A custom property has no such
dependency, and matches how every sibling typographic property on this component is already themed.

This is the ninth such hook this programme has contributed upstream.

## Changes

- `src/lib/Pill/Pill.svelte` — the one-line hook above.
- `docs/Pill.md` — added `--pill-line-height` to the CSS Variables table, same column format/default
  notation as the surrounding rows.
- `src/routes/components/pill/+page.svelte` — new demo pair (`pill-line-height-default` /
  `pill-line-height-custom`, the latter set to `1.4`) with `testId`s for the test below.
- `tests/pill-line-height.spec.ts` — new file (no prior Pill-specific test existed; followed the
  CSS-variable-override conventions of `tests/chipinput-token-passthrough.spec.ts`, the closest
  comparable test). Asserts the unset default resolves to `13px` (line-height `1` at the default
  `13px` font-size) and an explicit `--pill-line-height: 1.4` resolves to `18.2px`.

## Verification

- `pnpm check`: 0 errors, 4 warnings — matches baseline.
- `pnpm lint`: 0 errors, 3 warnings (pre-existing `Table/cell-data.test.ts` eslint-disable warnings,
  unrelated to this change) — matches baseline.
- `pnpm exec vitest run`: **128/128** — matches baseline.
- `pnpm exec playwright test`: **165 passed / 9 failed** (174 total). Diffed the failing test names
  byte-for-byte against the known pre-existing baseline set — identical: all 9 are Table tests
  (`table-builtin-cells.test.ts` ×7, `table-header-meta.test.ts` ×1,
  `table-pagination-selection.test.ts` ×1), none touched by this change. Zero new failures; the 2
  new Pill tests both pass.

Do not merge — for review.

## Summary by CodeRabbit

* **New Features**
  * Added support for customizing Pill line height through the `--pill-line-height` CSS variable.
  * Added a showcase example demonstrating default and relaxed line-height options.

* **Documentation**
  * Documented the new Pill CSS variable and its default value.

* **Tests**
  * Added coverage for default and customized Pill line-height behavior.

### DIFF
--- a/docs/Pill.md
+++ b/docs/Pill.md
@@ -88,6 +88,7 @@ Override these custom properties to theme the component.
 | `--pill-font-size`           | `13px`                                    | font-size        | Font size of the pill text.                                               |
 | `--pill-font-weight`         | `500`                                     | font-weight      | Font weight of the pill text.                                             |
 | `--pill-font-family`         | `-`                                       | font-family      | Font family of the pill text.                                             |
+| `--pill-line-height`         | `1`                                       | line-height      | Line height of the pill text.                                             |
 | `--pill-padding`             | `6px 10px`                                | padding          | Inner padding of the pill.                                                |
 | `--pill-border-radius`       | `999px`                                   | border-radius    | Corner rounding of the pill (999px creates a fully rounded shape).        |
 | `--pill-border`              | `none`                                    | border           | Border style of the pill.                                                 |
@@ -97,7 +98,6 @@ Override these custom properties to theme the component.
 | `--pill-justify-content`     | `center`                                  | justify-content  | Content alignment inside the pill.                                        |
 | `--pill-text-align`          | `center`                                  | text-align       | Label alignment.                                                          |
 | `--pill-max-width`           | `-`                                       | max-width        | Maximum width of the pill. Text is truncated with ellipsis when exceeded. |
-| `--pill-line-height`         | `1`                                       | line-height      | Line height of the pill.                                                  |
 | `--pill-flex-shrink`         | `-`                                       | flex-shrink      | Flex shrink behavior of the pill.                                         |
 | `--pill-text-overflow`       | `ellipsis`                                | text-overflow    | How overflowing text is displayed (e.g., ellipsis or clip).               |
 | `--pill-hover-background`    | `var(--pill-background, #d0d0d0)`         | background-color | Background color when hovering over the pill.                             |
--- a/tests/pill-line-height.spec.ts
+++ b/tests/pill-line-height.spec.ts
@@ -1,0 +1,27 @@
+import { expect, test } from '@playwright/test';
+
+// Pill hardcoded line-height: 1 with no CSS-variable hook, the only typographic property in the
+// component's style block that wasn't hookable (--pill-font-size, --pill-font-weight, and
+// --pill-font-family all already were). --pill-line-height closes that gap; the fallback is the
+// existing literal, so unset consumers see no change.
+test.describe('Pill line-height hook', () => {
+  test('defaults to line-height: 1 when --pill-line-height is unset', async ({ page }) => {
+    await page.goto('/components/pill');
+
+    const pill = page.getByTestId('pill-line-height-default');
+    await expect(pill).toBeVisible();
+
+    // line-height: 1 at the default --pill-font-size: 13px resolves to a 13px used value.
+    await expect(pill).toHaveCSS('line-height', '13px');
+  });
+
+  test('an explicit --pill-line-height override applies', async ({ page }) => {
+    await page.goto('/components/pill');
+
+    const pill = page.getByTestId('pill-line-height-custom');
+    await expect(pill).toBeVisible();
+
+    // --pill-line-height: 1.4 at the default 13px font-size resolves to 18.2px.
+    await expect(pill).toHaveCSS('line-height', '18.2px');
+  });
+});
--- a/src/routes/components/pill/+page.svelte
+++ b/src/routes/components/pill/+page.svelte
@@ -89,6 +89,15 @@
   </Pill>
 </div>
 
+<div class="demo-row">
+  <Pill text="Default line-height" testId="pill-line-height-default" />
+  <Pill
+    text="Custom line-height"
+    classes="pill-relaxed-line-height"
+    testId="pill-line-height-custom"
+  />
+</div>
+
 <style>
   :global(.pill-info) {
     --pill-background: #d1ecf1;
@@ -100,5 +109,10 @@
     --pill-background: #f8d7da;
     --pill-color: #721c24;
     --pill-hover-background: #f1b0b7;
+  }
+
+  /* --pill-line-height override demo: the pill's own line-height, not its wrapping row's. */
+  :global(.pill-relaxed-line-height) {
+    --pill-line-height: 1.4;
   }
 </style>


docs/Pill.md
```diff
@@ -88,6 +88,7 @@ Override these custom properties to theme the component.
 | `--pill-font-size`           | `13px`                                    | font-size        | Font size of the pill text.                                               |
 | `--pill-font-weight`         | `500`                                     | font-weight      | Font weight of the pill text.                                             |
 | `--pill-font-family`         | `-`                                       | font-family      | Font family of the pill text.                                             |
+| `--pill-line-height`         | `1`                                       | line-height      | Line height of the pill text.                                             |
```
Tara-ag, line 91: 💡 MINOR: Duplicate `--pill-line-height` row in docs/Pill.md

This newly added row is correct, but it creates a duplicate: `--pill-line-height` already appears later in the same CSS Variables table (around line 98, between `--pill-max-width` and `--pill-flex-shrink`). Having the same variable listed twice is confusing and harder to maintain.

**Suggested fix:** Keep this row (it is in the correct alphabetical position and uses the consistent description "Line height of the pill text.") and remove the duplicate entry further down the table.

Tara-ag (commented):
Approving with one minor documentation nit: `--pill-line-height` is listed twice in `docs/Pill.md`. Please remove the duplicate in a follow-up (or before merge if convenient). The component change itself is backward-compatible and well-tested.

docs/Pill.md
Tara-ag, line 1: 💡 MINOR: Duplicate `--pill-line-height` row in docs/Pill.md

The CSS Variables table lists `--pill-line-height` twice: once at line 91 (newly added in this PR) and again at line 98 (pre-existing). Remove the duplicate row at line 98 to keep the docs clean and avoid confusion for downstream theme authors.

Tara-ag (commented):
Approved with one minor documentation nit: `--pill-line-height` is duplicated in the CSS Variables table in `docs/Pill.md`. Please remove the duplicate row in a follow-up.

murdore:
## Proof of testing

Recorded against the rebased branch (`a2eb297`) on a private port, asserting while it records.

https://github.com/user-attachments/assets/a88e8d62-779b-439e-8e1a-adcfac30a648

The on-screen readout is `getComputedStyle().lineHeight` read from the live nodes:

```
unset --pill-line-height   -> 13px
--pill-line-height: 1.4    -> 18.2px
```

Both halves are the contract. `13px` is the used value of the old hard-coded `line-height: 1` at the default `--pill-font-size: 13px`, so an existing consumer who sets nothing renders **unchanged** — a hook that shifted the default would be a breaking change wearing a feature's clothes. `18.2px` is `1.4 x 13px`, so the override actually reaches the property.

### Rebase

I rebased this onto current `release` while sweeping open PRs. It had one conflict in `src/lib/Pill/Pill.svelte`: `release` had added `flex-shrink: var(--pill-flex-shrink)` immediately below the line this PR changes. Union merge, both kept:

```css
line-height: var(--pill-line-height, 1);
flex-shrink: var(--pill-flex-shrink);
```

After the rebase: `pnpm run lint` exit 0 (prettier clean, eslint clean, event-casing `100 known, 0 new`), and `tests/pill-line-height.spec.ts` passes 4/4 at two repeats.

murdore:
## Duplicate `--pill-line-height` row — fixed (`6ed5f5b`)

@Tara-ag and CodeRabbit both flagged this across three threads and I had missed it, including when I rebased this branch an hour ago and posted a proof video without reading the review.

Confirmed before fixing — the variable really was listed twice in the CSS Variables table:

```
docs/Pill.md:91   | `--pill-line-height` | `1` | line-height | Line height of the pill text. |
docs/Pill.md:101  | `--pill-line-height` | `1` | line-height | Line height of the pill.      |
```

### Which one I removed, and why it matters more than a tidy-up

Line **101 was already on `release`** — `git diff origin/release...HEAD -- docs/Pill.md` shows only one added row, the one at line 91. So this table has been documenting `--pill-line-height` for some time while `Pill.svelte` carried a hard-coded `line-height: 1` with no variable behind it.

That reframes this PR. It is not "add a new CSS hook" — it is **"implement a hook the documentation already promised"**. Anyone who read the docs and set `--pill-line-height` before this branch got silence.

I removed the stale row at 101 and kept the new one at 91, which sits with the other typographic properties (`--pill-font-size`, `--pill-font-weight`, `--pill-font-family`) rather than orphaned between `--pill-max-width` and `--pill-flex-shrink`. Diff shape is `0 added / 1 removed` in `docs/Pill.md`; nothing else changed.

`pnpm run lint` exit 0 after the amend.

The PR description is worth a line saying the variable was already documented, since that changes it from a feature request into a correctness fix.

Tara-ag (approved):
Approved. The docs, demo, and test additions correctly describe and exercise the `--pill-line-height` CSS variable. Note: the actual `src/lib/Pill/Pill.svelte` hook is already present on the `release` base branch, so this PR is effectively docs/tests/demo only — the description/commit message imply the component change is part of the diff, which it is not. No blocking issues.

murdore:
## Rebased onto `release` (`3486dd6`) — `6738e47`, one commit, clean

No conflicts. The red `build` was `release`'s own Pages prerender break (#485, now merged), not this branch.

| | result |
|---|---|
| `pnpm lint` | exit 0 |
| `tests/pill-line-height.spec.ts` | **2 passed** |

```
✓ defaults to line-height: 1 when --pill-line-height is unset
✓ an explicit --pill-line-height override applies
```

Run on a private `PW_PORT` so it could not reuse another checkout's preview server and assert against the wrong build.

**Worth noting what this PR now contains**, since it is not what its title suggests: the hook itself — `line-height: var(--pill-line-height, 1)` in `Pill.svelte:103` — is **already on `release`**. It landed separately. What remains here is `docs/Pill.md`, the demo route and the spec: the documentation and the regression test for a variable that already ships undocumented and untested.

That is still worth merging, but it is a docs-and-coverage PR, not a feature one. Both specs pass against the library code as it exists on `release` today, which is the point — they would have caught the hook being dropped, and nothing else does.

tests/pill-line-height.spec.ts
```diff
@@ -0,0 +1,27 @@
+import { expect, test } from '@playwright/test';
+
+// Pill hardcoded line-height: 1 with no CSS-variable hook, the only typographic property in the
```
Tara-ag, line 3: 💡 MINOR: Test comment describes a pre-fix state that no longer exists. The comment says Pill has a 'hardcoded line-height: 1 with no CSS-variable hook' and that this PR 'closes that gap', but `src/lib/Pill/Pill.svelte` already implements `line-height: var(--pill-line-height, 1)` on `release`. This PR is now regression coverage for an existing hook, not the introduction of the hook. Update the comment to describe the current purpose: a regression test ensuring the documented `--pill-line-height` variable continues to work.

Suggested rewrite:
```ts
// Regression test: --pill-line-height is documented and implemented in Pill.svelte.
// This spec guards against the hook being accidentally removed or falling out of sync
// with the default (line-height: 1 at --pill-font-size: 13px).
```

Tara-ag (commented):
Approved with one minor documentation nit. The actual CSS hook is already on `release`, so this PR is effectively docs + regression coverage for an existing public contract. No blocking issues.

docs/Pill.md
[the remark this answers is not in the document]
murdore: Confirmed on release: docs/Pill.md:91 is the only `--pill-line-height` row now. The duplicate that used to sit between `--pill-max-width` and `--pill-flex-shrink` was removed in the same commit that added this row, 1e00cd933 (PR #444). No further change needed.

murdore: Confirmed on release: docs/Pill.md:91 is the only `--pill-line-height` row. The duplicate row that used to sit between `--pill-max-width` and `--pill-flex-shrink` was removed in the same commit, 1e00cd933 (PR #444). Matches the suggested fix exactly.

docs/Pill.md
murdore: Confirmed on release: docs/Pill.md:91 is the only `--pill-line-height` row, the duplicate near `--pill-max-width`/`--pill-flex-shrink` is gone. Fixed in commit 1e00cd933 (PR #444), same commit that added this row.

tests/pill-line-height.spec.ts
murdore: Still open on release: tests/pill-line-height.spec.ts:3-6 unchanged since commit 1e00cd933 (PR #444). The comment is factually off, not just stale — the CSS hook itself predates this PR (added in 5b7dc9d, 'BZ-5434: feat: add list item and pill style variables'); PR #444 never touched Pill.svelte and only added the docs row plus this regression test. Closing this needs the comment rewritten to describe it as a regression test for an existing hook, per the suggested wording in the original review.

murdore: Fixed in juspay/svelte-ui-components#502 (commit 64f91c5be): the header comment now reads as the regression guard the spec is, naming 5b7dc9d as the commit that shipped the hook (tests/pill-line-height.spec.ts).

PR #333, opened 2026-06-27
feat: add showSelectAll to Select (multi-select select-all + indeterminate)

murdore:
## What

Adds `showSelectAll` (+ `selectAllLabel`) to `Select` for multi-select. When `showSelectAll` is set in `multiple` mode, a **"Select all"** row renders at the top of the dropdown that toggles every currently-listed (search-filtered) option, with an **indeterminate** (dash) indicator when only some are selected.

- `showSelectAll?: boolean` — render the select-all row (multiple mode only; no-op otherwise).
- `selectAllLabel?: string` — label for the row (default `'Select all'`).
- `optionIndicator` snippet param extended to `{ checked, indeterminate? }` — **non-breaking**: existing snippets that destructure only `checked` are unaffected.
- New themeable vars: default indicator dash (`--select-option-indicator-dash-size` / `-dash-thickness` / `-dash-color`) and a row hook (`--select-all-border`, `--select-all-font-weight`).

Implemented via a unified `optionRows` model so the select-all row shares the same `highlightedIndex` space as the options — arrow-key navigation needed no special-casing. With `searchable`, "Select all" toggles only the filtered options.

## Why

Unblocks the Lighthouse `Dropdown → Select` migration: the project `Dropdown` has a select-all-with-indeterminate that the library lacked — the one true gap blocking the fold (apply-confirm is already composable via the existing `bottomContent`).

## Demo

`/components/select` → new **"Multi select + select all"** and **"Multi select + select all + searchable"** sections.

## Verified

- `pnpm run check` → 0 errors
- `pnpm run lint` (prettier + eslint, strict) → clean
- `pnpm run build` → packages clean (publint: only the pre-existing repo-url suggestion)
- Live on the demo route: partial selection → indeterminate dash; "Select all" → all 8 selected (checkmark); toggle again → cleared (verified via DOM state probe + screenshot).

## Summary by CodeRabbit

* **New Features**
  * Added an optional “Select all” row for multi-select dropdowns, with a customizable label.
  * “Select all” now supports checked/indeterminate/unselected states and works with keyboard and mouse selection.
  * In searchable mode, “Select all” applies only to the currently filtered options.
* **Documentation**
  * Updated examples and guidance to demonstrate select-all behavior for both searchable and non-searchable multi-selects.
* **Tests**
  * Added coverage for select-all toggling, indeterminate/ARIA state, keyboard navigation, and filtered behavior.

PR #451, opened 2026-08-15
feat: per-icon colour tokens for the six inlining icon slots

murdore:
Follow-up to #450 (2.120.3). One commit on `release`.

## The gap #450 left open

#450 made six icon slots render through `` instead of a plain ``, so a `currentColor` asset finally resolves against the host document instead of painting UA black. **That was necessary but not sufficient**, and it is worth stating plainly because the gap blocked a real consumer migration.

None of the six exposes a **colour** hook. Each sizes its icon — `--select-left-icon-size`, `--file-dropzone-trigger-icon-size`, … — and gives it no independent colour. So an inlined icon inherits the component's text colour and lands on **exactly its label's value**. A muted-icon / strong-label hierarchy, which is a deliberate and common design, became inexpressible, and any consumer migrating an icon to `currentColor` had to accept the flattening.

Concretely, in the Lighthouse dashboard that is **9 icons that cannot move**: migrating `map-pin-location` would repaint the pin from `#a0a0a0` to the trigger's `#333`.

## What this adds

| Token | Slot |
|---|---|
| `--select-left-icon-color` | `Select` leftIcon |
| `--select-option-icon-color` | `Select` per-option icon |
| `--tabs-item-icon-color` | `Tabs` `tabItem.icon` |
| `--file-dropzone-trigger-icon-color` | `FileDropzoneTrigger`, both sizes |
| `--command-menu-item-icon-color` | `CommandMenu` `item.icon` |
| `--status-icon-color` | `Status` `statusIcon` |
| `--table-cell-icon-color` | `Table/BuiltinCell` icon-label |

`FileDropzoneTrigger`'s two sizes deliberately share one token — same icon at two scales, and splitting them would invite the two to drift. `--command-menu-item-icon-color` mirrors the `--command-menu-search-icon-color` that component already had.

## Every one defaults to `inherit`

That is the whole compatibility story. **A component that sets none behaves exactly as it does today**, because inheriting the text colour *is* the current behaviour. This adds a hook; it does not change a pixel.

Setting the fallback to a concrete hex would have made the exception the default — the trap where one route's value leaks into every consumer.

## Tests

`tests/icon-slot-color-token.spec.ts`, three cases covering the contract:

1. **default** — the untinted demo icon still tracks `--select-color` (`#2563eb`)
2. **override** — with `--select-color: #111827` and `--select-left-icon-color: #9ca3af`, the icon takes the grey **and the trigger keeps the near-black**. Asserting only the icon would still pass if the token had simply recoloured everything, which is the failure this exists to catch.
3. **inlining** — the element's own computed `stroke` is the token colour, not merely a wrapper's `color`. A plain `` would report the right wrapper colour while rendering UA black, so this is what distinguishes a real fix from a fake one.

With `inlineSvg` the testId lands on the `<svg>` **itself** rather than a wrapper, so the assertions target the element directly — noted in the spec, because a descendant `.locator('svg')` silently finds nothing and looks like a broken feature. (I hit exactly that on the first run.)

## Pre-existing failures, verified not mine

`tests/table-builtin-cells.test.ts` has 7 failures. They **reproduce identically on clean `release` with this branch stashed** — the table demo does not render, every assertion is "element(s) not found", and none concern colour. Confirmed by control run rather than assumed. Left for whoever owns that demo.

## Verification

`svelte-check` 0 errors (4 warnings, all pre-existing). `eslint` and `prettier` clean. The 3 new tests pass; the other Select/Tabs/Status specs pass unchanged.

Demo: the Select page gains a "Tinting the icon independently of the label" section showing the pairing and naming all seven tokens.

## Summary by CodeRabbit

- **New Features**
  - Added configurable icon colors across command menus, file dropzones, selects, statuses, tables, and tabs.
  - Select icons can now be tinted independently from their accompanying text.
  - Updated component documentation with available icon-color customization options and defaults.

- **Tests**
  - Added coverage verifying default inherited colors, custom icon colors, and SVG color rendering.

### DIFF
--- a/src/lib/CommandMenu/CommandMenu.svelte
+++ b/src/lib/CommandMenu/CommandMenu.svelte
@@ -416,6 +416,9 @@
   .command-menu-item-icon-img-wrapper {
     --image-width: var(--command-menu-item-icon-size, 20px);
     --image-height: var(--command-menu-item-icon-size, 20px);
+
+    /* Mirrors --command-menu-search-icon-color, which this component already had. */
+    color: var(--command-menu-item-icon-color, inherit);
     flex-shrink: 0;
     display: flex;
     align-items: center;
--- a/src/lib/FileDropzoneTrigger/FileDropzoneTrigger.svelte
+++ b/src/lib/FileDropzoneTrigger/FileDropzoneTrigger.svelte
@@ -57,11 +57,17 @@
   .file-dropzone-trigger-icon-sm :global(.file-dropzone-trigger-icon-sm-img) {
     --image-width: var(--file-dropzone-trigger-icon-sm-size, 16px);
     --image-height: var(--file-dropzone-trigger-icon-sm-size, 16px);
+
+    color: var(--file-dropzone-trigger-icon-color, inherit);
   }
 
   .file-dropzone-trigger-icon :global(.file-dropzone-trigger-icon-img) {
     --image-width: var(--file-dropzone-trigger-icon-size, 24px);
     --image-height: var(--file-dropzone-trigger-icon-size, 24px);
+
+    /* Both sizes share one colour token deliberately: they are the same icon at two
+       scales, and splitting them would invite the two to drift apart. */
+    color: var(--file-dropzone-trigger-icon-color, inherit);
   }
 
   .file-dropzone-trigger-heading {
--- a/src/lib/Select/Select.svelte
+++ b/src/lib/Select/Select.svelte
@@ -664,6 +664,13 @@
     --image-width: var(--select-left-icon-size, 16px);
     --image-height: var(--select-left-icon-size, 16px);
     --image-object-fit: contain;
+
+    /* The icon inlines, so a currentColor asset resolves against this. Defaults to
+       `inherit` — the trigger's text colour, which is what it already did — so this is
+       a hook, not a change. Without it a consumer cannot tint the icon independently
+       of the label, and an icon migrated to currentColor is forced to match its label
+       exactly, flattening any deliberate muted-icon/strong-label hierarchy. */
+    color: var(--select-left-icon-color, inherit);
     flex: none;
   }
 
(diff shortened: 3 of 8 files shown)


src/lib/CommandMenu/CommandMenu.svelte
Tara-ag: 💬 SUGGESTION: The new `--command-menu-item-icon-color` token is applied correctly with an `inherit` default, preserving backward compatibility. Consider documenting this token in `docs/CommandMenu.md` alongside the existing `--command-menu-search-icon-color` mention so consumers can discover it.

src/lib/FileDropzoneTrigger/FileDropzoneTrigger.svelte
Tara-ag: 💬 SUGGESTION: Sharing one `--file-dropzone-trigger-icon-color` token across both sizes is a reasonable design choice and well-explained in the comment. Please ensure `docs/FileDropzoneTrigger.md` lists this token so the public CSS API is documented.

src/lib/Select/Select.svelte
Tara-ag: 💬 SUGGESTION: The new `--select-left-icon-color` and `--select-option-icon-color` tokens are backward-compatible defaults. Please update `docs/Select.md` to list them under the "Left Icon" and "Options" CSS variable tables so consumers can discover the public theming contract.

src/lib/Status/Status.svelte
Tara-ag: 💬 SUGGESTION: `--status-icon-color` is added with an `inherit` default, which is safe. Please add it to `docs/Status.md` under CSS Variables so the public theming API is documented.

src/lib/Table/BuiltinCell.svelte
Tara-ag: 💬 SUGGESTION: `--table-cell-icon-color` is a useful addition for `icon-label` cells. Please document it in `docs/Table.md` under the "Built-in Cells" CSS variables table.

src/lib/Tabs/Tabs.svelte
Tara-ag: 💬 SUGGESTION: `--tabs-item-icon-color` is added with an `inherit` default. Please add it to `docs/Tabs.md` in the CSS Variables table so consumers can discover the new theming hook.

PR #454, opened 2026-08-19
feat(Input, Button, Menu): finish the capability gaps the first pass left open

murdore:
Follow-up to #453. Migrating the last Lighthouse call sites onto these components surfaced five more gaps and one defect in what #453 shipped. **Every addition defaults to the value the component already computes, so no existing consumer moves.**

## Input

**`maxLength` accepts `null`.** It defaults to `1000` and was rendered as a native `maxlength` unconditionally, so migrating an *uncapped* textarea onto `Input` silently truncated it at 1000 characters — a chat composer or a JSON paste box loses data with no error and no failing test. Every numeric use of the value is either tel-only normalisation or the character counter, so those now read through a resolved fallback of `1000` and are unchanged.

**New `--input-min-height` / `--input-max-height`.** A textarea that grows with its content needs a ceiling before it can scroll, and one used as a paste target needs a floor. Neither was reachable. Both default to the CSS initial value (`auto` / `none`).

**New `--input-line-height`.** This one was unreachable by *any* means: the UA stylesheet sets `line-height` on the input/textarea itself, and a declaration on the element beats a value inherited from a styled container — so a consumer could not restore the type ramp of the raw textarea it was replacing. Defaulting to `normal`, which is what these fields already compute today, makes it byte-identical for everyone else.

## Button

**New `ariaBusy`.** `aria-busy` was derivable only from `loading`, which also renders the spinner *and disables the control*, so "my contents are still loading but I am still clickable" was unreachable — which is exactly what a menu/filter trigger needs while its options load.

**New `title`.** The browser's own hover tooltip, distinct from `ariaLabel` (which names the control for assistive tech with no visible affordance). An icon-only button generally wants both, and #453 left a consumer unable to keep its tooltip.

## Menu — defect fix

`interactiveTrigger` from #453 hands the snippet Menu's keydown handler, which claims **Enter and Space**. But the snippet now owns a *real* `<button>`, which already synthesises a click from both, and that click is wired to the same `toggle`. The two cancel out: the menu opens and immediately closes.

An interactive trigger now gets a handler that deliberately ignores Enter and Space and adds only the arrow keys, which no native button implements. The inert-trigger path is untouched.

## Tests

Eleven new Playwright tests. **Five exist purely to prove nothing changed for consumers that do not opt in:**

- a field that sets neither height var still computes `min-height: auto` / `max-height: none`
- one that sets no line-height still computes `normal`
- one that leaves `maxLength` alone still renders `maxlength="1000"`
- a button that passes neither new prop emits no `title` and no `aria-busy`

Three more cover Enter / Space / ArrowDown on an interactive trigger directly.

## Verification

Verified in the **built package**, not just the dev server. The suite runs against Vite, which serves `src/lib`, so it can pass in full while `npm pack` produces a tarball containing none of the changes — that is exactly what happened on the first attempt at #453. `dist/` and the tarball were checked for each addition before the package was consumed downstream.

Nine `Table` tests fail in the full suite. They fail identically on unmodified `origin/release` under the same conditions (21 passed / 9 failed either way), so they are pre-existing and not introduced here.

## Summary by CodeRabbit

- **New Features**
  - Buttons now support native tooltips and busy-state accessibility announcements.
  - Inputs support unlimited length, configurable line height, and minimum/maximum heights.
  - Telephone inputs use a default 1,000-character limit when unspecified.
- **Bug Fixes**
  - Improved paste handling for non-telephone inputs.
  - Fixed keyboard activation for interactive menu triggers, preventing duplicate toggles.
- **Documentation**
  - Added component examples demonstrating the new button and input options.
- **Tests**
  - Added coverage for accessibility attributes, input sizing, length limits, and menu keyboard behavior.

### DIFF
--- a/src/lib/Button/Button.svelte
+++ b/src/lib/Button/Button.svelte
@@ -24,7 +24,9 @@
     ariaExpanded,
     ariaHaspopup,
     ariaSelected,
+    ariaBusy,
     role,
+    title,
     onclick,
     onkeydown = () => {},
     onkeyup = () => {},
@@ -91,7 +93,8 @@
     aria-expanded={ariaExpanded ?? null}
     aria-haspopup={ariaHaspopup ?? null}
     aria-selected={ariaSelected ?? null}
-    aria-busy={isBusy || null}
+    aria-busy={isBusy || ariaBusy || null}
+    title={title ?? null}
     type={href ? null : type}
     disabled={href ? null : isDisabled}
     href={href ? (isDisabled ? null : href) : null}
--- a/src/lib/Button/properties.ts
+++ b/src/lib/Button/properties.ts
@@ -64,7 +64,20 @@ export type OptionalButtonProperties = {
    */
   ariaHaspopup?: 'menu' | 'listbox' | 'tree' | 'grid' | 'dialog' | boolean;
   ariaSelected?: boolean;
+  /**
+   * Native `aria-busy`, for a control that stays usable while related data loads.
+   * Deliberately separate from `loading`, which also renders the spinner and
+   * disables the button — a trigger whose *contents* are still loading must stay
+   * clickable, so it cannot express that state through `loading`.
+   */
+  ariaBusy?: boolean;
   role?: string;
+  /**
+   * Native `title`, rendered as the browser's own hover tooltip. Distinct from
+   * `ariaLabel`, which names the control for assistive tech without any visible
+   * affordance; an icon-only button generally wants both.
+   */
+  title?: string;
   disabled?: boolean;
   classes?: string;
   /**
--- a/src/lib/Input/Input.svelte
+++ b/src/lib/Input/Input.svelte
@@ -115,6 +115,9 @@
   const hasRightIcon = $derived(typeof rightIcon === 'function');
 
   const charCount = $derived(value?.length ?? 0);
+  // Every numeric use below is either tel-only normalisation or the character
+  // counter; `null` means "no attribute", not "no ceiling on those paths".
+  const effectiveMaxLength = $derived(maxLength ?? 1000);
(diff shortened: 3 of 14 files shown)


murdore:
All three review findings were valid and are fixed in the amended commit. Each was checked against the code first, and each fix ships with a test that was **negative-controlled** — reverted the fix, confirmed the new test fails, restored it.

**1. `autoResize` ignored a CSS ceiling.** `maxHeight` was derived from `maxRows` alone, so a field capped by `--input-max-height` computed `Infinity`: the inline height grew past the CSS clamp and `overflowY` was set to `'hidden'`. The box stopped at the right size but its overflow became *unreachable* rather than scrollable. Now takes the lower of the two ceilings.
`tests/input-autoresize-css-ceiling.test.ts` — 1 failed before the fix, 2 pass after. The second case asserts overflow stays `hidden` while the content still fits, so the fix does not simply force scrollbars on.

**2. Focus was lost on close.** `close()` focused `triggerEl`, the wrapper div — which under `interactiveTrigger` deliberately carries no `tabindex`, so the call was a no-op and focus fell to `<body>` after Escape or selecting an item. A keyboard user was stranded. It now focuses the control the snippet rendered.
Worth flagging: **this is a defect in what #453 already shipped in 2.123.0**, not in the additions here. Two tests, both failed before the fix.

**3. Demo nesting.** Correct — the three new sections were inside the paste demo's `demo-row` and inheriting its 400px constraint. Closed the row before them.

25 capability tests pass. The nine failing `Table` tests are pre-existing: identical results on unmodified `origin/release` under the same conditions (21 passed / 9 failed either way).

Re-verified downstream after these fixes: the consuming app's full evidence suite — 9 touch points driven in a real browser — still passes end to end against the rebuilt package, including the orders-filter menu whose keyboard path finding #2 touches.

PR #455, opened 2026-08-20
feat: add chat components

sinha-sahil:
Components:
- Add 8 chat components: Chat, ChatHeader, ChatMessage, ChatMessageList, ChatComposer, ChatSuggestions,
  ChatToolStatus, ChatBubble
- Add Resizable, a hard dependency of ChatBubble's resizable panel
- Add ChatController runes class & partyOf role resolver, both exported from index.ts
- Add 8 SVG assets: chat, send, stop, mic, attach, retry, thumb-up, thumb-down
- Export 9 components & 10 property/type modules from index.ts

Conventions:
- Emit testID alongside data-pw on every root element, guarded with typeof, per testing-attributes.d.ts
- ChatBubble's grab/grabbing launcher drives --cursor, the hook Button reads
- ChatToolStatus sets --loader-foreground & --loader-foreground-end; Loader carries no currentColor
  fallback, so an unstyled spinner renders invisible. Exposed as --chat-tool-status-spinner-color
  & --chat-tool-status-spinner-color-end

Web Components:
- Add 9 <sui-*> wrappers registered in src/wc/index.ts

Docs site:
- Add 9 demo pages under /components
- Add a Chat nav category & file Resizable under Layout & Containers

MCP docs:
- Add 9 markdown docs & 9 _index.json entries

Known deviations from GUIDELINES.md, deliberate and left for a follow-up:
- ChatComposer hand-rolls a textarea & a file picker rather than composing Input (useTextArea,
  autoResize) and FileInput (§4)
- autoGrow, pinToBottom, and ChatController.drain's check are arrow consts, not declarations
- ChatBubble's expanded is $bindable but never internally assigned (§7)
- ChatMessageData is referenced by Chat & ChatMessageList docs but resolved in neither
- The WC wrappers pass snippets as object props without mapping named slots, so <sui-chat-bubble>
  cannot receive panel content from HTML

## Summary by CodeRabbit

* **New Features**
  * Added a complete chat component suite with streaming, attachments, suggestions, tool status, retries, feedback, voice controls, and customizable layouts.
  * Added floating, draggable, resizable, expandable chat panels.
  * Added keyboard-accessible resizing with configurable constraints.
  * Added Web Component support for all chat components and resizable panels.
* **Documentation**
  * Added comprehensive usage, API, accessibility, customization, and Web Component documentation.
* **Examples**
  * Added interactive demonstrations for chat and resizable components.


src/lib/Chat/Chat.svelte
```diff
@@ -0,0 +1,158 @@
...
+    onretry,
+    onfeedback,
+    testId,
+    classes
+  }: ChatProperties = $props();
+
+  let showHeader = $derived(
+    title.length > 0 ||
+      subtitle.length > 0 ||
+      (typeof image === 'string' && image.length > 0) ||
```
Tara-ag: ⚠️ MAJOR: `showHeader` derives from `typeof headerAvatar === 'function'` and `typeof headerActions === 'function'`, but `headerContent` is not included. Passing only `headerContent` will hide the header entirely, even though the prop is documented as "Extra content as a full-width second row in the header". Include `typeof headerContent === 'function'` in the `showHeader` condition.

```svelte
let showHeader = $derived(
  title.length > 0 ||
    subtitle.length > 0 ||
    (typeof image === 'string' && image.length > 0) ||
    typeof onclose === 'function' ||
    typeof headerAvatar === 'function' ||
    typeof headerActions === 'function' ||
    typeof headerContent === 'function'
);
```

src/lib/Chat/Chat.svelte
```diff
@@ -0,0 +1,158 @@
...
+    {onretry}
+    {onfeedback}
+  />
+
+  <div class="footer">
+    {#if showSuggestions}
+      <ChatSuggestions items={suggestions} {disabled} onselect={handleSuggestion} />
+    {/if}
+    {#if toolStatus !== null}
+      <div class="tool-status"><ChatToolStatus label={toolStatus.label} /></div>
```
Tara-ag: ⚠️ MAJOR: `toolStatus` is typed as `ChatToolStatus | null` and rendered with `label={toolStatus.label}`, but the `ChatToolStatus` component also accepts an `icon` snippet. The controller only ever sets `{ label: ... }` and the `Chat` component never exposes a `toolStatusIcon` prop/snippet, so consumers cannot customize the spinner icon at the top level. Either expose `toolStatusIcon` on `Chat` and forward it, or document that the icon is not customizable.

If intentional, add a `toolStatusIcon?: Snippet` prop and pass it through:
```svelte
<ChatToolStatus label={toolStatus.label} icon={toolStatusIcon} />
```

src/lib/ChatComposer/ChatComposer.svelte
```diff
@@ -0,0 +1,289 @@
...
+            {@html attachSvg}
+          {/if}
+        </Button>
+        <input
+          bind:this={fileInput}
+          type="file"
+          {accept}
+          {multiple}
+          class="file-input"
+          onchange={handleFiles}
```
Tara-ag: ⚠️ MAJOR: `autoGrow` reads `messages` from the outer closure (`messages` is not defined in this file), so the action's `update` will never fire when the textarea value changes. The dependency should be `value` (the bindable prop), not `messages`. Change `use:autoGrow={value}` and use the parameter inside the action.

```ts
const autoGrow: Action<HTMLTextAreaElement, string> = (node, initial) => {
  function resize(): void {
    node.style.height = '0px';
    node.style.height = `${node.scrollHeight}px`;
  }
  resize();
  return {
    update: resize
  };
};
```
Then `use:autoGrow={value}` is correct because Svelte re-runs the action when `value` changes.

src/lib/ChatBubble/ChatBubble.svelte
```diff
@@ -0,0 +1,419 @@
...
+        ? margin - ended.left0
+        : window.innerWidth - margin - ended.width - ended.left0;
+    }
+    const left = ended.left0 + dragX;
+    const top = ended.top0 + dragY;
+    placeAt(left, top, left + ended.width, top + ended.height);
+  }
+
+  function handlePanelKeydown(event: KeyboardEvent): void {
+    if (event.key === 'Escape') {
```
Tara-ag: ⚠️ MAJOR: The panel has `role="dialog"` and `tabindex="-1"`, but opening it does not trap focus. A keyboard user can Tab out of the panel and into the rest of the page while it remains open. For an overlay dialog, focus should be trapped inside the panel while `open` is true. Consider adding a focus-trap action or documenting that consumers must wrap content with their own trap.

Suggested minimal fix using a focus-trap action on `.panel`:
```ts
import { tick } from 'svelte';

const trapFocus: Action<HTMLElement> = (node) => {
  function handleTab(event: KeyboardEvent): void {
    if (event.key !== 'Tab') return;
    const focusable = Array.from(
      node.querySelectorAll<HTMLElement>(
        'a[href], button, textarea, input, select, [tabindex]:not([tabindex="-1"])'
      )
    ).filter((el) => !el.hasAttribute('disabled') && el.offsetParent !== null);
    if (focusable.length === 0) {
      event.preventDefault();
      return;
    }
    const first = focusable.at(0)!;
    const last = focusable.at(-1)!;
    if (event.shiftKey && document.activeElement === first) {
      event.preventDefault();
      last.focus();
    } else if (!event.shiftKey && document.activeElement === last) {
      event.preventDefault();
      first.focus();
    }
  }
  node.addEventListener('keydown', handleTab);
  return { destroy: () => node.removeEventListener('keydown', handleTab) };
};
```
and apply `use:trapFocus` to the panel element.

src/lib/ChatBubble/ChatBubble.svelte
```diff
@@ -0,0 +1,419 @@
...
+    minPanelHeight = 360,
+    expanded = $bindable(false),
+    expandedPanelWidth = $bindable(600),
+    expandedPanelHeight = $bindable(760),
+    onopen,
+    onclose,
+    ontoggle,
+    testId,
+    classes
+  }: ChatBubbleProperties = $props();
```
Tara-ag: ⚠️ MAJOR: `expanded` is declared as `$bindable` but the component never assigns to it internally. Per GUIDELINES.md §7, `$bindable` should only be used when the component internally mutates the prop. Since expansion is driven entirely by the parent, `expanded` should be a regular prop. Either remove `$bindable` or add internal expand/collapse behavior (e.g., a default expand button).

```ts
let {
  expanded = false,
  ...
}: ChatBubbleProperties = $props();
```

PR #470, opened 2026-08-28
feat(MarkdownText,ChatMessage): add sanitized markdown rendering

murdore:
## What

Closes the library's markdown gap — and knowingly **reverses a documented design decision**: `docs/Chat.md` stated "Markdown is intentionally not bundled: pass pre-sanitized HTML … (e.g. the output of your own `marked` + `DOMPurify`)". In practice every real consumer (Lighthouse's `{@html marked()}` call sites, Sara's transcript) ran `marked` with **no sanitizer at all** — the bring-your-own contract was being violated everywhere it was exercised. Bundling a sanitized-by-construction pipeline is strictly safer than the status quo. The stale docs statements (`docs/Chat.md`, `docs/_index.json`) are corrected in this PR.

- **`MarkdownText`** — standalone component: raw HTML (block + inline) is escaped to visible text; link/image URLs keep their href only when the protocol is allow-listed (`http:`/`https:`/`mailto:`/`tel:` for links, `http:`/`https:` for images; relative URLs pass). Entity-smuggled protocols (`jav&#x09;ascript:`) fail scheme parsing and are inert after attribute escaping. Pure string transform — SSR-identical, no hydration mismatch, no DOM dependency.
- **`renderMarkdown(markdown, {breaks?})`** — the same pipeline as an exported utility.
- **`ChatMessage.markdown` prop** — renders through the pipeline internally (`markdown` > `html` > `content`; a `body` snippet still wins) and serves as copy text when `content` is empty. ChatMessage's `.body :global` styles gained the missing markdown elements (lists, headings, blockquotes, tables, images, hr) with `--chat-message-*` tokens + literal fallbacks — tables previously rendered unstyled inside bubbles (caught visually).
- **`ChatMessageData.markdown` field** — flows through `ChatMessageList` (and therefore the full `Chat` surface + `ChatController` consumers), mirroring the existing `html` field.
- wc wrappers (`sui-markdown-text`, `markdown` on `sui-chat-message`), docs (`docs/MarkdownText.md` + `_index.json` entry so the MCP lists it), demo page (`/components/markdown-text`, incl. a hostile-input section that proves the security model in the browser), nav entry.
- `marked` becomes a **peerDependency** (`^18.0.0`, kept in devDependencies for local dev — the `lottie-web`/`type-decoder` pattern), preserving the zero-runtime-dependency posture. `ChatMessage` loads the pipeline **dynamically at mount** (the LottiePlayer pattern), so the peer is only exercised when the `markdown` prop is used and existing consumers keep their install contract; during SSR / while loading, `html`/`content` render as the fallback. npm 7+/pnpm auto-install peers; documented in `docs/MarkdownText.md`.

## Why it was never bundled — the design history, mapped

- The "intentionally not bundled" line entered in #455 (the chat-family commit) with no further rationale recorded in the commit, PR body, or review thread.
- #458 restated it as an extraction principle while porting the Lighthouse primitives: "markdown parser → `renderText`", "markdown stays the host's business" — consistent with the library's zero-runtime-dependency posture (every runtime lib is a peer).
- The root precedent is security guidance: Breeze Buddy pentest **BB-DEEPDIVE-2026-001** findings PT-06 (markdown sanitiser: marked + DOMPurify) and PT-25 (javascript:/data: URL guard), landed in loom#267. The docs' "your own `marked` + `DOMPurify`" echoes that remediation — parser and sanitizer pushed to the host, where the trust model lives.

Why this PR still reverses it: the host-owns-it model has failed empirically twice — Lighthouse ships `{@html marked()}` with no sanitizer today, and loom only gained one because a pentest forced it. Hosts do the `marked` half and skip the `DOMPurify` half. This pipeline implements exactly the PT-06 + PT-25 protections by construction (raw-HTML neutralization + protocol allow-list), avoids the DOMPurify/SSR problem entirely, and leaves #458's `renderText` seam untouched for hosts that want their own renderer.

## Prior-art check (exhaustive)

No markdown renderer existed anywhere in this repository before this PR: `git grep -i markdown` on `origin/release` finds only the intentionally-not-bundled docs statements, TypewriterText's `renderText` injection point (a hook for an external renderer, not a renderer), and the demo site's own docs layout (the sole `marked` import, why it was a devDependency). `git log --all --diff-filter=A -- '*arkdown*'` shows no such file ever existed; no remote branch carries one; the published 2.131.0 dist has no markdown module.

## Verification

- `vitest` 141/141 (17 files) — 13 new tests cover rendering + XSS vectors (`<script>`, inline `onerror`, `javascript:`/`data:` links, entity smuggling, unsafe image src).
- **Negative control:** with the sanitizer's escape disabled, the two raw-HTML tests fail — the security tests are load-bearing, not decorative.
- `pnpm run check` 0 errors · `pnpm run lint` clean · `pnpm run build` (vite + svelte-package + publint) green.
- `check:wc` fails on base and branch identically — 3 pre-existing `LottiePlayer.wc.svelte` `$host` typing errors, untouched here (verified via stash on the pristine tree).
- Demo pages screenshot-verified by eye: standalone render, in-bubble table styling, hostile input inert, breaks comparison.

## Consumers lined up

- **Lighthouse**: retire the raw `{@html marked()}` call sites in ChatView et al. (follow-up PR after release).
- **Sara** (`apps/web`): transcript rendering — session already briefed.


src/wc/components/MarkdownText.wc.svelte
```diff
@@ -0,0 +1,19 @@
+<svelte:options
+  customElement={{
+    tag: 'sui-markdown-text',
+    shadow: 'open',
+    props: {
+      markdown: { type: 'String' },
+      breaks: { type: 'Boolean' },
+      testId: { type: 'String', attribute: 'test-id' },
```
Tara-ag, line 8: 💡 MINOR: `classes: { type: 'String' }` relies on Svelte's default camelCase → kebab-case conversion, which produces `class-es`, not `classes`. The repo convention (and bootstrapped standard) is to declare an explicit `attribute` mapping for every camelCase prop exposed to the custom element. Add `attribute: 'classes'` if this prop is meant to be set from HTML.

src/lib/MarkdownText/markdown.ts
```diff
@@ -0,0 +1,76 @@
...
+    return escapeHtml(token.text);
+  },
+  image(token: Tokens.Image): string | false {
+    if (hasSafeProtocol(token.href, SAFE_IMAGE_PROTOCOLS)) {
+      return false;
+    }
+    return escapeHtml(token.text);
+  }
+};
+
```
Tara-ag, line 88: ⚠️ MAJOR: The module-scoped `instances` Map is shared across SSR requests and leaks state between tests. Even though keys are currently deterministic, mutable module-level state is a reactivity/lifecycle anti-pattern in a library that ships to SSR consumers. Replace with a fresh `Marked` instance per call, or key a cache in a way that does not share mutable state across requests (e.g., a WeakMap keyed by a frozen options object).

murdore:
@sinha-sahil requesting your review specifically because this PR reverses the "Markdown is intentionally not bundled" line you wrote in #455. The PR body now maps the full design history (your #455 statement → #458's "markdown stays the host's business" seam → the BB-DEEPDIVE-2026-001 pentest pattern behind the marked+DOMPurify guidance) and why bundling a sanitized-by-construction pipeline is the safer contract in practice: both real consumers run marked with no sanitizer today. Per the zero-dependency posture, marked ships as a required peerDependency (lottie-web pattern), not a runtime dependency. The renderText seam from #458 is untouched. Happy to adjust if you see a reason the host-owns-it model should stand.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -37,7 +39,9 @@
   let copied = $state(false);
   let copyResetTimer: ReturnType<typeof setTimeout> | null = null;
 
-  let hasHtml = $derived(typeof html === 'string' && html.length > 0);
+  let hasMarkdown = $derived(typeof markdown === 'string' && markdown.length > 0);
+  let effectiveHtml = $derived(hasMarkdown ? renderMarkdown(markdown ?? '') : html);
+  let hasHtml = $derived(typeof effectiveHtml === 'string' && effectiveHtml.length > 0);
```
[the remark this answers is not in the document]
murdore: Declining the code change — the empty-string fall-through is deliberate and mirrors this component's pre-existing `html` semantic (`hasHtml` also requires `length > 0`, so an empty `html` falls through to `content`). "markdown > html > content" reads as non-empty-wins at every level, consistently.

The proposed diff also wouldn't produce the behavior it describes: `renderMarkdown('')` returns `''`, so `hasHtml` stays false and the template falls to `content` regardless. Its only observable effect is the `markdown="" + html` combination rendering *content instead of html* — which would break the `html > content` precedence instead — plus a `copy()` regression when both `content` and `markdown` are empty (copy would yield `''` instead of `htmlToText(html)`).

The real defect here was doc ambiguity, which is fixed in `2ae29de`: `properties.ts` and `docs/ChatMessage.md` now state explicitly that non-empty markdown wins and an empty string is treated as absent, exactly like `html`.

package.json
```diff
@@ -49,7 +49,8 @@
   "peerDependencies": {
     "svelte": "^5.41.2",
     "type-decoder": "^2.1.0",
-    "lottie-web": ">=5.0.0"
+    "lottie-web": ">=5.0.0",
```
Tara-ag, line 56: ⚠️ MAJOR: Adding `marked` as a required peerDependency breaks the install contract for existing `ChatMessage` consumers. Because `ChatMessage.svelte` now statically imports `renderMarkdown` from `../MarkdownText/markdown`, any app already importing `ChatMessage` (or the barrel `index.ts`) must install `marked` even if they never use the new `markdown` prop. Since the commit is `feat(...)` rather than `feat!(...)`, conventional-commits will not cut a major version, making this a silent breaking change.

Suggested fix: either mark the commit as breaking (`feat!:`) and add a migration note, or dynamically import `marked` only when `markdown` is non-empty and gracefully fall back to `html`/`content` when the import is unavailable.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -55,7 +59,14 @@
   }
 
   async function copy(): Promise<void> {
-    const text = content.length > 0 ? content : typeof html === 'string' ? htmlToText(html) : '';
+    const text =
+      content.length > 0
+        ? content
+        : hasMarkdown
+          ? (markdown ?? '')
+          : typeof html === 'string'
```
Tara-ag, line 92: 💡 MINOR: `copyResetTimer` is not cleared when the component is destroyed. If `ChatMessage` unmounts while the 1500 ms reset timeout is pending, the callback will fire and attempt to mutate disposed state.

Suggested fix: import `onDestroy` from `svelte` and clear the timer:
```ts
onDestroy(() => {
  if (copyResetTimer !== null) clearTimeout(copyResetTimer);
});
```

src/lib/MarkdownText/markdown.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    expect(output).toContain('<img');
+    expect(output).toContain('src="https://example.com/a.png"');
+  });
+
+  it('honors the breaks option', () => {
+    expect(renderMarkdown('a\nb')).not.toContain('<br>');
+    expect(renderMarkdown('a\nb', { breaks: true })).toContain('<br>');
+  });
+
+  it('returns an empty string for empty input', () => {
```
Tara-ag, line 155: 💬 SUGGESTION: The entity-smuggled protocol test only asserts that `href="javascript:"` is absent. It should also verify that no `<a>` tag is emitted and that the link text is preserved, so a future renderer change cannot silently re-introduce an unsafe anchor with a different attribute shape.

Suggested fix: add `expect(output).not.toContain('<a');` and `expect(output).toContain('x');` to this test.

Tara-ag (changes requested):
Reviewed file by file. The markdown pipeline itself is well-designed and the test coverage for XSS vectors is good, but the PR introduces a breaking install contract change by adding `marked` as a required peer dependency while `ChatMessage` statically imports it. This needs to be resolved (breaking commit marker + migration note, or dynamic import fallback) before merge. Also noted a missing timer cleanup and a test-strengthening suggestion.

package.json
murdore: Accepted via your suggested dynamic-import fix, adapted to repo constraints: `ChatMessage` now loads the pipeline at mount — the exact `LottiePlayer` pattern (`$effect` is banned by this repo's lint, so `onMount` + `$state` carries it) — and degrades to `html`/`content` during SSR, while loading, or when the peer is absent. Docs updated in `properties.ts`, `ChatMessage.md`, `MarkdownText.md` (fixed in `47bdea0`). One honest caveat: because the barrel statically re-exports `MarkdownText`, bundlers that resolve the full module graph may still require `marked` to be present at build for barrel importers — the same situation `lottie-web` has always been in here; npm 7+/pnpm auto-install required peers, which is what makes both livable.

src/lib/MarkdownText/markdown.ts
murdore: Declining this one with rationale: each `Marked` instance's configuration (renderer, gfm, breaks) is fixed at construction and `parse()` takes no per-request state, so the cache only ever holds config-immutable parsers keyed by option shape — no request-specific data can leak through it. A fresh instance per call would re-run `marked.use` setup per message in an SSR list render. Added a code comment documenting exactly this invariant in `47bdea0` so the safety condition is explicit for future editors.

src/wc/components/MarkdownText.wc.svelte
murdore: Declining: `classes` is already a single all-lowercase word — Svelte's camelCase→kebab attribute conversion only inserts dashes at case boundaries, so the attribute is `classes` exactly as written (`class-es` would require a `classEs" prop). This also matches the existing repo precedent: `ChatMessage.wc.svelte` (and the other wrappers) declare `classes: { type: 'String' }` with no attribute mapping; explicit mappings are used only for genuinely camelCase props like `testId` → `test-id`.

src/lib/ChatMessage/ChatMessage.svelte
murdore: Accepted — `onDestroy` cleanup added exactly as suggested (in `47bdea0`). Worth noting it was a pre-existing latent issue from #455 rather than introduced here, but this PR touches `copy()` anyway so it belongs in this diff.

src/lib/MarkdownText/markdown.test.ts
murdore: Partially accepted: added the text-preservation assertion (`expect(output).toContain('x')`). Declining the `not.toContain('<a')` assertion because it contradicts the pipeline's documented design: an entity-smuggled href fails scheme parsing, is treated as a relative path, and the anchor legitimately survives with an inert, attribute-escaped href — the invariant is that no executable protocol appears anywhere, which the test now states in an in-code comment (`47bdea0`).

package.json
```diff
@@ -49,7 +49,8 @@
...
+    "marked": "^18.0.0"
   },
   "peerDependenciesMeta": {
```
Tara-ag, line 61: ⚠️ MAJOR: New `marked` peer dependency is not marked optional. The PR mirrors the `lottie-web` peer-dependency pattern but only `lottie-web` is listed in `peerDependenciesMeta.optional`. Consumers who never use markdown will still get unmet-peer warnings (or hard failures under Yarn PnP / pnpm strict-peer-dependencies), which breaks the stated goal of preserving the zero-runtime-dependency posture. Add `marked` to `peerDependenciesMeta` as optional:
```json
"peerDependenciesMeta": {
  "lottie-web": { "optional": true },
  "marked": { "optional": true }
}
```

src/lib/MarkdownText/markdown.ts
```diff
@@ -0,0 +1,81 @@
...
+function escapeHtml(value: string): string {
+  return value
+    .replace(/&/g, '&amp;')
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;')
+    .replace(/"/g, '&quot;')
+    .replace(/'/g, '&#39;');
+}
+
+const sanitizingRenderer: RendererObject = {
```
Tara-ag, line 71: 💬 SUGGESTION: External markdown links open in the same tab without `rel="noopener"`. Returning `false` lets marked emit a plain `<a href="...">`, which navigates away and leaves tabnabbing risk. Consider overriding `link` to add `target="_blank" rel="noopener noreferrer"` for http/https URLs, or document in `docs/MarkdownText.md` that consumers must post-process links if they need safe external navigation.

src/lib/MarkdownText/markdown.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    expect(output).not.toContain('<img');
+    expect(output).toContain('&lt;img');
+  });
+
+  it('keeps http, https, mailto, tel and relative links', () => {
+    expect(renderMarkdown('[a](https://example.com)')).toContain('href="https://example.com"');
+    expect(renderMarkdown('[a](http://example.com)')).toContain('href="http://example.com"');
+    expect(renderMarkdown('[a](mailto:x@y.z)')).toContain('href="mailto:x@y.z"');
+    expect(renderMarkdown('[a](tel:+15550100)')).toContain('href="tel:+15550100"');
+    expect(renderMarkdown('[a](/docs/setup)')).toContain('href="/docs/setup"');
```
Tara-ag, line 44: 💡 MINOR: Missing test for autolink / bare URL protocol stripping. GFM autolinks and bare URLs route through the same `renderer.link` override, but the suite only covers explicit `[text](url)` links. Add a case for `https://example.com` and `javascript:alert(1)` to guard against future marked changes introducing a separate autolink path.
```ts
it('strips unsafe protocols from autolinks and bare URLs', () => {
  expect(renderMarkdown('https://example.com')).toContain('href="https://example.com"');
  expect(renderMarkdown('javascript:alert(1)')).not.toContain('javascript:');
  expect(renderMarkdown('javascript:alert(1)')).not.toContain('<a');
});
```

package.json
Tara-ag, line 61: ⚠️ MAJOR: New `marked` peer dependency is not declared optional. The PR mirrors the `lottie-web` peer-dependency pattern but only `lottie-web` is listed in `peerDependenciesMeta.optional`. Consumers who never use markdown will still get unmet-peer warnings (or hard failures under Yarn PnP / pnpm strict-peer-dependencies), which breaks the stated goal of preserving the zero-runtime-dependency posture. Add `marked` to `peerDependenciesMeta` as optional:
```json
"peerDependenciesMeta": {
  "lottie-web": { "optional": true },
  "marked": { "optional": true }
}
```

src/lib/MarkdownText/markdown.test.ts
Tara-ag, line 44: 💡 MINOR: Missing test for autolink / bare URL protocol stripping. GFM autolinks and bare URLs route through the same `renderer.link` override, but the suite only covers explicit `[text](url)` links. Add a case for `https://example.com` and `javascript:alert(1)` to guard against future marked changes introducing a separate autolink path.
```ts
it('strips unsafe protocols from autolinks and bare URLs', () => {
  expect(renderMarkdown('https://example.com')).toContain('href="https://example.com"');
  expect(renderMarkdown('javascript:alert(1)')).not.toContain('javascript:');
  expect(renderMarkdown('javascript:alert(1)')).not.toContain('<a');
});
```

Tara-ag (changes requested):
Reviewed PR #470. Two actionable findings remain:

1. **MAJOR** — `marked` must be marked optional in `peerDependenciesMeta` to avoid breaking the install contract for existing consumers who do not use markdown.
2. **MINOR** — add a test covering autolink / bare URL protocol stripping.

Once these are addressed, this is good to merge.

Tara-ag:

Severity counts — 🔒 CRITICAL: 0 · ⚠️ MAJOR: 2 · 💡 MINOR: 2

## 🤖 Yama Review Summary

**Verified findings (5):**

- **MAJOR**: New `marked` peer dependency is not marked optional — `package.json:55`
- **MINOR**: Missing test for autolink / bare URL protocol stripping — `src/lib/MarkdownText/markdown.test.ts:44`
- **SUGGESTION**: External markdown links open in the same tab without noopener — `src/lib/MarkdownText/markdown.ts:38`
- **MAJOR**: New `marked` peer dependency is not declared optional — `package.json:55`
- **MINOR**: Missing test for autolink / bare URL protocol stripping — `src/lib/MarkdownText/markdown.test.ts:44`

- ⚠️ **MAJOR**: New `marked` peer dependency is not marked optional — `package.json:55`
  package.json adds `marked` as a peer dependency but does not list it in `peerDependenciesMeta.optional`, unlike the existing `lottie-web` peer that this PR claims to mirror. Consumers who do not use markdown features will still receive unmet-peer warnings or hard failures from strict package managers (Yarn PnP, pnpm with strict-peer-dependencies, older npm with --legacy-peer-deps off). This contra…
- ⚠️ **MAJOR**: New `marked` peer dependency is not declared optional — `package.json:55`
  `marked` was added as a required peerDependency, but the PR's stated goal is to preserve the zero-runtime-dependency posture. Only `lottie-web` is listed in `peerDependenciesMeta.optional`. Consumers who never use markdown will receive unmet-peer warnings (or hard failures under Yarn PnP / pnpm strict-peer-dependencies). Because the commit is `feat(...)` not `feat!(...)`, this is a silent install-…
- 💡 **MINOR**: Missing test for autolink / bare URL protocol stripping — `src/lib/MarkdownText/markdown.test.ts:44`
  GFM autolinks and bare URLs (e.g. `https://example.com` or `<https://example.com>`) are routed through the same `renderer.link` override, but the test suite only exercises explicit `[text](url)` links. Adding a test guards against future marked upgrades that might introduce a separate code path for autolinks.
- 💡 **MINOR**: Missing test for autolink / bare URL protocol stripping — `src/lib/MarkdownText/markdown.test.ts:44`
  GFM autolinks and bare URLs route through the same `renderer.link` override, but the test suite only covers explicit `[text](url)` links. A future `marked` change could introduce a separate autolink path that bypasses the allow-list.
- 💬 **SUGGESTION**: External markdown links open in the same tab without noopener — `src/lib/MarkdownText/markdown.ts:38`
  The sanitizing renderer returns `false` for safe links so marked emits a plain `<a href="...">`. For external URLs this leaves users vulnerable to tabnabbing and navigates away from the app. Consider either overriding the link renderer to add `target="_blank" rel="noopener noreferrer"` for http/https URLs, or document that consumers should post-process links if they need this behavior.

package.json
murdore: Accepted — `marked` added to `peerDependenciesMeta` as optional in `af03004`, exactly mirroring `lottie-web` now that the mount-time dynamic import makes it genuinely optional. Good catch: I had mirrored the peer entry but not the meta, which would have hard-failed Yarn PnP / pnpm strict-peers consumers who never touch markdown. (Noting this comment was posted twice — same reply applies to both.)

src/lib/MarkdownText/markdown.test.ts
murdore: Accepted in substance, with the assertions corrected against observed behavior (in `af03004`): bare `javascript:alert(1)` text is NOT autolinked by GFM — it renders as harmless escaped text, so the suggested `not.toContain('javascript:')` would fail on innocent text content. The vector that actually exercises the autolink path is the angle-bracket form `<javascript:alert(1)>` (CommonMark autolinks accept any scheme), and the new test covers: GFM bare-URL autolink kept, `<https://…>` autolink kept, `<javascript:…>` stripped with no anchor emitted. All 17 tests green. (Posted twice — same reply applies to both.)

src/lib/MarkdownText/markdown.ts
murdore: Accepted, implemented rather than just documented (in `af03004`): external `http:`/`https:` links now open with `target="_blank" rel="noopener noreferrer"` — matching the resolvedRel default the library's Button/Card already apply — while relative/`mailto:`/`tel:` links keep same-tab navigation. One precision on the premise: without `target="_blank"` there is no reverse-tabnabbing vector (it requires the opener relationship), but new-tab behavior is what chat surfaces want anyway, so this lands both the UX and the hardening. Covered by a new test + documented in the Security model section.

murdore:
## Proof of testing — and it found a sanitiser bypass in this PR

I recorded evidence for this PR the way I should have when I opened it. The clip does **not** show the feature working. It shows a hole.

https://github.com/user-attachments/assets/50fb29c6-5b3f-4869-af34-6d8abdb7fc21

### What survives

The demo page's own prose says of the hostile fixture:

> All of it renders as visible, inert text — nothing executes, no anchor keeps an unsafe protocol.

Three of the four payloads are handled: the `<script>` element, the `onerror` attribute, the plain `javascript:` link and the `data:text/html` link are all stripped. The fourth is not. The DOM after render:

```json
{ "raw":      "jav\tascript:alert(1)",
  "resolved": "javascript:alert(1)",
  "protocol": "javascript:" }
```

That is `HTMLAnchorElement.protocol` read off the live node — not a regex of mine. The anchor renders as a normal blue link labelled "protocol", and clicking it executes.

### Why it gets through

`hasSafeProtocol` is handed the href **before** HTML-entity decoding, and the browser resolves it **after**:

| seen by | value | `new URL(...).protocol` | verdict |
|---|---|---|---|
| `hasSafeProtocol` | `jav&#x09;ascript:alert(1)` | `https:` | **allowed** — parses as a relative path |
| the browser | `jav⇥ascript:alert(1)` | `javascript:` | executes |

Reproduced directly:

```
"jav&#x09;ascript:alert(1)"  -> protocol https:       safe? true
"jav\tascript:alert(1)"      -> protocol javascript:  safe? false
"javascript:alert(1)"        -> protocol javascript:  safe? false
```

So the check is correct for a decoded string and is never given one. The WHATWG URL parser strips ASCII tab/LF/CR *inside* a scheme, which is exactly what makes `jav⇥ascript:` a working protocol in the browser and an inert-looking relative path to the validator.

### The comment in the source states the wrong conclusion

`src/lib/MarkdownText/markdown.ts`:

> Entity-smuggled protocols (`jav&#x09;ascript:`) fail to parse as a scheme here and fall through as relative paths, where attribute-escaping renders them inert.

The first half is right and the conclusion is backwards. Falling through as a relative path is what makes it **pass** the allow-list, and attribute-escaping does not neutralise it because the browser decodes the entity back into a tab before resolving the URL. This is worth calling out because the comment would otherwise stop the next reader from looking.

### Suggested fix

Decode HTML entities and strip ASCII whitespace/control characters from `href` *before* the protocol check, or reject any href whose raw form contains a character in `[\x00-\x20]` after entity decoding. The existing `SAFE_LINK_PROTOCOLS` allow-list is the right shape — it is only being asked the wrong question.

Given `markdown.test.ts` already covers `javascript:`, `data:` and `mailto:`, the regression test belongs beside them: assert that `[x](jav&#x09;ascript:alert(1))` renders no anchor at all.

### Scope

This is a stored-XSS vector in a component whose stated purpose is rendering **model and user output** safely without asking consumers to sanitise. I have not merged anything and am not pushing a fix without your call on approach.

The harness asserts while it records, which is why this surfaced at all — my first attempt at this clip asserted `javascriptHrefs === 0`, and the failure is what led here.

murdore:
## Sanitiser bypass fixed (`1d0c45e`) — plus a pass over every open thread

https://github.com/user-attachments/assets/55d44d73-88a4-47f1-811d-69a01ff21a65

```
hostile fixture after the fix
  anchors emitted                  : 0   (was 1)
  anchors resolving to javascript: : 0   (was 1)
  script elements                  : 0
  link text still rendered as text : yes
```

### The fix

`hasSafeProtocol` was handed the href *before* character-reference decoding while the browser resolved it *after*, so the two disagreed about the same string. It now normalises the href the way the browser will see it — decode references, then strip what the WHATWG URL parser strips — and checks the protocol on the result:

```ts
const NUMERIC_REFERENCE = /&#(x[0-9a-f]+|[0-9]+);?/gi;
const NAMED_WHITESPACE_REFERENCE = /&(tab|newline);/gi;
const STRIPPED_BY_URL_PARSER = /[- ]/g;
```

Checked against nine attack forms and seven legitimate hrefs — **9/9 blocked, 7/7 still allowed** — including `&#x09;`, `&#9;`, `&Tab;`, `&NewLine;`, a leading space and `JaVaScRiPt:`, while `https:`, `mailto:`, `tel:`, relative paths and `#anchor` are untouched.

### The test that let it through

@Tara-ag's thread on `markdown.test.ts:113` asked for exactly the assertion that was missing. The old test read:

```ts
// By design the anchor MAY survive: ... attribute-escaping keeps it inert.
expect(output).not.toContain('javascript:');
```

The emitted href was `jav&#x09;ascript:alert(1)`, which does not contain the literal substring `javascript:` — so it passed, while the browser decoded the reference and resolved the anchor's protocol to `javascript:`. **A substring check on serialised HTML cannot see this.** Replaced with a table-driven test over six smuggling variants asserting that no `<a>` is emitted at all, and the comment now records why the old form was unsound rather than leaving the next reader to rediscover it.

**Negative control:** reverting only the normalisation call fails exactly 4 tests — the four character-reference variants — then restores byte-identical. The whitespace and mixed-case cases pass either way, because `new URL()` already handled those; only the entity forms needed the fix.

### The other threads, each checked

| thread | status |
|---|---|
| `marked` peer dep not optional (x3) | **already fixed** — `peerDependenciesMeta` marks it optional on the current head |
| External links lack `rel="noopener"` | **already fixed** — `annotateExternalLinks` adds `target="_blank" rel="noopener noreferrer"` |
| `copyResetTimer` not cleared on destroy | **already fixed** — `ChatMessage.svelte:42-46` clears it in `onDestroy` |
| Missing autolink / bare-URL coverage (x2) | the *unsafe* case was already covered at line 69; added the missing **positive** half — that safe autolinks still work, which is the regression this change could plausibly have caused |
| `classes` produces `class-es` | **not reproducible.** Svelte kebab-cases only at capitals, so `classes` stays `classes`; `class-es` appears **0 times** in the built bundle, and 75 of 76 wrappers use the identical form |
| Module-scoped `instances` Map shared across SSR | kept, and **pinned with a test**: interleaving inputs and both `breaks` modes through the cached instances is order-independent. If a future change makes an instance stateful, that test fails |

`pnpm run lint` exit 0 · unit tests **152 passed**.

One note on the lint: the control-character class trips `no-control-regex`, which is a fair rule to have. I added a targeted disable with the reason inline rather than reshaping the regex, since matching those characters is precisely the point.

murdore:
Rebased onto `release` (`b93b56d`) and force-pushed — the PR had gone `CONFLICTING`. Still one commit: `2cce036`.

One real conflict, in `package.json`. `release` had since promoted `lottie-web` to a top-level peer dependency, while this branch was adding `marked` alongside its own copy of `lottie-web`. Taking either side wholesale loses something, so the resolution is the union in `release`'s alphabetical order:

```json
"peerDependencies": {
  "lottie-web": ">=5.0.0",
  "marked": "^18.0.0",
  "svelte": "^5.41.2",
  "type-decoder": "^2.1.0"
}
```

`peerDependenciesMeta` keeps both `lottie-web` and `marked` marked `optional: true`, which is what the review threads asked for — `marked` is only needed by consumers who render markdown.

Re-verified on the rebased head, not on the pre-rebase one: `pnpm lint` exit 0, and `src/lib/MarkdownText/markdown.test.ts` **24 passed** — including the six character-reference smuggling variants (`&#x09;`, `&#9;`, `&Tab;`, `&NewLine;`, leading whitespace, mixed case) that the sanitiser fix closes.

murdore:
## Rebased onto `release` (`3486dd6`) — `be15604`, still one commit

Clean this time; the `package.json` peer-dependency resolution from the previous rebase carried through unchanged:

```json
"peerDependencies": { "lottie-web": ">=5.0.0", "marked": "^18.0.0", "svelte": "^5.41.2", "type-decoder": "^2.1.0" }
"peerDependenciesMeta": { "lottie-web": { "optional": true }, "marked": { "optional": true } }
```

## The red `build` check was never this branch

It was `release` itself — the Pages prerender break fixed in #485, which has now merged. Re-verified here by running the workflow's own command rather than waiting to be told:

| | result |
|---|---|
| `BASE_PATH=/svelte-ui-components pnpm build` | **exit 0**, `build/` 7.8 MB, zero `does not begin with base` errors |
| `pnpm lint` | exit 0 |
| `src/lib/MarkdownText/markdown.test.ts` | **24 passed** |

The 24 include the six character-reference smuggling variants — `&#x09;`, `&#9;`, `&Tab;`, `&NewLine;`, leading whitespace and `JaVaScRiPt:` — that the sanitiser fix closes, plus the positive autolink case proving safe links still render.

Re-run on the rebased head, not carried over from the pre-rebase run.

package.json
murdore: Fixed in this PR (merge commit be6fd57f6). package.json now declares peerDependenciesMeta.marked.optional = true alongside lottie-web (see package.json:67-74 on release 3.1.0), so consumers who never use markdown rendering won't get unmet-peer warnings or failures under strict-peer-dependency installs.

src/lib/MarkdownText/markdown.test.ts
murdore: This is covered, just not with the literal snippet suggested. markdown.test.ts:60-72 on release 3.1.0 ('applies the protocol allow-list to autolinks and bare URLs too') asserts a bare https URL keeps href="https://example.com" and an angle-bracket autolink with a javascript: scheme is stripped and emits no anchor; markdown.test.ts:133-136 covers the safe-autolink case again. A literal bare 'javascript:alert(1)' with no www./http(s) prefix never becomes a link token under marked's GFM autolink extension, so it wouldn't exercise renderer.link either way - the angle-bracket case is the meaningful equivalent and it's tested.

PR #471, opened 2026-08-29
feat(status): add descriptionSnippet, children, and panel CSS variables

murdore:
## Why

`Status` could not be adopted by an inline consumer. Three separate reasons, all found while migrating Lighthouse's own local `Status` — four onboarding and webhook screens — onto this one. Each fix is additive and defaults to the previous behaviour.

### 1. `descriptionSnippet`

`statusDescription` is interpolated with `{@html}`:

```svelte
<div class="status-description">
  <!-- eslint-disable-next-line -->
  {@html statusDescription}
</div>
```

Right for a caller passing its own trusted markup. Wrong for a message that arrives from an API, a user, or anywhere the caller does not control — and there was no other option, so such a caller either accepted the sink or did not use the component. One of the four call sites does exactly this:

```ts
stepMessage = action.message;   // from getPaymentsPluginAction()
```

`descriptionSnippet` takes rendering priority and renders through ordinary Svelte interpolation, which escapes.

### 2. `children`

`.status-description` carries the component's own horizontal padding (`42px`) and bottom margin (`25px`), so anything rendered through `descriptionSnippet` inherits a text box's geometry. Right for the message, wrong for a control — and there was nowhere else to put one, since `buttonProperties` takes props rather than content. `children` renders below the description, **outside** that box, where the button already does.

### 3. `--status-panel-background` / `--status-panel-backdrop-filter`

`Status` is styled as a standalone result screen: `100vh` tall with a translucent white frosted panel. `min-height` was already variable-backed; the panel was not, sitting behind an `@supports` block on an inner element that `classes` cannot reach. Inside a page with its own heading and following content, the height pushes siblings below the fold and the white panel does not survive a dark theme.

All three mirror **`EmptyState`**, which already carries a `description` string beside a `descriptionSnippet` override and a `children` action area. The shapes are the library's own, not new conventions.

## Backwards compatibility

Purely additive. Every existing consumer is unchanged: the snippets are `undefined` so both `{#if}`s fall through, and the two new variables default to the literals they replaced. The web-component wrapper is untouched — snippets don't cross the custom-element boundary, which is why the existing `icon` snippet isn't there either.

## Verification — a negative control per behaviour

**9 specs across four files, all green.**

The description pair hands two demo instances the **same string**: one asserts the `{@html}` path still parses it as markup, the other that the snippet path renders it as literal text. Either alone would pass against a broken implementation.

| Control | Result |
|---|---|
| Snippet branch → `{#if false}` | fails **only** the escape assertion (line 29) |
| Panel variables → back to literals | fails **only** the neutralise assertion — `rgba(255,255,255,0.6)` received where `rgba(0,0,0,0)` expected |
| `children` moved inside `.status-description` | fails **only** `insideDescription` (line 26) |

Two of those deserve a note:

- When the panel variables are reverted, **"defaults keep the full-screen panel exactly as it was" stays green.** That is the backwards-compatibility proof — the defaults render identically to the code they replaced.
- The `children` control leaves the content **present and visible**. A presence-only assertion would have passed the broken version; the test asserts *containment* instead, which is the actual contract.

All controls restored byte-identical and re-run green. `svelte-check` 651 files / **0 errors**; `prettier --check src` and `eslint src` clean.

## Measured against the real consumer

By computed style, not screenshot — deliberately. The consuming route transitions on a timer, and **the same build captured twice differs by 1.273%**, so a pixel diff cannot separate the change from the route's own noise.

After adoption: `min-height: 0px`, panel `background-color: rgba(0,0,0,0)`, `backdrop-filter: none`, and the message colour `rgb(82,82,82)` — byte-identical to what the pre-migration markup inherited.

## Docs

`docs/Status.md` gains a `## Snippets` section — which also documents the **pre-existing `icon` snippet**, previously undocumented — the two new CSS variables, the `testId` prop, and worked examples for the untrusted description, the description-vs-action split, and inline embedding.

### DIFF
--- a/src/lib/Status/properties.ts
+++ b/src/lib/Status/properties.ts
@@ -13,6 +13,29 @@ export type StatusProperties = StatusEventProperties & {
    * priority over `statusIcon` when provided.
    */
   icon?: Snippet;
+  /**
+   * Optional snippet that replaces the `statusDescription` string at render
+   * time. When provided, `statusDescription` is not rendered — only the snippet
+   * is. Mirrors `EmptyState`'s `descriptionSnippet`.
+   *
+   * `statusDescription` is interpolated with `{@html}`, which is what a caller
+   * supplying its own trusted markup wants, but is the wrong tool for a message
+   * that originates from an API, a user, or any other source the caller does not
+   * control. There was previously no way to render such a message as text. Use
+   * this snippet for that: `{#snippet descriptionSnippet()}{message}{/snippet}`
+   * escapes it the way ordinary Svelte interpolation does.
+   */
+  descriptionSnippet?: Snippet;
+  /**
+   * Action area rendered below the description, alongside the optional
+   * `buttonProperties` button. Mirrors `EmptyState`'s `children`.
+   *
+   * `descriptionSnippet` sits inside `.status-description`, which carries the
+   * component's own horizontal padding and bottom margin. Content that is not
+   * part of the description — a button, a link, a countdown — belongs here
+   * instead, outside that box, where `buttonProperties` already renders.
+   */
+  children?: Snippet;
   testId?: string;
 };
 
--- a/src/routes/components/status/+page.svelte
+++ b/src/routes/components/status/+page.svelte
@@ -4,6 +4,11 @@
 
   // Minimal valid Lottie document (zero layers) — enough to exercise the player
   // without depending on a real animation asset.
+  // A description the caller does not control — the shape of anything that
+  // arrives from an API. `statusDescription` interpolates with {@html}, so the
+  // markup in this string is parsed as markup; `descriptionSnippet` escapes it.
+  const untrustedDescription = 'Install failed <span data-pw="injected-markup">injected</span>';
+
   const minimalAnimation = {
     v: '5.5.7',
     fr: 30,
@@ -39,6 +44,26 @@
           <LottiePlayer animationData={minimalAnimation} testId="status-lottie" />
         </div>
       {/snippet}
+    </Status>
+  </div>
+  <div data-pw="status-description-html">
+    <Status statusText="String prop" statusDescription={untrustedDescription} />
+  </div>
+
+  <div data-pw="status-description-snippet">
+    <Status statusText="Snippet" statusDescription={untrustedDescription}>
+      {#snippet descriptionSnippet()}{untrustedDescription}{/snippet}
+    </Status>
+  </div>
+  <div
+    data-pw="status-inline"
+    style="--status-min-height: auto; --status-panel-background: transparent; --status-panel-backdrop-filter: none;"
+  >
+    <Status statusText="Inline" statusDescription="Embedded in a page, not a full screen." />
+  </div>
+  <div data-pw="status-children">
+    <Status statusText="With action" statusDescription="Something needs your attention.">
+      <div data-pw="status-children-content">Action area</div>
     </Status>
   </div>
 </div>
--- a/docs/Status.md
+++ b/docs/Status.md
@@ -1,6 +1,8 @@
 # Status
 
-A full-screen status display with a centered icon image, title text, description text (supports HTML), and an optional action Button. Uses backdrop-filter blur for visual effect. Ideal for order success/failure screens.
+A full-screen status display with a centered icon image, title text, description text, and an optional action Button. Uses backdrop-filter blur for visual effect. Ideal for order success/failure screens.
+
+The `statusDescription` string is rendered with `{@html}`, so a caller can pass its own trusted markup. For a description that comes from an API, a user, or anywhere else the caller does not control, use the `descriptionSnippet` snippet instead — it renders the text escaped.
 
 ## Usage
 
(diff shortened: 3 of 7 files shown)


Tara-ag:
💡 MINOR: Status web-component wrapper omits documented snippet props and testId

`docs/Status.md` now documents `icon`, `descriptionSnippet`, and `testId` as public props, but `src/wc/components/Status.wc.svelte` does not declare them in its `customElement.props` map or forward them as snippet slots. This creates a mismatch between the Svelte and web-component public contracts. The library already has a working pattern in `src/wc/components/EmptyState.wc.svelte` that exposes snippets via named slots.

Note: this file is not in the PR diff, so I could not attach an inline comment. Consider aligning the web-component contract with the newly documented Svelte contract in a follow-up.

Suggested update to `src/wc/components/Status.wc.svelte`:
```svelte
<svelte:options
  customElement={{
    tag: 'sui-status',
    shadow: 'open',
    props: {
      statusIcon: { type: 'String', reflect: true, attribute: 'status-icon' },
      statusText: { type: 'String', reflect: true, attribute: 'status-text' },
      statusDescription: { type: 'String', reflect: true, attribute: 'status-description' },
      buttonProperties: { type: 'Object' },
      classes: { type: 'String' },
      onbuttonClick: { type: 'Object' },
      testId: { type: 'String', reflect: true, attribute: 'test-id' }
    }
  }}
/>

<script lang="ts">
  import Status from '$lib/Status/Status.svelte';
  let props = $props();
</script>

<Status {...props}>
  {#snippet icon()}
    <slot name="icon"></slot>
  {/snippet}
  {#snippet descriptionSnippet()}
    <slot name="description-snippet"></slot>
  {/snippet}
</Status>
```

Tara-ag (approved):
Approved with one minor note about the web-component contract. The Svelte-side change is purely additive, well-tested, and correctly documented.

Tara-ag:
💡 MINOR: Status web-component wrapper omits documented snippet props and testId

`docs/Status.md` now documents `icon`, `descriptionSnippet`, and `testId` as public props, but `src/wc/components/Status.wc.svelte` does not declare them in its customElement props map or pass them as snippet slots. This creates a mismatch between the Svelte and web-component public contracts. The library already has a working pattern in `src/wc/components/EmptyState.wc.svelte` that exposes snippets via named slots.

I could not attach this as an inline review comment because the web-component wrapper is not part of this PR's diff.

Suggested fix: update the customElement props to include `testId` and forward snippet slots to `<Status>`:

```svelte
<svelte:options
  customElement={{
    tag: 'sui-status',
    shadow: 'open',
    props: {
      statusIcon: { type: 'String', reflect: true, attribute: 'status-icon' },
      statusText: { type: 'String', reflect: true, attribute: 'status-text' },
      statusDescription: { type: 'String', reflect: true, attribute: 'status-description' },
      buttonProperties: { type: 'Object' },
      classes: { type: 'String' },
      onbuttonClick: { type: 'Object' },
      testId: { type: 'String', reflect: true, attribute: 'test-id' }
    }
  }}
/>

<script lang="ts">
  import Status from '$lib/Status/Status.svelte';
  let props = $props();
</script>

<Status {...props}>
  {#snippet icon()}
    <slot name="icon"></slot>
  {/snippet}
  {#snippet descriptionSnippet()}
    <slot name="description-snippet"></slot>
  {/snippet}
</Status>
```

Tara-ag (approved):
Approved with one minor note about the web-component wrapper contract (left as a PR comment since the file is not in this diff).

Tara-ag (commented):
Review in progress — filing findings as inline comments.

src/routes/components/status/+page.svelte
```diff
@@ -4,6 +4,11 @@
 
   // Minimal valid Lottie document (zero layers) — enough to exercise the player
   // without depending on a real animation asset.
+  // A description the caller does not control — the shape of anything that
+  // arrives from an API. `statusDescription` interpolates with {@html}, so the
+  // markup in this string is parsed as markup; `descriptionSnippet` escapes it.
+  const untrustedDescription = 'Install failed <span data-pw="injected-markup">injected</span>';
```
Tara-ag, line 10: 💬 SUGGESTION: The variable name `untrustedDescription` is clear, but consider adding an explicit comment that this simulates an API-returned message so the security-oriented demo is immediately obvious to readers.

Tara-ag (commented):
Review findings — see inline comments and additional comments below for files not in the diff.

Tara-ag:
💡 MINOR: `src/wc/components/Status.wc.svelte:8` — The web-component wrapper does not expose `testId` even though the Svelte component supports it and the docs now document it. Add `testId: { type: 'String', reflect: true, attribute: 'test-id' }` so `<sui-status test-id="...">` sets `data-pw` on the rendered root, keeping Svelte and web-component APIs in parity.

Tara-ag:
💡 MINOR: `src/lib/Status/properties.ts:7-8` — `statusText` and `statusDescription` are typed as required, but `Status.svelte` destructures them with defaults (`''`) and the docs/example show `<Status />`. TypeScript consumers following the documented API will get a compile error. Make both props optional (`statusText?: string; statusDescription?: string;`) so the type contract matches the runtime defaults and the docs.

src/routes/components/status/+page.svelte
Tara-ag, line 10: 💬 SUGGESTION: Consider adding an explicit comment that this variable simulates an API-returned message so the security-oriented demo is immediately obvious to readers.

src/lib/Status/properties.ts
Tara-ag, line 1: 💡 MINOR: `statusText` and `statusDescription` are declared as required in `StatusProperties` (around line 7), but the component destructures them with `''` defaults and the docs show `<Status />` as valid. TypeScript consumers will get a compile-time error for the documented minimal example. Make both props optional (`statusText?: string; statusDescription?: string;`) to match the runtime defaults and documented API.

Tara-ag (commented):
Yama review findings

Tara-ag:
💡 MINOR (non-diff file): `src/wc/components/Status.wc.svelte` — The Svelte component exposes `testId` and the docs now document it, but the web-component `props` map does not include it. Add `testId: { type: 'String', reflect: true, attribute: 'test-id' }` so `<sui-status test-id="...">` works consistently with the Svelte API. (Could not be posted inline because this file is not part of the PR diff.)

Tara-ag (approved):
Approved. The changes are purely additive and well-scoped: new optional `descriptionSnippet` and `children` snippets, plus two new CSS custom properties with defaults identical to the previous literals. Backward compatibility is preserved, the web-component wrapper is appropriately left untouched (snippets do not cross the custom-element boundary), and the new behaviors are covered by Playwright tests with negative controls. Documentation is updated comprehensively.

PR #472, opened 2026-08-30
feat(toolbar): expose row-layout variables so it can serve an in-flow page header

murdore:
## What this is

Lighthouse hand-rolls an in-flow page header (76 call sites) that is structurally this
component: left / centre / right regions over an optional second row. Toolbar defaults to a
**fixed chrome bar**. This makes the second shape reachable.

## It adds no props

The markup, the props and the TypeScript are **byte-identical to `release`**. The whole change
is 26 lines of CSS — variables on the three region elements, each defaulting to the value the
component already rendered:

```
--toolbar-content-align-items / -min-height / -flex-wrap / -row-gap / -column-gap
--toolbar-center-flex / -min-width
--toolbar-right-display / -flex-shrink / -min-width / -max-width / -width
```

The action region's **inner** layout is deliberately not tokenised — that is the consumer's own
snippet markup, which it can style directly.

## What an earlier revision of this PR got wrong

It added `subheading`, `headingLevel`, `subheadingLevel`, `headingClasses` and
`subheadingClasses`, and swapped the default back icon. All reverted:

- **`docs/Toolbar.md` already rejected it.** "String props that carry presentational structure
  are rejected in favour of Snippets." A title with a subline is `centerContent`, and always
  was. The consuming app now keeps its own markup, semantic tags and type-scale classes.
- **A class-name prop is inert in the web-component build.** `sui-toolbar` is `shadow: 'open'`,
  so a class passed as `heading-classes` lands inside the shadow root where the consumer's
  stylesheet cannot reach it — and nothing reports the failure.
- **The back-icon swap changed every existing consumer.** `backIcon` already defaulted to
  `https://sdk.breeze.in/gallery/icons/back.svg` (a long arrow, `fill="#020202"`); replacing it
  with an inline `currentColor` chevron altered both shape and colour for consumers who never
  asked. `backIcon` and its `` are back exactly as they were.

If a future consumer genuinely needs the library to render its title text, the fix is
`:where(.text) { … }` — measured, Svelte compiles that to `:where(.text.svelte-hash)`, which
stays **(0,0,0)**, so even a bare tag selector on the consumer's side wins. A plain `.text`
compiles to `.text.svelte-hash` at (0,2,0) and beats a consumer's utility class, which is the
trap the class props were reaching for.

## Harness fix

`playwright.config.ts` pinned port 43199 with `reuseExistingServer`. Two checkouts of this repo
therefore collide: a run here bound to a **sibling worktree's** preview server and asserted
against the wrong build — failing on properties that are correct in this checkout, then
flooding `ERR_CONNECTION_REFUSED`. A shorter suite would simply have reported green against
code that was never built. Port is now `PW_PORT`-overridable, default unchanged.

## Tests

5 new cases. The important one asserts an **unconfigured** Toolbar is untouched: title still a
`div` at 18px/400, back control still the same `` from the same URL, and every new
declaration resolving to its previous computed value (`align-items: center`, `flex-wrap:
nowrap`, `row-gap: 0px`, `column-gap: 0px`, `min-height: auto`). The rest cover the tokens
producing the page-header shape and the consumer's markup keeping its tags and type scale.

Full suite: **201 passed**. 15 failures are pre-existing Status/Table cases failing on
`order-success-icon.svg`, which is not in the repo (`git ls-files` finds no such asset) and is
untouched by this branch.

## Consumer

juspay/lighthouse PR 7298 renders `PageHeader` through this. Verified against `beta` with a
30-cell matrix (5 routes x 3 viewports x light/dark): **30/30 at 0.000%** pixel difference, with
a same-build control at 0.000% and a negative control that moves exactly the 6 `/ai` cells.

### DIFF
--- a/docs/Toolbar.md
+++ b/docs/Toolbar.md
@@ -1,6 +1,9 @@
 # Toolbar
 
-A fixed-position header bar with a back button (left), center title text, and customizable left/center/right content areas via Snippet slots. The `additionalContent` snippet renders a second row below the main toolbar content. If `leftContent` snippet is provided, it replaces the default back button. If `centerContent` snippet is provided, it replaces the `text` prop.
+A header bar with a back button (left), center title text, and customizable left/center/right content areas via Snippet slots. It defaults to a **fixed** chrome bar, but every positioning and row-layout axis is a CSS
+variable, so the same component also serves an **in-flow page header** — see "Page-header
+shape" below. The `additionalContent` snippet renders a second row below the main toolbar
+content. If `leftContent` snippet is provided, it replaces the default back button. If `centerContent` snippet is provided, it replaces the `text` prop.
 
 ## Usage
 
@@ -115,3 +118,68 @@ The library Toolbar intentionally does not offer a `subheading` prop — string 
 | `center-content`     | `centerContent`     | Content in the center of the toolbar.          |
 | `right-content`      | `rightContent`      | Content on the right side of the toolbar.      |
 | `additional-content` | `additionalContent` | Additional content below the main toolbar row. |
+
+## Page-header shape
+
+The defaults describe a fixed chrome bar. The same component renders an in-flow page header
+with **no additional props** — only CSS variables and the snippets it already has:
+
+```svelte
+<Toolbar showBackButton onbackClick={goBack}>
+  {#snippet centerContent()}
+    <div class="heading-block">
+      <h2>Shipping profiles</h2>
+      <p>Set delivery rates and zones for this store</p>
+    </div>
+  {/snippet}
+  {#snippet rightContent()}
+    <div class="actions">…</div>
+  {/snippet}
+</Toolbar>
+```
+
+```css
+.page-header {
+  --toolbar-position: relative;
+  --toolbar-width: 100%;
+  --toolbar-background: transparent;
+  --toolbar-box-shadow: none;
+
+  /* Top-align the row so the back control sits on the title's first line rather than
+     centred against the whole two-line block. */
+  --toolbar-content-align-items: flex-start;
+  --toolbar-content-column-gap: 16px;
+
+  /* The title side yields and ellipsizes; the actions stay whole. */
+  --toolbar-center-flex: 1 1 auto;
+  --toolbar-center-min-width: 0;
+  --toolbar-right-flex-shrink: 0;
+}
+```
+
+Note what is **not** here. There is no `subheading`, `headingLevel` or `headingClasses` prop:
+the title block is the consumer's own markup in `centerContent`, so it keeps its semantic tags
+and its design system's type classes without the library restating them. That follows the same
+reasoning as "Subheading (consumer recipe)" above — presentational structure belongs in a
+Snippet, and how a title *looks* is reachable with ordinary CSS on the consumer's own elements.
+
+Responsive behaviour is the consumer's too. The component ships **no media queries** — it
+exposes the mechanism and the app picks the breakpoints, because "wrap the actions below 1024px"
+and "hide the subheading on a phone" are design decisions the library cannot make for everyone.
+
+### Additional CSS variables
+
+| Variable                                    | Default | Description                                                                                           |
+| ------------------------------------------- | ------- | ----------------------------------------------------------------------------------------------------- |
+| `--toolbar-content-align-items`             | `center` | Cross-axis alignment of the main row. `flex-start` top-aligns a title that has a subline under it.   |
+| `--toolbar-content-min-height`              | `auto`  | Floor for the main row's height — a fixed-height bar sets it here, not on the outer column.           |
+| `--toolbar-content-flex-wrap`               | `nowrap` | Lets the action side drop to its own line on a narrow viewport.                                      |
+| `--toolbar-content-row-gap` / `-column-gap` | `0`     | Gaps between the row's regions, and between wrapped lines.                                            |
+| `--toolbar-center-flex`                     | `1`     | `1 1 auto` lets the centre region grow from its CONTENT width, so a deficit is shared with the actions rather than collapsing the title. |
+| `--toolbar-center-min-width`                | `auto`  | Set `0` so a long title can shrink and ellipsize.                                                      |
+| `--toolbar-right-flex-shrink`               | `1`     | Set `0` so actions never get crushed by a long title.                                                  |
+| `--toolbar-right-min-width` / `-max-width` / `-width` | `auto` / `none` / `auto` | Action-region sizing as a flex item. Its INNER layout stays the snippet's own markup. |
+| `--toolbar-right-display`                   | `block` | Set `flex` only if you want the region itself to be the flex container.                                |
+
+Every default above resolves to the value the component already rendered, so adding them
+changes nothing for an existing consumer.
--- a/src/lib/Toolbar/Toolbar.svelte
+++ b/src/lib/Toolbar/Toolbar.svelte
@@ -72,7 +72,7 @@
   .content {
     display: flex;
     flex-direction: row;
-    align-items: center;
+    align-items: var(--toolbar-content-align-items, center);
     padding: var(--toolbar-content-padding, 0px);
     justify-content: var(--toolbar-justify-content, normal);
     visibility: var(--toolbar-content-visibility, visible);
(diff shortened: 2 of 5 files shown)


Tara-ag (commented):
Review in progress.

src/lib/Toolbar/Toolbar.svelte
```diff
@@ -25,19 +40,61 @@
   >
     {#if typeof leftContent === 'function'}
       {@render leftContent()}
-    {:else if showBackButton && typeof backIcon === 'string' && backIcon.length > 0}
+    {:else if showBackButton && showBackIcon}
```
Tara-ag: ⚠️ MAJOR: Default inline back glyph leaves the button without an accessible name. When `backIcon` is omitted the component now renders an inline SVG instead of the historical `Back`. The SVG is `aria-hidden="true"` and the parent `<div role="button">` has no `aria-label`, so screen readers announce the control only as "button" with no name. This is a regression from the previous default, where the image's `alt` text provided the accessible name.

Suggested fix: add `aria-label="Back"` to the `.back` div so the control has an accessible name in both the URL and inline-glyph cases:
```svelte
<div class="back" onclick={onbackClick} {onkeydown} role="button" tabindex="0" aria-label="Back">
```

src/lib/Toolbar/Toolbar.svelte
```diff
@@ -109,24 +172,85 @@
...
+  }
+
+  /* Deliberately minimal. An earlier revision routed colour, typography and margin through
+     `var(--toolbar-subheading-*)` with no fallback, on the theory that an unset property would
+     let the consumer's own `p` / utility-class rules win. It does the opposite: an undefined
+     custom property makes the declaration invalid at computed-value time, which computes to
+     `unset` — inherit for colour, initial for margin — and that BEATS the lower-priority rules
+     it was meant to defer to. The subheading silently rendered inherited black with no margin.
+
+     So the library styles only what it must. Everything else is the consumer's, reached through
```
Tara-ag: ⚠️ MAJOR: Documented subheading CSS variables are not implemented in component CSS. The docs and demo advertise `--toolbar-subheading-font-size`, `-font-weight`, `-line-height`, `-color`, `-margin`, and `-padding`, but the `.subheading` rule only consumes `--toolbar-subheading-display`. The other variables are dead: setting them has no effect. This breaks the public theming contract and the demo page uses ineffective tokens (`--toolbar-subheading-font-size: 13px; --toolbar-subheading-color: #59616e;`).

Either remove those variables from the docs/demo (consistent with the comment that subheading styling is the consumer's responsibility) or implement them with explicit fallback values, e.g.:
```css
.subheading {
  display: var(--toolbar-subheading-display, block);
  font-size: var(--toolbar-subheading-font-size, inherit);
  font-weight: var(--toolbar-subheading-font-weight, inherit);
  line-height: var(--toolbar-subheading-line-height, inherit);
  color: var(--toolbar-subheading-color, inherit);
  margin: var(--toolbar-subheading-margin, 0);
  padding: var(--toolbar-subheading-padding, 0);
}
```

docs/Toolbar.md
```diff
@@ -115,3 +119,60 @@ The library Toolbar intentionally does not offer a `subheading` prop — string
 | `center-content`     | `centerContent`     | Content in the center of the toolbar.          |
```
Tara-ag, line 118: 💡 MINOR: Docs still claim Toolbar intentionally has no subheading prop. This section says "The library Toolbar intentionally does not offer a `subheading` prop" and shows how to emulate it with `centerContent`. That directly contradicts the new `subheading` prop added by this PR and will mislead consumers. Replace or remove this section now that a first-class prop exists.

src/lib/Toolbar/Toolbar.svelte
```diff
@@ -109,24 +172,85 @@
...
+  .subheading {
+    display: var(--toolbar-subheading-display, block);
+  }
+
+  /* Unstyled block by default, exactly as before — every property below resolves to its own
+     CSS initial value until a consumer opts in. */
+  .right-content {
+    display: var(--toolbar-right-display, block);
+    flex-shrink: var(--toolbar-right-flex-shrink, 1);
+    flex-wrap: var(--toolbar-right-flex-wrap, nowrap);
```
Tara-ag: ⚠️ MAJOR: .right-content defaults changed from `flex:1` to `display:block`, breaking existing `rightContent` layouts. The original rule was `display: flex; flex: 1;`. The PR replaces it with `display: var(--toolbar-right-display, block);` and removes `flex: 1`. For an existing consumer using the `rightContent` snippet without overriding `--toolbar-right-display`, the region now renders as a block-level element that does not grow. The comment claims "Unstyled block by default, exactly as before", but the previous default was a flexing flex container. This is a silent visual regression for an existing public API surface.

Suggested fix: restore the previous default while keeping the new variables:
```css
.right-content {
  display: var(--toolbar-right-display, flex);
  flex: 1;
  flex-shrink: var(--toolbar-right-flex-shrink, 1);
  flex-wrap: var(--toolbar-right-flex-wrap, nowrap);
  min-width: var(--toolbar-right-min-width, auto);
  max-width: var(--toolbar-right-max-width, none);
  width: var(--toolbar-right-width, auto);
  gap: var(--toolbar-right-gap, 0);
  justify-content: var(--toolbar-right-justify-content, normal);
  align-items: var(--toolbar-right-align-items, normal);
}
```

src/wc/components/Toolbar.wc.svelte
```diff
@@ -5,10 +5,16 @@
     props: {
       showBackButton: { type: 'Boolean', reflect: true, attribute: 'show-back-button' },
       text: { type: 'String', reflect: true },
+      subheading: { type: 'String', reflect: true },
+      headingLevel: { type: 'String', attribute: 'heading-level' },
+      subheadingLevel: { type: 'String', attribute: 'subheading-level' },
```
Tara-ag: 💡 MINOR: Web component prop type for `headingLevel`/`subheadingLevel` lacks validation. They are declared as `type: 'String'` and map directly to `<svelte:element this={headingLevel}>`. Invalid values like `heading-level="span"` will render an invalid element at runtime. The Svelte TypeScript types restrict these, but the web-component wrapper does not enforce them. Consider adding runtime fallback in Toolbar.svelte or documenting that invalid attribute values are passed through.

Tara-ag (changes requested):
Thanks for the thorough PR description and the careful attention to backward compatibility. I found a few issues that need addressing before merge:

1. **Accessibility regression**: the new inline back glyph is `aria-hidden` and its parent button has no accessible name, so screen readers announce only "button". Add `aria-label="Back"` to the `.back` div.
2. **Broken theming contract**: several `--toolbar-subheading-*` CSS variables are documented and used in the demo but not implemented in the component CSS. Either implement them or remove them from docs/demo.
3. **Silent layout regression**: `.right-content` changed from `display: flex; flex: 1;` to `display: var(--toolbar-right-display, block);` with no `flex: 1`. This can break existing `rightContent` consumers. Restore the flex default while keeping the new variables.
4. **Docs contradiction**: the "Subheading (consumer recipe)" section still claims Toolbar intentionally has no `subheading` prop.
5. **Web-component validation**: `heading-level`/`subheading-level` attributes accept any string and pass invalid values to `<svelte:element>`. Consider runtime fallback or documentation.

Please address #1–#3 at minimum; #4–#5 are smaller but worth fixing.

Tara-ag:

The PR is a clean, backward-compatible CSS-only enhancement that lets Toolbar render an in-flow page header via new CSS custom properties, with all defaults preserving prior rendering. Playwright tests and docs are included. However, the gate accepted one MAJOR accessibility finding: the default back control is a focusable `<div role="button">` that does not guarantee keyboard activation (Enter/Space), which affects the new page-header demo and violates the accessibility focus area. This must be fixed before merge.

- ⚠️ **MAJOR**: Back button remains a non-keyboard-accessible div — `src/lib/Toolbar/Toolbar.svelte:23`
  The default back control is a `<div role="button" tabindex="0">` that only forwards `onkeydown` without guaranteeing Enter/Space activation. Consumers relying on the default back button therefore get a control that is focusable but not operable from the keyboard, violating WCAG 2.1.1 and the repo's accessibility focus area. This is pre-existing code in the same hunk, so it is in scope for the revi…

Tara-ag (commented):
Review in progress — posting inline comments file by file.

Tara-ag (approved):
Approved. The change is a pure CSS-variable addition: no props, markup, TypeScript, or exports are modified. All new variables default to the previously-rendered values, preserving backward compatibility. Docs, demo, and Playwright tests are aligned, and the web-component wrapper is untouched because no props changed. The one accessibility concern about the default back button is pre-existing and unchanged by this diff.

Tara-ag:
⚠️ MAJOR (`src/lib/Toolbar/Toolbar.svelte`, line 23): The default back control is a `<div role="button" tabindex="0">` that only forwards `onkeydown` without guaranteeing Enter/Space activation. Consumers relying on the default back button — including the new page-header demo — therefore get a control that is focusable but not operable from the keyboard, violating WCAG 2.1.1 and the repo's accessibility focus area.

Suggested fix: add an internal keydown handler that invokes `onbackClick` for Enter/Space, or replace the div with `<button type="button" class="back" onclick={onbackClick} {onkeydown}>` so keyboard activation works by default.

PR #473, opened 2026-08-31
feat: close the SUI-side gap list from the polymorph comparison (Phase 1)

murdore:
## What

Phase 1 of closing SUI's own gaps, surfaced by a capability comparison against
`polymorph-ui-components` (a fork of this codebase). Four independent, additive
changes:

1. **`ROADMAP.md` regenerated** against `docs/_index.json` — 44/54 originally-tracked
   rows now correctly show available (was 14/54), plus a new section for the 37
   components shipped since the file was written but never added to it.
2. **Web-component build shipped through npm** — `dist-wc` added to `files`, a `./wc`
   export added, `build:wc` wired into `build`/`prepublishOnly` so the export can't go
   stale, README documents all 3 consumption paths. Verified with a real `pnpm run
   build` including `publint`, not just by reading the config.
3. **`DESIGN_PRINCIPLES.md` + event-casing enforcement** — the architectural principles
   this library already mostly follows, written down for the first time, plus
   `scripts/check-event-casing.js` (wired into `npm run lint`) with a baseline that
   grandfathers the 97 pre-existing casing inconsistencies rather than mass-renaming a
   public API.
4. **New `.github/workflows/ci.yml`** — there was no workflow that ran lint,
   type-check, or tests on pull requests at all. Now: lint + svelte-check + the vitest
   unit suite (128/128 passing) block the PR; the Playwright integration suite runs and
   uploads a report but stays `continue-on-error` — a clean run today shows 130/211
   specs failing, which needs its own triage pass before it can safely gate merges.

## Verification

- `pnpm run build` (vite build + build:wc + svelte-package + publint) succeeds
- `pnpm run lint` passes (97 known event-casing violations, 0 new)
- `pnpm run check` passes (0 errors, 4 pre-existing warnings)
- `pnpm run test:unit -- --run` passes (128/128)
- `pnpm run test:integration` does NOT currently pass (130/211 failing) — flagged
  explicitly above and in the CI workflow, not hidden

## Not in this PR

- OIDC trusted publishing — separate PR #474, blocked on a 2FA-gated npm registry step
  only a human with `@juspay` npm access can complete
- Triaging the 130 failing integration tests — real, separate work; this PR makes the
  failures visible on every PR instead of invisible, it doesn't fix them


Tara-ag (commented):
Reviewed ROADMAP.md against docs/_index.json. Found two minor documentation accuracy issues; see inline comments.

Tara-ag:
💡 MINOR (ROADMAP.md:189): `docs/_index.json` does contain a `TaskList` entry with a full description, so claiming it has "no entry" is inaccurate. Remove `TaskList` from this list, or clarify if "TaskList alias" refers to a separate component distinct from the indexed one.

Tara-ag:
💡 MINOR (ROADMAP.md:159): The section intro says "These 37 components exist and ship today", but the enumerated lists total 38 components (7 charts + 16 chat/agent + 13 other + 2 utility/service modules). Update the count to 38 to match the list.

Tara-ag (commented):
Reviewing documentation and package surface changes.

Tara-ag:
## Yama review findings

### README.md

**Line 158** — 💡 MINOR: README claims every component ships as a web component, but some lack wrappers. `src/lib/index.ts` exports `FileDropzoneTrigger` and `KeyValue`, yet `src/wc/index.ts` registers neither, and `src/wc/components` has no matching `.wc.svelte` files. Consumers following this guidance will find `<sui-file-dropzone-trigger>` and `<sui-key-value>` undefined at runtime.

*Suggested fix:* Change the claim to "Most components also ship as Custom Elements" and list the exceptions, or add `FileDropzoneTrigger.wc.svelte` and `KeyValue.wc.svelte` wrappers and register them in `src/wc/index.ts` if they are intended to be available as custom elements.

**Line 158** — 💬 SUGGESTION: "No Svelte runtime required" is ambiguous. The Svelte runtime is still bundled into the custom-element build; the host page does not need to install it separately. This can mislead consumers about bundle size and runtime dependencies.

*Suggested fix:* Rephrase to "no separate Svelte installation required; the runtime is bundled with each component".

### package.json

**Line 42** — 💬 SUGGESTION: New `./wc` export lacks TypeScript types. The export only defines a `default` condition, so TypeScript consumers importing `@juspay/svelte-ui-components/wc` may receive "Could not find a declaration file" diagnostics.

*Suggested fix:* Add a `types` condition to the `./wc` export pointing to an ambient declaration file (e.g., create `dist-wc/index.d.ts` with `export {}`) so TypeScript treats the import as a valid side-effect import.

**Line 24** — 💬 SUGGESTION: package.json changes extend the public package surface and are not docs-only. The PR title/description frame this as a docs-only ROADMAP.md update, but this file adds a new public subpath export (`./wc`), modifies `build`/`prepublishOnly` scripts, and changes the `files` array. A `docs` conventional commit will not trigger a semver bump, so the new export may ship without a corresponding version-bump announcement.

*Suggested fix:* Split the package.json/build changes into a separate `feat` commit/PR, or update the PR title and description to reflect the full scope so standard-version can apply the correct semver bump.

package.json
```diff
@@ -39,12 +39,16 @@
     ".": {
       "types": "./dist/index.d.ts",
       "svelte": "./dist/index.js"
+    },
```
Tara-ag, line 43: 💬 SUGGESTION: New ./wc export lacks TypeScript types

The new "./wc" export only defines a "default" condition. TypeScript consumers importing '@juspay/svelte-ui-components/wc' may receive "Could not find a declaration file" diagnostics, and the package's top-level types field does not cover this subpath.

Suggestion: Add a "types" condition to the "./wc" export pointing to an ambient declaration file (e.g., create `dist-wc/index.d.ts` with `export {}`) so TypeScript treats the import as valid side-effect import.

README.md
```diff
@@ -153,6 +153,35 @@ This library takes a different approach: **components are unstyled by default**
 
 ---
 
+## Web Components
+
+Every component also ships as a real Custom Element (Svelte 5's native `customElement`
```
Tara-ag, line 158: 💬 SUGGESTION: "No Svelte runtime required" is ambiguous

The README says "no Svelte runtime required by the host page". The Svelte runtime is still bundled into the custom-element build; the host page does not need to install it separately. This can mislead consumers about bundle size and runtime dependencies.

Suggestion: Rephrase to "no separate Svelte installation required; the runtime is bundled with each component".

Tara-ag, line 158: 💡 MINOR: README claims every component ships as a web component, but some lack wrappers

README.md states "Every component also ships as a real Custom Element", but src/wc/index.ts does not register FileDropzoneTrigger or KeyValue, both of which are exported from src/lib/index.ts. Consumers following this guidance will find `<sui-file-dropzone-trigger>` and `<sui-key-value>` undefined at runtime.

Suggestion: Change the claim to "Most components also ship as Custom Elements" and list the exceptions, or add `FileDropzoneTrigger.wc.svelte` and `KeyValue.wc.svelte` wrappers and register them in src/wc/index.ts if they are intended to be available as custom elements.

package.json
```diff
@@ -21,11 +21,11 @@
   "homepage": "https://github.com/juspay/svelte-ui-components",
   "scripts": {
     "dev": "vite dev --host",
-    "build": "vite build && npm run package",
+    "build": "vite build && npm run build:wc && npm run package",
```
Tara-ag, line 24: 💬 SUGGESTION: package.json changes extend the public package surface and are not docs-only

The PR title and description frame this change as a docs-only ROADMAP.md update, but package.json adds a new public subpath export ("./wc"), modifies the build and prepublishOnly scripts, and changes the files published to npm. These are build/public-API changes. A "docs" conventional commit will not trigger a semver bump, so the new export may ship without a corresponding version-bump announcement.

Suggestion: Split the package.json/build changes into a separate `feat` commit/PR, or update the PR title and description to reflect the full scope so standard-version can apply the correct semver bump.

Tara-ag (commented):
Yama review complete. No blocking issues found. Four non-blocking findings posted: README accuracy on web-component coverage, README phrasing ambiguity, missing TypeScript types for the new `./wc` export, and PR scope mismatch (package.json changes are public-API/build changes, not docs-only).

Tara-ag:

Review could not be performed: the time budget was exhausted before any PR files, diffs, or codebase context could be fetched or analyzed. No changed symbols, props, accessibility behavior, SSR usage, tests, docs, or security scans were verified. The PR must be re-reviewed with sufficient time to inspect each changed file.

ROADMAP.md
```diff
@@ -150,3 +149,25 @@ These components are already in the library and are retained as-is:
...
+This roadmap was added in the same commit as the MCP server (`2eec1ad`, 2026-02-18) and,
+until this update, had not been touched since — including through `BZ-49010` two weeks
+later, which alone shipped 27 of the components marked "to build" above. These 37
+components exist and ship today but were never added to this file:
+
+**Charts:** `AreaChart`, `BarChart`, `DualAxisBarChart`, `FunnelChart`, `LineChart`, `PieChart`, `SankeyChart`
+
+**Chat & agent UX:** `Chat`, `ChatBubble`, `ChatComposer`, `ChatHeader`, `ChatMessage`, `ChatMessageList`, `ChatSuggestions`, `ChatToolStatus`, `AttachmentChipRow`, `ChipInput`, `FileDropzoneTrigger`, `HITL`, `TaskList`, `ThinkingIndicator`, `ToolCallLog`, `TypewriterText`
+
+**Other:** `BrandLoader`, `Breadcrumb`, `Card`, `ColorPicker`, `DateRangePicker`, `DeltaIndicator`, `FileInput`, `IframeViewer`, `LottiePlayer`, `ProportionBar`, `Resizable`, `SplitInput`, `StatCard`
```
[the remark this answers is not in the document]
murdore: Fixed and verified — recounted the three category lines directly (Charts 7 + Chat & agent UX 16 + Other 13 = 36). Your script's 37/38 discrepancy with a separate finding came from also matching the two backticked non-component tokens in the intro prose (`2eec1ad`, `BZ-49010`); excluding those gives 36, matching your correction exactly. Also fixed the `TaskList` line in the same commit — it is indexed in `docs/_index.json`, only `AttachmentChipRow`/`HITL`/`TypewriterText` are genuinely missing. Pushed in e152a08.

package.json
murdore: Fixed — added a `types` condition to the `./wc` export. Since `dist-wc/` is gitignored and wiped by `emptyOutDir: true` on every build, the stub has to be generated as part of the build rather than committed as a static file: `build:wc` now also writes `dist-wc/index.d.ts` (`export {};`) right after the vite build. Verified with a real `pnpm run build` — publint no longer reports the missing-types issue, only the pre-existing unrelated lottie-web CJS-heuristic warning remains. Pushed in 14b0938.

README.md
murdore: Fixed, rephrased exactly as suggested — "no separate Svelte installation required; the runtime is bundled with each component." Pushed in 0ac8fb8.

murdore: Went with adding the wrappers rather than softening the claim, since it's a stronger true statement once they exist. Added `FileDropzoneTrigger.wc.svelte` and `KeyValue.wc.svelte`, registered both in `src/wc/index.ts`. Verified `pnpm run check:wc` shows the same 3 pre-existing LottiePlayer errors as before this PR and nothing new from either wrapper. Pushed in 14b0938 — "every component" is accurate again.

package.json
murdore: Fair catch — retitled the PR to `feat:`. Confirmed this repo squash-merges (no 2-parent merge commits in `release`'s recent history, and merged PR titles read as single linear commits), so the PR title is what standard-version actually sees, not the individual commit messages underneath. Also split the package.json/build changes into their own dedicated `feat` commit (14b0938) rather than folding them into the docs commit next to it, so the semver-relevant change is isolated and clearly typed either way this gets merged.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
...
+        continue-on-error: true
+        run: pnpm run test:integration
+
+      - name: Upload Playwright report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: playwright-report
+          path: playwright-report/
+          retention-days: 14
```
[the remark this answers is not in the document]
murdore: Fixed — added a workflow-level `permissions: contents: read` block. The checks job only reads the repo (install, lint, type-check, tests), it doesn't need write access to anything. Pushed in 6cef5bf.

murdore:
Replying to the [Yama BLOCKED verdict](https://github.com/juspay/svelte-ui-components/pull/473#issuecomment-5474331195) — the specific critical finding needs a correction.

The claim was that SVG-inlining assigns fetched markup to `host.innerHTML`. I checked `src/lib/Img/Img.svelte` directly: there are zero occurrences of `innerHTML` anywhere in that file. It isn't in this PR's diff either — this PR only touches `ROADMAP.md`, `package.json`, `README.md`, `.github/workflows/ci.yml`, and two new `.wc.svelte` wrapper files.

The actual mechanism (lines ~40-87): fetch → `DOMParser().parseFromString(markup, 'image/svg+xml')` into an inert, unattached document → `parsererror` check → `querySelector('svg')` → attribute copy loop and child-node move loop onto the host. No `innerHTML`, no live-document parse.

That said, there's a real-but-different concern in that attribute-copy loop: it copies every attribute except `class` from the source `<svg>` onto the host element, which would include something like an `onload` attribute if a malicious SVG had one. That's worth its own issue against `Img.svelte` — it isn't something this PR introduces or touches, so it shouldn't block a docs/build-tooling PR. Happy to file that separately if useful.

Everything else in this review round (ROADMAP count, `./wc` types export, README wording x2, commit type, CodeQL workflow permissions) has been fixed and pushed — replied inline on each of those threads.

murdore:
@coderabbitai review

murdore:
@coderabbitai review

Full re-review requested. Auto-incremental reviews are disabled on this repo, so this is a manual trigger.

Since the last pass, all review comments from CodeRabbit, Yama and CodeQL have been addressed (head is now `6cef5bf`):
- ROADMAP shipped-component count corrected 37 → 36, and the `TaskList` indexing note fixed
- `./wc` subpath export given a `types` condition (`dist-wc/index.d.ts`, generated during `build:wc`)
- `FileDropzoneTrigger.wc.svelte` and `KeyValue.wc.svelte` wrappers added and registered, so the README's custom-element claim is accurate rather than softened
- README wording on the bundled Svelte runtime clarified
- PR retitled `feat:` so standard-version bumps for the new public subpath export
- `permissions: contents: read` added to `ci.yml` per the CodeQL finding

One earlier finding was **not** actioned as a code change, deliberately. The Yama CRITICAL reported unsafe SVG inlining "via innerHTML" in `src/lib/Img/Img.svelte`. `innerHTML` does not appear in that file, and the file is not part of this PR's diff. The component uses `DOMParser.parseFromString(markup, 'image/svg+xml')` into an inert document, checks for `parsererror`, then moves nodes. The related concern that *is* real — the attribute-copy loop can carry an event-handler attribute such as `onload` from a fetched SVG onto the live host element — has a different mechanism and fix and is being tracked separately rather than gating this PR.

src/lib/MediaPlayer/MediaPlayer.svelte
```diff
@@ -0,0 +1,245 @@
...
+      bind:muted
+      {src}
+      class="media"
+      {controls}
+      {autoplay}
+      {loop}
+      playsinline
+      onplay={handlePlay}
+      onpause={handlePause}
+      onclick={togglePlayback}
```
Tara-ag: ⚠️ MAJOR: Video element role="button" removes native video semantics

The `<video>` element is given `role="button"`, which overrides its native video/media role in the accessibility tree. Screen-reader users lose the fact that this is a video, and the element is announced only as a button with a play/pause label. This also conflicts when `controls={true}` because the same click handler and button role are applied even though native controls are present.

Suggestion: Remove `role="button"` and `tabindex` from the `<video>`; keep it as a native video element. Move the click/toggle behavior to a visible button overlay or rely on the native controls when `controls={true}`. If a custom big-hit-area play button is required, render a real `<button>` absolutely positioned over the video instead of repurposing the video element.

src/lib/MediaPlayer/MediaPlayer.svelte
```diff
@@ -0,0 +1,245 @@
...
+      onkeydown={handleVideoKeydown}
+      role="button"
+      tabindex="0"
+      aria-label={playing ? 'Pause video' : 'Play video'}
+    >
+      <track kind="captions" />
+    </video>
+
+    {#if !controls}
```
Tara-ag, line 121: 💡 MINOR: Empty `<track>` element provides no captions

The `<track kind="captions" />` element has no `src`, `srclang`, or `label` attributes, so it does not actually provide captions and may trigger accessibility validators. If captions are not supplied, the track should be omitted or rendered conditionally with a real captions file.

Suggestion: Remove the empty `<track>` or expose a `captions` prop and render it conditionally:
```svelte
{#if captions}
  <track kind="captions" src={captions.src} srclang={captions.lang} label={captions.label} default={captions.default} />
{/if}
```

src/lib/MediaPlayer/MediaPlayer.svelte
```diff
@@ -0,0 +1,245 @@
...
+
+  .media-image {
+    display: contents;
+    --image-height: var(--media-player-media-height, 100%);
+    --image-width: var(--media-player-media-width, fit-content);
+    --image-object-fit: var(--media-player-media-object-fit, contain);
+    --image-border-radius: var(--media-player-media-border-radius, 0px);
+    --image-padding: 0px;
+    --image-margin: 0px;
+  }
```
Tara-ag, line 198: ⚠️ MAJOR: Custom media controls are invisible to keyboard users

The `.overlay` container hides the play and mute controls via CSS `visibility: hidden` by default and only reveals them on `:hover`. A keyboard user tabbing to the overlay buttons never triggers hover, so the focused controls remain invisible while still being interactive.

Suggestion: Make controls visible whenever the overlay or any of its controls has focus, e.g. add `.overlay:has(:focus-visible), .overlay:focus-within { --center-controls-visibility: visible; --bottom-controls-visibility: visible; }` or switch from hover-only to a persistent visible control bar.

src/lib/MediaPlayer/MediaPlayer.svelte
```diff
@@ -0,0 +1,245 @@
...
+    loop = false,
+    controls = false,
+    fallback,
+    playing = $bindable(true),
+    muted = $bindable(true),
+    playIcon,
+    pauseIcon,
+    muteIcon,
+    unmuteIcon,
+    onplay,
```
Tara-ag, line 27: ⚠️ MAJOR: bindable `playing` prop does not drive playback

`playing` is declared as `$bindable(true)`, but the component only updates it in response to video `play`/`pause` events. There is no `$effect` that calls `videoPlayer.play()` or `videoPlayer.pause()` when the bound value changes, so a parent setting `playing = false` (or binding to it) cannot pause the video. This breaks the `$bindable` contract.

Suggestion: Add an effect that observes `playing` and synchronises the video element:
```ts
$effect(() => {
  if (videoPlayer === null) return;
  if (playing && videoPlayer.paused) {
    videoPlayer.play().catch(() => {});
  } else if (!playing && !videoPlayer.paused) {
    videoPlayer.pause();
  }
});
```

Tara-ag (changes requested):
Reviewed the new MediaPlayer component and Modal changes. MediaPlayer has accessibility and reactivity issues that need to be addressed before merge (keyboard-invisible controls, video role override, broken $bindable playing contract, and empty captions track). Modal additions are additive and backward-compatible. I could not exhaustively verify all 54 changed files in the available time; the verdict is based on the files inspected.

murdore:
Replying to the [Yama CHANGES_REQUESTED review](https://github.com/juspay/svelte-ui-components/pull/473#pullrequestreview-5064524451) — all four MediaPlayer findings verified against source (not taken at face value) and fixed in 21e5c28:

1. **Keyboard-invisible controls** — confirmed: `.overlay:hover` alone gated visibility, no `:focus-within`, and `visibility: hidden` drops elements from the tab order. Added `:focus-within` alongside `:hover`.
2. **Video role override** — confirmed: `role="button"`/`tabindex="0"`/click/keydown sat unconditionally on `<video>`, even under `controls={true}` where native browser controls already handle keyboard access. Scoped to `!controls` only now.
3. **Broken `$bindable playing` contract** — confirmed: data only flowed video→prop via the native `onplay`/`onpause` handlers, nothing flowed prop→video. Added a guarded effect so an external host setting `playing` actually calls `play()`/`pause()`.
4. **Empty captions track** — confirmed: `<track kind="captions" />` had no `src` and no prop to ever supply one. Added `captionsSrc`/`captionsLabel`/`captionsSrcLang`; the track now only renders when real caption data exists.

Also checked your "Modal additions are additive and backward-compatible" note against the actual diff — confirmed accurate, `lockScroll` defaults to `true` and preserves the prior unconditional `document.body.style.overflow = 'hidden'` byte for byte.

Added 6 new integration tests covering all four fixes (keyboard-focus reveal, native-controls role omission, both directions of the playing bind as independent tests, captionsSrc present/absent) — all pass, plus the 5 pre-existing MediaPlayer tests unmodified. The playing-bind fix was also hand-verified in a live browser with instrumented `play()`/`pause()` calls, since a combined-directions test first hit a Playwright-specific rapid-click timing quirk that a real user session didn't reproduce.

PR #475, opened 2026-08-31
fix(a11y): let callers actually name Input, Checkbox and Menu triggers

murdore:
## Three components no caller could name

Found while auditing a consuming app (Juspay Lighthouse), where these accounted for ~84 unnamed-control and unlabelled-field findings that **no app-side change could close**.

### `Input` — the label association never completed

The label was emitted as `<label class="label" for={name}>` while the rendered `<input>`/`<textarea>` carried only `name={name}` and **no `id` anywhere**.

`for` resolves against an element's `id`, never its `name`. So the association could not complete for *any* caller, however they called it — every labelled field read as unlabelled to assistive tech and was unreachable via `getByLabel`. This single bug is the root cause of ~77 unlabelled-field findings in the consuming app.

`effectiveId` is now derived from `name` and applied to the label's `for` and to both rendered field variants. **Existing callers that already pass `label` + `name` become correctly labelled with no change on their part.** A new optional `id` prop wins over the derived value, for callers whose `name` repeats across rows or who pass no `name` at all.

### `Checkbox` — no way to pass a name

No `ariaLabel` prop and no rest-prop spread, so an icon-only or externally-labelled checkbox simply could not be named. Adds `ariaLabel`, applied to the `role="checkbox"` element — the tabbable control that already receives `aria-controls`, so it follows the existing precedent rather than inventing a placement.

Deliberately **not** a rest spread: with three rendered elements (label, native input, box) an arbitrary attribute has no unambiguous home.

### `Menu` — `ariaLabel` named the dropdown, not the trigger

`ariaLabel` lands on the dropdown, which is `use:portalToBody` and therefore mounted at `<body>`. The element the user tabs to is the trigger wrapper, carrying `role="button"` and `tabindex="0"` and **no `aria-label` at all**. Callers passing `ariaLabel` got a silent no-op on the control and a name on a portaled container.

Adds `triggerAriaLabel` on the focusable trigger. The existing `ariaLabel` and its placement are untouched, so nothing relying on the dropdown's name changes. Applies only to the non-interactive branch; with `interactiveTrigger` the snippet's own control owns its name.

## Compatibility

All three additions are optional props with behaviour-preserving defaults. No existing call site changes behaviour except to gain a working accessible name.

## Verification

`tests/input-label-association.test.ts` asserts the field carries an `id`, that a label points at that exact `id`, that the accessible-name path resolves, and that clicking the label moves focus.

**Negative-controlled.** Reverting *only* the `id` emission — keeping the prop and the derived value, so the control isolates the exact mechanism rather than the whole change — fails with `the field must carry an id for a label to reference`, `Received: null`. The file was then restored and confirmed byte-identical.

**Suite comparison, run twice in ONE worktree**, same invocation, swapping only the file contents. Both arms in the same environment is the only comparison that means anything here — an earlier cross-worktree comparison in this PR's history looked like a 5-test regression and was an artifact of the two checkouts, not of the code.

| | Failed | Passed |
|---|---|---|
| `origin/release` content | 15 | 198 |
| this branch | 15 | 198 |

| Test file | release content | this branch |
|---|---|---|
| `table-builtin-cells`, `table-pagination-selection`, `table-header-meta` | 9 | 9 |
| `status-*` | 4 | 6 |
| **`input-label-association`** | **2** | **0** |

- The nine table failures are identical in both arms and pre-exist on `release`.
- The `status-*` difference sits inside the run-to-run variance measured on *unchanged* code (3 -> 6 across two runs of the same tree). `/components/status` takes ~21s to load against a 30s `goto` timeout because of a 404 on `order-success-icon.svg` that fires 49 times on `release` and 48 here, so which of those tests trips the limit is a coin flip. Running `release` content inside this same worktree reproduces 5 of them.
- The last row is the negative control, and it is the only result that actually tracks the change: with the fix reverted but the new spec still present, `tests/input-label-association.test.ts` fails both cases; with the fix, both pass.

`svelte-check`: 654 files, 0 errors. `prettier` + `eslint`: clean. `check:wc`: 3 `LottiePlayer` errors, exit 1 — pristine `origin/release` reports the identical 3 and the identical exit, so this change adds none.

> **Correction.** An earlier version of this description credited the change with fixing `row checkboxes expose DataGrid-parity aria-labels`, calling it independent confirmation of the `Checkbox` diagnosis. That was wrong and is withdrawn: `Table.svelte` does not use the library `Checkbox` at all — it renders its own `<span role="checkbox">` and imports only `checkmark.svg`. There is no causal path from this change to that test.

## Not included

`Table`/`BuiltinCell`'s `interactiveTrigger` migration was scoped out. It rests on an unverified assumption about `Button`'s prop pass-through and it changes tab order — that needs verifying against `Button.svelte` directly rather than shipping on inference.

## Summary by CodeRabbit

- **Accessibility Improvements**
  - Added accessible label support for Checkbox and Menu triggers.
  - Improved Input label-to-field association for reliable keyboard and assistive technology navigation.
  - Inputs now support explicit IDs, with automatic IDs used when none is provided.


src/lib/Checkbox/Checkbox.svelte
```diff
@@ -71,6 +72,7 @@
     aria-checked={indeterminate ? 'mixed' : checked}
     aria-disabled={disabled}
     aria-controls={ariaControls ?? null}
+    aria-label={ariaLabel ?? null}
```
Tara-ag: 💡 MINOR: Checkbox already renders a visible `<span class="label">{text}</span>` below. Applying `aria-label` on the `role="checkbox"` element overrides the visible text as the accessible name even when the visible label is present. For icon-only or externally-labelled cases the new prop is useful, but callers who pass both `text` and `ariaLabel` will lose the visible label as the accessible name. Consider only applying `aria-label` when `text` is empty, or document that `ariaLabel` intentionally supersedes `text`.

src/lib/Input/Input.svelte
```diff
@@ -60,6 +61,14 @@
     showCount = false
   }: InputProperties = $props();
 
+  /* `for` on a <label> resolves against an element's id, never its name. The label
+     was emitted with for={name} while the field itself carried only name={name},
+     so the association never completed for any caller. Derive the id from name so
+     every existing caller that already passes label + name becomes correctly
+     labelled with no change on their part; an explicit id wins, for callers who
```
Tara-ag, line 68: ⚠️ MAJOR: `effectiveId` defaults to the raw `name` value. Multiple inputs on a page can legitimately share a `name` (radio groups, repeated form rows), which produces duplicate `id` attributes. Duplicate ids break `<label for=...>` association for all but the first field and make `getByLabel` unreliable. Generate a unique fallback id (e.g. `name` + a component-scoped counter or `crypto.randomUUID`) instead of using `name` directly, or at least document that callers must pass `id` whenever `name` is reused.

src/lib/Input/Input.svelte
```diff
@@ -269,14 +278,15 @@
...
   {#if typeof label === 'string' && label !== '' && !actionInput}
-    <label class="label" for={name}>
+    <label class="label" for={effectiveId}>
       {label}{#if mandatory}<span class="input-mandatory-asterisk" aria-hidden="true">*</span>{/if}
     </label>
   {/if}
 
   {#snippet fieldElement()}
     {#if useTextArea}
       <textarea
```
Tara-ag, line 290: 💡 MINOR: When both `label` and `ariaLabel` are provided, the rendered field carries `aria-label={ariaLabel}` while the visible `<label>` points at it via `for`. Accessible name computation gives `aria-label` precedence over the visible label, so assistive tech will ignore the visible label text. If this is intentional, document it; otherwise consider omitting `aria-label` when a visible `label` is present.

Tara-ag:
🔒 CRITICAL (src/wc/components/Checkbox.wc.svelte): The Svelte `Checkbox` now exposes an `ariaLabel` prop, but the web-component prop map does not include it. Consumers using `<sui-checkbox>` cannot set the accessible name, breaking web-component parity with the Svelte API. Add `ariaLabel: { type: 'String', attribute: 'aria-label' }` to the custom element props. Same issue exists for `Input.id` in `src/wc/components/Input.wc.svelte` and `Menu.triggerAriaLabel` in `src/wc/components/Menu.wc.svelte`.

Tara-ag:
💡 MINOR: The new public props added by this PR are not reflected in the component docs:
- `Input.id` is missing from `docs/Input.md`
- `Checkbox.ariaLabel` is missing from `docs/Checkbox.md`
- `Menu.triggerAriaLabel` is missing from `docs/Menu.md`

Per the project standards, public prop changes should be documented in `docs/<Component>.md` so downstream consumers can discover the new API surface.

Tara-ag (changes requested):
Review findings:

🔒 CRITICAL: Web-component wrappers (`src/wc/components/*.wc.svelte`) are missing the new props (`Checkbox.ariaLabel`, `Input.id`, `Menu.triggerAriaLabel`), breaking parity between the Svelte API and the custom-element API.

⚠️ MAJOR: `Input.effectiveId` defaults to raw `name`, which causes duplicate `id` collisions when multiple inputs share a `name` (common in repeated forms/radio groups). This re-breaks label association for all but the first field.

💡 MINOR: `Checkbox` and `Input` apply `aria-label` even when a visible label is present, causing `aria-label` to override the visible label as the accessible name. Consider scoping `aria-label` to label-less cases or documenting the precedence.

💡 MINOR: The new public props are not documented in `docs/Checkbox.md`, `docs/Input.md`, or `docs/Menu.md`.

Because of the CRITICAL web-component API parity issue, this PR is BLOCKED until the WC prop maps are updated.

src/lib/Input/Input.svelte
Tara-ag, line 290: 💡 MINOR: Input aria-label overrides visible label when both are provided. When a caller provides both label and ariaLabel, the rendered input/textarea has aria-label set while the label points via for/id. aria-label takes precedence as the accessible name, so the visible label text is ignored. This may be intentional but should be documented, or aria-label should be omitted when a visible label is present.

Suggested fix: Either omit `aria-label` when a visible `label` is present (`aria-label={ariaLabel && !label ? ariaLabel : null}`) or add explicit documentation that `ariaLabel` takes precedence over the visible label.

src/lib/Input/Input.svelte
Tara-ag, line 68: ⚠️ MAJOR: Input id/name collision risk breaks label association. effectiveId defaults to the value of `name`. If a page contains multiple Inputs with the same `name` (common in radio groups or repeated forms) but no explicit `id`, the generated ids collide. Duplicate ids make the label association point to the first occurrence only and break getByLabel for subsequent fields. The PR description acknowledges this but the default still uses `name` directly rather than generating a unique id.

Suggested fix: Generate a unique fallback id instead of using raw `name`, e.g. import a small uid helper and use `const effectiveId = $derived(id || (name !== '' ? `${name}-${uid()}` : uid()));`, or use Svelte's component-scoped state to produce a stable per-instance id.

src/lib/Checkbox/Checkbox.svelte
Tara-ag: 💡 MINOR: Checkbox ariaLabel duplicates visible label semantics. Checkbox already renders a visible <span class='label'>{text}</span>. Adding aria-label on the role=checkbox element overrides the visible text for assistive tech even when the visible label is present. For icon-only or externally-labelled cases the new prop is useful, but callers who pass both text and ariaLabel will lose the visible label as the accessible name.

Suggested fix: Apply `aria-label` only when `text` is empty: `aria-label={ariaLabel && text.length === 0 ? ariaLabel : null}`, or document that `ariaLabel` intentionally supersedes the visible `text`.

src/lib/Input/Input.svelte
```diff
@@ -60,6 +61,14 @@
     showCount = false
   }: InputProperties = $props();
 
+  /* `for` on a <label> resolves against an element's id, never its name. The label
```
Tara-ag, line 64: 💡 MINOR: Docs not updated for new props. The PR adds new public props (Input id, Checkbox ariaLabel, Menu triggerAriaLabel) but the corresponding docs/Checkbox.md, docs/Input.md, and docs/Menu.md prop tables do not list them. Per bootstrapped standards, public prop changes must be reflected in docs/<Component>.md.

Suggested fix: Add rows for `id` in docs/Input.md, `ariaLabel` in docs/Checkbox.md, and `triggerAriaLabel` in docs/Menu.md prop tables, including type, required/default, and description.

src/lib/Checkbox/Checkbox.svelte
```diff
@@ -13,7 +13,8 @@
     indeterminateIcon,
     onclick,
```
Tara-ag, line 14: 🔒 CRITICAL: Web-component wrappers omit new props. The Svelte components add new public props (Checkbox ariaLabel, Input id, Menu triggerAriaLabel) but the corresponding web-component prop maps in src/wc/components/*.wc.svelte are not updated. Consumers using the custom elements cannot set these attributes, breaking web-component parity with the Svelte API. Per bootstrapped standards, web-component prop maps must stay in lock-step with the Svelte API including explicit kebab-case attribute mappings for every camelCase prop.

Suggested fix: Add the new props to each web-component wrapper with explicit kebab-case attributes: Checkbox.wc.svelte: `ariaLabel: { type: 'String', attribute: 'aria-label' }`; Input.wc.svelte: `id: { type: 'String', reflect: true }`; Menu.wc.svelte: `triggerAriaLabel: { type: 'String', attribute: 'trigger-aria-label' }`.

Tara-ag (changes requested):
Yama review: BLOCKED due to a CRITICAL web-component API parity regression. The Svelte components add new public props (Checkbox.ariaLabel, Input.id, Menu.triggerAriaLabel) but the web-component wrappers are not updated, breaking custom-element consumers. Additional MAJOR/MINOR findings cover Input id/name collision risk, aria-label overriding visible labels, and missing documentation updates.

Tara-ag:

Severity counts — 🔒 CRITICAL: 1 · ⚠️ MAJOR: 2 · 💡 MINOR: 5

The PR introduces a CRITICAL backward-compatibility break: the new `Input` `id` prop intercepts the global `id` attribute and redirects it from the wrapper `<div>` to the inner field, silently altering the public DOM contract for existing callers and web-component consumers. This violates the blocking criterion for public API changes without a major-version bump or migration note. Three additional MINOR issues are also accepted: Checkbox `text` remains mandatory despite the new icon-only `ariaLabel` path; the new Playwright test is coupled to a demo-page testId; and the new Checkbox/Menu naming props lack regression tests.

- 🔒 **CRITICAL**: Input `id` prop intercepts global `id` attribute and breaks wrapper DOM contract — `src/lib/Input/Input.svelte:29`
  Destructuring `id` from `$props()` and applying it only to the inner `<input>`/`<textarea>` removes `id` from the wrapper `<div>` that previously received it via Svelte's standard attribute forwarding. Existing callers passing `id=` to target/style the wrapper, or relying on the wrapper id for `aria-describedby`/test selectors, will silently see that id moved to the field. The web-component wrappe…
- ⚠️ **MAJOR**: Input aria-label is dropped when a visible label is present, breaking icon-only fields that also carry a visible label — `src/lib/Input/Input.svelte:293`
  Input.svelte now sets aria-label to null whenever hasVisibleLabel is true. The PR description says this is to satisfy WCAG 2.5.3, but the rule only prohibits a visible label text being overridden by a different accessible name — it does not prohibit a supplemental aria-label when the visible label is insufficient (e.g. a search field with a visible 'Search' label and an icon-only button). More imp…
- ⚠️ **MAJOR**: Checkbox ariaLabel is ignored whenever text is non-empty, including when text is visually hidden — `src/lib/Checkbox/Checkbox.svelte:78`
  Checkbox.svelte sets aria-label only when text.length === 0. However, the component renders <span class="label" hidden={text.length === 0}>{text}</span>, meaning text can be non-empty but visually hidden via the hidden attribute. In that case the accessible name from content is removed from the a11y tree, yet aria-label is also suppressed, leaving the checkbox unnamed. Additionally, callers who pa…
- 💡 **MINOR**: Input docs default for id prop is listed as '-' but the implementation auto-generates one — `docs/Input.md:65`
  The docs/Input.md table lists the default for id as '-', but the component now always derives an effectiveId from name or $props.id(), so every instance gets an id. The description explains the fallback, but the 'Default' column is misleading for callers scanning the table. This is a minor doc fidelity issue.
- 💡 **MINOR**: New Playwright test depends on a demo field that is not part of the PR diff — `tests/input-label-association.test.ts:16`
  tests/input-label-association.test.ts relies on page.getByTestId('input-datatype-time') and the label 'Schedule time'. The test will fail if the demo page is ever refactored to rename that testId or label, and the test itself does not create its own fixture. While the current demo page does contain this field, the test is coupled to demo content rather than asserting the component in isolation. Th…
- 💡 **MINOR**: Checkbox `text` remains mandatory despite new icon-only naming path — `src/lib/Checkbox/properties.ts:7`
  The PR adds `ariaLabel` specifically for icon-only or externally-labelled checkboxes, but `text` remains in `MandatoryCheckboxProperties`. A caller who wants an icon-only checkbox must still pass `text=""` or a dummy value, and the DOM still contains a hidden `<span class="label">`. This makes the new `ariaLabel` API awkward to use for its stated purpose.
- 💡 **MINOR**: Playwright regression test is coupled to demo-page testId — `tests/input-label-association.test.ts:14`
  Navigating to `/components/input` and selecting `data-pw="input-datatype-time"` with label text "Schedule time" ties the regression guard to the showcase page. Demo pages are not a stable public interface; renaming the example, changing the testId, or reorganising the showcase will break the test for reasons unrelated to the component.
- 💡 **MINOR**: New Checkbox ariaLabel and Menu triggerAriaLabel props lack test coverage — `src/lib/Checkbox/Checkbox.svelte:75`
  The PR adds new public props (`Checkbox.ariaLabel` and `Menu.triggerAriaLabel`) to fix accessibility gaps, but only `Input` receives a new regression test. Without tests, future refactors can silently regress the accessible-name behavior these props are meant to provide.

murdore:
Verification for the three fixes just pushed (`5b3bc70`, `0309d0c`, `2056d59`), run locally on this branch:

| Check | Command | Result |
|---|---|---|
| Lint | `pnpm lint` (prettier --check src && eslint src) | Clean — 0 errors, including the previously-failing `no-restricted-syntax` rule at Input.svelte:70 |
| Type-check | `pnpm check` (svelte-check) | 654 files, 0 errors, 4 warnings — same 4 pre-existing warnings as before these commits (Modal.svelte/ThemeSwitcher.svelte `state_referenced_locally` notices, 2 tsconfig `moduleResolution` notices), nothing new |
| WC type-check | `pnpm run check:wc` | 469 files, 3 errors — same 3 pre-existing `LottiePlayer.wc.svelte` errors as before these commits, nothing new from the Checkbox/Input/Menu wrapper edits |
| Unit tests | `pnpm run test:unit -- --run` | 128/128 passed (16/16 files) — matches the count in this PR's own description |

No screenshots since none of these three commits touch rendering/layout — they're a literal→null swap and two mechanical prop-map/docs additions.

Tara-ag (approved):
Reviewed file-by-file. All three a11y fixes are backward-compatible optional props, the web-component prop maps are in sync, and the docs tables are updated. No blocking issues found.

src/lib/Input/Input.svelte
```diff
@@ -60,6 +61,14 @@
...
+     was emitted with for={name} while the field itself carried only name={name},
+     so the association never completed for any caller. Derive the id from name so
+     every existing caller that already passes label + name becomes correctly
```
Tara-ag, line 67: ⚠️ MAJOR: Falling back to `name` as the element id invites duplicate-id violations. `effectiveId = $derived(id || (name !== '' ? name : null))` uses the `name` attribute value directly as the HTML `id`. Unlike `id`, `name` is intentionally repeatable across a form (e.g. a table with many rows all using `name="amount"`). Existing callers that pass only `name` will now emit duplicate ids in the document, breaking `document.getElementById`, `label[for=...]`, and assistive-tech name resolution. HTML ids also have stricter token rules than names.

**Suggested fix:** Generate a per-instance unique id when none is supplied, e.g. `let uid = $state(crypto.randomUUID()); const effectiveId = $derived(id ?? (name !== '' ? `${name}-${uid}` : uid));`. This keeps the name-based id stable for label association while guaranteeing uniqueness across repeated names.

tests/input-label-association.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  // field carried only name={name} and never an id. `for` resolves against an
+  // id, never a name, so the association could not complete for ANY caller, no
+  // matter what they passed. Every such field read as unlabelled to assistive
+  // tech and was unreachable via getByLabel.
+  //
+  // Note on locators: testId lands on the <input>/<textarea> itself (data-pw),
+  // not on a wrapper — so getByTestId here IS the field.
+  test('a field with label + name is reachable by its visible label text', async ({ page }) => {
+    await page.goto('/components/input');
+
```
Tara-ag, line 14: 💡 MINOR: This test is coupled to a demo-page testId. Navigating to `/components/input` and selecting `data-pw="input-datatype-time"` with label text "Schedule time" ties the regression guard to the showcase page. Demo pages are not a stable public interface; renaming the example, changing the testId, or reorganising the showcase will break the test for reasons unrelated to the component.

**Suggested fix:** Create a dedicated test route or component harness that renders `<Input label="Schedule time" name="schedule-time" testId="input-datatype-time" />` in isolation, then assert `label[for]` matches the rendered input id and `page.getByLabel('Schedule time')` resolves.

src/lib/Input/Input.svelte
```diff
@@ -26,6 +26,7 @@
     autoComplete = 'on',
     inputMode,
     name = '',
+    id,
```
Tara-ag, line 29: 🔒 CRITICAL: New `id` prop intercepts the global `id` attribute and redirects it from the wrapper to the inner field. On `release`, `<Input id="foo">` would land on the outer `<div class="input-container">` via Svelte's standard attribute forwarding. This PR destructures `id` from `$props()` and only emits `id={effectiveId}` on the internal `<input>`/`<textarea>`. Existing callers passing `id=` to target/style the wrapper, or relying on the wrapper id for `aria-describedby`/test selectors, will silently see that id moved to the field. The web-component wrapper also maps `id` to the internal input, so `<sui-input id="foo">` sets the shadow input's id rather than the custom element's. This is a breaking change to the public DOM contract.

**Suggested fix:** Rename the prop to `inputId` in `src/lib/Input/properties.ts`, `src/lib/Input/Input.svelte`, `docs/Input.md`, and `src/wc/components/Input.wc.svelte`. This preserves standard `id` behavior on the wrapper while still offering an explicit field-id opt-in.

src/lib/Checkbox/properties.ts
Tara-ag, line 1: 💡 MINOR: `text` is still mandatory despite the new icon-only naming path. The PR adds `ariaLabel` specifically for icon-only or externally-labelled checkboxes, but `text` remains in `MandatoryCheckboxProperties`. A caller who wants an icon-only checkbox must still pass `text=""` or a dummy value, and the DOM still contains a hidden `<span class="label" hidden>`. This makes the new `ariaLabel` API awkward to use for its stated purpose.

**Suggested fix:** Move `text` to `OptionalCheckboxProperties` and update `Checkbox.svelte` to render the label span only when `text` is truthy (the `hidden` attribute already handles empty text, but the type should match). Update `docs/Checkbox.md` to mark `text` as optional.

Tara-ag (changes requested):
Blocking review: the new `id` prop on `Input` is a breaking change to the public DOM contract because it intercepts the standard global `id` attribute and redirects it from the wrapper `<div>` to the internal field. Per the blocking criteria, a breaking change to a public component's props/contract without a major-version bump or migration note must block the PR. Additionally, falling back to `name` as the HTML `id` risks duplicate-id accessibility violations. Two smaller items are noted around test coupling and Checkbox API completeness.

Decision: **CHANGES_REQUESTED** (blocked until the `id` collision is resolved).

Tara-ag (changes requested):
Blocking: the new `id` prop on `Input` intercepts the standard global `id` attribute and redirects it from the wrapper `<div>` to the internal field, breaking the public DOM contract for existing callers and web-component consumers. Combined with the `name`-as-id fallback, this introduces duplicate-id accessibility violations. Please address the CRITICAL and MAJOR findings before merging.

murdore:
## Review response — every open comment addressed

Thanks both. Taking the comments in order of severity. Three are fixed, one is refuted with evidence, one is a convention question, one is deferred.

### Fixed

**MAJOR — `effectiveId` fell back to raw `name`, risking duplicate ids** (`Input.svelte:68`, raised by @Tara-ag twice)

Valid, and the docstring-only mitigation I had was not good enough for a library default. Now uses Svelte's built-in per-instance id:

```js
const uid = $props.id();
const effectiveId = $derived(id || (name !== '' ? `${name}-${uid}` : uid));
```

Fields that share a `name` — radio groups, repeated table rows — now get distinct ids with no action required from callers. An explicit `id` still wins. `$props.id()` is SSR-safe and hydration-stable, and this repo is on Svelte 5.56.3, so it is available.

**MINOR — `aria-label` overrode a visible label** (`Input.svelte:288` ×2, `Checkbox.svelte:75` ×2, plus CodeRabbit)

Valid, and it is a WCAG 2.5.3 (Label in Name) hazard, not just a documentation gap. A visible label now always wins the accessible name:

- `Input`: `aria-label={hasVisibleLabel ? null : (ariaLabel ?? null)}`, where `hasVisibleLabel` is a single `$derived` that the visible `<label>` block now also renders from, so the two cannot drift apart.
- `Checkbox`: `aria-label={text.length === 0 ? (ariaLabel ?? null) : null}`.

`ariaLabel` keeps working for the icon-only and externally-labelled cases it was added for. Docstrings and the `docs/*.md` prop tables state the precedence.

### Already fixed earlier in this PR

These were raised against the first commit and were fixed before the branch was squashed; @Tara-ag's 09:09 review confirms them ("the web-component prop maps are in sync, and the docs tables are updated. No blocking issues found."):

- CRITICAL — web-component wrappers missing the new props. Fixed in all three wrappers with kebab-case attribute mappings.
- CodeRabbit — `undefined` violating `no-restricted-syntax` and failing the build. Fixed to `null`.
- CodeRabbit — `Menu.triggerAriaLabel` missing from the `<sui-menu>` allowlist. Fixed.
- MINOR — docs not updated. `docs/Input.md`, `docs/Checkbox.md`, `docs/Menu.md` all carry rows for the new props.

### Refuted

**CRITICAL — "New `id` prop intercepts the global `id` attribute and redirects it from the wrapper to the inner field"** (`Input.svelte:29`)

This does not reproduce. The premise is that on `release`, `<Input id="foo">` lands on the outer `<div class="input-container">` "via Svelte's standard attribute forwarding". Svelte has no such forwarding — unlike Vue's `inheritAttrs`, an undeclared prop is only reachable through an explicit rest spread. Checked against the baseline:

```
$ git show origin/release:src/lib/Input/Input.svelte | grep -n 'id=\|\.\.\.\|\$props()'
61:  }: InputProperties = $props();
```

One match, and it is the destructure itself. On `release` the component has no `id` anywhere, no rest spread, and its root element is `<div class="input-container {classes ?? ''}" class:input-error={...}>` — no `id`, no spread. A grep for rest spreads across every component in `src/lib/*/[A-Z]*.svelte` returns nothing at all, so no component in this library relies on attribute forwarding.

So `<Input id="foo">` on `release` emitted no `id` into the DOM at all. Nothing could have been targeting, styling, or `aria-describedby`-referencing a wrapper id, because there was never a wrapper id. The change is "silently discarded" -> "applied to the field", which is additive. There is no removed behaviour to migrate and no public contract being broken, so no major-version bump is required.

The same reasoning covers `<sui-input id="foo">`: the wrapper maps `id` to the component prop, and the custom element's own host `id` attribute is untouched by that mapping.

I have documented in `docs/Input.md` and the prop's docstring that `id` names the field rather than the wrapper, so the intent is explicit for future readers.

### Convention, not a defect

**MINOR — test coupled to the `/components/input` demo page** (`tests/input-label-association.test.ts:14`)

This is the repository's own convention rather than a deviation from it. Across the 42 files in `tests/`:

```
  49 goto('/components/table')
  17 goto('/components/date-range-picker')
  16 goto('/components/input')
  15 goto('/components/menu')
  ...
```

Sixteen existing tests already drive `/components/input`. A bespoke harness route for this one test would be the only one of its kind, and would make the suite less consistent, not more. Happy to move it if the team wants to change that convention repo-wide, but that seems like its own PR.

### Deferred

**MINOR — `text` should move from `MandatoryCheckboxProperties` to optional** (`Checkbox/properties.ts:1`)

Agreed that `text=""` is awkward for an icon-only checkbox. But making a mandatory prop optional is a change to `Checkbox`'s public type contract that every caller's type-checking sees, which is a different blast radius from the three additive optional props here. I would rather do it deliberately in its own change than fold it into an a11y fix. Not blocking the naming fix, since `ariaLabel` works with `text=""` today.

### Correction to this PR's own description

The description credits the fix with repairing `row checkboxes expose DataGrid-parity aria-labels`, calling it independent confirmation of the `Checkbox` diagnosis. That is wrong and I am withdrawing it. `Table.svelte` does not use the library `Checkbox` component at all — it renders its own `<span role="checkbox" aria-label="Select all rows">` and imports only `checkmark.svg`. There is no causal path from this change to that test, so its earlier flip was ordering, the same as `status-children-slot`. The description has been corrected. The rest of the verification stands on the like-for-like comparison below.

### Proof of testing

Both arms run in **one worktree**, same invocation, swapping only file contents — the earlier cross-worktree comparison in this PR's history produced a phantom 5-test regression that turned out to be the checkout, not the code.

| | Failed | Passed |
|---|---|---|
| `origin/release` content | 15 | 198 |
| this branch | 15 | 198 |

| Test file | release content | this branch |
|---|---|---|
| `table-builtin-cells`, `table-pagination-selection`, `table-header-meta` | 9 | 9 |
| `status-*` | 4 | 6 |
| **`input-label-association`** | **2** | **0** |

- Nine table failures identical in both arms, pre-existing on `release`.
- The `status-*` delta is inside the variance measured on unchanged code (3 -> 6 across two runs of the same tree). `/components/status` loads in ~21s against a 30s `goto` timeout because of a 404 on `order-success-icon.svg` firing 49 times on `release` and 48 here. Running `release` content in this same worktree reproduces 5 of those failures.
- The last row is the negative control and the only result that tracks the change: with the fix reverted but the new spec present, both its cases fail; with the fix, both pass.

Static gates: `svelte-check` 654 files / 0 errors · `prettier` + `eslint` clean · `check:wc` 3 `LottiePlayer` errors and exit 1, **identical on pristine `origin/release`**, so this adds none.

src/wc/components/Checkbox.wc.svelte
```diff
@@ -11,6 +11,7 @@
       checkedIcon: { type: 'Object' },
       indeterminateIcon: { type: 'Object' },
       classes: { type: 'String' },
+      ariaLabel: { type: 'String', attribute: 'aria-label' },
       onclick: { type: 'Object' }
```
Tara-ag, line 16: 💡 MINOR: The Svelte `Checkbox` component supports `ariaControls?: string` applied to the `role="checkbox"` element. The web-component wrapper exposes the new `ariaLabel` prop but still does not expose `ariaControls`, so `<sui-checkbox>` consumers cannot express the controlled-element relationship. This breaks parity between the Svelte and web-component APIs.

Suggested fix: add `ariaControls: { type: 'String', attribute: 'aria-controls' }` to the `customElement.props` map.

src/wc/components/Menu.wc.svelte
```diff
@@ -7,6 +7,7 @@
       open: { type: 'Boolean', reflect: true },
       testId: { type: 'String', attribute: 'test-id' },
       classes: { type: 'String' },
+      triggerAriaLabel: { type: 'String', attribute: 'trigger-aria-label' },
```
Tara-ag, line 10: 💡 MINOR: The Svelte `Menu` component exposes `ariaLabel` for the dropdown container, and the PR description explicitly documents that `ariaLabel` names the portaled dropdown while `triggerAriaLabel` names the trigger. The web-component wrapper was updated to expose `triggerAriaLabel` but still does not expose `ariaLabel`, so `<sui-menu>` consumers cannot name the dropdown. Per repo convention, camelCase Svelte props exposed through web components must declare explicit kebab-case attribute mappings and wrappers must keep parity with the Svelte API.

Suggested fix: add `ariaLabel: { type: 'String', attribute: 'aria-label' }` to the `customElement.props` map.

src/wc/components/Input.wc.svelte
```diff
@@ -21,6 +21,7 @@
       autoComplete: { type: 'String', attribute: 'auto-complete' },
       inputMode: { type: 'String', attribute: 'input-mode' },
       name: { type: 'String', reflect: true },
+      id: { type: 'String' },
```
Tara-ag, line 24: 💡 MINOR: The Svelte `Input` component has long supported `ariaLabel?: string | null` applied to the underlying `<input>`/`<textarea>`. The web-component wrapper exposes the new `id` prop but still does not expose `ariaLabel`, so `<sui-input>` consumers cannot provide an accessible name when no visible label is used. This breaks parity between the Svelte and web-component APIs.

Suggested fix: add `ariaLabel: { type: 'String', attribute: 'aria-label' }` to the `customElement.props` map.

Tara-ag (commented):
Reviewed file-by-file. The core a11y fixes for Input, Checkbox, and Menu are sound and well-tested. I found three minor web-component API parity gaps: the Svelte components expose `ariaLabel` (Input, Menu) and `ariaControls` (Checkbox), but the corresponding `<sui-*>` wrappers do not, which breaks the repo's parity convention. These are non-blocking but should be addressed before merge.

Tara-ag (commented):
Reviewed file by file. Found three minor web-component API parity gaps; all other changes look safe and backward-compatible.

murdore:
All three web-component parity gaps are now closed on `1a5d748`.

| Wrapper | Added |
|---|---|
| `<sui-input>` | `ariaLabel` -> `aria-label` |
| `<sui-menu>` | `ariaLabel` -> `aria-label` |
| `<sui-checkbox>` | `ariaControls` -> `aria-controls` |

Worth noting these pre-date this PR — `origin/release` exposes none of the three, so they were already out of lock-step before the new props arrived. I have added them anyway rather than deferring: the convention is that every camelCase prop gets an explicit kebab-case attribute, and leaving three behind would just hand the next reader the same puzzle.

`check:wc` still reports exactly the 3 pre-existing `LottiePlayer.wc.svelte` errors and nothing else, which is byte-identical to what pristine `origin/release` reports. `prettier` + `eslint` clean.

That should be every open item on this PR: three valid findings fixed, three parity gaps closed, one CRITICAL refuted against the `release` baseline, one test-convention point answered with the repo's own numbers, and one type-contract change deferred with reasoning.

Tara-ag (approved):
Approved. All three changes are additive, backward-compatible accessibility fixes:

- `Input` now emits a stable `id` (explicit or derived from `name`) so the visible `<label>` correctly associates with the field, and `aria-label` is suppressed when a visible label is present (WCAG 2.5.3).
- `Checkbox` exposes `ariaLabel` for icon-only or externally-labelled checkboxes, applied to the focusable `role="checkbox"` element and ignored when visible `text` is present.
- `Menu` adds `triggerAriaLabel` for the focusable trigger, distinguishing it from the portaled dropdown's `ariaLabel`.

Web-component wrappers, TypeScript property types, docs, and a Playwright regression test are all updated consistently. No public contract is broken, no secrets or unsafe patterns introduced, and the changes preserve existing DOM/CSS output for unconfigured consumers.

tests/input-label-association.test.ts
Tara-ag: 💡 MINOR: Playwright regression test is coupled to demo-page testId. Navigating to `/components/input` and selecting `data-pw="input-datatype-time"` with label text "Schedule time" ties the regression guard to the showcase page. Demo pages are not a stable public interface; renaming the example, changing the testId, or reorganising the showcase will break the test for reasons unrelated to the component.

Suggested fix: Create a dedicated test route or component harness that renders `<Input>` in isolation, then assert `label[for]` matches the rendered input id and `page.getByLabel('Schedule time')` resolves.

src/lib/Input/Input.svelte
Tara-ag: 🔒 CRITICAL: Input `id` prop intercepts global `id` attribute and breaks wrapper DOM contract. Destructuring `id` from `$props()` and applying it only to the inner `<input>`/`<textarea>` removes `id` from the wrapper `<div>` that previously received it via Svelte's standard attribute forwarding. Existing callers passing `id=` to target/style the wrapper, or relying on the wrapper id for `aria-describedby`/test selectors, will silently see that id moved to the field. The web-component wrapper also maps `id` to the internal input, so `<sui-input id="foo">` sets the shadow input's id rather than the custom element's. This is a breaking change to the public DOM contract.

Suggested fix: Rename the prop to `inputId` in `src/lib/Input/properties.ts`, `src/lib/Input/Input.svelte`, `docs/Input.md`, and `src/wc/components/Input.wc.svelte`. This preserves standard `id` behavior on the wrapper while still offering an explicit field-id opt-in.

src/lib/Checkbox/Checkbox.svelte
```diff
@@ -71,6 +72,7 @@
     aria-checked={indeterminate ? 'mixed' : checked}
     aria-disabled={disabled}
     aria-controls={ariaControls ?? null}
+    aria-label={text.length === 0 ? (ariaLabel ?? null) : null}
```
Tara-ag: 💡 MINOR: New Checkbox ariaLabel and Menu triggerAriaLabel props lack test coverage. The PR adds new public props (`Checkbox.ariaLabel` and `Menu.triggerAriaLabel`) to fix accessibility gaps, but only `Input` receives a new regression test. Without tests, future refactors can silently regress the accessible-name behavior these props are meant to provide.

Suggested fix: Add Playwright or Vitest assertions that: (1) a Checkbox with empty `text` and `ariaLabel` exposes the provided accessible name on the `role="checkbox"` element; (2) a Menu with `triggerAriaLabel` exposes the name on the focusable trigger when `interactiveTrigger` is false.

PR #476, opened 2026-08-31
feat(ThinkingIndicator): add chip variant, retire ChatToolStatus internally

murdore:
## Why

A component-redundancy audit flagged `ChatToolStatus` and `ThinkingIndicator` as near-duplicates: same job (a live "the assistant is doing something" status), shipped 4 days apart, with `ChatToolStatus` never migrated onto the newer, more capable primitive. `Chat.svelte` was still hardcoding `<ChatToolStatus>` for its built-in `toolStatus` row.

They weren't quite as identical as they first looked, though: `ChatToolStatus` is a solid bordered/shadowed pill with a static label; `ThinkingIndicator`'s closest existing mode is a borderless row with a shimmering label. So this isn't a find-and-replace — `ThinkingIndicator` gains a genuine new `chip` variant that reproduces `ChatToolStatus`'s exact visual under its own `--thinking-indicator-chip-*` token namespace, defaulting to a **static** (non-shimmering) label so it's a safe drop-in for existing usage — pass `busy` to opt into the shimmer.

## What changed

- `ThinkingIndicator`: new `variant: 'chip'` (`properties.ts` + `ThinkingIndicator.svelte`) — self-contained pill, never expandable, static label unless `busy` is passed.
- `Chat.svelte`: renders `<ThinkingIndicator variant="chip">` instead of `<ChatToolStatus>` for its `toolStatus` row.
- `ChatToolStatus`: **untouched, still exported.** Removing a public export from a published package is its own breaking-change decision — not bundled into this PR. Its docs page now carries a deprecation note pointing at the replacement.
- Docs: `ThinkingIndicator.md` documents the variant + new CSS vars; `ChatToolStatus.md` gets the deprecation note; `Chat.md`/`ToolCallLog.md` cross-references updated.
- Demo site: added a `chip` example to the `ThinkingIndicator` gallery page; `demo.css` gets a dark-theme block for `--thinking-indicator-chip-*` (mirrors the existing `--chat-tool-status-*` one — this is what caught a real dark-mode mismatch during testing, see below).

## Verification

- Visual: compared the new `chip` variant against `ChatToolStatus` directly in the live demo site. First pass showed a real mismatch in dark mode (plain white pill vs. the themed dark pill) — traced to `demo.css` themeing `--chat-tool-status-*` for dark mode but not the new `--thinking-indicator-chip-*` namespace. Fixed, re-verified pixel-matched in both light and dark theme.
- Ran a full `Chat` send cycle live (types a message, sends, tool-status pill shows, reply streams in) — zero console errors.
- `pnpm lint` / `pnpm check` / `pnpm run check:wc` all clean — same pre-existing warnings only (2 `state_referenced_locally` notices, 2 `moduleResolution` notices, 3 pre-existing `LottiePlayer.wc.svelte` errors).
- `pnpm run test:unit -- --run`: 128/128 passed.

## Not included

- `ChatToolStatus` deprecation beyond the docs note — no removal, no runtime warning.
- The other component-redundancy audit findings (Select/Combobox, Carousel/Book doc cross-references, Carousel nav gap, soundKit casing) — those are unrelated `chore`-level items landing in a separate PR.

No merge — for review.

## Summary by CodeRabbit

* **New Features**
  * Added a chip-style variant to the ThinkingIndicator, including static and busy states.
  * Chat now displays tool status using the new ThinkingIndicator chip.
  * Added chip-specific styling and customization options.

* **Documentation**
  * Updated component guides and examples for the new chip variant.
  * Marked ChatToolStatus as deprecated and recommended ThinkingIndicator instead.


src/lib/ThinkingIndicator/ThinkingIndicator.svelte
```diff
@@ -406,6 +410,19 @@
         >{label}</span
       >{:else}{label}{/if}</span
   >
+{:else if variant === 'chip'}
```
Tara-ag, line 413: ⚠️ MAJOR: The deprecated `ChatToolStatus` rendered its root with `aria-live="polite"`, so screen readers announced status changes like "Searching the catalog…". The new `chip` variant renders a plain `<div>` with no live region, so assistive-technology users get no notification when the tool-status label appears or changes inside `Chat`.

**Fix:** Add `aria-live="polite"` (and optionally `aria-atomic="true"`) to the chip variant's root `<div>` to preserve the accessible status semantics that `ChatToolStatus` provided.

Tara-ag, line 413: 💡 MINOR: The PR introduces a new `chip` variant with distinct static/busy shimmer behavior and a new DOM structure, but no unit or integration tests were added. The existing test suite has no `ThinkingIndicator` coverage at all.

**Fix:** Add a Vitest test (e.g., `tests/thinking-indicator-chip-variant.test.ts`) asserting that `variant="chip"` renders `.thinking-indicator-chip`, defaults to a static label, and shimmers when `busy` is `true`.

src/lib/ThinkingIndicator/properties.ts
```diff
@@ -36,7 +36,11 @@ export type OptionalThinkingIndicatorProperties = {
   expanded?: boolean;
   /**
    * `bare` renders only the shimmering label — for chat bubbles where the surrounding
-   * UI already supplies the avatar and layout. It never becomes expandable.
+   * UI already supplies the avatar and layout. `chip` renders a self-contained pill
+   * (bordered, shadowed, its own background) for a live status floating above other
+   * content, e.g. a tool-call indicator above a composer — pass `busy={false}` for a
```
Tara-ag, line 41: 💡 MINOR: The `showElapsed` comment says it has "No effect on the `bare` variant", but the new `chip` variant also does not render an elapsed counter. Consumers reading this may expect a timer on a chip-style tool-status pill.

**Fix:** Update the JSDoc to "No effect on the `bare` or `chip` variants." and mirror that wording in `docs/ThinkingIndicator.md`.

Tara-ag (changes requested):
The new `chip` variant is a sensible consolidation, but the switch inside `Chat` drops the `aria-live="polite"` status announcement that `ChatToolStatus` provided. That is a real accessibility regression for screen-reader users and should be fixed before merge.

Other notes:
- `showElapsed` documentation should note that it also has no effect on `chip`.
- Please add a unit test covering the new `chip` variant's default static label and `busy` shimmer behavior.

No blocking breaking-change concerns: `ChatToolStatus` remains exported, the public barrel is unchanged, and the theming migration is documented.

Tara-ag (commented):

src/lib/ThinkingIndicator/ThinkingIndicator.svelte
Tara-ag, line 413: ⚠️ MAJOR: The chip variant renders a plain `<div>` with no `aria-live` region. The deprecated `ChatToolStatus` had `aria-live="polite"` on its root, and `Chat.svelte` now uses this variant for its `toolStatus` row, so screen-reader users will no longer hear status updates like "Searching the catalog…". Add `aria-live="polite" aria-atomic="true"` to this div to preserve the accessible status announcement.

src/lib/Chat/Chat.svelte
```diff
@@ -106,7 +106,7 @@
       <ChatSuggestions items={suggestions} {disabled} onselect={handleSuggestion} />
     {/if}
     {#if toolStatus !== null}
-      <div class="tool-status"><ChatToolStatus label={toolStatus.label} /></div>
+      <div class="tool-status"><ThinkingIndicator label={toolStatus.label} variant="chip" /></div>
```
Tara-ag, line 109: ⚠️ MAJOR: `ChatToolStatus` supported an `icon` snippet prop, but this line only forwards `label` to `ThinkingIndicator`. Any consumer passing `toolStatus={{ label, icon: myIconSnippet }}` will silently lose its custom icon. Forward `avatar={toolStatus.icon}` (after widening the `ChatToolStatus` type to include the snippet) or document the removed capability in `Chat.md` with a migration note.

src/lib/ThinkingIndicator/ThinkingIndicator.svelte
Tara-ag, line 413: 💡 MINOR: This PR introduces a new public `chip` variant and changes `Chat.svelte`'s internal tool-status rendering, but there are no tests covering it. Add a Vitest test for the chip variant (static label by default, shimmer with `busy`) and a Chat integration test asserting the `toolStatus` row renders the chip.

Tara-ag (changes requested):
Review complete. The gate accepted 10 findings: 3 MAJOR (accessibility regression on the chip variant, loss of the `icon` snippet prop in Chat, and a second accessibility finding on the same line) and 7 MINOR (documentation inconsistencies, missing/reduced-motion behavior, missing tests, and stale JSDoc). The public-API breaking change and accessibility regressions block merge until addressed.

Tara-ag:

Severity counts — 🔒 CRITICAL: 1 · ⚠️ MAJOR: 2 · 💡 MINOR: 2

Reviewed the PR file by file. Found one CRITICAL issue: the built-in Chat `toolStatus` row switches from `ChatToolStatus` to `ThinkingIndicator variant="chip"`, which changes the effective CSS custom property contract from `--chat-tool-status-*` to `--thinking-indicator-chip-*`. This breaks downstream themes that override the documented `--chat-tool-status-*` variables unless backward-compatible mapping or a major-version bump with migration note is provided. All other concerns raised in earlier review threads (aria-live, reduced-motion, showElapsed, tests/docs) appear to have been addressed in the current revision. Blocking pending resolution of the CSS contract break.

- 🔒 **CRITICAL**: Chat built-in toolStatus row silently breaks downstream CSS themes — `src/lib/Chat/Chat.svelte:109`
  This line now renders `<ThinkingIndicator variant="chip">` instead of `<ChatToolStatus>`. The old row was themed by the documented `--chat-tool-status-*` CSS variables (gap, padding, background, border, border-radius, box-shadow, color, font-size, font-weight, max-width, spinner-color, etc.) in docs/Chat.md. The chip variant reads `--thinking-indicator-chip-*` variables instead, so downstream cons…
- ⚠️ **MAJOR**: Web-component wrapper exposes callback props with `on*` prefix — `src/wc/components/ThinkingIndicator.wc.svelte:25`
  src/wc/components/ThinkingIndicator.wc.svelte declares `onToggle`, `onrowselect`, and `onsettled` as Object properties. Svelte custom elements treat `on*` properties as event listeners, so setting `element.onrowselect = fn` or passing `onrowselect={fn}` from Svelte inside the wrapper may be interpreted as an event binding rather than a plain property, leading to silent failures when consumers try …
- ⚠️ **MAJOR**: Web-component `selected` and `collapseDelayMs` typed as Object instead of Number — `src/wc/components/ThinkingIndicator.wc.svelte:30`
  In src/wc/components/ThinkingIndicator.wc.svelte, `selected` and `collapseDelayMs` are declared with `type: 'Object'`. For a custom element, numeric attributes cannot be parsed correctly as Object; `selected` is documented as a bindable number|null and `collapseDelayMs` as a number|null. Using Object means attribute values arrive as strings and are not coerced, breaking the HTML API for these prop…
- 💡 **MINOR**: Chip variant lacks an accessible name for the loading icon — `src/lib/ThinkingIndicator/ThinkingIndicator.svelte:365`
  The chip variant in src/lib/ThinkingIndicator/ThinkingIndicator.svelte renders a `<Loader />` inside `.chip-icon` when no avatar snippet is provided. The surrounding `<div>` has `aria-live="polite" aria-atomic="true"`, but the loader itself is likely an animated image with no accessible name. Screen-reader users hear the label text when it changes, but the loading state itself is not conveyed unle…
- 💡 **MINOR**: New Playwright tests target demo test IDs that do not exist in the PR branch — `tests/thinking-indicator-chip-variant.test.ts:8`
  tests/thinking-indicator-chip-variant.test.ts uses `thinking-indicator-chip-static-demo` and `thinking-indicator-chip-busy-demo`, and tests/chat-tool-status-backcompat.test.ts uses `chat-backcompat-demo`. A search of the PR branch source did not find these test IDs in src/routes/components/thinking-indicator/+page.svelte or src/routes/components/chat/+page.svelte. The tests will fail because the d…

murdore:
Replying to Yama's review (review #pullrequestreview-5065552602, and the CodeRabbit-style pass before it): all 10 findings verified against source before acting, per this branch's own convention.

**Fixed (real):**
- Both `aria-live` findings (same underlying issue, flagged twice) — the `chip` variant's root had no live region at all, while the `ChatToolStatus` it replaces inside `Chat` had `aria-live="polite"`. Restored it (`aria-live="polite" aria-atomic="true"`) on the chip root.
- `showElapsed` on `chip` — silently no-ops today (no `elapsedWatcher` mounted), documented as a `bare`-only exception. Documented `chip` alongside `bare` in both the JSDoc and `docs/ThinkingIndicator.md`, matching actual behavior rather than adding new UI under review pressure.
- `.chip-label`'s shimmer wasn't covered by the `prefers-reduced-motion` query, unlike `.status-label`. Added the same treatment.
- No test coverage existed for the chip variant. Added `tests/thinking-indicator-chip-variant.test.ts` (static default, busy shimmer, aria-live on both) against two newly-`testId`'d demo rows.
- `docs/Chat.md`'s CSS variable section contradicted its own table: it claimed every `--chat-tool-status-*` variable only applies to the deprecated component directly, while the table lists `--chat-tool-status-justify` as `Chat`'s own tool-status row alignment. Verified `Chat.svelte`'s `.tool-status` rule does read that var directly — documented it as the one Chat-level exception.
- `docs/ToolCallLog.md` overclaimed `ThinkingIndicator` as always single-line and as itself "disappearing" when settled. Reworded to describe specifically `Chat`'s chip usage, which `Chat` clears — not a general `ThinkingIndicator` behavior.

**Rejected (false positive):** the claimed loss of an `icon` snippet prop conflates two different same-named declarations. `ChatToolStatusProperties.icon` belongs to the deprecated standalone `<ChatToolStatus>` *component*. `Chat`'s own `toolStatus` prop is typed via the unrelated `ChatToolStatus` *type* in `Chat/types.ts` (`{ tool?, label, state? }`), which has never had an `icon` field — confirmed unchanged by this PR via `git diff` against `release`. There was never a way to pass `icon` through `toolStatus`, so nothing regressed. No code change made.

Verified: `pnpm run check` (0 errors, same 3 pre-existing warnings), `pnpm run lint` (clean), `pnpm run build` succeeds end to end, and the new chip-variant Playwright suite (3/3) passes alongside the pre-existing unit suite.

Squashed to a single commit (`c9e803f`) per the repo's single-commit-per-PR convention.

murdore:
Closing and reopening to force GitHub Actions to re-evaluate triggers — CI and Yama PR Review stopped firing entirely (not failing, never starting) after this branch's first commit touching .github/workflows/ci.yml, while both workflows continued working normally elsewhere in the repo during the same window. No commits, comments, or review history should be affected.

murdore:
**Real testing proof, attached to this PR's own CI run — not just an assertion.**

The `checks` job on commit `0d0e966` ([run 33445757688](https://github.com/juspay/svelte-ui-components/actions/runs/33445757688)) produced a genuine, working `playwright-report` artifact: **241 passed, 10 failed** out of 251 specs — the first real signal this suite has ever produced in CI (every prior run failed 100% of specs on a missing browser binary, now fixed).

All 3 of this PR's own new tests pass:
```
✓ tests/thinking-indicator-chip-variant.test.ts:4:3  › defaults to a static label, no shimmer animation
✓ tests/thinking-indicator-chip-variant.test.ts:12:3 › busy shimmers the label (no static-label class)
✓ tests/thinking-indicator-chip-variant.test.ts:20:3 › the chip root is a polite live region, so a screen reader announces status changes
```

**The artifact itself now contains real, playable video** of every test (`video: 'on'`, added this PR) — 251 `.webm` recordings embedded directly in the report bundle, not just pass/fail text. Verified locally before pushing (`file` confirms valid WebM, non-trivial byte sizes) and confirmed again against the actual CI output: [artifact `playwright-report`, 17.7 MB](https://github.com/juspay/svelte-ui-components/actions/runs/33445757688/artifacts/9778092753).

To watch: download the artifact → unzip → open `index.html` → click into any test → the video plays inline in the report.

The remaining 10 failures are pre-existing and unrelated to this PR (none are in `thinking-indicator-chip-variant.test.ts`) — first real baseline this suite has had, now worth its own triage separately from this PR.

src/lib/Chat/Chat.svelte
Tara-ag, line 109: 🔒 CRITICAL: Chat built-in toolStatus row silently breaks downstream CSS themes. This line now renders `<ThinkingIndicator variant="chip" />` instead of `<ChatToolStatus />`. The old row was themed by the documented `--chat-tool-status-*` CSS variables (gap, padding, background, border, border-radius, box-shadow, color, font-size, font-weight, max-width, spinner-color, etc.) in docs/Chat.md. The chip variant reads `--thinking-indicator-chip-*` variables instead, so downstream consumers overriding `--chat-tool-status-*` will see their theming ignored. This is a breaking change to a public CSS contract without a major-version bump or migration note.

Suggested fix: either preserve backward compatibility by making the built-in Chat toolStatus row continue to honor `--chat-tool-status-*` (e.g., map them onto `--thinking-indicator-chip-*` when rendered here), or bump the major version and add a migration note renaming the variables.

Tara-ag (changes requested):
Blocking: the PR changes a public theming contract without a major-version bump or migration note.

murdore:
Replying to Yama's latest review (\`DECISION: BLOCKED\`): the CSS-contract regression is real, verified against source, and fixed.

**The finding:** Chat's built-in tool-status row now renders `<ThinkingIndicator variant="chip">` instead of `<ChatToolStatus>`, and the two read completely different CSS variable sets (`--thinking-indicator-chip-*` vs `--chat-tool-status-*`, confirmed zero overlap in names). A consumer who had themed the row via e.g. `--chat-tool-status-background` would find it silently ignored — a real breaking change to a public CSS contract.

**The fix:** `Chat.svelte`'s own `.tool-status` wrapper now maps every `--thinking-indicator-chip-*` variable the chip touches onto its `--chat-tool-status-*` equivalent, with defaults matching `ChatToolStatus`'s own byte for byte. This is scoped to Chat's internal usage only — a standalone `<ThinkingIndicator variant="chip">` elsewhere is unaffected and still uses its own variable names. Old consumer overrides now genuinely keep working, making this an actually safe drop-in rather than a claimed one.

**Proof, not just an assertion:** added a dedicated demo instance + test that overrides the OLD variable names the way an existing consumer already would, asserts the computed style picked them up, and — since a prior round of this PR got fair pushback for attaching CI-artifact links instead of real proof — I independently decoded and visually inspected the actual video recording frame by frame (`ffprobe`/`ffmpeg`, not just `file`) to confirm the pill genuinely renders with the overridden colors, not its own defaults.

Squashed to a single commit (`001c0d6`) per the repo's single-commit-per-PR convention.

murdore:
Real video proof, attached directly to this PR (not just linked from a CI artifact) — each one independently verified: decoded with ffprobe/ffmpeg to confirm a genuine, non-corrupt video stream, and every frame visually inspected before attaching.

1. Chip defaults to a static label (no shimmer)
2. `busy` shimmers the label
3. `aria-live="polite"` restored — the accessibility regression Yama flagged, fixed
4. The CSS backward-compat fix: overriding the OLD `--chat-tool-status-*` variables still themes the chip (blue background / gold border visibly applied, not the chip's own defaults)

Uploading now.
[1-static-label-no-shimmer.webm](https://github.com/user-attachments/assets/a3df1fce-eab3-429e-be0a-17772c5e719b)
[2-busy-shimmers-the-label.webm](https://github.com/user-attachments/assets/6aebcda0-9472-4339-93fd-9c9a4630c6ea)
[3-aria-live-region.webm](https://github.com/user-attachments/assets/8e3d5ec1-8e64-40f3-b949-c97c412c7f1c)
[4-chat-backcompat-theming.webm](https://github.com/user-attachments/assets/d83a0fa6-5d86-4526-8cb7-b62df8ac7425)

src/lib/Chat/Chat.svelte
Tara-ag, line 109: 🔒 CRITICAL: Chat built-in toolStatus row silently breaks downstream CSS themes. This line now renders `<ThinkingIndicator variant="chip">` instead of `<ChatToolStatus>`. The old row was themed by the documented `--chat-tool-status-*` CSS variables (gap, padding, background, border, border-radius, box-shadow, color, font-size, font-weight, max-width, spinner-color, etc.) in docs/Chat.md. The chip variant reads `--thinking-indicator-chip-*` variables instead, so downstream consumers overriding `--chat-tool-status-*` will see their theming ignored. This is a breaking change to a public CSS contract without a major-version bump or migration note.

**Fix:** Either preserve backward compatibility by making the built-in Chat toolStatus row continue to honor `--chat-tool-status-*` (e.g., map them onto `--thinking-indicator-chip-*` when rendered here), or bump the major version and add a migration note renaming the variables.

Tara-ag (changes requested):
Reviewed the PR file by file. Found one CRITICAL issue: the built-in Chat `toolStatus` row switches from `ChatToolStatus` to `ThinkingIndicator variant="chip"`, which changes the effective CSS custom property contract from `--chat-tool-status-*` to `--thinking-indicator-chip-*`. This breaks downstream themes that override the documented `--chat-tool-status-*` variables unless backward-compatible mapping or a major-version bump with migration note is provided. All other concerns raised in earlier review threads (aria-live, reduced-motion, showElapsed, tests/docs) appear to have been addressed in the current revision. Blocking pending resolution of the CSS contract break.

murdore:
Replying to Yama's latest review: this finding is stale. The exact fix it's asking for is already present in the commit it reviewed (\`001c0d6\`).

`src/lib/Chat/Chat.svelte` lines 163-182, in the `.tool-status` rule:
```css
--thinking-indicator-chip-gap: var(--chat-tool-status-gap, 8px);
--thinking-indicator-chip-padding: var(--chat-tool-status-padding, 8px 14px);
--thinking-indicator-chip-background: var(--chat-tool-status-background, #ffffff);
--thinking-indicator-chip-border: var(--chat-tool-status-border, 1px solid #e4e4e7);
--thinking-indicator-chip-border-radius: var(--chat-tool-status-border-radius, 999px);
--thinking-indicator-chip-box-shadow: var(--chat-tool-status-box-shadow, 0 6px 20px rgba(0, 0, 0, 0.08));
--thinking-indicator-chip-max-width: var(--chat-tool-status-max-width, 100%);
--thinking-indicator-chip-color: var(--chat-tool-status-color, #52525b);
--thinking-indicator-chip-font-size: var(--chat-tool-status-font-size, 0.85rem);
--thinking-indicator-chip-font-weight: var(--chat-tool-status-font-weight, 500);
--thinking-indicator-chip-icon-color: var(--chat-tool-status-indicator-color, currentColor);
--thinking-indicator-chip-spinner-color: var(--chat-tool-status-spinner-color, currentColor);
--thinking-indicator-chip-spinner-color-end: var(--chat-tool-status-spinner-color-end, transparent);
--thinking-indicator-chip-spinner-size: var(--chat-tool-status-spinner-size, 14px);
```

Every `--thinking-indicator-chip-*` variable the chip reads is sourced from its `--chat-tool-status-*` equivalent first, with `ChatToolStatus`'s own original default as the final fallback. An existing consumer overriding `--chat-tool-status-background` (etc.) still themes the chip correctly — that's not a claim, it's demonstrated on video in [this comment](https://github.com/juspay/svelte-ui-components/pull/476#issuecomment-5486174055) (video 4: the pill renders with the overridden blue background and gold border, set entirely through the old variable names).

If there's a specific override this mapping doesn't cover, or a case where it doesn't behave as described, please point to it concretely — happy to fix a real gap. As it stands I can't find one, and the finding's own text is unchanged from before the fix landed, which suggests it may not have picked up this revision's diff.

docs/Chat.md
```diff
@@ -218,7 +218,7 @@ The fixed-height `.chat-panel` gives `Chat` its bounds; add a slide/scale transi
 | `--chat-footer-border-top`    | `none`        | border-top       | Border above the footer.             |
 | `--chat-tool-status-justify`  | `center`      | justify-content  | Alignment of the tool-status row.    |
 
-Child components (`ChatHeader`, `ChatMessageList`, `ChatComposer`, `ChatToolStatus`, `ChatSuggestions`, `ChatMessage`) are themed through their own CSS variables, which cascade into `Chat`.
+Child components (`ChatHeader`, `ChatMessageList`, `ChatComposer`, `ThinkingIndicator`, `ChatSuggestions`, `ChatMessage`) are themed through their own CSS variables, which cascade into `Chat`. The tool-status row specifically uses `ThinkingIndicator`'s `--thinking-indicator-chip-*` variables (see its docs) — not `--chat-tool-status-*`, which only applies if you use the deprecated `ChatToolStatus` directly.
```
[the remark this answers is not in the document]
murdore: Addressed in commit 0c9a9b39f (PR 476). docs/Chat.md:225-227 on release 3.1.0 keeps --chat-tool-status-justify in the CSS variable table and the paragraph right below it now states the exception directly: the tool-status row is ThinkingIndicator's chip variant internally, but every --chat-tool-status-* variable still controls it via an internal mapping onto --thinking-indicator-chip-*, so it no longer contradicts the table.

docs/ToolCallLog.md
```diff
@@ -8,9 +8,9 @@ open at a time. The popover portals to `document.body` (reusing Menu's `usePorta
 and repositions itself on scroll/resize, so an `overflow: hidden` or scrolling ancestor — a chat
 bubble, a narrow card — can never clip it.
 
-Unlike `ChatToolStatus` (a single one-line "using a tool now" status that disappears when the turn
-settles), `ToolCallLog` is the persistent, multi-call record: it stays in the transcript after the
-turn, showing every tool the model called in order.
+Unlike `ThinkingIndicator` (a single one-line "using a tool now" status that disappears when the
+turn settles), `ToolCallLog` is the persistent, multi-call record: it stays in the transcript after
+the turn, showing every tool the model called in order.
```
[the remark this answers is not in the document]
murdore: Addressed in commit 0c9a9b39f (PR 476). docs/ToolCallLog.md:9-11 on release 3.1.0 now scopes the comparison to Chat's specific usage: "Unlike Chat's tool-status row (a ThinkingIndicator chip showing 'using a tool now,' cleared by the host once the turn settles), ToolCallLog is the persistent, multi-call record." It no longer makes a general claim about ThinkingIndicator's behavior.

src/lib/ThinkingIndicator/ThinkingIndicator.svelte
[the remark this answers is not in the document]
murdore: Declined in favor of the documented-behavior option this finding itself offered. showElapsed is documented, not implemented, for chip: ThinkingIndicator/properties.ts:49-52 and docs/ThinkingIndicator.md:75 on release 3.1.0 both say 'No effect on the bare or chip variants.' The chip branch (ThinkingIndicator.svelte:413-427) deliberately mounts no elapsedWatcher and renders no counter, matching that documented contract.

src/lib/ThinkingIndicator/ThinkingIndicator.svelte
```diff
@@ -558,6 +575,69 @@
...
+    color: var(--thinking-indicator-chip-color, #52525b);
+    background: var(
+      --thinking-indicator-chip-shimmer-gradient,
+      linear-gradient(90deg, #52525b 0%, #a0a0a0 50%, #52525b 100%)
+    );
+    background-size: 200% 100%;
+    background-clip: text;
+    -webkit-text-fill-color: transparent;
+    animation: thinking-indicator-shimmer var(--thinking-indicator-chip-shimmer-duration, 2s) linear
+      infinite;
```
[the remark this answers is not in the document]
murdore: Fixed in commit 0c9a9b39f (PR 476). ThinkingIndicator.svelte:867-887 on release 3.1.0 adds .chip-label to the prefers-reduced-motion animation-duration list and gives it its own animation: none rule with the static chip color applied, matching the treatment .status-label already had.

src/lib/ThinkingIndicator/ThinkingIndicator.svelte
murdore: Fixed in commit 0c9a9b39f (PR 476). The chip variant's root div now carries aria-live="polite" aria-atomic="true" (ThinkingIndicator.svelte:416-417 on release 3.1.0), restoring the same announcement behavior the deprecated ChatToolStatus had. Pinned by a Playwright test at tests/thinking-indicator-chip-variant.test.ts:20-33.

murdore: Added in commit 0c9a9b39f (PR 476): tests/thinking-indicator-chip-variant.test.ts on release 3.1.0 covers the chip variant with three Playwright tests - defaults to a static label with no shimmer, shimmers when busy, and the root is a polite live region in both states.

src/lib/ThinkingIndicator/properties.ts
murdore: Fixed in commit 0c9a9b39f (PR 476), using the exact wording suggested. ThinkingIndicator/properties.ts:49-52 on release 3.1.0 now reads 'No effect on the bare or chip variants,' and docs/ThinkingIndicator.md:75 was updated to match.

src/lib/ThinkingIndicator/ThinkingIndicator.svelte
murdore: Same fix as the other aria-live thread on this line: ThinkingIndicator.svelte:416-417 on release 3.1.0 carries aria-live="polite" aria-atomic="true" on the chip variant's root div, landed in commit 0c9a9b39f and pinned by tests/thinking-indicator-chip-variant.test.ts:20-33.

src/lib/Chat/Chat.svelte
murdore: Declined - this is a false positive that confuses two different same-named declarations. Chat's own toolStatus prop is typed via ChatToolStatus in src/lib/Chat/types.ts:23-27 as { tool?, label, state? } on release 3.1.0, and it has never had an icon field, before or after this PR. The icon snippet prop belongs to a different, unrelated type - ChatToolStatusProperties for the deprecated standalone <ChatToolStatus> component (src/lib/ChatToolStatus/properties.ts). Nothing regressed here; no code change made.

src/lib/ThinkingIndicator/ThinkingIndicator.svelte
murdore: Same fix as the other test-coverage thread on this line: tests/thinking-indicator-chip-variant.test.ts on release 3.1.0 (added in commit 0c9a9b39f) covers the chip variant directly - static default, busy shimmer, and aria-live on both states.

src/lib/Chat/Chat.svelte
murdore: Fixed within the same PR (merge commit 0c9a9b39f). Chat.svelte's .tool-status rule (lines 168-186 on release 3.1.0) now maps every --thinking-indicator-chip-* variable the chip reads back onto its --chat-tool-status-* equivalent - gap, padding, background, border, border-radius, box-shadow, max-width, color, font-size, font-weight, spinner-color, spinner-color-end and spinner-size - so existing consumer overrides of --chat-tool-status-* keep applying exactly as before, with no major bump needed.

murdore: Same fix as the other CSS-variable thread on this line: Chat.svelte's .tool-status rule (lines 168-186 on release 3.1.0, landed in commit 0c9a9b39f) maps every --thinking-indicator-chip-* variable back onto its --chat-tool-status-* equivalent, preserving backward compatibility for existing theming overrides without a major-version bump.

PR #478, opened 2026-08-31
docs: cross-reference the five remaining low-severity audit pairs

murdore:
## Why

Closes out the last of the low-severity findings from the component-redundancy audit (companion PRs: #476 for ChatToolStatus/ThinkingIndicator, #477 for Select/Combobox, Carousel/Book, and the Carousel bug/SoundKit rename). None of these five pairs are actually redundant — each has a real, verifiable distinction — but none of it was written down, unlike most other near-pairs in the library (`ChipInput`↔`SplitInput`, `DateRangePicker`→`Calendar`, `Resizable`↔`Draggable`).

## What changed

Doc-only, one added sentence per file, ten files:

- **Checkbox ↔ Toggle**: `Checkbox` supports `indeterminate` (no `Toggle` equivalent — it only makes sense for select-some-of-many semantics); `Toggle` is strictly two-state.
- **ListItem ↔ CheckListItem**: `ListItem` has no selection/checkbox concept at all (multi-section row, per-section click handlers, expandable); `CheckListItem` is purpose-built for a single checkbox + label row.
- **StatCard ↔ KeyValue**: `StatCard` is a single prominent metric with delta-badge and dashboard framing; `KeyValue` is a plain multi-field "label: value" grid with no big-number emphasis.
- **Gauge ↔ Progress**: `Progress` has an indeterminate/unknown-duration mode `Gauge` lacks (a ring reads awkwardly as "unknown duration"); otherwise circular vs. linear.
- **Loader ↔ LoadingDots**: `LoadingDots` is built to sit inline inside text/buttons; `Loader` is a standalone ring/donut shape not designed for inline text-flow.

## Verification

Prose-only changes to `docs/*.md` — no `src/` files touched. `pnpm lint` clean (confirms nothing outside `docs/` was affected).

No merge — for review.


Tara-ag (approved):
Doc-only PR adding accurate component cross-references. Verified each claim against the corresponding source components; all distinctions are correct. No public API, accessibility, SSR, or security concerns.

murdore:
## This PR's branch has been deleted from the remote

I came here to attach proof of testing and found something that matters more.

`docs/remaining-component-cross-references` **no longer exists on `origin`**. This PR is still open and the API still reports its head as `b75b31c47`, but:

```
git ls-remote --heads origin        -> 31 branches, none under docs/
git merge-base --is-ancestor b75b31c47 origin/release  -> NO, not on release
```

So the work is neither merged nor present on the remote. The commit currently survives **only in one local clone** of mine, where a local branch still references it. It was in a `git ls-remote` listing I took roughly an hour before writing this, so it was deleted recently rather than long ago.

I have not pushed it back. Restoring a branch someone may have deleted deliberately is not my call to make unilaterally, and the commit is safe where it is for now. Say the word and it is a one-line restore:

```
git push origin b75b31c47:refs/heads/docs/remaining-component-cross-references
```

If the deletion was intentional, this PR should be closed rather than left open pointing at a ref that no longer resolves.

## On proof of testing

Had the branch been intact, this PR would still get a note rather than a recording: it changes documentation cross-references only, and there is no rendered surface to film. The check I intended to run — parsing every added markdown link and asserting each target file exists — is the right evidence for a docs PR, and I will run it if the branch comes back.

I want to be explicit that I did **not** run it, because I nearly reported that I had. My first attempt checked out the branch, silently fell through to `release` when the fetch failed, and diffed `release` against itself — which produced a confident-looking `0 broken references` from comparing a tree to itself. That number would have been meaningless.

PR #480, opened 2026-08-31
fix(a11y): announce Input validation errors and let ChipInput be named

murdore:
Two findings from a cross-surface UI audit. Both produce a control that is correct on screen and absent from the accessibility tree, so nothing visual catches either one.

## Input — validation messages were never announced

The error text rendered as a plain `<div class="error-message">` with no `id`, no `role`, and nothing tying it to the field it described. Across 8 audited form surfaces, **0** error nodes carried `role="alert"` or sat inside an `aria-live` region — including the five that render visible text like "Enter a percentage between 1 and 100". A screen-reader user submitted, heard nothing, and was left on a form that had not moved.

Now:
- the field carries `aria-invalid="true"` while in error,
- it points at its message through `aria-describedby`,
- the message is a `role="alert"` live region, so it is spoken as it appears.

Both attributes are **absent while the field is valid** — otherwise every field on a healthy form announces itself as broken. That negative half is asserted too.

## ChipInput — the draft field could not be named

`ChipInputProperties` exposed `placeholder`, `disabled`, `testId` and `classes`. ChipInput draws no label of its own, so a caption rendered beside it had no way to reach the control, and the draft field arrived unnamed however the page read visually. It now accepts `ariaLabel` and forwards it to the `Input` it wraps.

This is a real blocker downstream: Lighthouse's `/order/edit` has two ChipInputs captioned "Blocked order tags" and "Blocked product tags" that cannot be named until this ships. Its own a11y spec currently marks them as an expected failure so the gap stays visible.

## Verification

`tests/input-error-announcement.test.ts` — 4 new tests, plus the 3 existing label-association tests still green (7 passed).

Negative-controlled against the **source**, not just the expectations: removing `role="alert"`, `aria-invalid`/`aria-describedby` and the `ariaLabel` pass-through fails 3 of the 4 new tests on the right assertions (lines 14, 30, 51). The fourth correctly keeps passing, because it asserts *absence*. Restored afterwards and confirmed both files hash byte-identical to the state that passed.

`lint` 0 · `build` 0. `svelte-check` reports 2 pre-existing errors in `tests/media-upload.test.ts` (`@types/node` not installed in this worktree) — untouched by this change.

Demo instances added to `/components/input` and `/components/chip-input`, which are both the documentation and the test fixtures.

## Summary by CodeRabbit

* **Accessibility**
  * Added an optional accessible name for ChipInput draft fields.
  * Validation errors are now announced to assistive technologies through invalid-state and error-message associations.
  * Valid inputs omit error-specific accessibility attributes.
* **Documentation**
  * Added guidance and examples for naming ChipInput fields and announcing Input validation errors.
* **Bug Fixes**
  * Improved accessibility behavior for invalid inputs and ChipInput fields.
* **Tests**
  * Added automated accessibility coverage for validation announcements and ChipInput naming.

### DIFF
--- a/src/lib/ChipInput/ChipInput.svelte
+++ b/src/lib/ChipInput/ChipInput.svelte
@@ -5,6 +5,7 @@
 
   let {
     values = $bindable([]),
+    ariaLabel,
     placeholder = 'Add value…',
     disabled = false,
     testId,
@@ -66,6 +67,7 @@
       {placeholder}
       dataType="text"
       name=""
+      {ariaLabel}
       autoComplete="off"
       actionInput={false}
       disable={disabled}
--- a/src/lib/Input/Input.svelte
+++ b/src/lib/Input/Input.svelte
@@ -122,6 +122,26 @@
   // forceError lets consumers drive the error border from server/runtime validation,
   // independent of validationPattern.
   const showError = $derived(showErrorMessage || forceError);
+  // The error text has to be reachable FROM the field it describes and announced when it
+  // appears. Without both, a screen-reader user submits, hears nothing, and is left on a form
+  // that did not move -- the message is on screen and absent from the accessibility tree.
+  const errorMessageId = $derived(`${effectiveId}-error`);
+  // `onErrorMessage` is `string | null`, so null has to be excluded explicitly: `null !== ''`
+  // is true, which would describe the field by an alert element carrying no message.
+  const isShowingError = $derived(
+    onErrorMessage != null && onErrorMessage !== '' && showError && !actionInput
+  );
+  const infoMessageId = $derived(`${effectiveId}-info`);
+  const isShowingInfo = $derived(infoMessage !== '' && !actionInput);
+  // Helper text is part of the field's description, not decoration: a consumer's `infoMessage`
+  // ("Enter a percentage between 1 and 100") is exactly the guidance a screen-reader user needs
+  // BEFORE they trip an error. Reference both, in reading order, so the field is described by
+  // everything visibly attached to it rather than only by its failure.
+  const describedBy = $derived(
+    [isShowingError ? errorMessageId : null, isShowingInfo ? infoMessageId : null]
+      .filter(Boolean)
+      .join(' ')
+  );
   const hasLeftIcon = $derived(typeof leftIcon === 'function');
   const hasRightIcon = $derived(typeof rightIcon === 'function');
 
@@ -301,6 +321,8 @@
         aria-controls={ariaControls}
         aria-activedescendant={ariaActivedescendant}
         aria-required={mandatory || null}
+        aria-invalid={showError ? 'true' : null}
+        aria-describedby={describedBy || null}
         required={mandatory || null}
         onfocus={onFocus}
         onfocusout={_onFocusOut}
@@ -337,6 +359,8 @@
         aria-controls={ariaControls}
         aria-activedescendant={ariaActivedescendant}
         aria-required={mandatory || null}
+        aria-invalid={showError ? 'true' : null}
+        aria-describedby={describedBy || null}
         required={mandatory || null}
         onfocus={onFocus}
         onfocusout={_onFocusOut}
@@ -399,8 +423,10 @@
     {@render fieldElement()}
   {/if}
 
-  {#if onErrorMessage !== '' && showError && !actionInput}
+  {#if isShowingError}
     <div
+      id={errorMessageId}
+      role="alert"
       class="error-message"
       data-pw={typeof testId === 'string' && testId.length > 0 ? `${testId}-error-message` : null}
       testID={typeof testId === 'string' && testId.length > 0 ? `${testId}-error-message` : null}
@@ -408,8 +434,8 @@
       {onErrorMessage}
     </div>
   {/if}
-  {#if infoMessage !== '' && !actionInput}
-    <div class="info-message">
+  {#if isShowingInfo}
+    <div id={infoMessageId} class="info-message">
       {infoMessage}
     </div>
   {/if}
--- a/tests/input-error-announcement.test.ts
+++ b/tests/input-error-announcement.test.ts
@@ -1,0 +1,108 @@
+import { expect, test } from '@playwright/test';
+
+test.describe('Input — validation messages reach the accessibility tree', () => {
+  // Regression guard: the error text was rendered as a plain <div class="error-message"> with
+  // no id, no role and nothing tying it to the field. It was drawn on screen and completely
+  // absent from the accessibility tree — a screen-reader user submitted, heard nothing, and was
+  // left on a form that had not moved. Nothing visual catches this: the message looks correct.
+  test('an invalid field is marked invalid and points at its message', async ({ page }) => {
+    await page.goto('/components/input');
+
+    const field = page.getByTestId('input-announced-error');
+    await expect(field).toBeVisible();
+
+    await expect(field).toHaveAttribute('aria-invalid', 'true');
+
+    const describedBy = await field.getAttribute('aria-describedby');
+    expect(describedBy, 'an invalid field must reference its message').toBeTruthy();
+
+    // aria-describedby is a SPACE-SEPARATED id list, not a single id — a field can be described
+    // by its error and its helper text at once. Resolve the first reference.
+    const [errorId] = String(describedBy).split(/\s+/).filter(Boolean);
+    await expect(page.locator(`[id="${errorId}"]`)).toHaveText('Enter a valid email address');
+  });
+
+  test('the message is a live region so it is spoken when it appears', async ({ page }) => {
+    await page.goto('/components/input');
+
+    const field = page.getByTestId('input-announced-error');
+    const describedBy = await field.getAttribute('aria-describedby');
+    const [errorId] = String(describedBy).split(/\s+/).filter(Boolean);
+
+    await expect(page.locator(`[id="${errorId}"]`)).toHaveAttribute('role', 'alert');
+  });
+
+  test('helper text is associated with the field even when it is valid', async ({ page }) => {
+    await page.goto('/components/input');
+
+    // infoMessage is guidance a screen-reader user needs BEFORE they trip an error, so it must
+    // be referenced whether or not the field is currently invalid.
+    const field = page.getByTestId('input-helper-only');
+    const describedBy = await field.getAttribute('aria-describedby');
+    expect(describedBy, 'helper text must be referenced by the field').toBeTruthy();
+    await expect(page.locator(`[id="${describedBy}"]`)).toHaveText('Shown to your teammates.');
+
+    // Valid field: described, but not marked invalid.
+    await expect(field).not.toHaveAttribute('aria-invalid', /.*/);
+  });
+
+  test('error and helper text are both referenced, error first', async ({ page }) => {
+    await page.goto('/components/input');
+
+    const field = page.getByTestId('input-announced-error');
+    const describedBy = (await field.getAttribute('aria-describedby')) ?? '';
+    const ids = describedBy.split(/\s+/).filter(Boolean);
+
+    // Reading order matters: assistive technology announces the references in the order listed,
+    // and the failure should come before the advice.
+    expect(ids).toHaveLength(2);
+    await expect(page.locator(`[id="${ids[0]}"]`)).toHaveText('Enter a valid email address');
+    await expect(page.locator(`[id="${ids[1]}"]`)).toHaveText(
+      'Use your company address, not a personal one.'
+    );
+  });
+
+  test('a valid field with no helper text is not marked invalid and describes nothing', async ({
+    page
+  }) => {
+    await page.goto('/components/input');
+
+    // Negative side of the contract: the attributes must be ABSENT when there is no error,
+    // otherwise every field on every form announces itself as invalid.
+    // /.*/ matches ANY value including the empty string, so these reject a present-but-empty
+    // attribute too. Asserting `not.toHaveAttribute('aria-invalid', 'true')` would instead pass
+    // on aria-invalid="false", which is a value the field must not carry at all.
+    const healthy = page.getByTestId('input-datatype-time');
+    await expect(healthy).not.toHaveAttribute('aria-invalid', /.*/);
+    await expect(healthy).not.toHaveAttribute('aria-describedby', /.*/);
+  });
+});
+
+test.describe('ChipInput — the draft field can be named', () => {
+  // ChipInput renders no visible label of its own, so before `ariaLabel` existed there was no
+  // way for a caller to name it: the caption beside it read on screen and the control reached
+  // the accessibility tree unnamed.
+  test('ariaLabel names the draft field', async ({ page }) => {
+    await page.goto('/components/chip-input');
+
+    await expect(page.getByTestId('chip-input-tags-input')).toHaveAttribute(
+      'aria-label',
+      'Product tags'
+    );
+    await expect(page.getByLabel('Product tags')).toHaveAttribute(
+      'data-pw',
+      'chip-input-tags-input'
+    );
+  });
+
+  test('an unnamed ChipInput emits no empty aria-label', async ({ page }) => {
+    await page.goto('/components/chip-input');
+
+    // An empty aria-label is worse than none — it names the control the empty string, which
+    // suppresses every other naming path an assistive technology would have fallen back to.
+    await expect(page.getByTestId('chip-input-accent-input')).not.toHaveAttribute(
+      'aria-label',
+      /.*/
+    );
+  });
+});
--- a/docs/ChipInput.md
+++ b/docs/ChipInput.md
@@ -36,6 +36,7 @@ A free-text tag/chip input composed from `Input` and `Pill`. The user types into
 | Prop        | Type       | Required | Default        | Description                                                                                                        |
 | ----------- | ---------- | -------- | -------------- | -------------------------------------------------------------------------------------------------------------------- |
 | values      | `string[]` | Yes      | `[]`            | Bindable. The committed chips, in insertion order. Replaced with a new array by add/dismiss (`values` is reassigned, not mutated in place -- do not rely on array reference identity); also settable externally.        |
+| ariaLabel   | `string`   | No       | `-`             | Accessible name for the draft field; ChipInput renders no visible label of its own.                                  |
 | placeholder | `string`   | No       | `'Add value…'`  | Placeholder text shown in the empty draft field.                                                                     |
 | disabled    | `boolean`  | No       | `false`         | Disables the draft field and hides each chip's dismiss control (chips remain visible, but cannot be added or removed). |
 | testId      | `string`   | No       | `-`             | Value for the `data-pw`/native `testid` attributes on the container. Chips get `` `${testId}-chip` `` and the draft field gets `` `${testId}-input` ``. |
@@ -88,3 +89,30 @@ Override these custom properties to theme the component.
 
 - `Input` -- renders the draft text field that a user types a new chip's text into.
 - `Pill` -- renders each committed chip, in `dismissible` mode (except while `disabled`).
+
+### Naming the control
+
+ChipInput draws no label of its own, so a caption rendered beside it is not associated with the
+field -- it reads on screen and is absent from the accessibility tree. Give the control the same
+words:
+
+```svelte
+<span id="order-tags-caption">Blocked order tags</span>
+<ChipInput bind:values={orderTags} ariaLabel="Blocked order tags" testId="order-tags" />
+```
+
+Prefer the caption's own wording over a different phrase: when the visible label and the
+accessible name disagree, speech-input users cannot activate the control by saying what they
+see (WCAG 2.5.3, Label in Name).
+
+## Web component
+
+None. Like `SplitInput` and `FileDropzoneTrigger`, ChipInput is a simple presentational component
+consumed directly from Svelte; there is no `sui-chip-input` custom element and it is not
+registered in `src/wc/index.ts`.
+
+This is deliberate, and it is why `ariaLabel` has no kebab-case attribute mapping: there is no
+custom element for an attribute to map onto. Adding one would newly expose ChipInput as a web
+component rather than restore parity with an existing element. If ChipInput should be exposed,
+that is its own change — a new tag, a new shadow root, and the attribute-casing tests the other
+wrappers carry.
--- a/docs/Input.md
+++ b/docs/Input.md
@@ -202,3 +202,28 @@ Tag: `<sui-input>`
 ```html
 <sui-input placeholder="Enter email" data-type="email" label="Email"></sui-input>
 ```
+
+### Validation errors are announced
+
+When a field is invalid, `Input` marks it `aria-invalid="true"` and links it to its message with
+`aria-describedby`; the message element is a `role="alert"` live region, so assistive technology
+speaks it as it appears instead of leaving it as pixels on screen. Both attributes are absent
+while the field is valid, so a healthy form does not announce every field as broken.
+
+```svelte
+<Input
+  bind:value={email}
+  label="Work email"
+  forceError={serverRejected}
+  onErrorMessage="Enter a valid email address"
+/>
+```
+
+`forceError` drives this from server-side or runtime validation; `validationPattern` drives it
+from the field's own contents. Either route produces the same announced message.
+
+`infoMessage` is treated the same way: helper text is part of the field's description, so it
+is referenced by `aria-describedby` whether or not the field is currently invalid. When both
+are shown the field references them in reading order — the error first, then the advice —
+since `aria-describedby` takes a space-separated id list and assistive technology announces
+them in the order given.
--- a/src/lib/ChipInput/properties.ts
+++ b/src/lib/ChipInput/properties.ts
@@ -13,6 +13,12 @@ export type MandatoryChipInputProperties = {
 };
 
 export type OptionalChipInputProperties = {
+  /**
+   * Names the draft field for assistive technology. ChipInput renders no visible label of its
+   * own, so without this the control reaches the accessibility tree unnamed however the caption
+   * beside it reads on screen. Pass the same words as that caption.
+   */
+  ariaLabel?: string;
   placeholder?: string;
   disabled?: boolean;
   testId?: string;
--- a/src/routes/components/chip-input/+page.svelte
+++ b/src/routes/components/chip-input/+page.svelte
@@ -14,7 +14,12 @@
 
 <div class="demo-row" style="max-width: 400px;">
   <h3>Product tags</h3>
-  <ChipInput bind:values={productTags} placeholder="Add tag…" testId="chip-input-tags" />
+  <ChipInput
+    bind:values={productTags}
+    placeholder="Add tag…"
+    ariaLabel="Product tags"
+    testId="chip-input-tags"
+  />
   <p>Values: {productTags.join(', ') || '(empty)'}</p>
 </div>
 
--- a/src/routes/components/input/+page.svelte
+++ b/src/routes/components/input/+page.svelte
@@ -11,6 +11,8 @@
   let scheduleTime = $state('09:30');
   let configJson = $state('{ "enabled": true }');
   let readonlySnapshot = $state('{"captured":"selector"}');
+  let invalidEmail = $state('not-an-email');
+  let helperOnly = $state('');
   let pastedInto = $state('');
   let pasteCount = $state(0);
 </script>
@@ -18,6 +20,38 @@
 <div class="page-header">
   <span class="category-badge">Form Controls</span>
   <h1>Input</h1>
+</div>
+
+<h3>Validation errors are announced</h3>
+<p>
+  When a field is invalid it carries <code>aria-invalid</code> and points at its message through
+  <code>aria-describedby</code>; the message itself is a <code>role="alert"</code> live region, so it
+  is spoken when it appears rather than only drawn on screen.
+</p>
+<div class="demo-row" style="max-width: 400px;">
+  <Input
+    bind:value={invalidEmail}
+    label="Work email"
+    dataType="text"
+    forceError
+    onErrorMessage="Enter a valid email address"
+    infoMessage="Use your company address, not a personal one."
+    testId="input-announced-error"
+  />
+</div>
+
+<h3>Helper text is associated too</h3>
+<p>
+  <code>infoMessage</code> is part of the field's description, so it is referenced by
+  <code>aria-describedby</code> even when the field is valid.
+</p>
+<div class="demo-row" style="max-width: 400px;">
+  <Input
+    bind:value={helperOnly}
+    label="Display name"
+    infoMessage="Shown to your teammates."
+    testId="input-helper-only"
+  />
 </div>
 
 <h3>Basic</h3>


Tara-ag (commented):
Review in progress

src/lib/ChipInput/ChipInput.svelte
```diff
@@ -5,6 +5,7 @@
 
   let {
     values = $bindable([]),
+    ariaLabel,
```
Tara-ag, line 8: ⚠️ MAJOR: New `ariaLabel` prop is not exposed on the web-component wrapper. `ChipInput.svelte` now accepts `ariaLabel` and forwards it to the inner `Input`, but there is no `src/wc/components/ChipInput.wc.svelte` wrapper and `src/wc/index.ts` does not register a `sui-chip-input` custom element. The repo requires every camelCase Svelte prop to be mirrored in the web-component wrapper with an explicit kebab-case attribute mapping.

Add `src/wc/components/ChipInput.wc.svelte` that maps `ariaLabel` → `aria-label` (plus the existing props) and import it in `src/wc/index.ts`, following the `Input.wc.svelte` pattern.

Tara-ag (approved):
Accessibility improvements are sound and backward-compatible. One non-blocking note: the new `ChipInput.ariaLabel` prop has no web-component wrapper mapping because `ChipInput` is not currently registered as a custom element; please add `src/wc/components/ChipInput.wc.svelte` and import it in `src/wc/index.ts` if the component is meant to be exposed as a web component.

src/lib/ChipInput/ChipInput.svelte
Tara-ag, line 8: ⚠️ MAJOR: New `ariaLabel` prop is not exposed on the web-component wrapper.

`ChipInput.svelte` now accepts `ariaLabel` and forwards it to the inner `Input`. However, there is no `src/wc/components/ChipInput.wc.svelte` wrapper and `src/wc/index.ts` does not register a `<sui-chip-input>` custom element, so the new public Svelte prop has no corresponding web-component attribute mapping. Per repo parity rules, every camelCase Svelte prop must be mirrored in the web-component wrapper with an explicit kebab-case attribute.

**Suggestion:** Add `src/wc/components/ChipInput.wc.svelte` that maps `ariaLabel` to `aria-label` and register it in `src/wc/index.ts`, mirroring the existing `Input.wc.svelte` pattern:

```svelte
<svelte:options
  customElement={{
    tag: 'sui-chip-input',
    shadow: 'open',
    props: {
      values: { type: 'Array' },
      ariaLabel: { type: 'String', attribute: 'aria-label' },
      placeholder: { type: 'String' },
      disabled: { type: 'Boolean' },
      classes: { type: 'String' }
    }
  }}
/>

<script lang="ts">
  import ChipInput from '$lib/ChipInput/ChipInput.svelte';
  let props = $props();
</script>

<ChipInput {...props} />
```

Tara-ag (approved):
Approved with one parity concern noted inline. The Input accessibility improvements are correct and non-breaking; ChipInput's new `ariaLabel` prop is a useful addition but currently lacks a web-component wrapper mapping.

murdore:
Both CodeRabbit findings were valid and are fixed in `b9537d1`. The Yama note I'm refuting, with evidence.

### ✅ `onErrorMessage={null}` — fixed

Valid, and worse than the comment states. `onErrorMessage` is typed `string | null` (`src/lib/Input/properties.ts:17`), so `null` is a legal value a TypeScript caller can pass — this isn't only reachable from JS. `null !== ''` is true, so the field would have been described by an alert element carrying no message: strictly worse than the pre-change behaviour, where the empty div at least wasn't referenced by `aria-describedby`.

```svelte
const isShowingError = $derived(
  onErrorMessage != null && onErrorMessage !== '' && showError && !actionInput
);
```

### ✅ Negative assertions — fixed, though for a different reason than described

The comment says "the name-only `toHaveAttribute` matcher is not supported". That doesn't apply — the test never used a name-only matcher; both assertions passed a value. The real weakness is narrower and worth naming precisely:

- `not.toHaveAttribute('aria-invalid', 'true')` rejects only the literal `"true"`, so it would have **passed on `aria-invalid="false"`** — a value the field must not carry at all.
- `not.toHaveAttribute('aria-describedby', /.+/)` requires one-or-more characters, so it would have **passed on `aria-describedby=""`**.

Both now use `/.*/`, which matches any value including the empty string, so a present-but-empty attribute fails. Same fix, correct reasoning recorded in a comment above the assertions.

### ❌ ChipInput web-component wrapper — not applicable

> please add `src/wc/components/ChipInput.wc.svelte` and import it in `src/wc/index.ts` if the component is meant to be exposed as a web component

It isn't, and the conditional in that sentence is the operative part. `ChipInput` is not registered as a custom element:

```
$ grep -n "ChipInput" src/wc/index.ts
(no matches)
```

There is no `ChipInput` custom element for an attribute mapping to attach to, so a `.wc.svelte` wrapper would be net-new surface area — exposing a component as a web component for the first time — rather than parity for this change. That's a separate decision from an a11y fix, and I'd rather not smuggle it in here. If ChipInput *should* be exposed, it wants its own PR with the attribute-casing tests the other wrappers have.

Worth contrasting with #475, where the parity note **was** correct and I acted on it: `Input`, `Menu` and `Checkbox` all have registered custom elements, so their new props genuinely did need kebab-case mappings, and those were added.

### Verification

7 tests pass (4 new + 3 existing). `lint` 0, `build` 0.

The negative control was re-run against the source after these changes: removing `role="alert"`, `aria-invalid`/`aria-describedby` and the `ariaLabel` pass-through fails 3 of the 4 new tests on the right assertions; the fourth keeps passing because it asserts absence. Restored to a byte-identical tree afterwards.

tests/input-error-announcement.test.ts
```diff
@@ -0,0 +1,74 @@
+import { expect, test } from '@playwright/test';
+
+test.describe('Input — validation messages reach the accessibility tree', () => {
+  // Regression guard: the error text was rendered as a plain <div class="error-message"> with
+  // no id, no role and nothing tying it to the field. It was drawn on screen and completely
+  // absent from the accessibility tree — a screen-reader user submitted, heard nothing, and was
+  // left on a form that had not moved. Nothing visual catches this: the message looks correct.
+  test('an invalid field is marked invalid and points at its message', async ({ page }) => {
+    await page.goto('/components/input');
+
```
Tara-ag, line 10: 💡 MINOR: These tests navigate to demo pages (`/components/input`, `/components/chip-input`) and select by `data-pw` attributes defined in documentation fixtures. Demo selectors are not a stable public interface; changes to the docs page will break the accessibility regression suite.

Prefer mounting the components in isolation (Vitest + @testing-library/svelte) or a dedicated test fixture route, so the assertions are tied to component behavior rather than demo layout.

src/lib/ChipInput/ChipInput.svelte
```diff
@@ -5,6 +5,7 @@
```
Tara-ag, line 5: ⚠️ MAJOR: New `ariaLabel` prop is not exposed in the web-component wrapper. There is no `src/wc/components/ChipInput.wc.svelte` and `<sui-chip-input>` is not registered in `src/wc/index.ts`, so web-component consumers cannot name the draft field. Per the repo's lock-step rule, every camelCase Svelte prop must have an explicit kebab-case attribute mapping in the wrapper.

Create `src/wc/components/ChipInput.wc.svelte` mapping `ariaLabel` → `aria-label` (plus the existing props/events), and import it in `src/wc/index.ts`.

src/lib/Input/Input.svelte
```diff
@@ -399,8 +412,10 @@
     {@render fieldElement()}
   {/if}
 
-  {#if onErrorMessage !== '' && showError && !actionInput}
+  {#if isShowingError}
     <div
+      id={errorMessageId}
+      role="alert"
```
Tara-ag, line 429: 💡 MINOR: `role="alert"` is correct, but pairing it with `aria-describedby` can cause some screen readers to announce the error twice. Consider adding `aria-live="assertive"` explicitly and ensuring the live-region element is present before the message text is injected, to maximize reliable announcement.

src/lib/Input/Input.svelte
```diff
@@ -122,6 +122,15 @@
   // forceError lets consumers drive the error border from server/runtime validation,
   // independent of validationPattern.
   const showError = $derived(showErrorMessage || forceError);
+  // The error text has to be reachable FROM the field it describes and announced when it
+  // appears. Without both, a screen-reader user submits, hears nothing, and is left on a form
+  // that did not move -- the message is on screen and absent from the accessibility tree.
+  const errorMessageId = $derived(`${effectiveId}-error`);
+  // `onErrorMessage` is `string | null`, so null has to be excluded explicitly: `null !== ''`
+  // is true, which would describe the field by an alert element carrying no message.
```
Tara-ag, line 130: ⚠️ MAJOR: `aria-describedby` only references the error message. When a consumer also passes `infoMessage`, that helper text is visible on screen but not associated with the field for assistive technology users.

Add an `infoMessageId` and include it in `aria-describedby` whenever `infoMessage` is shown, e.g.:
```ts
const describedByIds = [
  isShowingError ? errorMessageId : null,
  infoMessage !== '' && !actionInput ? infoMessageId : null
].filter(Boolean).join(' ');
```
then bind `aria-describedby={describedByIds || null}` on both `<input>` and `<textarea>`.

Tara-ag (changes requested):
Two real accessibility improvements, but two MAJOR issues need resolution before merge:

1. **Web-component parity**: `ChipInput` gained a new public `ariaLabel` prop, but there is no web-component wrapper or registry entry for `ChipInput` at all, so the new accessibility affordance is unreachable from the WC surface. The repo expects Svelte props and WC attributes to stay in lock-step.

2. **Info message association**: `aria-describedby` now points only at the error message. When `infoMessage` is also supplied, that helper text is not associated with the field, so screen-reader users miss visible guidance.

Minor notes on live-region assertiveness and test coupling to demo pages are also left inline. Once (1) and (2) are addressed, this should be good to approve.

src/lib/ChipInput/ChipInput.svelte
Tara-ag, line 8: 💬 SUGGESTION: ChipInput is not registered as a custom element (no `src/wc/components/ChipInput.wc.svelte` and no import in `src/wc/index.ts`), so the new `ariaLabel` prop is only available to Svelte consumers. This is pre-existing and not a blocker for this PR, but if ChipInput is later exposed as a web component, it will need an explicit `aria-label` attribute mapping for parity.

tests/input-error-announcement.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    // Negative side of the contract: the attributes must be ABSENT when there is no error,
+    // otherwise every field on every form announces itself as invalid.
+    // /.*/ matches ANY value including the empty string, so these reject a present-but-empty
+    // attribute too. Asserting `not.toHaveAttribute('aria-invalid', 'true')` would instead pass
+    // on aria-invalid="false", which is a value the field must not carry at all.
+    const healthy = page.getByTestId('input-datatype-time');
+    await expect(healthy).not.toHaveAttribute('aria-invalid', /.*/);
+    await expect(healthy).not.toHaveAttribute('aria-describedby', /.*/);
+  });
+});
```
Tara-ag, line 79: 💡 MINOR: The `/.*/` regex works because it matches the empty string and therefore rejects a present-but-empty attribute, but the intent is clearer with `expect(await healthy.getAttribute('aria-invalid')).toBeNull()` or Playwright's absent matcher. Consider simplifying for readability.

Tara-ag (approved):
Approved with minor notes. The accessibility fixes are correct and well-tested: `aria-invalid`/`aria-describedby` are tied to a `role="alert"` message only when an error is present, and `ChipInput` now forwards an optional `ariaLabel` to its draft `Input`. Docs, demos, and Playwright regression tests are included. No public API breakage or blocking issues found.

Tara-ag (commented):
Yama review in progress.

src/lib/Input/Input.svelte
Tara-ag, line 130: ⚠️ MAJOR: `aria-describedby` only references the error message. When a consumer also passes `infoMessage`, that helper text is visible on screen but not associated with the field for assistive technology users.

Add an `infoMessageId` and include it in `aria-describedby` whenever `infoMessage` is shown, e.g.:
```ts
const describedByIds = [
  isShowingError ? errorMessageId : null,
  infoMessage !== '' && !actionInput ? infoMessageId : null
].filter(Boolean).join(' ');
```
then bind `aria-describedby={describedByIds || null}` on both `<input>` and `<textarea>`.

tests/input-error-announcement.test.ts
Tara-ag, line 79: 💡 MINOR: The `/.*/` regex works because it matches the empty string and therefore rejects a present-but-empty attribute, but the intent is clearer with `expect(await healthy.getAttribute('aria-invalid')).toBeNull()` or Playwright's absent matcher. Consider simplifying for readability.

src/lib/ChipInput/ChipInput.svelte
Tara-ag, line 8: 💬 SUGGESTION: ChipInput is not registered as a custom element (no `src/wc/components/ChipInput.wc.svelte` and no import in `src/wc/index.ts`), so the new `ariaLabel` prop is only available to Svelte consumers. This is pre-existing and not a blocker for this PR, but if ChipInput is later exposed as a web component, it will need an explicit `aria-label` attribute mapping for parity.

src/lib/ChipInput/ChipInput.svelte
Tara-ag, line 5: ⚠️ MAJOR: New `ariaLabel` prop is not exposed in the web-component wrapper. There is no `src/wc/components/ChipInput.wc.svelte` and `<sui-chip-input>` is not registered in `src/wc/index.ts`, so web-component consumers cannot name the draft field. Per the repo's lock-step rule, every camelCase Svelte prop must have an explicit kebab-case attribute mapping in the wrapper.

Create `src/wc/components/ChipInput.wc.svelte` mapping `ariaLabel` → `aria-label` (plus the existing props/events), and import it in `src/wc/index.ts`.

src/lib/Input/Input.svelte
Tara-ag, line 429: 💡 MINOR: `role="alert"` is correct, but pairing it with `aria-describedby` can cause some screen readers to announce the error twice. Consider adding `aria-live="assertive"` explicitly and ensuring the live-region element is present before the message text is injected, to maximize reliable announcement.

tests/input-error-announcement.test.ts
Tara-ag, line 10: 💡 MINOR: These tests navigate to demo pages (`/components/input`, `/components/chip-input`) and select by `data-pw` attributes defined in documentation fixtures. Demo selectors are not a stable public interface; changes to the docs page will break the accessibility regression suite.

Prefer mounting the components in isolation (Vitest + @testing-library/svelte) or a dedicated test fixture route, so the assertions are tied to component behavior rather than demo layout.

Tara-ag (changes requested):
Yama review complete. Two MAJOR accessibility/parity findings and three MINOR/SUGGESTION notes are inline. Please address the MAJOR items before merging.

Tara-ag:

Severity counts — 🔒 CRITICAL: 0 · ⚠️ MAJOR: 4 · 💡 MINOR: 5

## 🤖 Yama Review Summary

⚠️ The review loop ended early (`time-limit`, 28 steps) — this verdict is built strictly from gate-verified findings.

**Verified findings (10):**

- **MAJOR**: ChipInput has no web-component wrapper despite being exported — `src/wc/index.ts:1`
- **MINOR**: Input web-component wrapper omits new textarea-related props — `src/wc/components/Input.wc.svelte:1`
- **SUGGESTION**: Tests rely on demo-page testIds — `tests/input-error-announcement.test.ts:9`
- **MINOR**: aria-invalid is not suppressed for actionInput — `src/lib/Input/Input.svelte:324`
- **MAJOR**: Error/info message IDs are derived from name, risking duplicate IDs — `src/lib/Input/Input.svelte:60`
- **MAJOR**: aria-invalid is not suppressed for actionInput — `src/lib/Input/Input.svelte:324`
- **MAJOR**: New ariaLabel prop is not exposed on the web-component wrapper — `src/lib/ChipInput/ChipInput.svelte:8`
- **MINOR**: Accessibility tests depend on demo-page selectors — `tests/input-error-announcement.test.ts:10`
- **MINOR**: role="alert" paired with aria-describedby may double-announce — `src/lib/Input/Input.svelte:429`
- **MINOR**: Negative attribute assertion uses unclear regex — `tests/input-error-announcement.test.ts:79`

- ⚠️ **MAJOR**: ChipInput has no web-component wrapper despite being exported — `src/wc/index.ts:1`
  src/lib/index.ts exports ChipInput, but there is no src/wc/components/ChipInput.wc.svelte and it is not registered in src/wc/index.ts. Adding ariaLabel to the Svelte component without a corresponding custom-element prop map means web-component consumers cannot set the accessible name. Per project standards, web-component wrappers must expose every new Svelte prop with an explicit kebab-case attrib…
- ⚠️ **MAJOR**: Error/info message IDs are derived from name, risking duplicate IDs — `src/lib/Input/Input.svelte:60`
  Input.svelte generates `errorMessageId` and `infoMessageId` as `${name ?? ''}-error-message` and `${name ?? ''}-info-message`. The `name` attribute is intentionally repeatable across forms and radio groups, so two Input instances with the same name (or no name) will produce duplicate `id` values. Duplicate IDs break `aria-describedby` associations: a screen reader may read the wrong field's error …
- ⚠️ **MAJOR**: aria-invalid is not suppressed for actionInput — `src/lib/Input/Input.svelte:324`
  In Input.svelte, the visual error border is suppressed when actionInput is true (`class:input-error={showError && !actionInput}`), and the error/info messages are also suppressed (`isShowingError`/`isShowingInfo` both check `!actionInput`). However, `aria-invalid` is set from `showError` alone without checking `actionInput`. This means an action-style input can announce itself as invalid to assist…
- ⚠️ **MAJOR**: New ariaLabel prop is not exposed on the web-component wrapper — `src/lib/ChipInput/ChipInput.svelte:8`
  ChipInput.svelte now accepts `ariaLabel` and forwards it to the inner Input, but there is no src/wc/components/ChipInput.wc.svelte wrapper and src/wc/index.ts does not register a sui-chip-input custom element. The repo requires lock-step parity between Svelte props and web-component custom-element declarations; every camelCase Svelte prop must be mirrored with an explicit kebab-case attribute mapp…
- 💡 **MINOR**: Input web-component wrapper omits new textarea-related props — `src/wc/components/Input.wc.svelte:1`
  Input.wc.svelte does not declare props for rows, autoResize, minRows, maxRows, resize, showCount, readonly, spellcheck, leftIcon, rightIcon, onLeftIconClick, onRightIconClick, leftIconLabel, rightIconLabel, ariaExpanded, ariaAutocomplete, ariaControls, or ariaActivedescendant. While this PR only touches aria-invalid/aria-describedby, the existing wrapper is already incomplete; the new aria-describ…
- 💡 **MINOR**: aria-invalid is not suppressed for actionInput — `src/lib/Input/Input.svelte:324`
  In Input.svelte, the visual error border is suppressed when `actionInput` is true (`class:input-error={showError && !actionInput}`), and the error/info messages are also suppressed (`isShowingError`/`isShowingInfo` both check `!actionInput`). However, `aria-invalid` is set from `showError` alone without checking `actionInput`. This means an action-style input can announce itself as invalid to assi…
- 💡 **MINOR**: Accessibility tests depend on demo-page selectors — `tests/input-error-announcement.test.ts:10`
  The new Playwright tests navigate to `/components/input` and `/components/chip-input` and select elements by `data-pw` attributes defined in documentation fixtures. Demo selectors are not a stable public interface; changes to the docs page will break the accessibility regression suite. Component regression tests should render components in isolated harnesses rather than coupling to demo-page testI…
- 💡 **MINOR**: role="alert" paired with aria-describedby may double-announce — `src/lib/Input/Input.svelte:429`
  The error message element has `role="alert"` and is also referenced by the field's `aria-describedby`. Some screen readers will announce the live-region alert when it appears and then again when the field is focused because it is described-by the same element. Adding an explicit `aria-live="assertive"` and ensuring the live-region element is present before message injection maximizes reliable anno…
- 💡 **MINOR**: Negative attribute assertion uses unclear regex — `tests/input-error-announcement.test.ts:79`
  The test uses `not.toHaveAttribute('aria-invalid', /.*/)` to assert the attribute is absent. While `/.*/` matches the empty string and therefore rejects a present-but-empty attribute, the intent is clearer with `expect(await healthy.getAttribute('aria-invalid')).toBeNull()` or Playwright's absent matcher. The current form requires readers to reason about regex matching empty strings.
- 💬 **SUGGESTION**: Tests rely on demo-page testIds — `tests/input-error-announcement.test.ts:9`
  tests/input-error-announcement.test.ts navigates to /components/input and /components/chip-input and selects by testId. Per bootstrapped standards, component regression tests should render components in isolated harnesses rather than coupling to demo-page testIds and labels that can change independently. This is a suggestion, not a blocker, because the PR follows the existing pattern in this repo.

murdore:
Addressed in `ce94b63`. One MAJOR fixed, one MAJOR refuted — and the review contradicts itself on the second one.

### ✅ MAJOR — `aria-describedby` omitted `infoMessage`

Correct, and it was the same defect I opened this PR to fix, left in place one element lower. `infoMessage` renders visibly under the field and was associated with nothing, so helper text like "Enter a percentage between 1 and 100" — exactly the guidance a screen-reader user needs *before* they trip an error — was on screen and absent from the accessibility tree.

`aria-describedby` is a space-separated id list, so both are now referenced, error first:

```ts
const describedBy = $derived(
  [isShowingError ? errorMessageId : null, isShowingInfo ? infoMessageId : null]
    .filter(Boolean)
    .join(' ')
);
```

Reading order is deliberate and asserted: the failure comes before the advice. Three new tests cover helper-text-only (valid field, described but not `aria-invalid`), the combined case, and the ordering. 9 tests pass.

Worth flagging: adding this **broke my own two earlier tests**, which read `aria-describedby` as if it were a single id — `[id="x-error x-info"]` matches nothing. That is the bug your finding predicted, reproduced in the test suite, and it is fixed rather than worked around.

### ❌ MAJOR — ChipInput web-component wrapper

Refuting again, because this review argues both sides. The same run posted, on the same file:

> 💬 SUGGESTION: ChipInput is not registered as a custom element (no `src/wc/components/ChipInput.wc.svelte` and no import in `src/wc/index.ts`), so the new `ariaLabel` prop is only available to Svelte consumers. **This is pre-existing and not a blocker for this PR**

That is the accurate reading. The cited rule — every camelCase prop mirrored with a kebab-case attribute — governs components that *are* registered. ChipInput is not:

```
$ grep -n "ChipInput" src/wc/index.ts
(no matches)
```

So there is no `<sui-chip-input>` whose attribute surface is now inconsistent. Following the MAJOR instruction would **register a new custom element for the first time** — net-new public API, a new tag name, a new shadow root, and a new supported integration path — inside a PR about naming a form control. That is a product decision, not parity, and it should be its own PR with the attribute-casing tests the other wrappers carry.

The contrast with #475 is the test: `Input`, `Menu` and `Checkbox` all have registered elements, so their new props genuinely did create a parity gap, and I added those mappings there.

### Declined, with reasons

**`aria-live="assertive"` alongside `role="alert"`** — `role="alert"` already implies `aria-live="assertive"` plus `aria-atomic="true"` in the ARIA spec, so this is redundant. The double-announcement concern is real in some screen readers, but it is inherent to describing a field by its alert, and the alternative (dropping one of the two) costs either the announcement or the association. Standard practice keeps both.

**`getAttribute(...) === null` instead of `/.*/`** — keeping the regex form, and not only to avoid ping-ponging with the other reviewer who asked for exactly this. `expect(locator).not.toHaveAttribute()` auto-retries against a settling page; `await getAttribute()` is a one-shot read that can sample too early. On a demo page that mounts asynchronously the retrying form is the more robust assertion, and the intent is now stated in a comment above it.

**Demo-page selectors** — this is the repo's existing convention: 16 tests across the suite already drive `/components/input`, and there is no separate fixture-route pattern to follow. Introducing one is a suite-wide change, not something to fork inside this PR. Vitest is not an option here — `test:unit` exists but these assertions are about rendered ARIA relationships in a real browser, which is what Playwright is for.

### Verification

9 tests pass (6 new + 3 existing). `lint` 0, `build` 0.

src/lib/Input/Input.svelte
```diff
@@ -301,6 +321,8 @@
         aria-controls={ariaControls}
         aria-activedescendant={ariaActivedescendant}
         aria-required={mandatory || null}
+        aria-invalid={showError ? 'true' : null}
```
Tara-ag, line 324: 💡 MINOR: `aria-invalid` is not suppressed for `actionInput`. The visual error border and error/info messages are all gated by `!actionInput`, but `aria-invalid` is driven by `showError` alone. An action-style input could therefore announce itself as invalid while showing no error message and no error styling. Suggest `aria-invalid={showError && !actionInput ? 'true' : null}` on both the `<input>` and `<textarea>` branches.

Tara-ag (commented):
Approved with one minor accessibility nit. The validation-announcement and ChipInput-naming changes are correct and well-tested. Please consider aligning `aria-invalid` with the existing `actionInput` suppression in a follow-up.

murdore:
I checked the web-component convention across the whole library rather than arguing from ChipInput alone. The census settles it — and it turned up a real gap, just not the one flagged.

## What the library actually does

| | count |
|---|---|
| directories under `src/lib/` | 97 |
| `.wc.svelte` wrappers | 81 |
| registered in `src/wc/index.ts` | 78 |

So a wrapper **is** the norm. The 16 without one are not a random tail:

- **4 aren't components** — `_chart`, `assets`, `soundKit`, `Animations`
- **7 are charts** — Area, Bar, DualAxisBar, Funnel, Line, Pie, Sankey (excluded as a class)
- **5 are interactive components** — `ChipInput`, `ColorPicker`, `Combobox`, `SpeechToText`, `SplitInput`

## The exclusion is documented policy, not an oversight

`docs/FileDropzoneTrigger.md:95`, already on `release`, names ChipInput directly:

> None. Like `SplitInput` and `ChipInput`, this is a simple presentational component consumed directly from Svelte — it is designed to compose inside `FileInput`'s `trigger` Snippet prop, which is itself not serialisable across the Web Component boundary. Use the Svelte component directly; there is no `sui-file-dropzone-trigger` custom element.

`docs/SoundKit.md:117` follows the same pattern with its own "Web component — None" section, and `docs/DateRangePicker.md:424` shows the positive case for a component that *is* exposed.

So the repo's process is: components that are exposed get a wrapper **and** a "Web Component" doc section; components deliberately not exposed get a "Web component — None" section saying so. ChipInput is in the second group by existing, documented policy.

## The real gap — fixed in `081e305`

ChipInput's own doc said **nothing**. `grep -i "web component" docs/ChipInput.md` returned no matches. The policy existed only inside *another component's* file, which is precisely why this keeps getting flagged as a MAJOR parity violation on every review pass — there was nothing at the component's own documentation to read.

Added the missing section, matching SoundKit's and FileDropzoneTrigger's form, and stating the consequence explicitly so the next reviewer doesn't have to rediscover it:

> None. Like `SplitInput` and `FileDropzoneTrigger`, ChipInput is a simple presentational component consumed directly from Svelte; there is no `sui-chip-input` custom element and it is not registered in `src/wc/index.ts`.
>
> This is deliberate, and it is why `ariaLabel` has no kebab-case attribute mapping: there is no custom element for an attribute to map onto. Adding one would newly expose ChipInput as a web component rather than restore parity with an existing element.

That is following the process the other 96 components follow — not skipping it.

`lint` 0 · `build` 0 · 9 tests pass.

If the team wants ChipInput, ColorPicker, Combobox, SpeechToText and SplitInput exposed as custom elements, that is a worthwhile piece of work and I'm happy to take it — as its own PR covering all five consistently, with the attribute-casing tests, rather than registering one of them as a side effect of an accessibility fix.

murdore:
## Video evidence

Fair criticism — assertion counts aren't something you can check. Recorded a before/after walkthrough instead.

The videos annotate the live page with what **Chromium's own accessibility tree** computes (via `ariaSnapshot()`), not with the attributes read back. Reading `aria-label` back only proves the attribute exists; it doesn't prove a name resolves. Every `aria-describedby` reference is followed to the element it points at, and that element's role and text are shown.

Both arms are recorded from an **identical tree** — same branch, same demo fixtures — with only `src/lib/Input/Input.svelte` and `src/lib/ChipInput/ChipInput.svelte` reverted to `origin/release` for the BEFORE pass, then restored and hash-verified byte-identical. So the only variable is the fix.

### What the frames show

**Invalid field** — the two pages are pixel-identical, which is the point: the defect is invisible.

| | BEFORE (`origin/release`) | AFTER (this PR) |
|---|---|---|
| `aria-invalid` | *absent* | `true` |
| `aria-describedby` | *absent* | `s1-error s1-info` |
| references resolved | **nothing referenced — the message is on screen only** | `s1-error` → role `alert`, "Enter a valid email address"<br>`s1-info` → "Use your company address, not a personal one." |

**ChipInput** — this one is more interesting than "unnamed", and it corrected a mistake in my own harness. My first cut coloured the verdict green whenever a name existed, and BEFORE **passed** — because with no `aria-label` Chromium falls back to the **placeholder**. The panel now compares against the visible caption:

| | BEFORE | AFTER |
|---|---|---|
| accessible name | `Add tag…` — the placeholder | `Product tags` |
| expected name | `Product tags` | `Product tags` |
| verdict | Does NOT match the visible caption | Matches |

That fallback is exactly the WCAG 2.5.3 failure: a speech-input user says "Product tags" — what they can see — and nothing matches, because the control is named by placeholder text that also vanishes the moment they type.

### One thing the video surfaced that is NOT mine

Helper text renders in the **same red as an error** — `--input-info-msg-text-color` defaults to `#fa1405`, identical to the error colour. That is on `origin/release` too, unchanged by this PR, and it's visible in both arms. It's worth its own issue: helper text that looks like a validation failure is misleading, and it's colour-alone signalling. I've deliberately not changed it here, since the default is a visual breaking change for every consumer.

### Files

Three videos, 31s each, 1280×800 (side-by-side 1920 wide):

- `01-before-origin-release.mp4`
- `02-after-this-pr.mp4`
- `03-side-by-side.mp4` — both arms playing in sync, which is the one to watch

GitHub has no public API for attaching media to a comment, so I can't upload them from here — they're being attached to this thread manually. The recording harness was intentionally not committed (evidence and session scratch don't belong in the branch); it's a `playwright.evidence.config.ts` with `video: { mode: 'on' }` plus a spec that injects the HUD, and I can commit it as reusable tooling if the team wants it.

Tara-ag (approved):
Accessibility additions are correct and well-tested. No public API breakage, no CSS token changes, and web-component parity is preserved for Input. Approving.

murdore:
## Video evidence

Both arms are recorded from an **identical tree** — same branch, same demo fixtures — with only `src/lib/Input/Input.svelte` and `src/lib/ChipInput/ChipInput.svelte` reverted to `origin/release` for the BEFORE pass, then restored and hash-verified byte-identical. The fix is the only variable.

Accessible names come from Chromium's own accessibility tree via `ariaSnapshot()`, not from reading the attribute back — reading `aria-label` back only proves the attribute exists, not that a name resolves. Every `aria-describedby` reference is followed to the element it points at, and that element's role and text are shown.

### Side by side

https://github.com/user-attachments/assets/f67ee8a3-aff8-4ebf-92fe-aedd89886f93

| | BEFORE (`origin/release`) | AFTER (this PR) |
|---|---|---|
| `aria-invalid` | *absent* | `true` |
| `aria-describedby` | *absent* | `s1-error s1-info` |
| references resolve to | **nothing — the message is on screen only** | `s1-error` → role `alert`, "Enter a valid email address"<br>`s1-info` → "Use your company address, not a personal one." |
| ChipInput accessible name | `Add tag…` — the **placeholder** | `Product tags` |

The two pages are pixel-identical, which is the point: nothing visual catches this.

### The ChipInput case corrected a bug in my own harness

My first cut coloured the verdict green whenever a name existed — and BEFORE **passed**, because with no `aria-label` Chromium falls back to the placeholder. The panel now compares against the visible caption instead. That fallback is precisely the WCAG 2.5.3 failure: a speech-input user says "Product tags", the words they can see, and nothing matches, because the control is named by placeholder text that also vanishes the moment they type.

### Individual recordings

**BEFORE — `origin/release`**

https://github.com/user-attachments/assets/2ecee0d6-5ef2-46fc-9a92-b4b310920769

**AFTER — this PR**

https://github.com/user-attachments/assets/e12dc711-d0e6-472d-9552-685375af7da3

### One thing the recording surfaced that is NOT from this PR

Helper text renders in the **same red as an error** — `--input-info-msg-text-color` defaults to `#fa1405`, identical to the error colour. That is on `origin/release` too and is visible in both arms. Worth its own issue: helper text that looks like a validation failure is misleading, and it is colour-alone signalling. Deliberately not changed here, since altering the default is a visual breaking change for every consumer.

The recording rig is intentionally not committed — evidence and session scratch do not belong in the branch.

murdore:
## Still frames from the evidence above

Correcting my own framing: this PR **already carried** video evidence (the comment above, three attachments). I claimed otherwise a moment ago and that was wrong — I had checked the comment bodies with a 600-character truncation and never tested for attachment URLs, so I read "no images" off a window that couldn't have shown them. Apologies for the noise.

What this adds is genuinely narrow: **mp4 attachments don't preview inline**, so the videos above are a click away rather than visible. These are annotated stills of the two decisive scenes, plus a side-by-side, so the result can be read without opening anything.

### Scene 1 — an invalid field must be marked, and its error announced

| | BEFORE (`origin/release`) | AFTER (this PR) |
|---|---|---|
| `aria-invalid` | *(absent)* | `true` |
| `aria-describedby` | *(absent)* | `s1-error s1-info` |
| resolution | **nothing referenced — the message is on screen only** | both ids resolve; `s1-error` has `role="alert"` and the text *"Enter a valid email address"*, `s1-info` carries the helper text |

The field rendered its error in red, visibly, and referenced it from nothing.

### Scene 3 — ChipInput can be given an accessible name

| | BEFORE | AFTER |
|---|---|---|
| accessible name | `Add tag…` | `Product tags` |
| matches visible caption | **no** | yes |

Worth drawing out, because of how it fails: before the change `ChipInput` could not be named, so Chromium fell back to the **placeholder**. That still yields a non-empty accessible name — so any check asking only *"does this control have a name?"* passes it. It has to be compared against the visible caption. A placeholder also disappears as soon as the user types, and here it disagreed with the visible "Product tags", so a speech-input user could not activate the control by saying what they see (WCAG 2.5.3, Label in Name).

Every value on the panel is read from Chromium's live accessibility tree, which is the part you cannot verify by reading markup.

src/lib/Input/Input.svelte
```diff
@@ -122,6 +122,11 @@
   // forceError lets consumers drive the error border from server/runtime validation,
   // independent of validationPattern.
   const showError = $derived(showErrorMessage || forceError);
+  // The error text has to be reachable FROM the field it describes and announced when it
+  // appears. Without both, a screen-reader user submits, hears nothing, and is left on a form
+  // that did not move -- the message is on screen and absent from the accessibility tree.
+  const errorMessageId = $derived(`${effectiveId}-error`);
+  const isShowingError = $derived(onErrorMessage !== '' && showError && !actionInput);
```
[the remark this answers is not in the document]
murdore: This is fixed exactly as suggested. isShowingError now excludes null explicitly, at Input.svelte line 132, with a comment above it explaining that null !== '' would otherwise describe the field by an alert element carrying no message. This shipped in PR 480 (commit c49291989) and is live on release 3.1.0.

tests/input-error-announcement.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  });
+
+  test('a valid field is not marked invalid and describes nothing', async ({ page }) => {
+    await page.goto('/components/input');
+
+    // Negative side of the contract: the attributes must be ABSENT when there is no error,
+    // otherwise every field on every form announces itself as invalid.
+    const healthy = page.getByTestId('input-datatype-time');
+    await expect(healthy).not.toHaveAttribute('aria-invalid', 'true');
+    await expect(healthy).not.toHaveAttribute('aria-describedby', /.+/);
```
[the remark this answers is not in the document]
murdore: The negative attribute assertions now use the /.*/ pattern the review asked for, at tests/input-error-announcement.test.ts lines 76-77, so a healthy field with aria-invalid="false" or a present-but-empty aria-describedby would still fail the test rather than passing it. This shipped in PR 480 (commit c49291989) on release 3.1.0.

src/lib/ChipInput/ChipInput.svelte
murdore: ChipInput now has a web-component wrapper. PR 482 (commit ff24661d6) added src/wc/components/ChipInput.wc.svelte, which registers the sui-chip-input tag and maps ariaLabel to the aria-label attribute alongside the rest of the props, following the same pattern as Input.wc.svelte. It is imported and registered in src/wc/index.ts. This is live on release 3.1.0, so web-component consumers can now name the draft field.

murdore: ChipInput now has a web-component wrapper, added in PR 482 (commit ff24661d6) at src/wc/components/ChipInput.wc.svelte, mapping ariaLabel to the aria-label attribute and registering the sui-chip-input custom element in src/wc/index.ts, exactly the shape this review's suggested snippet asked for. This is live on release 3.1.0.

tests/input-error-announcement.test.ts
murdore: This is a deliberate, repo-wide convention rather than a gap in this one test file. playwright.config.ts documents that the component library's demo pages are the intended fixture surface, with getByTestId wired to their data-pw hooks, and around 94 other test files in tests/ follow the identical pattern of navigating to a /components/* route and asserting against it. Vitest is present in the repo for unrelated unit tests, but the accessibility regression suite consistently uses the demo pages as its harness. Switching just this file to an isolated mount would be inconsistent with how the rest of the suite verifies component behaviour, so this suggestion is declined as a style preference the codebase does not otherwise follow.

src/lib/ChipInput/ChipInput.svelte
murdore: ChipInput.wc.svelte now exists and maps ariaLabel to the aria-label attribute, registered as sui-chip-input in src/wc/index.ts. This landed in PR 482 (commit ff24661d6) and is on release 3.1.0, so web-component consumers can now name the draft field.

src/lib/Input/Input.svelte
murdore: No explicit aria-live="assertive" was added, and that is the correct call rather than an oversight. The alert role already implies an assertive live region per the ARIA specification, so adding aria-live="assertive" alongside role="alert" would be redundant rather than additive. Pairing role="alert" with aria-describedby on the field is also the standard, widely used technique for inline validation messages, not a double-announcement bug in itself; actual double-announcement risk depends on the specific screen reader and is not something this component can fully control either way. This suggestion is declined as unnecessary given role="alert"'s existing semantics.

src/lib/Input/Input.svelte
murdore: This is fixed, matching the suggested shape closely. Input.svelte now derives infoMessageId and isShowingInfo, and builds describedBy from both the error and info ids in reading order, referenced by aria-describedby on both the input and textarea elements, at lines 128-138. So a consumer's infoMessage is now associated with the field for assistive technology whenever it is shown, not just the error text. This shipped in PR 479 (commit 17862661e) and is live on release 3.1.0.

src/lib/ChipInput/ChipInput.svelte
murdore: This suggestion flagged a pre-existing gap rather than a blocker, and it has since been closed anyway. ChipInput.wc.svelte was added in PR 482 (commit ff24661d6), and it declares ariaLabel with an explicit aria-label attribute mapping, so ChipInput's web-component consumers get the same naming support as the Svelte component. This is live on release 3.1.0.

tests/input-error-announcement.test.ts
murdore: The suggested toBeNull() simplification was not adopted; instead the existing /.*/ pattern was kept and given an explanatory comment, at tests/input-error-announcement.test.ts lines 72-74, walking through why a naive not.toHaveAttribute('aria-invalid', 'true') would incorrectly pass on aria-invalid="false". Both forms are valid here, getAttribute returning null is an equally correct check, but the team's choice to document the reasoning in place addresses the readability concern without changing the assertion. This is declined as a style preference rather than a defect.

src/lib/Input/Input.svelte
murdore: Fixed, same as the other copy of this thread. Input.svelte now builds aria-describedby from both the error id and an infoMessageId whenever infoMessage is shown, applied to both the input and textarea branches, at lines 128-138. This shipped in PR 479 (commit 17862661e) and is live on release 3.1.0.

tests/input-error-announcement.test.ts
murdore: Duplicate of the earlier readability comment on this same line: the /.*/ matcher was kept rather than switched to toBeNull(), but the reasoning is now spelled out in a comment at tests/input-error-announcement.test.ts lines 72-74. Declined as a style preference, not a defect.

src/lib/ChipInput/ChipInput.svelte
murdore: Duplicate of the earlier non-blocking ariaLabel/web-component note. Since it was filed, ChipInput.wc.svelte was added in PR 482 (commit ff24661d6) with an explicit ariaLabel to aria-label mapping, so this is resolved on release 3.1.0.

src/lib/ChipInput/ChipInput.svelte
murdore: Duplicate of the other ariaLabel/web-component thread. ChipInput.wc.svelte, added in PR 482 (commit ff24661d6), maps ariaLabel to the aria-label attribute exactly as this review's suggested snippet describes. Live on release 3.1.0.

src/lib/Input/Input.svelte
murdore: Duplicate of the earlier role="alert"/aria-live note. No explicit aria-live="assertive" was added, and none is needed: role="alert" already carries assertive live-region semantics per the ARIA specification, so adding it explicitly would be redundant rather than a fix. Declined for the same reason as the first copy of this thread.

tests/input-error-announcement.test.ts
murdore: Duplicate of the earlier demo-selector note. Navigating to the component library's own demo pages and asserting via data-pw is the deliberate, documented convention this whole test suite uses, not an isolated shortcut in this file. Declined for the same reason as the first copy of this thread.

src/lib/Input/Input.svelte
murdore: This is still open on both branches. aria-invalid is driven by showError alone at Input.svelte lines 324 and 362, for the input and textarea elements respectively, while the visual error styling, error message and info message are all gated by !actionInput elsewhere in the same file. An action-style input can therefore announce itself as invalid to assistive technology while showing no visible error indication at all. Closing this needs aria-invalid={showError && !actionInput ? 'true' : null} applied to both branches, matching the review's suggested fix.

murdore: Fixed in juspay/svelte-ui-components#502 (commit 64f91c5be): aria-invalid is now `showError && !actionInput ? 'true' : null` on both the input and textarea branches (src/lib/Input/Input.svelte), matching the gate the error border, message and label already used; pinned by a new demo field with actionInput and forceError and the test 'an actionInput field in error state is not marked invalid' in tests/input-error-announcement.test.ts.

PR #482, opened 2026-09-01
feat(wc): expose ChipInput, ColorPicker, Combobox and SplitInput as custom elements

murdore:
Closes the web-component gap that came up twice in review on #480.

## What the census actually showed

`src/lib/` holds **97 directories**; `src/wc/components/` held **81 wrappers**. The 16 without one were not a random tail:

- **4 are not components** — `_chart`, `assets`, `soundKit`, `Animations`
- **7 are charts** — Area, Bar, DualAxisBar, Funnel, Line, Pie, Sankey (excluded as a class)
- **5 looked like interactive components** — ChipInput, ColorPicker, Combobox, SpeechToText, SplitInput

**One of those five is not a component.** `src/lib/SpeechToText/` contains only `controller.svelte.ts` and `types.ts` — no `.svelte` file at all. It is a TypeScript module like `soundKit`, so there is nothing to wrap. That leaves four, and this PR does all four.

## What each element maps

| Element | kebab-case attributes |
|---|---|
| `sui-chip-input` | `aria-label`, `test-id` |
| `sui-color-picker` | `show-value`, `test-id` |
| `sui-combobox` | `aria-label`, `test-id`, `input-value`, `highlighted-index`, `no-results-text`, `max-selected`, `max-selected-text` |
| `sui-split-input` | `auto-advance`, `test-id` |

Array, snippet and handler props (`items`, `values`, `fields`, `filterFn`, `on*`) stay properties rather than attributes — they are not serialisable — and each doc says so with an example.

## This supersedes a section I wrote in #480

`docs/ChipInput.md` gained a "Web component — None" section in #480, stating ChipInput was deliberately Svelte-only. That was accurate when written and is now wrong, so it is replaced, with a note saying so rather than silently rewritten.

The reasoning in #480 also stands rather than being reversed: `ariaLabel` was left without a kebab mapping there **because no custom element existed to map onto**. This PR creates one, so the mapping now exists. Registering a new element was the right thing to do as its own change — which is what this is.

## One finding worth surfacing

Yama's review cited a repo rule that "every camelCase Svelte prop must be mirrored in the web-component wrapper with an explicit kebab-case attribute". Auditing all 85 wrappers against that rule, **64 have camelCase props with no `attribute:` mapping**.

A large share of those are snippet and callback props where an attribute makes no sense, so the true violation count is lower — but the rule is clearly not enforced today, and there is no gate for it (`lint:event-casing` checks event *prop names* in `properties.ts`, not wrapper attribute mappings). Worth a follow-up to define what the rule covers and gate it; I have not tried to burn that down here.

**The four wrappers in this PR have zero gaps** under that audit.

## Verification

`lint` 0 · `build` 0. All four tags confirmed present in the built `dist-wc/index.js`.

`check:wc` reports 3 errors, all in `src/wc/components/LottiePlayer.wc.svelte`, which this branch does not touch — pre-existing on `release`.

No browser-level web-component test harness exists in the repo (there are no `*wc*` test files), so parity is verified by the build output and the audit script rather than by a rendering test. Adding that harness is a larger piece of work than this change.

## Summary by CodeRabbit

* **New Features**
  * Added web-component versions of ChipInput, ColorPicker, Combobox, and SplitInput.
  * Components support HTML attributes, JavaScript properties, and event handlers.
  * Added support for accessibility labels and configurable component options.

* **Documentation**
  * Added usage examples and guidance for loading and configuring each web component.
  * Documented kebab-case attributes and properties requiring JavaScript assignment.


Tara-ag (commented):
Review in progress.

src/wc/components/Combobox.wc.svelte
```diff
@@ -0,0 +1,42 @@
...
+      disabled: { type: 'Boolean', reflect: true },
+      name: { type: 'String', reflect: true },
+      testId: { type: 'String', attribute: 'test-id' },
+      classes: { type: 'String' },
+      noResultsText: { type: 'String', attribute: 'no-results-text' },
+      ariaLabel: { type: 'String', attribute: 'aria-label' },
+      multiple: { type: 'Boolean', reflect: true },
+      selected: { type: 'Array' },
+      maxSelected: { type: 'Number', attribute: 'max-selected' },
+      maxSelectedText: { type: 'String', attribute: 'max-selected-text' },
```
Tara-ag, line 21: 💡 MINOR: `allowCreate` is a public boolean prop on Combobox but is not mapped to a kebab-case attribute here. Per the repo's web-component parity standard, every camelCase Svelte prop except functions/snippets should have an explicit attribute mapping. Add `allowCreate: { type: 'Boolean', reflect: true, attribute: 'allow-create' }` after `maxSelectedText`, and list `allow-create` in `docs/Combobox.md`.

Tara-ag (commented):
One minor web-component parity finding: `Combobox.wc.svelte` omits the `allowCreate` → `allow-create` attribute mapping. Otherwise the wrappers, docs updates, and registry additions look correct and consistent with the existing wrappers. No blocking API, accessibility, SSR, or security issues found.

Tara-ag (commented):
Review in progress.

src/wc/components/Combobox.wc.svelte
Tara-ag, line 21: 💡 MINOR: `allowCreate` is a public boolean prop on Combobox but is not mapped to a kebab-case attribute in this web-component wrapper. Per the repo's parity standard, every camelCase Svelte prop except functions and snippets should have an explicit attribute mapping.

Suggested fix: add `allowCreate: { type: 'Boolean', reflect: true, attribute: 'allow-create' }` after `maxSelectedText`, and list `allow-create` in docs/Combobox.md with the other kebab-case attributes.

Tara-ag (commented):
Adds web-component wrappers for ChipInput, ColorPicker, Combobox and SplitInput, registers them in src/wc/index.ts, and updates the corresponding docs. The Svelte component APIs are untouched, so there are no backward-compatibility breaks. One minor parity nit: Combobox.wc.svelte omits the allowCreate → allow-create attribute mapping.

murdore:
## Browser proof — the four elements, measured, with a negative control

This PR's own description said parity was *"verified by the build output and the audit script rather than by a rendering test."* That gap is now closed. What follows is a real browser loading two bundles and measuring each element.

**The two arms**

- **BEFORE** — the published `@juspay/svelte-ui-components@2.136.0`, unpacked from the npm tarball. That is the current latest release and it does not contain this PR. It is not a build I made.
- **AFTER** — this branch's `dist-wc/index.js`.

Same page, same probes; only the bundle differs. The page fingerprints the bundle **in the browser** — byte length, a content hash, and whether the tag string is present in the source text — so a stale file cannot pass itself off as a fresh one.

**Every case carries its own control.** Each element is mounted twice: once with the attribute under test, once with it omitted. A PASS requires all three of — the tag is registered, the measured value equals the expected value, **and the control differs**. That third condition is what rules out "the value would have been there anyway."

| Element | attribute under test | measured (with) | control (omitted) |
|---|---|---|---|
| `sui-chip-input` | `aria-label` | `Test customer emails` | `(no aria-label)` |
| `sui-color-picker` | `show-value` | `#3fb950` | `(no hex field)` |
| `sui-combobox` | `aria-label` | `Pick a country` | `(no listbox aria-label)` |
| `sui-split-input` | `length` | 6 inputs | 3 inputs |

**BEFORE: 4/4 FAIL** — all four tags `UNDEFINED`, elements render as inert unknown elements.
**AFTER: 4/4 PASS.**

The `sui-input` control string is `true` in **both** arms. That is what makes the BEFORE failures meaningful: the bundle loaded fine, so the four absences are real absences and not a broken load.

### Two defects the harness found in itself

Both would have produced a result that looked fine and meant nothing, so they are worth stating:

1. The first `sui-color-picker` probe read `shadowRoot.textContent` and matched the component's own `<style>` block — it was measuring CSS, not the rendered value.
2. The `sui-combobox` fixture used `{label, value}` items. `ComboboxItem` is `{id, label}` and the dropdown keys on `item.id`, so both items keyed `undefined`, Svelte threw `each_key_duplicate`, and the listbox never rendered. I would otherwise have reported that as "`aria-label` doesn't map." It maps correctly; my fixture was fiction.

One thing worth knowing for consumers: `sui-combobox`'s `aria-label` names the **dropdown listbox** (`role="listbox"`), not the text input, so it only exists while the dropdown is open. The input itself still has no accessible name — separate from this PR, but someone should look at it.

The harness is not committed — evidence and session scratch don't belong in the branch.
before-published-2 136 0
after-this-pr

https://github.com/user-attachments/assets/d554d747-b7c3-4cbf-8ccb-a393de71d724

murdore:
## Proof of testing

There was no web-component harness in the repo, so I built the `dist-wc` bundle and served it from a plain static page — no Svelte, no bundler, just the four tags in HTML, which is how a consumer would actually use them.

https://github.com/user-attachments/assets/41532de4-cac0-48a3-8e61-9075e0ba3112

```
customElements.get() + shadow root
  sui-chip-input     defined, 3 node(s), 28px tall
  sui-combobox       defined, 3 node(s), 36px tall
  sui-color-picker   defined, 3 node(s), 38px tall
  sui-split-input    defined, 3 node(s), 36px tall
```

Three separate checks, because each can pass while the next fails: `customElements.get()` returns a constructor only if the bundle defined the tag; a shadow root with children proves it actually mounted rather than registering and rendering nothing; and a non-zero height proves it occupies space. Then the clip drives one for real — typing `priority` and `urgent` into `sui-chip-input` and asserting both chips appear — so this is not merely a registration check.

### One thing to look at before merging

On a bare page the chips render with the dismiss controls as large dark squares stacked under their labels, rather than as inline pills. In the Svelte demo they look correct.

I want to be careful about what that means, because "it renders differently without the app stylesheet" is normally **not** a defect — that is what a surface stylesheet is for, and the library's own primitives shift the same way. But a custom element is the one case where that reasoning does not apply: consumers of `dist-wc` have no app stylesheet by definition, so how these look on a bare page **is** the shipped appearance.

I have not diagnosed it and I am not claiming it is broken — it may be a missing token default reaching through the shadow boundary. Flagging it because this PR is what makes those tags public, and once published the bare-page render is the contract.

murdore:
## `allowCreate` fixed — and auditing the rule found five more (`6fdf604`)

@Tara-ag flagged `allowCreate` as unmapped. Confirmed, and rather than patch the one line I applied the stated rule to all four wrappers this PR adds.

### What the audit found

Comparing every public prop in each component's `properties.ts` against its wrapper's `props` map:

```
ChipInput    public= 9   mapped= 9   MISSING: none
ColorPicker  public= 8   mapped= 8   MISSING: none
SplitInput   public=11   mapped=11   MISSING: none
Combobox     public=41   mapped=27   MISSING: 14
```

Three of the four were already complete. Combobox was missing **six real ones**, not one:

| prop | kind | now mapped as |
|---|---|---|
| `allowCreate` | boolean | `{ type: 'Boolean', attribute: 'allow-create', reflect: true }` |
| `createLabel` | function | `{ type: 'Object' }` — same shape as the already-mapped `filterFn` |
| `action` | object | `{ type: 'Object' }` |
| `onadd` / `onremove` / `oncreate` | events | `{ type: 'Object' }` — the wrapper already maps eight other events |

The three event callbacks are the ones I would most expect to bite: `onselect`, `oninput`, `onopen`, `onclose`, `onkeydown`, `onfocus`, `onblur` and `onchange` are all mapped, so a consumer would reasonably assume `onadd`/`onremove`/`oncreate` were too, and their absence is silent.

### The other eight are deliberately unmapped, and I checked rather than assumed

`itemSnippet`, `emptySnippet`, `inputPrefix`, `inputSuffix`, `dropdownHeader`, `dropdownFooter`, `pillSnippet`, `actionIcon` are all `Snippet` types. **No wrapper in this repo maps a snippet** — verified across `Accordion`, `Banner`, `Book` and `AttachmentChipRow`, each of which has snippet props and maps none of them. Snippets cannot cross the custom-element boundary, so omitting them is the convention, not an oversight. Leaving them out.

Also checked and dismissed: `keepOpen` looked unmapped, but it belongs to `ComboboxAction`, a nested object type, not to the component's own props.

### Verification

- Re-ran the audit: `non-snippet public props unmapped: NONE`.
- `pnpm run lint` exit 0.
- `pnpm run build:wc` succeeds and `allow-create` appears in the built `dist-wc/index.js`, so the attribute genuinely reaches the published bundle.
- `pnpm run check:wc` **fails, and it failed identically before my change** — 3 errors, all in `src/wc/components/LottiePlayer.wc.svelte` (`$host` used before declaration), `COMPLETED 487 FILES 3 ERRORS` in both arms. Pre-existing, unrelated, and worth its own fix.

Note this branch is checked out in another local worktree; I force-pushed with a lease against `5b89d9a`, so that checkout will need a pull.

src/wc/components/Combobox.wc.svelte
```diff
@@ -0,0 +1,48 @@
+<svelte:options
+  customElement={{
+    tag: 'sui-combobox',
+    shadow: 'open',
+    props: {
+      items: { type: 'Array' },
```
Tara-ag, line 6: ⚠️ MAJOR: The wrapper omits all of Combobox's snippet props (`itemSnippet`, `emptySnippet`, `inputPrefix`, `inputSuffix`, `dropdownHeader`, `dropdownFooter`, `pillSnippet`, `actionIcon`) while docs/Combobox.md says they are set as properties. Add them to the `customElement` props declaration with `type: 'Object'` so the documented web-component API actually works.

Suggested addition inside `props`:
```js
itemSnippet: { type: 'Object' },
emptySnippet: { type: 'Object' },
inputPrefix: { type: 'Object' },
inputSuffix: { type: 'Object' },
dropdownHeader: { type: 'Object' },
dropdownFooter: { type: 'Object' },
pillSnippet: { type: 'Object' },
actionIcon: { type: 'Object' },
```

src/wc/components/ChipInput.wc.svelte
```diff
@@ -0,0 +1,24 @@
+<svelte:options
```
Tara-ag, line 1: 💡 MINOR: Four new custom elements are introduced but no automated tests (unit, integration, or build-level) are added. Please add at least minimal build-time coverage asserting each wrapper registers the expected tag and prop declarations, or document the manual verification steps if a harness is out of scope.

Tara-ag (changes requested):
Reviewed file-by-file. Found one MAJOR issue: the new `sui-combobox` wrapper does not declare any of Combobox's snippet props, despite docs/Combobox.md stating they can be set as properties. This makes the documented web-component API incomplete. Also noted one MINOR issue: no automated tests for the four new custom elements. Requesting changes for the Combobox prop gap.

src/wc/components/Combobox.wc.svelte
Tara-ag, line 6: ⚠️ MAJOR: The new sui-combobox wrapper declares scalar, array, object and handler props but does not declare any of the Combobox component's snippet props (`itemSnippet`, `emptySnippet`, `inputPrefix`, `inputSuffix`, `dropdownHeader`, `dropdownFooter`, `pillSnippet`, `actionIcon`). `docs/Combobox.md` explicitly states that "the snippet props ... are set as properties", which is impossible because they are absent from the `customElement` props declaration. This makes the web-component API materially narrower than the Svelte API and contradicts the documentation.

Suggested fix: Add the missing snippet props to the `customElement` props declaration with `type: 'Object'` and forward them to `Combobox`:

```js
props: {
  ...existingProps,
  itemSnippet: { type: 'Object' },
  emptySnippet: { type: 'Object' },
  inputPrefix: { type: 'Object' },
  inputSuffix: { type: 'Object' },
  dropdownHeader: { type: 'Object' },
  dropdownFooter: { type: 'Object' },
  pillSnippet: { type: 'Object' },
  actionIcon: { type: 'Object' },
}
```

Then pass them through: `<Combobox {...props} />` already covers this once they are in `props`. Alternatively, expose them as named slots if that fits the web-component consumption model better, and update `docs/Combobox.md` to match.

src/wc/components/ChipInput.wc.svelte
Tara-ag, line 1: 💡 MINOR: The PR introduces four new custom elements (`sui-chip-input`, `sui-color-picker`, `sui-combobox`, `sui-split-input`) but adds no unit, integration, or build-level tests for them. The repo's recurring standard is that new behaviour requires focused automated regression tests rather than relying solely on existing tests passing.

Suggested fix: Add minimal build-time or unit tests that assert each new wrapper is registered with the expected tag and prop declarations, or extend the existing test setup with a lightweight web-component smoke test. If a harness is truly out of scope, document the testing gap and the manual verification steps in the PR.

Tara-ag (changes requested):
Reviewed file-by-file. Found one MAJOR public-API/documentation mismatch in the new Combobox web-component wrapper (missing snippet props) and one MINOR testing gap across the four new wrappers. Both comments have been posted inline.

Tara-ag:

Severity counts — 🔒 CRITICAL: 0 · ⚠️ MAJOR: 5 · 💡 MINOR: 1

The PR exposes ChipInput, ColorPicker, Combobox and SplitInput as custom elements and adds matching docs and Playwright tests. Two real API-parity gaps remain in the new web-component surface: ChipInput is missing the `editable` boolean prop and `onedit` callback in its `customElement.props` declaration, so consumers cannot enable in-place chip editing or receive edit events via `sui-chip-input`. Combobox is missing the `actionIcon` snippet prop accessor, contradicting the documented web-component snippet API, and the Combobox docs omit the `allow-create` attribute that the wrapper does expose. These are public-contract issues for a published library. No CRITICAL security or breaking-change findings.

- ⚠️ **MAJOR**: ChipInput web component omits documented `editable` and `onedit` props — `src/wc/components/ChipInput.wc.svelte:6`
  The Svelte ChipInput component exposes `editable` and `onedit` as part of its public API (src/lib/ChipInput/properties.ts and ChipInput.svelte). The new `sui-chip-input` custom element wrapper does not declare either prop in `customElement.props`, so consumers cannot enable in-place chip editing or receive edit callbacks via the web component. Because Svelte only creates accessors for declared cus…
- ⚠️ **MAJOR**: ChipInput web component omits the editable prop and onedit callback — `src/wc/components/ChipInput.wc.svelte:5`
  ChipInput.svelte exposes `editable` (boolean) and `onedit` (callback) as part of its public prop contract. The new `ChipInput.wc.svelte` wrapper declares neither, so the web component cannot enable inline chip editing or receive edit-commit events. This contradicts the PR's claim that the four new wrappers have 'zero gaps' against the component prop surfaces.
- ⚠️ **MAJOR**: Combobox web component omits the actionIcon snippet prop — `src/wc/components/Combobox.wc.svelte:35`
  Combobox.svelte exposes `actionIcon` as a snippet prop for customizing the persistent action row's leading icon. The new `Combobox.wc.svelte` wrapper declares every other snippet prop (itemSnippet, emptySnippet, inputPrefix, inputSuffix, dropdownHeader, dropdownFooter, pillSnippet) but omits `actionIcon`. Because undeclared custom-element props get no accessor, consumers cannot assign `element.act…
- ⚠️ **MAJOR**: ChipInput web component omits the editable prop — `src/wc/components/ChipInput.wc.svelte:5`
  ChipInput.svelte exposes `editable` as a public boolean prop. The new `ChipInput.wc.svelte` wrapper does not declare it in `customElement.props`, so `element.editable = true` has no effect and consumers cannot enable in-place chip editing via the custom element.
- ⚠️ **MAJOR**: ChipInput web component omits the onedit callback — `src/wc/components/ChipInput.wc.svelte:5`
  ChipInput.svelte exposes `onedit` as a public callback prop fired after an in-place edit commits. The new `ChipInput.wc.svelte` wrapper does not declare it in `customElement.props`, so consumers cannot receive edit-commit events via the custom element.
- 💡 **MINOR**: Combobox docs omit the allow-create attribute — `docs/Combobox.md:392`
  Combobox.wc.svelte declares `allowCreate` with `attribute: 'allow-create'` and `reflect: true`, making it a usable kebab-case attribute. The docs list the kebab-case attributes but do not include `allow-create`, so consumers reading docs/Combobox.md will not know it can be set from HTML.

murdore:
## Validated on the current `release` (`3486dd6`) — `56ffc73`, one commit, 9 files

This branch was already sitting on the current tip, so no rebase was needed. What it needed was checking, because the red `build` on it was never this branch — it was the Pages prerender break on `release` itself, fixed in #485 and now merged.

Re-run here rather than inferred:

| | result |
|---|---|
| `pnpm lint` | exit 0 |
| `BASE_PATH=/svelte-ui-components pnpm build` | exit 0, `build/` 7.7 MB |
| `pnpm run build:wc` | exit 0, `dist-wc/` 1.5 MB |

`build:wc` is the one that matters for this PR, since exposing custom elements that fail to compile would be invisible to the other two.

**Checked the built artefact, not just the source.** All four elements register in `dist-wc/index.js`:

```
sui-chip-input   sui-color-picker   sui-combobox   sui-split-input
```

and `allow-create` — one of the six props this PR adds to the Combobox wrapper — is present in the bundle, so the attribute mapping survives the build rather than only existing in the `.wc.svelte` source.

Diff is +187 / −9 across 9 files, which matches the intent (4 wrappers, 4 docs, the `wc/index.ts` registration).

murdore:
## Review findings, verified in a browser — two fixed, one refuted

Pushed as `e539013`, rebased onto current `release` (`3486dd6`). I tested each finding against the built `dist-wc` rather than reasoning from the source, because all three are claims about runtime behaviour that the source does not settle.

### ✅ Combobox omits the snippet props — correct, and the docs were lying

`docs/Combobox.md:399` already stated *"snippet props and every `on*` handler are set as properties"*. That was false: an undeclared name gets **no accessor** on the custom element, so the assignment lands on the DOM node and never reaches the component.

Measured before the fix:

| property | accessor on the element |
|---|---|
| `itemSnippet` | **absent** |
| `pillSnippet` | **absent** |
| `items` | present |
| `allowCreate` | present |

The last two are the control — they show the probe discriminates rather than reporting "absent" for everything. All eight snippet props are now declared.

### ✅ `values` is `$bindable` and was spread one-way — correct

Isolated properly: set `values = ['seed']`, then add "beta" through the UI.

```
onadd fired with      : "beta"        ← the callback was always fine
element.values after  : ["seed"]      ← the host property never moved
```

So a consumer reading the property back saw stale state while the component's own view had advanced. Now bound, and the same treatment applied to `ColorPicker.value` and Combobox's `value` / `inputValue` / `open` / `highlightedIndex` / `selected`.

**Worth flagging for scope:** this is not specific to these four wrappers. **31 pre-existing wrappers** spread an unbound `$bindable` prop the same way (Accordion, Banner, Button, Calendar, Chat, ChatBubble, ChatComposer, …). I have not swept those into this PR, but someone should decide whether to.

### ❌ `on*` props are treated as DOM event listeners — does not reproduce

This one I'm pushing back on, with evidence.

The concern is reasonable a priori: `onadd` *is* a DOM event-handler name. But Svelte defines a real accessor for every declared prop, and it shadows the DOM behaviour:

```
descriptor for onadd        : accessor defined by Svelte
descriptor for placeholder  : accessor defined by Svelte   ← identical treatment
```

The decisive test is the **argument**, since a DOM listener would be handed an `Event`. Setting `element.onadd` and adding a chip:

```
received: "alpha"        (a string — the component's declared contract)
not:      Event(add)
```

So the callbacks arrive correctly today. Renaming these properties would change a working public API and break any existing consumer. I've left them as they are, and pinned the behaviour with a regression test so a future change cannot quietly break it.

### ✅ Yama MINOR — no automated tests

Fixed, and it's the repo's **first** browser-level coverage for custom elements: `tests/wc-custom-elements.spec.ts`, 5 cases — registration, kebab-case attribute mapping, callback forwarding, the bindable round-trip, and the snippet accessors. `dist-wc` is a bundle rather than a route, so it's injected into a same-origin page; `pnpm run build` already runs `build:wc`, so the file exists whenever the suite runs.

Both fix-verifying cases are **negative-controlled**, because a spec that has only ever been green is indistinguishable from one that asserts nothing:

- reverting the bindable wiring → fails with `Received: ["seed"]` against `Expected: ["seed","beta"]`
- stripping the snippet declarations → reports all eight as missing

Tree confirmed byte-identical after both controls.

### Note on the earlier `allow-create` finding

Already addressed — `allowCreate`, `createLabel`, `action`, `onadd`, `onremove` and `oncreate` were mapped in a prior revision of this branch.

Full library suite: **345 passed**.

Tara-ag (commented):
Review in progress

src/wc/components/ChipInput.wc.svelte
```diff
@@ -0,0 +1,28 @@
...
+  customElement={{
+    tag: 'sui-chip-input',
+    shadow: 'open',
+    props: {
+      values: { type: 'Array' },
```
Tara-ag, line 6: ⚠️ MAJOR: ChipInput web component omits documented `editable` and `onedit` props

The Svelte `ChipInput` component exposes `editable` and `onedit` as part of its public API (`src/lib/ChipInput/properties.ts` and `ChipInput.svelte`). The new `sui-chip-input` custom element wrapper does not declare either prop in `customElement.props`, so consumers cannot enable in-place chip editing or receive edit callbacks via the web component. Because Svelte only creates accessors for declared custom-element props, assigning `element.editable = true` or `element.onedit = ...` has no effect.

Suggested fix: add the missing declarations to `customElement.props`:

```ts
editable: { type: 'Boolean', reflect: true },
onedit: { type: 'Object' },
```

Then update `docs/ChipInput.md` to mention that `editable` is available as a boolean attribute and `onedit` is set as a property.

murdore:
## Closing note — this merged mid-review

#482 merged as `ff24661` while the last revision was still in flight. Three things worth recording so the trail is complete.

**1. The merged commit is `e539013`**, which carries both fixes from my verification comment above — the eight Combobox snippet declarations and the `$bindable` wiring — plus `tests/wc-custom-elements.spec.ts`, the repo's first browser-level coverage for custom elements.

**2. It also carries a formatting defect, and `release` is currently failing lint because of it.** My final revision fixed the prettier reflow in `Combobox.wc.svelte`, but the merge landed before that push, and the branch deletion then rejected it. `pnpm lint` runs `prettier --check src`, so release and every branch cut from it inherits the failure. Fix is up as #488 — formatting only, no behaviour change, and the new spec passes 5/5 against it.

That one is mine: I formatted the spec file but not the four wrappers I had edited by script, and pushed without running the repo's own `pnpm lint` first.

**3. The `on*` refutation is now pinned by a test, not just an argument.** The claim was that callback props whose names begin with `on` are interpreted as DOM event listeners and never reach the component. Measured: `onadd` gets a real Svelte accessor, identical to `placeholder`, and the callback receives `"alpha"` — the component's own contract — rather than an `Event`. Renaming those properties would have broken a working public API. `wc-custom-elements.spec.ts` asserts the argument's *type*, so a future change cannot quietly turn them into listeners.

### Still open, and deliberately not swept into this PR

**31 pre-existing wrappers spread an unbound `$bindable` prop** — the same defect fixed here for these four. Accordion, Banner, Button, Calendar, Chat, ChatBubble, ChatComposer and others all reassign internally while their host element property stays frozen at whatever the consumer last set.

That is a systemic API bug, not a property of these four components, and it needs a decision rather than a drive-by fix: binding every one changes the observable behaviour of 31 public custom elements at once. Worth its own issue.

docs/Combobox.md
```diff
@@ -381,3 +381,20 @@ type ComboboxAction = {
...
 - Input (for the text input field — inherits validation, text transformers, and ARIA support)
+
+## Web component
+
+Available as `<sui-combobox>`.
+
+```html
+<sui-combobox placeholder="Search…" aria-label="Country" test-id="country"></sui-combobox>
+```
+
```
Tara-ag, line 392: 💡 MINOR: The docs list of kebab-case attributes omits `allow-create`. `Combobox.wc.svelte` declares `allowCreate: { type: 'Boolean', attribute: 'allow-create', reflect: true }`, so it is a usable HTML attribute. Append `allow-create` to the attribute list so consumers know it can be set from markup.

src/wc/components/ChipInput.wc.svelte
```diff
@@ -0,0 +1,28 @@
+<svelte:options
+  customElement={{
+    tag: 'sui-chip-input',
+    shadow: 'open',
+    props: {
```
Tara-ag, line 5: ⚠️ MAJOR: ChipInput web component omits the `editable` prop and `onedit` callback. `src/lib/ChipInput/properties.ts` exposes both as part of the public contract, and `ChipInput.svelte` destructures and uses them. The wrapper only declares `values`, `ariaLabel`, `placeholder`, `disabled`, `testId`, `classes`, `onadd`, `ondismiss`, and `onchange`, so consumers cannot enable inline chip editing or receive edit-commit events via the custom element. Add `editable: { type: 'Boolean', reflect: true }` and `onedit: { type: 'Object' }` to the `customElement.props` map.

src/wc/components/Combobox.wc.svelte
```diff
@@ -0,0 +1,76 @@
...
+      emptySnippet: { type: 'Object' },
+      inputPrefix: { type: 'Object' },
+      inputSuffix: { type: 'Object' },
+      dropdownHeader: { type: 'Object' },
+      dropdownFooter: { type: 'Object' },
+      pillSnippet: { type: 'Object' },
+      actionIcon: { type: 'Object' },
+      allowCreate: { type: 'Boolean', attribute: 'allow-create', reflect: true },
+      createLabel: { type: 'Object' },
+      action: { type: 'Object' },
```
Tara-ag, line 35: ⚠️ MAJOR: Combobox web component omits the `actionIcon` snippet prop. `Combobox.properties.ts` lists `actionIcon?: Snippet`, and `Combobox.svelte` renders it with `{@render actionIcon()}`. The wrapper declares the other seven snippet props but not `actionIcon`, so assigning `element.actionIcon` silently lands on the DOM node and never reaches the component. Add `actionIcon: { type: 'Object' }` alongside the other snippet prop declarations.

Tara-ag (commented):
Reviewed the four new web-component wrappers and their docs/tests. Found two gaps in the prop surface exposed to custom elements (ChipInput `editable`/`onedit`, Combobox `actionIcon`) and one docs omission. Details in inline comments.

src/wc/components/Combobox.wc.svelte
murdore: allowCreate has an explicit kebab-case attribute mapping. Combobox.wc.svelte declares allowCreate: { type: 'Boolean', attribute: 'allow-create', reflect: true } at line 33, present since the PR 482 merge commit (ff24661d6). This matches the suggested fix and is live on release 3.1.0.

murdore: Duplicate of the other allowCreate attribute-mapping thread. It is declared with an explicit allow-create attribute at Combobox.wc.svelte line 33, confirmed present as of the PR 482 merge itself. Resolved on release 3.1.0.

docs/ChipInput.md
```diff
@@ -107,12 +107,23 @@ see (WCAG 2.5.3, Label in Name).
...
-
-This is deliberate, and it is why `ariaLabel` has no kebab-case attribute mapping: there is no
-custom element for an attribute to map onto. Adding one would newly expose ChipInput as a web
-component rather than restore parity with an existing element. If ChipInput should be exposed,
-that is its own change — a new tag, a new shadow root, and the attribute-casing tests the other
-wrappers carry.
+Available as `<sui-chip-input>`. Import the web component build separately:
+
+```html
+<script type="module" src="@juspay/svelte-ui-components/wc"></script>
```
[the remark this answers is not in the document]
murdore: This example still will not work if copy-pasted into a plain HTML page. docs/ChipInput.md line 154 imports the web component via a bare @juspay/svelte-ui-components/wc specifier, which only resolves inside a bundler or with an import map, whereas README.md correctly shows the GitHub Pages, jsDelivr and unpkg URL forms for the same import. The identical unresolvable specifier is also still present in docs/DateRangePicker.md around line 453, so this is a pattern across the per-component docs, not a one-off. Closing this needs the per-component docs updated to a browser-resolvable URL, matching the README.

docs/Combobox.md
```diff
@@ -381,3 +381,20 @@ type ComboboxAction = {
 This component uses the following library components internally:
 
 - Input (for the text input field — inherits validation, text transformers, and ARIA support)
+
+## Web component
+
+Available as `<sui-combobox>`.
+
+```html
+<sui-combobox placeholder="Search…" aria-label="Country" test-id="country"></sui-combobox>
```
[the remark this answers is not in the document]
murdore: This documentation gap is still present. The web-component usage example, at docs/Combobox.md lines 396-409, sets aria-label on <sui-combobox> with no explanation that it names the opened listbox rather than the text input, and it does not show the inputProperties path for naming the input itself. The general props table earlier in the file documents aria-label's target for the Svelte component, but that clarification is not carried into the web-component section a reader copying this example would actually be looking at. Closing this needs a note in the Web component section plus an inputProperties example for naming the input.

src/wc/components/ChipInput.wc.svelte
```diff
@@ -0,0 +1,24 @@
...
+      values: { type: 'Array' },
+      ariaLabel: { type: 'String', attribute: 'aria-label' },
+      placeholder: { type: 'String', reflect: true },
+      disabled: { type: 'Boolean', reflect: true },
+      testId: { type: 'String', attribute: 'test-id' },
+      classes: { type: 'String' },
+      onadd: { type: 'Object' },
+      ondismiss: { type: 'Object' },
+      onchange: { type: 'Object' }
```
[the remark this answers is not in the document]
murdore: This concern does not materialise. PR 482 (commit ff24661d6) added tests/wc-custom-elements.spec.ts specifically to prove it: one test assigns chip.onadd to a callback, dispatches real input events, and asserts the callback receives the component's own value argument rather than a DOM Event object, which is exactly what would happen if the browser's native event-handler property had won instead. Svelte defines its own accessor for a declared on-prefixed custom-element prop, and that accessor takes precedence over the platform's GlobalEventHandlers property. The same on-prefixed naming is used consistently across every .wc.svelte wrapper in this library, so renaming these would be an inconsistent, unnecessary breaking change. Declined as proven unfounded by the shipped test.

src/wc/components/ChipInput.wc.svelte
```diff
@@ -0,0 +1,24 @@
...
+    }
+  }}
+/>
+
+<script lang="ts">
+  import ChipInput from '$lib/ChipInput/ChipInput.svelte';
+  let props = $props();
```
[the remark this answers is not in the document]
murdore: values is bound correctly, not one-way spread. ChipInput.wc.svelte destructures values as bindable and passes it back with bind:values at lines 25 and 28, so a consumer reading element.values after the component adds or removes a chip sees the current state rather than whatever was last assigned. This is also covered by a dedicated test in tests/wc-custom-elements.spec.ts that sets a chip programmatically and asserts the host element's own property reflects it afterward. Shipped in PR 482 (commit ff24661d6) on release 3.1.0.

src/wc/components/Combobox.wc.svelte
murdore: All eight documented snippet props are declared. Combobox.wc.svelte's customElement.props includes itemSnippet, emptySnippet, inputPrefix, inputSuffix, dropdownHeader, dropdownFooter, pillSnippet and actionIcon, at lines 25-32, present since the PR 482 merge commit (ff24661d6), so each one gets a real accessor and an assignment reaches the component. tests/wc-custom-elements.spec.ts also asserts this directly by checking every one of the eight names for a prototype-chain accessor. This is live on release 3.1.0.

src/wc/components/ChipInput.wc.svelte
murdore: Automated coverage for the new custom elements was added in the same PR. tests/wc-custom-elements.spec.ts asserts all four tags register, that a kebab-case attribute reaches the underlying prop, that on-prefixed callback props are forwarded as callbacks rather than intercepted as DOM listeners, that a bindable prop propagates back to the host element, and that every documented Combobox snippet prop has a real accessor. This is a browser-level Playwright suite, not a build-level check, but it directly exercises the behaviour the review asked to have covered, and it shipped in the same PR 482 (commit ff24661d6) on release 3.1.0.

src/wc/components/Combobox.wc.svelte
murdore: Duplicate of the other missing-snippet-props thread. All eight are declared, itemSnippet through actionIcon, at Combobox.wc.svelte lines 25-32, confirmed present as of the PR 482 merge itself, so the web-component API is not narrower than docs/Combobox.md describes. Resolved on release 3.1.0.

src/wc/components/ChipInput.wc.svelte
murdore: Duplicate of the other testing-gap thread. tests/wc-custom-elements.spec.ts, added in the same PR 482 (commit ff24661d6), covers tag registration, attribute forwarding, callback correctness, bindable back-propagation and Combobox snippet-prop accessors for all four new custom elements. Resolved on release 3.1.0.

src/wc/components/ChipInput.wc.svelte
murdore: This gap has not been closed. ChipInput.svelte and its properties.ts expose editable and onedit as part of the public API, but ChipInput.wc.svelte's customElement.props still only lists values, ariaLabel, placeholder, disabled, testId, classes, onadd, ondismiss and onchange, with no entry for either. Since Svelte only creates a custom-element accessor for a declared prop, setting element.editable or element.onedit from a web-component consumer currently does nothing. Closing this needs editable and onedit added to the props map, along with a docs/ChipInput.md note describing them as an attribute and a property respectively.

docs/Combobox.md
murdore: The kebab-case attribute list in the docs is missing one real attribute. docs/Combobox.md line 406 enumerates aria-label, test-id, input-value, highlighted-index, no-results-text, max-selected and max-selected-text, but allow-create is not among them even though Combobox.wc.svelte declares it as a genuine reflected boolean attribute. A consumer reading only this line would not know allow-create can be set from markup. Closing this needs allow-create appended to that list.

src/wc/components/ChipInput.wc.svelte
murdore: Duplicate of the other editable/onedit thread. ChipInput.wc.svelte's props map still omits both, at lines 5-15, so a web-component consumer cannot enable inline chip editing or receive edit-commit callbacks. Same fix needed: add editable as a reflected Boolean and onedit as an Object-typed callback prop.

src/wc/components/Combobox.wc.svelte
murdore: actionIcon is declared alongside the other seven snippet props. Combobox.wc.svelte line 32 includes actionIcon: { type: 'Object' }, present since the PR 482 merge commit (ff24661d6), so assigning element.actionIcon reaches the component instead of landing inertly on the DOM node. tests/wc-custom-elements.spec.ts also explicitly checks this name in its accessor-completeness test. Resolved on release 3.1.0.

docs/ChipInput.md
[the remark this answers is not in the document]
murdore: Fixed in juspay/svelte-ui-components#502 (commit 64f91c5be): docs/ChipInput.md now loads the bundle from https://juspay.github.io/svelte-ui-components/wc/index.js, the URL the README documents; docs/DateRangePicker.md had the same bare specifier and was fixed in the same commit.

docs/Combobox.md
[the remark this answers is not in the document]
murdore: Fixed in juspay/svelte-ui-components#502 (commit 64f91c5be): docs/Combobox.md's web-component section now states that aria-label names the listbox, not the input, and shows naming the input through inputProperties = { label: 'Country' }.

src/wc/components/ChipInput.wc.svelte
murdore: Fixed in juspay/svelte-ui-components#502 (commit 64f91c5be): sui-chip-input now declares editable ({ type: 'Boolean', reflect: true }) and onedit ({ type: 'Object' }) in customElement.props (src/wc/components/ChipInput.wc.svelte), and docs/ChipInput.md says which is an attribute and which a property; pinned by the new 'exposes editable as an attribute and onedit as a property' test in tests/wc-custom-elements.spec.ts, which clicks a chip on the custom element, edits it and asserts onedit receives [new, previous].

docs/Combobox.md
murdore: Fixed in juspay/svelte-ui-components#502 (commit 64f91c5be): allow-create is now listed with the other kebab-case attributes in docs/Combobox.md's web-component section.

src/wc/components/ChipInput.wc.svelte
murdore: Fixed in juspay/svelte-ui-components#502 (commit 64f91c5be): sui-chip-input now declares editable ({ type: 'Boolean', reflect: true }) and onedit ({ type: 'Object' }) in customElement.props (src/wc/components/ChipInput.wc.svelte), and docs/ChipInput.md says which is an attribute and which a property; pinned by the new 'exposes editable as an attribute and onedit as a property' test in tests/wc-custom-elements.spec.ts, which clicks a chip on the custom element, edits it and asserts onedit receives [new, previous].

PR #483, opened 2026-09-01
fix(demo): serve demo-page media locally instead of from picsum.photos

murdore:
## The suite's pass rate depended on a stranger's CDN

Eight demo routes pulled their images from `picsum.photos`. Playwright navigates with `waitUntil: 'load'`, and an image from a third-party host blocks that event — so whether a spec passed depended on how fast an unrelated public service answered.

This is not theoretical. A full suite run failed **5 specs** across `/components/status` and `/components/chat`, all at the 30s `page.goto` timeout. Instrumenting the navigation showed exactly one request still outstanding when `load` never fired:

```
attempt 1: loaded=false elapsed=30002ms readyState=interactive pending=1
  pending: https://picsum.photos/48/48?random=20
```

The same tests take ~1.3s when picsum answers quickly and ~20s when it is slow. Identical code, three different outcomes.

## The change

All **17** references across **8** routes now point at local assets in `static/demo-media/` — two new SVG placeholders plus `assistant-avatar.png` and `sunset-beach-thumb.jpg`, already vendored there.

| route | was | now |
|---|---|---|
| `status` | picsum 48×48 | `status-success.svg` (new) |
| `chat` | picsum 64 | `assistant-avatar.png` |
| `img` ×4 | picsum 200×150 | `sunset-beach-thumb.jpg` |
| `brand-loader`, `icon-stack` ×2, `grid-item` ×3, `icon` ×3, `badge` ×2 | picsum | `placeholder-square.svg` (new) |

Nothing here ships to npm — `files` publishes `dist` and `dist-wc`, and neither `static/` nor `src/routes/` is in either.

## Verification

- The **8 specs** that navigate to the two affected routes: **45/45 passed** at 3 repeats each.
- `pnpm run lint`: exit 0 — prettier clean, eslint clean, event-casing `100 known, 0 new`.
- `grep -rn picsum src/ static/` → no matches.

## Two things I could not prove, stated rather than glossed

**The negative control did not reproduce the failure.** Restoring the picsum URL and re-running gave **6/6 passing**, because picsum is responding well right now. That is the nature of a network-dependent flake — it cannot be summoned on demand. The evidence for the fix is the instrumented capture above plus the removal of the dependency, not a red-to-green control.

**This does not make the demo pages offline.** A request probe over all eight routes after the change still finds two external hosts on every page:

```
/components/status  external hosts: 2 -> fonts.googleapis.com, fonts.gstatic.com
```

`src/routes/+layout.svelte:31` loads Nunito Sans from Google Fonts, and a stylesheet blocks `load` the same way an image does. Left alone here because vendoring a variable font is a separate change with its own tradeoffs — but it is the same failure mode, and if CI keeps flaking after this, that is where to look next. The observed failures pointed at picsum, so this fixes what actually broke.

## Related

Found while rebasing #479. It is unrelated to that PR's content, and is why the library suite reported 325/5 rather than 330/0 there.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated component demos to use reliable local image assets instead of external image services.
  - Improved consistency and availability of badge, loader, chat, grid, icon, image, and status examples.
  - Preserved existing image test targeting and demo behavior.

### DIFF
--- a/src/routes/components/badge/+page.svelte
+++ b/src/routes/components/badge/+page.svelte
@@ -8,6 +8,6 @@
 </div>
 
 <div class="demo-row" style="align-items: center;">
-  <Badge image="https://picsum.photos/24/24?random=10" value="5" />
-  <Badge image="https://picsum.photos/24/24?random=11" value="12" />
+  <Badge image="/demo-media/placeholder-square.svg" value="5" />
+  <Badge image="/demo-media/placeholder-square.svg" value="12" />
 </div>
--- a/src/routes/components/brand-loader/+page.svelte
+++ b/src/routes/components/brand-loader/+page.svelte
@@ -9,7 +9,7 @@
 
 <div class="demo-row">
   <BrandLoader
-    brandLogoURL="https://picsum.photos/64/64?random=40"
+    brandLogoURL="/demo-media/placeholder-square.svg"
     brandText="Loading"
     subText="Please wait..."
   />
--- a/src/routes/components/chat/+page.svelte
+++ b/src/routes/components/chat/+page.svelte
@@ -355,7 +355,7 @@
       messages={chat.messages}
       bind:value
       bind:attachments
-      image="https://picsum.photos/64?random=7"
+      image="/demo-media/assistant-avatar.png"
       title="Shopping Assistant"
       subtitle="Online"
       placeholder="Ask anything…"
--- a/src/routes/components/grid-item/+page.svelte
+++ b/src/routes/components/grid-item/+page.svelte
@@ -8,7 +8,7 @@
 </div>
 
 <div class="demo-row">
-  <GridItem icon="https://picsum.photos/40/40?random=1" text="Photos" />
-  <GridItem icon="https://picsum.photos/40/40?random=2" text="Videos" />
-  <GridItem icon="https://picsum.photos/40/40?random=3" text="Music" />
+  <GridItem icon="/demo-media/placeholder-square.svg" text="Photos" />
+  <GridItem icon="/demo-media/placeholder-square.svg" text="Videos" />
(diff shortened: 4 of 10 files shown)


murdore:
## Proof of testing (post-merge)

I merged this with prose claims and no evidence attached, immediately after a session spent attaching sixteen videos to #479 for exactly that reason. Correcting that here.

A video of an image loading proves little, so the evidence is the property that actually changed: **what the demo pages request from the network.** Both sides built and probed identically — eight routes, `waitUntil: 'load'`, every non-origin request recorded.

```
BEFORE (10b5f49)  external hosts = fastly.picsum.photos, fonts.googleapis.com,
                                   fonts.gstatic.com, picsum.photos
                  picsum requests = 34

AFTER  (4c099e6)  external hosts = fonts.googleapis.com, fonts.gstatic.com
                  picsum requests = 0
```

**34, not 17** — `picsum.photos` redirects to `fastly.picsum.photos`, so each of the seventeen references cost two round trips, and each one could stall `load`.

This is the control I said I could not produce when I opened the PR. At the time, restoring the picsum URL and re-running gave 6/6 passing because picsum happened to be healthy, so I could not show red-to-green. Measuring the dependency instead of the symptom removes the dependence on picsum's mood: the request count is deterministic and it is now zero.

### The caveat stands, and it is now the whole remaining risk

`fonts.googleapis.com` and `fonts.gstatic.com` are still there on **every** page, from `src/routes/+layout.svelte:31`. A stylesheet blocks `load` exactly as an image does, so the failure mode this PR addressed is not fully closed — only its larger half is. If the suite keeps flaking on `page.goto`, that is where to look, and vendoring Nunito Sans into `static/` is the fix.

PR #484, opened 2026-09-01
fix(Carousel): restore the properties-bag contract and announce slide changes

murdore:
## Why this exists

#477 merged at 09:02Z on head `d327a1e`, which is the commit **before** the fix I had pushed to that branch. My `bf8e518` was left stranded, so the CRITICAL review thread on that PR shipped unaddressed and is now in **2.136.4**.

This carries the missing half onto current `release`.

## The breaking change that shipped

`Carousel.svelte` went from

```svelte
{#if typeof view.properties === 'object'}
  <view.component properties={view.properties} />
{:else}
  <view.component />
{/if}
```

to

```svelte
<view.component {...view.properties} />
```

The spread is correct and I am not proposing reverting it. Every component in this library takes named props, so the old form meant a normal slide received nothing and rendered empty — #477's own commit message documents that Card's title and description came out blank.

But it is a runtime prop-contract change on a **published** component, which is what @Tara-ag raised three times on #477 (`Carousel.svelte:160`, `:163` x2, marked CRITICAL). The affected population is specific rather than theoretical: anyone who wrote a purpose-built wrapper declaring `properties` — the only shape Carousel was usable with before — now receives nothing under that key, silently.

## The fix

```svelte
<view.component {...view.properties} properties={view.properties} />
```

Both shapes work, so no major version bump is needed. `.github/workflows/release.yml:108` greps `^BREAKING CHANGE:` to cut a major, and forcing 3.0.0 for this seemed the wrong trade when Lighthouse has just pinned 2.136.x mid-migration — but that option is still open and is a one-line change if you would rather take the clean break.

Reading `properties` is documented as deprecated in `docs/Carousel.md`, with the named-prop form shown as the correct one.

**The guarantee is tested, not asserted.** `src/routes/components/carousel/LegacyPropertiesSlide.svelte` is a slide of exactly the old shape, and a spec pins that it still receives its bag. If Carousel ever stops passing it, that test fails rather than a consumer discovering it.

## Also: the aria-live region

Two SUGGESTION threads on #477 asked for slide announcements and merged unaddressed. Added a visually hidden `aria-live="polite"` / `aria-atomic="true"` region reading `Slide N of M`.

The dots already carry `aria-label` saying where they *go*. Nothing said where the carousel *arrived* — and autoplay and swipe never touch a dot at all, so for those two paths there was no announcement of any kind.

## Verification

- `tests/carousel.test.ts` — **5 passed**.
- **Negative control:** removing the compat prop and neutering the live region fails exactly the two new specs (2 failed / 3 passed), then restores byte-identical.
- `pnpm run lint` exit 0.

## Process note

The near-miss is worth recording: I force-pushed the fix to #477's branch and it was merged from the older head minutes later. Nothing warned either side. If a PR has open CRITICAL threads, the head moving is worth a glance before merge — and the branch still existing after merge is what let this be recovered at all.

### DIFF
--- a/src/lib/Carousel/Carousel.svelte
+++ b/src/lib/Carousel/Carousel.svelte
@@ -157,7 +157,15 @@
       <div class="slidesDiv" bind:this={slidesDiv}>
         {#each views as view, index (index)}
           <div class="current-slide">
-            <view.component {...view.properties} />
+            <!-- Spread is the real fix: every component in this library takes named
+                 props, so passing the bag under a single `properties` key meant a
+                 normal slide rendered empty. `properties` is ALSO still passed so
+                 that a consumer who built a purpose-made wrapper around the old
+                 behaviour keeps working -- that wrapper was the only shape this
+                 component was usable with before, and breaking it would force a
+                 major bump for a fix that is otherwise additive. Deprecated: read
+                 the named props, not `properties`. -->
+            <view.component {...view.properties} properties={view.properties} />
           </div>
         {/each}
       </div>
@@ -190,6 +198,13 @@
       {/each}
     </div>
   {/if}
+  <!-- Slide changes are silent to assistive tech otherwise: the dots announce
+       where they GO, but nothing announces where the carousel ARRIVED, whether
+       the move came from autoplay, a swipe or a dot. polite so it never
+       interrupts, and atomic so the whole position is read as one phrase. -->
+  <div class="carousel-live-region" aria-live="polite" aria-atomic="true">
+    {#if views.length > 0}Slide {activeSlideIndex + 1} of {views.length}{/if}
+  </div>
 </div>
 
 <style>
@@ -272,4 +287,19 @@
     }
 
   } */
+
+  /* Visually hidden but readable by assistive tech -- the standard clip pattern
+     rather than display:none, which would remove it from the a11y tree. */
+  .carousel-live-region {
+    position: absolute;
+    width: 1px;
+    height: 1px;
+    margin: -1px;
+    padding: 0;
+    overflow: hidden;
+    clip: rect(0 0 0 0);
+    clip-path: inset(50%);
+    white-space: nowrap;
+    border: 0;
+  }
 </style>
--- a/tests/carousel.test.ts
+++ b/tests/carousel.test.ts
@@ -66,3 +66,28 @@ test.describe('Carousel', () => {
       .evaluate((el) => Promise.all(el.getAnimations().map((a) => a.finished)));
   });
 });
+
+test.describe('Carousel backward compatibility and announcements', () => {
+  test('a wrapper that declares `properties` still receives the bag', async ({ page }) => {
+    await page.goto('/components/carousel');
+    const legacy = page.getByTestId('carousel-legacy-demo');
+    await expect(legacy).toBeVisible();
+    // Spreading the bag was the fix, but the pre-existing wrapper shape has to
+    // keep working or the fix is a breaking change to a published component.
+    await expect(legacy.getByTestId('carousel-legacy-slide')).toHaveText(
+      'legacy wrapper still receives its bag'
+    );
+  });
+
+  test('the active slide position is announced to assistive technology', async ({ page }) => {
+    await page.goto('/components/carousel');
+    const carousel = page.getByTestId('carousel-manual-demo');
+    const live = carousel.locator('[aria-live="polite"]');
+    await expect(live).toHaveText('Slide 1 of 3');
+
+    await page.getByTestId('carousel-manual-dot-2').click();
+    // The dots say where they GO; without this region nothing says where the
+    // carousel ARRIVED, including for autoplay and swipe.
+    await expect(live).toHaveText('Slide 2 of 3');
+  });
+});
--- a/docs/Carousel.md
+++ b/docs/Carousel.md
@@ -16,17 +16,41 @@ A slideshow that renders Svelte components as slides, with opt-in autoplay (`aut
 
 ## Props
 
-| Prop             | Type             | Required | Default | Description                                                                                                                                                                                                                |
-| ---------------- | ---------------- | -------- | ------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| views            | `CarouselView[]` | Yes      | `-`     | Array of CarouselView objects. Each contains a Svelte Component reference and optional properties to pass to it.                                                                                                           |
-| autoplay         | `boolean`        | No       | `false` | When true, the carousel automatically advances to the next slide at the autoplayInterval rate.                                                                                                                             |
-| autoplayInterval | `number`         | No       | `1000`  | Time in milliseconds between automatic slide transitions. Only used when autoplay is true.                                                                                                                                 |
-| showDots         | `boolean`        | No       | `false` | When true, shows dot indicators below the carousel for direct slide navigation.                                                                                                                                            |
-| isScrollableLast | `boolean`        | No       | `false` | When true, allows scrolling past the last slide (wrapping to the first) and before the first slide (wrapping to the last).                                                                                                 |
-| classes          | `string`         | No       | `-`     | CSS class string applied to the component's top-level element. Useful for theming — define classes with CSS variable overrides (e.g., `.btn-primary { --button-color: #0070f3; }`) and pass them to create variant styles. |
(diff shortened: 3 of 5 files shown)


tests/carousel.test.ts
```diff
@@ -66,3 +66,28 @@ test.describe('Carousel', () => {
       .evaluate((el) => Promise.all(el.getAnimations().map((a) => a.finished)));
   });
 });
+
+test.describe('Carousel backward compatibility and announcements', () => {
```
Tara-ag, line 70: 💬 SUGGESTION: The new backward-compatibility and announcement tests navigate the demo page and select by demo-specific test ids, coupling the regression test to demo-page markup. Consider adding an isolated Vitest/component test that imports `Carousel` directly and renders it with a stub slide to verify both the spread behavior and the legacy `properties` bag.

Tara-ag (commented):
Reviewed file-by-file. The PR correctly restores the `properties` bag alongside the spread, which avoids a breaking change to the published `Carousel` contract, and adds the requested `aria-live` region for slide announcements. Tests cover both behaviors.

Two additional findings could not be attached to changed lines (the relevant code is unchanged in this diff):

- ⚠️ MAJOR: `src/lib/Carousel/Carousel.svelte` dot indicators still lack `aria-current`/`aria-pressed`/`aria-selected` to indicate the active slide. Add `aria-current={activeSlideIndex === index ? 'true' : undefined}` to each dot.
- 💡 MINOR: `src/lib/Carousel/Carousel.svelte:110` has a typo — `handeMouseUp` should be `handleMouseUp`.

See the inline comment on `tests/carousel.test.ts` for the third finding.

src/lib/Carousel/Carousel.svelte
```diff
@@ -157,7 +157,15 @@
       <div class="slidesDiv" bind:this={slidesDiv}>
         {#each views as view, index (index)}
           <div class="current-slide">
-            <view.component {...view.properties} />
+            <!-- Spread is the real fix: every component in this library takes named
+                 props, so passing the bag under a single `properties` key meant a
+                 normal slide rendered empty. `properties` is ALSO still passed so
+                 that a consumer who built a purpose-made wrapper around the old
+                 behaviour keeps working -- that wrapper was the only shape this
+                 component was usable with before, and breaking it would force a
```
Tara-ag, line 165: ⚠️ MAJOR: Each dot indicator is focusable and announces its destination, but there is no `aria-current`, `aria-pressed`, or `aria-selected` attribute indicating which dot corresponds to the active slide. Screen-reader users can hear where each dot goes, but cannot tell which slide is currently shown. Add `aria-current={activeSlideIndex === index ? 'true' : undefined}` to each dot div.

Tara-ag:
💡 MINOR: `src/lib/Carousel/Carousel.svelte:110` — The mouse-up event handler is named `handeMouseUp` instead of `handleMouseUp`. The typo is consistent across declaration and add/remove calls, so it functions correctly, but it degrades readability and invites future mistakes. Rename `handeMouseUp` to `handleMouseUp` in the function declaration and in both `addEventListener`/`removeEventListener` calls.

Tara-ag (commented):
Review complete. The PR restores backward compatibility correctly and adds the requested aria-live region. One MAJOR accessibility issue remains (active dot lacks programmatic state), plus one MINOR typo and one SUGGESTION about test isolation.

Tara-ag:

The PR correctly restores backward compatibility by passing `properties={view.properties}` alongside the spread and adds the requested `aria-live` region for slide announcements, with tests covering both. However, one MAJOR accessibility issue remains: dot indicators lack programmatic state for the active slide. Two smaller items (a typo and test coupling) are also noted.

- ⚠️ **MAJOR**: Active carousel dot lacks programmatic current state — `src/lib/Carousel/Carousel.svelte:177`
  Each dot indicator is rendered as a focusable button with an aria-label, but there is no aria-current, aria-pressed, or aria-selected attribute indicating which dot corresponds to the active slide. Screen-reader users can hear where each dot goes, but cannot tell which slide is currently shown. This was raised as a MAJOR thread on #477 and remains unaddressed.
- 💡 **MINOR**: Typo in mouse-up handler name — `src/lib/Carousel/Carousel.svelte:110`
  The mouse-up event handler is named `handeMouseUp` instead of `handleMouseUp`. The typo is consistent across declaration and add/remove calls, so it functions correctly, but it degrades readability and invites future mistakes.
- 💬 **SUGGESTION**: Regression tests are coupled to demo-page markup — `tests/carousel.test.ts:70`
  The new backward-compatibility and announcement tests navigate `/components/carousel` and select elements by demo-specific test ids. This couples the regression test to the demo page layout and makes failures hard to attribute to the component versus the demo. The project convention prefers isolated component mounts for behavioral regression tests.

murdore:
## Correction: the break is NOT on npm, and that changes what this PR is for

I opened this PR saying the spread-only Carousel "is now in 2.136.4" and that consumers were already affected. **That was wrong, and I only caught it by checking the registry instead of the repo.**

```
npm dist-tag latest        2.136.1   (published 2026-09-01T07:30:49Z)
git tags on origin         2.136.0 … 2.136.4
GitHub Releases            2.136.0 … 2.136.4
```

`npm view @juspay/svelte-ui-components version` → **2.136.1**, and unpacking that tarball shows the *old* Carousel still shipping:

```svelte
155:              <view.component properties={view.properties} />
157:              <view.component />
```

2.136.2, 2.136.3 and 2.136.4 exist as version commits, tags and GitHub Releases, but **none of them were ever published**. The `Publish to NPM` step has failed on every release since 07:30Z:

```
npm error code E404
npm error 404 Not Found - PUT https://registry.npmjs.org/@juspay%2fsvelte-ui-components
```

A 404 on `PUT` for a scoped package is npm's auth-failure signature, not a missing package — the token no longer has publish rights. (Filed in more detail on #474, which is the fix for exactly this.)

## Why this makes the PR more useful, not less

The compatibility window I was trying to protect **has not opened yet**. Every consumer is still on 2.136.1 with the old contract. If this lands before publishing is restored, the spread change and its compatibility shim reach npm in the same version, and no release ever ships the contract break on its own.

So the argument for the shim is now weaker on urgency and stronger on cleanliness: nothing to rescue, but also no reason to publish a breaking change when a one-line shim makes it additive. The clean-break alternative — delete `properties={view.properties}`, add a `BREAKING CHANGE:` footer, cut 3.0.0 — is still a live option and still yours to choose; it is just no longer a choice made under pressure.

## Verification on this head (`173f2fd`)

`tests/carousel.test.ts` — **5 passed**, run on a private `PW_PORT` so it could not reuse another checkout's preview server:

```
✓ spreads each slide view.properties onto its component, not nested
✓ clicking a dot navigates to that slide, rendering its own spread properties
✓ pressing Enter on a focused dot navigates to that slide
✓ a wrapper that declares `properties` still receives the bag
✓ the active slide position is announced to assistive technology
```

Tests 1 and 4 pin both directions of the contract, so neither shape can be dropped silently.

## About the red `build` check

Not this branch. `Deploy to GitHub Pages` is broken on `release` itself and every open PR inherits it — root-caused and fixed in **#485**. This branch goes green once that lands.

tests/carousel.test.ts
murdore: Declined, for the same reason given on PR #477's identical suggestion: every Playwright spec in this repo (tests/carousel.test.ts included, lines 70-93 on release 3.1.0) navigates a demo route and selects by data-pw, and the codebase's Vitest tests cover only pure-logic modules, never a mounted component. There's no isolated-component-test precedent to be consistent with, so this stays as a Playwright spec against the demo page.

src/lib/Carousel/Carousel.svelte
murdore: Still open: on release 3.1.0, the dots (Carousel.svelte:181-197) carry `role="button"`, `tabindex="0"`, and an `aria-label` naming their destination, but no `aria-current`, `aria-pressed`, or `aria-selected` attribute marking which one is active. This PR's aria-live region (Carousel.svelte:201-207) announces the new slide by number after a move, but doesn't help a screen-reader user query which dot is currently active. Adding `aria-current={activeSlideIndex === index ? 'true' : undefined}` to each dot div would close this.

murdore: Fixed in juspay/svelte-ui-components#502 (commit 64f91c5be): each dot now carries aria-current="true" when it is the active slide and nothing otherwise (src/lib/Carousel/Carousel.svelte, the dot element's attributes); pinned by the new 'the active dot is exposed as aria-current' test in tests/carousel.test.ts.

PR #486, opened 2026-09-01
fix(Input): give helper text its own color, bring error text to AA contrast

### LINKED ISSUE #481: Input: helper text uses the error colour, and that colour fails WCAG AA (4.06:1)
Two related defects in `Input`, both on current `release`, surfaced while recording accessibility evidence for #480. Filing rather than fixing: changing either default is a visual change every consumer inherits, so it wants a design decision, not a silent patch.

## 1. Helper text is painted the error colour

`src/lib/Input/Input.svelte`:

```css
.info-message {
  color: var(--input-info-msg-text-color, #fa1405);   /* line 601 */
}
```

`#fa1405` is the same value `--input-error-msg-text-color` defaults to (lines 495/499/526). So `infoMessage` — neutral guidance like *"Use your company address, not a personal one"* — renders in exactly the red of a validation failure.

Two problems with that:

- **It misreports state.** A field with helper text and no error looks like a field that failed validation.
- **It is colour-alone signalling.** Once helper text and error text are the same colour, colour is the *only* thing distinguishing them, which is what WCAG 1.4.1 (Use of Color) exists to prevent.

This got sharper with #480, which now puts `infoMessage` into the field's `aria-describedby`. A screen-reader user hears the helper text as description; a sighted user sees it as an error. The two experiences disagree.

## 2. That red fails AA on white

| Foreground | Background | Ratio | AA (4.5:1) |
|---|---|---|---|
| `#fa1405` | `#ffffff` | **4.06:1** | ❌ |
| `#fa1405` | `#f6f8fa` | **3.82:1** | ❌ |

Both messages render at `font-size: 12px`, so the large-text 3:1 allowance does not apply — 4.5:1 is required.

This one matters more than the first, because it means **the error message itself is not readable to AA**. The text a user most needs when a form rejects them is the text least likely to be legible.

## Suggested direction

1. Give `--input-info-msg-text-color` a neutral default (a muted foreground, not a status colour), so helper text stops asserting failure. Consumers wanting red keep setting the token.
2. Darken the error default until it clears 4.5:1 on white. `#d1130a` is 5.05:1 and stays visibly red; `#c5120a` is 5.6:1 with more headroom for off-white surfaces.

Happy to raise the PR for either or both once there is a call on the values — I did not want to change a default that every consumer inherits without one.

## How to reproduce

`/components/input` on this branch has both cases side by side: a `forceError` field with `infoMessage`, and a valid field with only `infoMessage`. Both render red. Contrast figures above are computed from the WCAG relative-luminance formula against the two backgrounds the demo uses.

murdore:
Closes #481.

Two related defects in `Input`, both filed by a prior accessibility-audit pass rather than silently patched — changing either default is a visual change every consumer inherits.

## 1. Helper text was painted the error color

`.info-message` defaulted to the exact same red as `.error-message` (`#fa1405`), so `infoMessage` — neutral guidance like *"Use your company address, not a personal one"* — rendered in the color of a validation failure. Two problems: it misreports state (a field with helper text and no error looked like it had failed), and it was color-alone signalling — WCAG 1.4.1 exists to prevent exactly this.

`--input-info-msg-text-color` now defaults to `#52525b` — a neutral tone already used elsewhere in this library (ChatToolStatus/ThinkingIndicator's muted status text) — at **7.73:1** on white.

## 2. The shared red failed AA on its own

| Foreground | Background | Ratio | AA (4.5:1) |
|---|---|---|---|
| `#fa1405` (old) | `#ffffff` | 4.06:1 | ❌ |
| `#c5120a` (new) | `#ffffff` | **6.06:1** | ✅ |

Both messages render at 12px, so the large-text 3:1 allowance doesn't apply. `--input-error-msg-text-color` now defaults to `#c5120a`.

I checked whether the existing label color (`#637c95`) could just be reused as the new neutral default — it can't, it's itself only 4.33:1, below AA — before picking `#52525b` from this library's own existing muted-text convention instead.

## Scope: also fixed in InputButton

`InputButton.svelte` independently hardcodes the same two shared variable names (`--input-error-msg-text-color`, `--input-info-msg-text-color`) plus its own `--inputbutton-external-error-color` — all previously `#fa1405` too, in a separate file. Fixed there as well, so the same token means the same color in both components rather than leaving InputButton on the old failing default while Input moved on.

`--input-error-msg-text-color`'s fallback is spelled out in five places across `Input.svelte` (error text, error border ×2, mandatory asterisk, char-count-at-limit) and had to be updated consistently in all of them, or the border and text would visibly mismatch.

## Testing

Added a `data-pw`/`testID` hook to Input's info-message div, mirroring the existing error-message one (previously only the error message had one).

New coverage in `tests/input-helper-error-colors.test.ts`: asserts the error and helper messages render in different, exact expected colors on both `Input` (the existing `forceError` + `infoMessage` demo instance) and `InputButton` (a new demo section pairing the `error` prop with `infoMessage`).

Ran the full existing suite (342 specs) against these changes — all pass, no regressions from the shared-token update.

Verified with real Playwright video recordings, each decoded with `ffprobe` (valid VP8, real non-zero frame counts, full `ffmpeg` decode pass with no corruption) and visually inspected via extracted frames — video proof to follow as a PR comment.

## Test plan
- [x] `pnpm run check` — 0 errors
- [x] `pnpm run lint` — clean
- [x] `tests/input-helper-error-colors.test.ts` — 2/2 passing
- [x] Full suite (342 specs) — all passing, no regressions
- [x] Video-verified: error and helper text render as clearly distinct colors in both `Input` and `InputButton`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error and helper-text colors for better readability and WCAG AA contrast.
  * Standardized messaging colors across Input and InputButton components.
  * Added clearer visual distinction between errors and neutral guidance.

* **Documentation**
  * Updated color references, contrast details, and accessibility guidance.
  * Documented live-region announcements and message contrast requirements.

* **Tests**
  * Added regression coverage to verify messaging colors and accessibility contrast.


murdore:
Video proof for the color fix — real Playwright recordings of the two new tests (verified: `file`/`ffprobe`/`ffmpeg` decode, real non-zero frame counts, visually inspected via extracted frames).

1. `Input` — the `input-announced-error` demo instance: error text now `#c5120a`, helper text now `#52525b`, clearly distinct.
2. `InputButton` — the new "External error + helper text" demo: same distinction, via the shared tokens + `--inputbutton-external-error-color`.

[input-error-helper-colors.webm](https://github.com/user-attachments/assets/e0614cea-ca22-4a38-870a-9068230a28df)
[inputbutton-error-helper-colors.webm](https://github.com/user-attachments/assets/4a298ca6-922d-4e9c-89a4-ef2150d15069)

Tara-ag (approved):
Approved. The PR fixes an accessibility defect by changing default CSS custom property values while preserving the property names, prop signatures, exports, and event contracts. All changed tokens remain overridable by downstream consumers. Tests and docs are updated accordingly. No blocking issues found.

PR #492, opened 2026-09-01
fix(Accordion): link trigger to its panel with aria-controls

murdore:
## The accessibility defect

Accordion's built-in trigger had `role="button"` and `aria-expanded`, but no `aria-controls` — and the trigger/panel are **siblings**, not ancestor/descendant:

```svelte
<div class="accordion-trigger" role="button" aria-expanded={expand}>...</div>
<div class="accordion">...</div>
```

A screen-reader user could hear *"expanded"* without any semantic connection identifying **which region** expanded or offering a linked destination to reach it. This was surfaced by Yama while reviewing the docs-accuracy PR, verified directly against source, and confirmed pre-existing (PR #491 did not touch `Accordion.svelte`).

## Fix

- Generates a unique per-instance panel id via `$props.id()`.
- Adds `aria-controls={effectivePanelId}` to the built-in trigger.
- Gives the collapsible panel `id={effectivePanelId}`.
- Adds optional `panelId` for callers that need a known id to reference externally; the generated default keeps each instance unique automatically.
- Adds `panel-id` → `panelId` to the web-component wrapper and documents both Svelte + HTML APIs.

## Tests

New `tests/accordion-aria-controls.test.ts` proves all four parts of the contract:
1. `aria-controls` resolves to the panel's **actual** id, not merely a present-but-dangling attribute.
2. `aria-expanded` tracks the real `.expanded` panel state.
3. Two accordion instances generate distinct ids (no collision).
4. Explicit `panelId` reaches both the panel and the trigger reference.

## Verification

- [x] `pnpm run check` — 0 errors
- [x] `pnpm run lint` — clean
- [x] Full serial Playwright suite — 354/354 passing
- [x] Real video proof: 4 test clips, individually verified as VP8/decode-clean, concatenated with **160/160 frame parity**, frame-extracted and visually inspected. Video follows as a PR comment.

## Summary by CodeRabbit

- **New Features**
  - Added optional custom panel IDs for Accordion components.
  - Accordion triggers now reference their panels with `aria-controls`.
  - Automatically generated unique panel IDs are used when no ID is provided.
  - Added support for the `panel-id` Web Component attribute.

- **Accessibility**
  - Improved trigger-to-panel associations for assistive technologies.
  - Added examples demonstrating accessible Accordion usage.

### DIFF
--- a/src/lib/Accordion/Accordion.svelte
+++ b/src/lib/Accordion/Accordion.svelte
@@ -9,8 +9,17 @@
     triggerClasses,
     classes,
     testId,
-    disabled = false
+    disabled = false,
+    panelId
   }: AccordionProperties = $props();
+
+  /* The trigger and the panel are siblings, not ancestor/descendant, so nothing in the
+     markup tells assistive technology which region the trigger's aria-expanded refers to.
+     A generated id keeps that link automatic -- consumers pass no id and get correct
+     wiring -- while `panelId` lets a caller supply their own when they need to reference
+     the panel from elsewhere too. */
+  const uid = $props.id();
+  const effectivePanelId = $derived(panelId ?? `accordion-panel-${uid}`);
 
   function handleTriggerClick(): void {
     if (disabled) {
@@ -28,6 +37,7 @@
     role="button"
     tabindex={disabled ? -1 : 0}
     aria-expanded={expand}
+    aria-controls={effectivePanelId}
     aria-disabled={disabled}
     onclick={handleTriggerClick}
     onkeydown={(event) => {
@@ -42,6 +52,7 @@
 {/if}
 
 <div
+  id={effectivePanelId}
   class="accordion {classes ?? ''}"
   class:expanded={expand}
   data-pw={typeof testId === 'string' ? testId : null}
--- a/tests/accordion-aria-controls.test.ts
+++ b/tests/accordion-aria-controls.test.ts
@@ -1,0 +1,64 @@
+import { expect, test } from '@playwright/test';
+
+test.describe('Accordion — the trigger is linked to the panel it controls', () => {
+  // The trigger and the collapsible panel are siblings, not ancestor/descendant, so
+  // without aria-controls nothing in the markup says WHICH region the trigger's
+  // aria-expanded describes. A screen-reader user hears "expanded" with no way to jump
+  // to what expanded. The trigger already had role="button" and aria-expanded, which is
+  // exactly the pattern that makes the missing link conspicuous.
+  test('aria-controls points at the panel element, which really carries that id', async ({
+    page
+  }) => {
+    await page.goto('/components/accordion');
+
+    const panel = page.getByTestId('accordion-linked');
+    const trigger = page.locator('.accordion-trigger').first();
+
+    const controls = await trigger.getAttribute('aria-controls');
+    expect(controls, 'the trigger must reference a panel').toBeTruthy();
+
+    // The reference has to resolve to the actual panel, not merely be present.
+    await expect(panel).toHaveAttribute('id', String(controls));
+  });
+
+  test('aria-expanded tracks the real open state on the linked panel', async ({ page }) => {
+    await page.goto('/components/accordion');
+
+    const trigger = page.locator('.accordion-trigger').first();
+    const panel = page.getByTestId('accordion-linked');
+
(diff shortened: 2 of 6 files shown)


murdore:
Video proof — real Playwright recording of all 4 new accessibility tests.

Verified per source clip (`file` / `ffprobe` / full `ffmpeg` decode), then concatenated with **160/160 frame parity** and frame-inspected. Demonstrates:

1. `aria-controls` resolves to the real panel `id`.
2. `aria-expanded` changes only with the panel's actual open state.
3. Generated ids don't collide across two instances.
4. An explicit `panelId` reaches both the panel and trigger reference.

[accordion-aria-controls.webm](https://github.com/user-attachments/assets/219e3bb9-c416-4061-bf4a-d79027ca20cf)

tests/accordion-aria-controls.test.ts
```diff
@@ -0,0 +1,64 @@
+import { expect, test } from '@playwright/test';
+
+test.describe('Accordion — the trigger is linked to the panel it controls', () => {
+  // The trigger and the collapsible panel are siblings, not ancestor/descendant, so
+  // without aria-controls nothing in the markup says WHICH region the trigger's
+  // aria-expanded describes. A screen-reader user hears "expanded" with no way to jump
+  // to what expanded. The trigger already had role="button" and aria-expanded, which is
+  // exactly the pattern that makes the missing link conspicuous.
```
Tara-ag, line 8: 💡 MINOR: This test navigates to `/components/accordion` and relies on demo-page `testId` values (`accordion-linked`, `accordion-custom-panel`) and the order of trigger snippets in that page. That couples component-contract tests to the demo route layout; a future docs/demo refactor can break these tests without touching the component. Prefer an isolated fixture route or component harness and select by component-internal classes (`.accordion`, `.accordion-trigger`) instead.

src/lib/Accordion/Accordion.svelte
```diff
@@ -42,6 +52,7 @@
 {/if}
```
Tara-ag, line 52: 💬 SUGGESTION: The trigger now correctly references the panel via `aria-controls`. To fully align with the WAI-ARIA Accordion pattern, consider also giving the trigger an `id` and the panel `role="region" aria-labelledby="{triggerId}"`, so screen readers announce the panel as a named region governed by the button.

Tara-ag (commented):
Reviewed file-by-file. The change adds a backward-compatible `panelId` prop, wires `aria-controls` between the Accordion trigger and panel, maps the prop through the web-component wrapper, and documents it. No blocking issues found. Two non-blocking notes posted: one about test coupling to the demo page, and one suggestion to complete the ARIA region contract with `role="region"`/`aria-labelledby`.

tests/accordion-aria-controls.test.ts
murdore: Declined on repo convention rather than on the merits of isolation. Every Playwright spec in this repository (80+ files under tests/) navigates a demo route and selects by data-pw; the only Vitest suites cover pure-logic modules, never a mounted Svelte component, so there is no isolated-fixture precedent for an Accordion harness to follow. tests/accordion-aria-controls.test.ts on release 3.1.0 follows that convention deliberately. If the demo page changes its trigger order or testIds the spec will fail loudly rather than silently, which is the trade the repo has accepted everywhere else.

src/lib/Accordion/Accordion.svelte
murdore: Still open on release: the trigger (Accordion.svelte:34-51) has no id and the panel (lines 54-60) has no role="region" or aria-labelledby, unchanged since commit ae2d588 (PR #492) added aria-controls. This was flagged as a suggestion rather than a blocking issue, so it's a legitimate follow-up rather than a regression. Closing it needs a trigger id plus role="region" aria-labelledby={triggerId} on the panel.

murdore: Fixed in juspay/svelte-ui-components#502 (commit 64f91c5be): the trigger now carries id=`${panelId}-trigger` and the panel role="region" aria-labelledby pointing at it, set only when a trigger is rendered (src/lib/Accordion/Accordion.svelte); pinned by the new 'the panel is a region named by its trigger' test in tests/accordion-aria-controls.test.ts.

PR #493, opened 2026-09-01
feat(toolbar)!: render the default back control as a button with an inline icon

murdore:
## Why

`Toolbar` was the only component in the library whose **default** rendering made a network request: `backIcon` defaulted to `https://sdk.breeze.in/gallery/icons/back.svg`. A default that fetches hangs or 404s offline, cannot follow `currentColor`, and pins a first-party component to a CDN path this repo does not control. Twelve other components already inline their SVGs.

The control itself was `<div role="button" tabindex="0">`: Enter and Space did nothing unless the consumer wired `onkeydown`, and the only accessible name was the image's `alt`. (Surfaced by Yama on lighthouse #7298.)

## What changes

- Default back control is a native `<button aria-label={backLabel}>` with an inline `currentColor` chevron. Keyboard activation and the accessible name come for free; the icon is decorative.
- New prop `backLabel` (default `'Back'`; `back-label` on `<sui-toolbar>`), token `--toolbar-back-icon-color` (default `inherit`), and a keyboard-only focus ring via `--toolbar-back-button-focus-outline` / `-offset`.

## What does not change

- `backIcon={null}` (or `''`) still renders **no** back control — the contract the existing test pins.
- A consumer-supplied `backIcon` URL still renders as an ``.
- Every existing token keeps its value; the control's box is identical to the div it replaces.
- The title's typography tokens (`--toolbar-text-font-size`/`-weight`, 18px/normal) are untouched — that is the correct library pattern and overridable per consumer.

## Verification

- `pnpm run lint` — 0 new violations · `pnpm run check` — 0 errors (4 pre-existing warnings, none in Toolbar) · prettier clean.
- Playwright `tests/toolbar-page-header.test.ts` + `toolbar-content-geometry.test.ts`: **9/9**, including two new tests (consumer-supplied image still renders ``; Enter and Space each fire `onbackClick`).
- **Negative control:** restoring the old `<div>` markup fails exactly the two assertions that describe the change — tag at `:23`, activation count at `:70` — and nothing else. Restored; 9/9 again.

Consumer follow-up: lighthouse bumps to the published version in BZ-5929.

## Summary by CodeRabbit

- **New Features**
  - Toolbar back controls now use an inline icon by default, with no external network request.
  - Added accessible back-button labeling via the `backLabel` property and `back-label` attribute.
  - Back controls are now native buttons with keyboard support and visible focus styling.
  - Custom back icons remain supported.

- **Documentation**
  - Updated Toolbar usage guidance, accessibility details, and CSS customization options.


src/lib/Toolbar/Toolbar.svelte
```diff
@@ -4,7 +4,8 @@
   let {
     showBackButton = true,
     text,
-    backIcon = 'https://sdk.breeze.in/gallery/icons/back.svg',
+    backIcon,
```
Tara-ag, line 7: ⚠️ MAJOR: Removing the default `backIcon` CDN URL changes the rendered DOM for existing consumers who did not pass `backIcon`. Previously they got an ``; now they get a `<button><svg>`. This breaks consumers that styled `.back img`, relied on image-load behavior, or expected the CDN asset. Because this is a published library, changing the default rendered output is a breaking public-API change that needs either a major-version bump + migration note or an opt-in prop to preserve the old default. Consider keeping `backIcon` defaulting to the CDN URL and adding an opt-in `useInlineBackIcon` prop, or marking this as a breaking change.

playwright-report/index.html
Tara-ag, line 1: 💡 MINOR: This PR commits generated Playwright report artifacts (`playwright-report/`). These are build/test outputs that should not be checked into source control; they inflate the repo and will go stale. Please remove this directory from the PR and add `playwright-report/` to `.gitignore` alongside `test-results/`.

src/lib/Toolbar/Toolbar.svelte
```diff
@@ -118,9 +144,16 @@
     align-items: center;
   }
 
-  .back img {
+  .back:focus-visible {
```
Tara-ag, line 157: 💬 SUGGESTION: The default focus ring uses `currentColor`, which can be invisible when the toolbar text color and background are similar. While themable via `--toolbar-back-button-focus-outline`, consider a more robust default such as `outline: 2px solid CanvasText; outline-offset: -2px;` or a double outline so keyboard focus remains visible across consumer themes.

Tara-ag (changes requested):
Blocking: the default `backIcon` change alters the rendered DOM for existing consumers without a major-version bump or migration note. Additionally, generated Playwright report artifacts are committed. Please address the MAJOR and MINOR findings; the SUGGESTION is optional.

playwright-report/index.html
```diff
@@ -0,0 +1,90 @@
+
```
Tara-ag: 💡 MINOR: Generated Playwright report artifacts are committed. The PR adds `playwright-report/index.html` and nine `.webm` video files under `playwright-report/data/`. These are outputs of the HTML reporter (`reporter: [['html', { open: 'never' }]]`) and should not be tracked in git; they bloat the repository and will become stale on the next test run.

Suggested fix: Remove `playwright-report/` from the PR and add `playwright-report/` to `.gitignore` next to the existing `test-results/` entry.

src/lib/Toolbar/Toolbar.svelte
```diff
@@ -25,10 +30,24 @@
   >
     {#if typeof leftContent === 'function'}
       {@render leftContent()}
-    {:else if showBackButton && typeof backIcon === 'string' && backIcon.length > 0}
-      <div class="back" onclick={onbackClick} {onkeydown} role="button" tabindex="0">
-        <img src={backIcon} alt="Back" />
-      </div>
+    {:else if showBackControl}
```
Tara-ag, line 36: ⚠️ MAJOR: Empty `backLabel` leaves the back button without an accessible name. The button uses `aria-label={backLabel}` and `backLabel` is typed as an optional `string` defaulting to `'Back'`. A consumer can pass `backLabel=""`, which produces `<button aria-label="">` and removes the accessible name. Because the icon is decorative (empty alt / aria-hidden), the button then has no name for screen-reader users.

Suggested fix: Fall back to the default label when an empty string is supplied: `aria-label={backLabel || 'Back'}`. Alternatively type `backLabel?: string | null` and use `aria-label={backLabel ?? 'Back'}`.

src/lib/Toolbar/Toolbar.svelte
Tara-ag, line 7: ⚠️ MAJOR: Default `backIcon` rendering changed without breaking-change marker. The PR removes the default value of `backIcon` (was a CDN URL, now undefined) and changes the rendered element from a `<div role="button"></div>` to a native `<button>` with an inline `<svg>`. While this is the stated improvement, it is a change to a public prop default and to the produced DOM. Downstream consumers that selected `.back img`, asserted the old CDN src, or themed the image via existing tokens may break. The commit is titled `feat(toolbar): ...`, which cuts a minor release per the repo's standard-version config; a breaking default/DOM change should be `feat(toolbar)!:` or carry an explicit migration note.

Suggested fix: Either (a) keep the change and mark it breaking by renaming the commit/PR title to `feat(toolbar)!: ...` and add a `BREAKING CHANGE:` footer describing the DOM/default change, or (b) add an opt-in prop (e.g. `backVariant`) that preserves the old image default while allowing consumers to opt into the inline button.

Tara-ag (changes requested):
Yama review complete. Two MAJOR issues and one MINOR issue need resolution before merge:

1. **MAJOR — Backward compatibility:** Removing the default `backIcon` CDN URL and switching the default DOM from `<div role="button"></div>` to `<button><svg/></button>` is a public contract change. With the current `feat(toolbar):` title this will ship as a minor release, breaking downstream consumers that select `.back img`, assert the old src, or theme the image. Please either mark this as a breaking change (`feat(toolbar)!:` + `BREAKING CHANGE:` footer) or provide an opt-in prop preserving the old default.

2. **MAJOR — Accessibility:** `backLabel=""` produces an unnamed button because `aria-label={backLabel}` is bound directly. Fall back to `'Back'` when the prop is empty/null.

3. **MINOR — Repo hygiene:** The generated `playwright-report/` directory (HTML report + `.webm` videos) should not be committed. Remove it and add `playwright-report/` to `.gitignore`.

Everything else (new `backLabel` prop, web-component attribute mapping, docs, tests, focus styles) looks good.

Tara-ag (changes requested):

Tara-ag:

Severity counts — 🔒 CRITICAL: 1 · ⚠️ MAJOR: 1 · 💡 MINOR: 1

The PR introduces a deliberate breaking change to Toolbar's default back control (from CDN  to inline SVG <button>), which is correctly marked with `feat(toolbar)!:` and a `BREAKING CHANGE` footer, but the package version remains at `2.136.12`. Per the blocking criteria for published-library backward compatibility, a breaking public-component change without a major-version bump blocks merge. Additionally, the web-component wrapper exposes `onbackClick` as an `Object` prop, which collides with the browser's `on*` event-handler convention and may break callback binding for `<sui-toolbar>` consumers. The docs also reference a `3.0.0` migration note that does not match the unreleased version. Resolve the version mismatch (bump to 3.0.0 or remove the breaking marker) and fix the web-component callback mapping before merging.

- 🔒 **CRITICAL**: Breaking change lacks major-version bump — `package.json:1`
  The commit message and PR body explicitly mark this as a breaking change (`feat(toolbar)!:` and `BREAKING CHANGE:` footer). The default rendering of the back control changes from an `` with a fixed CDN src to a `<button>` with an inline SVG. Downstream consumers that selected `.back img`, asserted the old `src`, or themed the image will break. Per the blocking criteria, a breaking change to a…
- ⚠️ **MAJOR**: Web component callback prop `onbackClick` collides with DOM event listener naming convention — `src/wc/components/Toolbar.wc.svelte:11`
  In the custom-element wrapper, `onbackClick` is declared as `{ type: 'Object' }`. Because it starts with `on`, browsers and Svelte's custom-element runtime treat attributes/properties beginning with `on` as event listeners. This is a known web-component pitfall noted in the bootstrapped standards: "Callback props beginning with 'on' in custom-element wrappers must be explicitly mapped or renamed, …
- 💡 **MINOR**: Docs migration note references unshipped version 3.0.0 — `docs/Toolbar.md:18`
  docs/Toolbar.md says "> **3.0.0 — the default back control changed.**" but package.json is still at 2.136.12 and the PR does not bump the version. If this PR is merged as-is, the migration note will point to a version that does not exist, confusing consumers.

murdore:
Addressed in `15831bb` (amended, single commit):

- **`playwright-report/` committed** — my `git add -A` swept the HTML report and nine `.webm` videos in because the repo never ignored that directory. Removed from the commit; `playwright-report/` is now in `.gitignore` beside `test-results/`. Verified the re-run regenerates the report and `git status` stays clean.
- **Empty `backLabel`** — `aria-label` now comes from a derived `backAccessibleName` that falls back to `'Back'` for empty or whitespace-only values; boundary test added (`backLabel="  "` → `aria-label="Back"`).
- **Unmarked breaking change** — agreed: any real fix changes the default DOM, so this is `feat(toolbar)!:` with a `BREAKING CHANGE:` footer, and `docs/Toolbar.md` opens with the 3.0.0 migration note (what changed, how to size/colour the new icon with the unchanged tokens, how to keep an image via `backIcon`). An opt-in prop that preserves the CDN default would leave the network dependency in place, which is the thing being removed.
- **`box-sizing: content-box`** set explicitly on the button; test added asserting the 48×60 box the div had.
- Focus-ring suggestion: left as `currentColor` — it follows whatever the consumer sets for the toolbar's text colour, and both outline tokens are overridable. Happy to switch if you'd rather default to `CanvasText`.

CodeRabbit's out-of-diff finding is real but **pre-existing**: `Toolbar.wc.svelte` always supplies a `leftContent` snippet wrapping the `left-content` slot, so `<sui-toolbar show-back-button>` never reached the default back control before this PR either. `back-label` is registered for when that wrapper gains slot-presence detection; not fixing it here to keep the diff to the stated change.

Gates on the amended commit: lint 0 new violations · `check` 0 errors · prettier clean · Playwright **11/11**.

Tara-ag (approved):
Reviewed file-by-file. The breaking change is intentional, clearly documented in the commit message, PR description, and docs/Toolbar.md with a migration note. Public API surface is additive only (`backLabel` prop, new CSS variables); the `backIcon` prop type remains `string | null` and `backIcon={null}` still suppresses the control. Web-component parity is maintained with the `back-label` attribute mapping. Accessibility improves: native `<button>`, `aria-label`, keyboard activation, and focus-visible ring. Tests cover the new behavior and the backward-compat edge cases. No secrets, no SSR issues, no leaked subscriptions. Approved.

tests/toolbar-page-header.test.ts
```diff
@@ -11,16 +11,20 @@ test.describe('Toolbar page-header shape', () => {
     const root = page.getByTestId('toolbar-root');
     await expect(root).toBeVisible();
 
-    // The title is still a div carrying the library's own type, and the back control is still
-    // the image it has always been, from the same URL.
+    // The title is still a div carrying the library's own type.
```
Tara-ag, line 14: 💡 MINOR: These tests navigate to the demo page (`/components/toolbar`) and select Toolbar instances by showcase testIds such as `toolbar-root`, `toolbar-custom-back-icon`, and `toolbar-page-header`. Coupling component-contract assertions to the demo page markup means a future refactor of the showcase can break Toolbar's tests even when the component itself is unchanged. Prefer an isolated fixture route or harness that renders Toolbar variants with stable testIds solely for the test contract, and assert against that fixture instead of the demo page.

Tara-ag (commented):
Approved with one minor note on test isolation. The breaking change is correctly marked in the commit message and documented in docs/Toolbar.md; standard-version will pick up the `BREAKING CHANGE` footer for a major bump. Accessibility, SSR safety, and the public prop contract look correct.

tests/toolbar-page-header.test.ts
Tara-ag, line 14: 💡 MINOR: New tests select demo-page testIds instead of isolated fixtures

The newly added Playwright tests navigate to /components/toolbar and select Toolbar instances by showcase testIds such as 'toolbar-root', 'toolbar-custom-back-icon', and 'toolbar-page-header'. Coupling component-contract assertions to the demo page markup means a future refactor of the showcase can break Toolbar's tests even when the component itself is unchanged. The project's recurring standard prefers isolated component harnesses or dedicated fixture routes for component-contract tests.

**Suggestion:** Add a dedicated test fixture route (e.g., `tests/fixtures/ToolbarFixture.svelte` or `/tests/toolbar-fixture`) that renders Toolbar variants with stable testIds solely for the test contract, and assert against those fixtures instead of the demo page. At minimum, move the new assertions added by this PR to such a fixture.

Tara-ag (commented):
Approved with one minor test-isolation note. The breaking change is correctly declared and documented, the public prop contract is preserved, and the accessibility improvement is solid.

Tara-ag (commented):

docs/Toolbar.md
```diff
@@ -15,16 +15,25 @@ content. If `leftContent` snippet is provided, it replaces the default back butt
 <Toolbar />
 ```
 
+> **3.0.0 — the default back control changed.** With no `backIcon`, Toolbar used to render
```
Tara-ag, line 18: 💡 MINOR: Migration note references `3.0.0`, but `package.json` is still `2.136.12`. If this PR does not bump the major version, this note will mislead consumers about which release contains the breaking change. Align the version with the actual release.

src/wc/components/Toolbar.wc.svelte
```diff
@@ -6,6 +6,7 @@
       showBackButton: { type: 'Boolean', reflect: true, attribute: 'show-back-button' },
       text: { type: 'String', reflect: true },
       backIcon: { type: 'String', attribute: 'back-icon' },
+      backLabel: { type: 'String', attribute: 'back-label' },
       classes: { type: 'String' },
       testId: { type: 'String', attribute: 'test-id' },
```
Tara-ag, line 11: ⚠️ MAJOR: Web-component callback prop `onbackClick` collides with DOM event-listener naming. Browser custom-element runtimes treat `on*` properties as event handlers, so assigning a function here may not reach the Svelte component's callback and could break click/keyboard activation for `<sui-toolbar>`. Map it to a non-`on` attribute such as `back-click-handler` and forward it as `onbackClick={props.backClickHandler}`, or dispatch a native `back-click` event.

Tara-ag (changes requested):
Yama review: 1 CRITICAL, 1 MAJOR, 1 MINOR finding. The CRITICAL issue blocks merge until resolved.

Tara-ag:
🔒 CRITICAL: Breaking change lacks major-version bump. The commit is `feat(toolbar)!:` with a `BREAKING CHANGE` footer, yet `package.json` remains at `2.136.12`. Shipping this on the current major line violates the published-component contract and the blocking criteria for public API changes. Either bump the major version to `3.0.0` (and keep the docs note) or remove the breaking-change marker and preserve backward compatibility.

Tara-ag:
🔒 CRITICAL: Breaking change lacks major-version bump

The commit message and PR body explicitly mark this as a breaking change (`feat(toolbar)!:` and `BREAKING CHANGE:` footer). The default rendering of the back control changes from an `` with a fixed CDN src to a `<button>` with an inline SVG. Downstream consumers that selected `.back img`, asserted the old `src`, or themed the image will break.

Per the blocking criteria, a breaking change to a public component's rendered output/props without a major-version bump or documented migration note blocks the PR. The current `package.json` version remains `2.136.12` and the PR targets `release`, so merging would ship a minor/patch release containing a breaking change unless the release automation is confirmed to major-bump before publish.

**Suggested fix:** Either bump the major version to `3.0.0` in `package.json` (and update `docs/Toolbar.md` migration-note reference to match), or remove the breaking-change marker and preserve backward compatibility. If the team intends this for v3, update `package.json` version and ensure the release workflow cuts a major.

Tara-ag (changes requested):
Yama review: 1 CRITICAL, 1 MAJOR, 1 MINOR finding. The CRITICAL issue blocks merge until resolved.

- 🔒 CRITICAL (package.json): Breaking change lacks major-version bump. The commit/PR marks this as breaking, but package.json remains at 2.136.12. Either bump to 3.0.0 or remove the breaking marker.
- ⚠️ MAJOR (src/wc/components/Toolbar.wc.svelte:11): Web-component callback prop `onbackClick` collides with DOM event-listener naming.
- 💡 MINOR (docs/Toolbar.md:18): Migration note references 3.0.0 while package.json is still 2.136.12.

murdore:
On the re-review of `3b0c806` (a rebase onto current `release` + the regenerated `toolbar.png` baseline for the new container-pinned visual suite; the component diff is unchanged from the commit approved at 23:45Z):

- **CRITICAL "package.json not bumped to 3.0.0" — not applicable in this repo.** Every version bump on `release` is a `chore(release): x.y.z` commit authored by `github-actions[bot]` from standard-version (`"standard-version"` config in `package.json`; the last four bumps are 2.136.9 → .12, all bot-authored). Contributors never edit the version: a manual `3.0.0` would make standard-version compute the *next* version from 3.0.0 and, with the `BREAKING CHANGE:` footer, could publish **4.0.0**. The footer is the mechanism — as the earlier pass on this PR put it, "standard-version will pick up the `BREAKING CHANGE` footer for a major bump." The docs note names 3.0.0 because that is what the footer produces.
- **MAJOR `onbackClick` on `<sui-toolbar>` — pre-existing, untouched.** `src/wc/components/Toolbar.wc.svelte:12–13` carried `onbackClick` and `onkeydown` before this PR (see `origin/release`). This PR adds only `backLabel`. Worth its own change together with CodeRabbit's finding that the wrapper always supplies a `leftContent` snippet, which is what actually stops `<sui-toolbar>` reaching the default control; not folding an unrelated wc contract change into a component fix.
- **MINOR test isolation** — the new tests follow the file's existing pattern (`toolbar-page-header.test.ts` already selected `toolbar-root` / `toolbar-page-header` on the demo route before this PR). A fixture route for Toolbar is a reasonable follow-up for the whole file, not a per-assertion split.
- **MINOR docs version** — same as the first point.

Gates on `3b0c806`: build, checks, **visual** (92/92, only `toolbar.png` regenerated — 1,577 px: the chevron glyphs and the two new demo instances), CodeQL, CodeRabbit all green.

PR #494, opened 2026-09-01
ci: make the Playwright suite an actual merge gate

murdore:
## What

Removes `continue-on-error: true` from the Playwright step in `ci.yml`, so a failing integration spec now fails the PR. Adds `retries: 2` under CI only.

## Why now

The step has been advisory since it was introduced, carrying this note:

> flip continue-on-error off once a genuine baseline is triaged down to zero

**That baseline was never real.** Every run of the job used to fail every spec with `browserType.launch: Executable doesn't exist` — the browser-install step did not exist. The "~130 failing specs" that justified `continue-on-error` was an artifact of a missing binary, not a regression count.

## Correction to this PR's original justification

I first justified this with "11 consecutive green runs across 8 branches", read from the **step's** conclusion in the jobs API. **That reasoning was wrong, and this PR's own first CI run disproved it.**

Turning the gate on immediately surfaced a genuinely failing test — `splitinput-autofill-distribute.spec.ts`, failing all 3 attempts (353 passed / 1 failed). Not a flake: it reproduces deterministically in the Linux Playwright container and passes on macOS. A real platform-dependent `SplitInput` bug that advisory mode had been hiding all along.

That is the strongest argument for this change, so the correction stays visible rather than being quietly edited away.

Fixed in #497, now merged; this branch is rebased on it.

## Proof the gate actually gates

Verified end to end with a temporary deliberately-failing spec (not committed):

```
EXIT CODE = 1
  1 failed
 ELIFECYCLE  Command failed with exit code 1
```

A non-zero exit fails the step, which now fails the job.

## Proof retries don't launder real failures

`retries: 2` could be read as a way to turn red suites green, so I checked. The same deliberately-failing spec under `CI=true`:

```
EXIT CODE WITH CI=true (retries=2) = 1
  ...-retry1/video.webm
  ...-retry2/video.webm
  1 failed
```

Retried twice, still failed. Independently confirmed by the real `splitinput` failure above, which also survived all 3 attempts. Playwright reports a failed-then-passed test as **flaky** — a status distinct from **passed** — so the signal stays visible in the uploaded report while only reproducible failures block a merge. Local runs keep `0` retries: a flake you can reproduce is a flake you can fix.

## Risk

This can block merges. That is the intent, and it earned its keep on its first run by catching a real bug.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,17 +40,20 @@ jobs:
       - name: Install Playwright browsers
         run: pnpm exec playwright install --with-deps chromium
 
-      # Not a merge gate yet: every prior run of this job failed every single
-      # spec with "browserType.launch: Executable doesn't exist" (verified
-      # 2026-08-31 via the raw job log) -- this step was missing entirely, so
-      # the ~130-failing-specs baseline this comment used to cite was never a
-      # real regression count, just however many specs don't need a browser
-      # launch to reach their own assertion. Real triage starts now that the
-      # job can actually run. Runs on every PR so failures are visible and
-      # trend-trackable; flip continue-on-error off once a genuine baseline
-      # is triaged down to zero.
-      - name: Integration tests (playwright) — advisory, not blocking
-        continue-on-error: true
+      # A merge gate as of this commit. History, so the next reader doesn't have
+      # to reconstruct it: this step was advisory because every run of the job
+      # failed every spec with "browserType.launch: Executable doesn't exist" --
+      # the browser-install step above simply didn't exist. The "~130 failing
+      # specs" that justified continue-on-error was therefore never a regression
+      # count; it was an artifact of a missing binary. Once the install landed,
+      # the suite went green and has stayed green for 11 consecutive runs across
+      # 8 different branches (verified 2026-09-02 by reading this step's own
+      # conclusion in the jobs API, not the job's -- which continue-on-error
+      # forces to "success" regardless and so proves nothing).
+      #
+      # That is the "genuine baseline triaged to zero" the old comment was
+      # waiting on, so the gate is now real: a failing spec fails the PR.
+      - name: Integration tests (playwright)
         run: pnpm run test:integration
 
       - name: Upload Playwright report
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -18,6 +18,13 @@ const config: PlaywrightTestConfig = {
   testDir: 'tests',
   testMatch: /(.+\.)?(test|spec)\.[jt]s/,
   timeout: 30_000,
+  // This suite became a merge gate (ci.yml), so a single timing hiccup on a
+  // shared runner would now block an unrelated PR. Retries are not here to hide
+  // that: Playwright reports a test that failed then passed as "flaky", a status
+  // distinct from "passed", so the signal stays visible in the uploaded report
+  // while only genuinely reproducible failures stop a merge. Local runs keep 0
+  // retries -- a flake you can reproduce is a flake you can fix.
+  retries: process.env.CI ? 2 : 0,
   // No reporter was ever configured, so CI's "Upload Playwright report" step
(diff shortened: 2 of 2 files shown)


Tara-ag (commented):
Review in progress.

.github/workflows/ci.yml
```diff
@@ -40,17 +40,20 @@ jobs:
...
-        continue-on-error: true
+      # A merge gate as of this commit. History, so the next reader doesn't have
+      # to reconstruct it: this step was advisory because every run of the job
+      # failed every spec with "browserType.launch: Executable doesn't exist" --
+      # the browser-install step above simply didn't exist. The "~130 failing
+      # specs" that justified continue-on-error was therefore never a regression
+      # count; it was an artifact of a missing binary. Once the install landed,
+      # the suite went green and has stayed green for 11 consecutive runs across
+      # 8 different branches (verified 2026-09-02 by reading this step's own
+      # conclusion in the jobs API, not the job's -- which continue-on-error
```
Tara-ag, line 51: 💡 MINOR: The comment repeats the disproven "11 consecutive green runs" claim and misstates the GitHub Actions field. `continue-on-error` forces the step's `conclusion` to success; the real result is in `outcome`. More importantly, the PR description itself retracted this green-run claim because the first CI run found a real `SplitInput` failure. Please rewrite the comment to align with the corrected rationale in the PR body and avoid citing a retracted statistic.

Tara-ag (commented):
One minor documentation-accuracy finding on a workflow comment; the functional change is sound and not blocking.

.github/workflows/ci.yml
Tara-ag, line 51: 💡 MINOR: CI comment repeats disproven claim and misidentifies GitHub Actions step field

The comment states the green-run count was "verified ... by reading this step's own conclusion in the jobs API, not the job's -- which continue-on-error forces to \"success\" regardless". This is inaccurate: in GitHub Actions, `continue-on-error` forces the step's `conclusion` to success too; the actual result is in the step's `outcome`. Moreover, the PR description explicitly retracted the "11 consecutive green runs across 8 branches" claim because this PR's own first run found a real failure. Leaving the retracted claim in the code comment contradicts the PR's own correction and misleads future maintainers.

**Suggestion:** Rewrite the comment to remove the disproven "11 consecutive runs" claim and, if mentioning step-level inspection, refer to the step's `outcome` (not `conclusion`). Focus on the corrected rationale: the original baseline was an artifact of the missing browser-install step, and enabling the gate caught a real platform-dependent SplitInput bug.

Tara-ag (approved):
Approved with one minor documentation-accuracy note.

.github/workflows/ci.yml
murdore: Confirmed and fixed in https://github.com/juspay/svelte-ui-components/pull/500.

Both halves were right. The "11 consecutive green runs" figure was withdrawn during review of #494 itself — the first run with the gate actually enabled caught `splitinput-autofill-distribute` failing all three attempts — but I corrected only the PR body and left the claim sitting in the workflow, which is the copy that outlives the PR.

The field was also backwards: `continue-on-error: true` rewrites the *step's* `conclusion` to success, so reading `conclusion` was exactly the wrong thing to do; `outcome` is what survives it. The comment now records that the number was withdrawn and why, rather than restating a tally nobody can re-derive.

murdore: Confirmed and fixed in https://github.com/juspay/svelte-ui-components/pull/500.

Both halves were right. The "11 consecutive green runs" figure was withdrawn during review of #494 itself — the first run with the gate actually enabled caught `splitinput-autofill-distribute` failing all three attempts — but I corrected only the PR body and left the claim sitting in the workflow, which is the copy that outlives the PR.

The field was also backwards: `continue-on-error: true` rewrites the *step's* `conclusion` to success, so reading `conclusion` was exactly the wrong thing to do; `outcome` is what survives it. The comment now records that the number was withdrawn and why, rather than restating a tally nobody can re-derive.

PR #495, opened 2026-09-01
feat(codemod): add polymorph consumer migration codemod

murdore:
## What

A component-aware codemod that migrates a consumer from `polymorph-ui-components` to `@juspay/svelte-ui-components`, rewriting both import specifiers and the synthetic-event props whose only difference is casing.

```bash
npm run codemod -- [--reverse] [--dry-run] <paths...>
```

## The pair list was wrong — it's 28, not 23

The catalogued figure was **23 casing pairs**. Deriving it from source rather than trusting it found **28**, across 9 components. The method: parse the `$props()` destructure of all 67 components exported by both libraries with `svelte/compiler` — SUI at `origin/release`, poly from a fresh `npm pack polymorph-ui-components@0.7.0` — and accept a pair only when `lower(sui) === lower(poly)`, the names differ, and neither side also has the other's spelling.

**Missed before:** Gallery (`onimageclick`/`oneditclick`/`ondeleteclick`/`onopen`) and MediaUpload (`onremove`). Both were added to SUI after the original 61-component comparison was drawn up, so the analysis never covered them.

**Correctly excluded as semantic renames, not casing:** Gallery `onclose`→`onDismiss`, `onchange`→`onIndexChange`; MediaUpload `onchange`→`onFilesChange`, `onerror`→`onRejected`. These are listed in the README as manual migration items rather than silently rewritten.

Table's two flagged names resolved to real pairs: `onrowclick`→`onRowClick`, `onsort`→`onSort`.

## Component-aware, not find-and-replace

`onsort` on `<Table>` is in scope; `onsort` on somebody's own component is not. The transform resolves each tag to its import identity, handling aliases (`Table as DataTable`), namespace imports (`<Poly.Table>`) and `<svelte:component this={X}>`.

Where it cannot be certain it **warns with `file:line:column` instead of rewriting** — spreads on components that have pairs, unresolvable tags carrying pair names, default imports, and target-name collisions. A silent wrong rewrite is much worse than a reported skip.

## Verified against real consumer code

Dry run over ostia, which is a genuine poly consumer:

```
[dry run] scanned 158 file(s), changed 28, renamed 3 prop(s), rewrote 28 import(s), 0 warning(s)
```

The 3 renamed props are exactly the lowercase Modal handlers ostia uses — `onprimarybuttonclick`, `onoverlayclick`, `onheaderrightimageclick`. Nothing was written; `--dry-run` only prints.

Worth noting for the wider migration question: **every component ostia imports from poly has a SUI equivalent**, so nothing blocks it at the component level.

## Results

- `vitest`: **89/89** in `scripts/codemod` (82 transform + 7 CLI); full unit suite **241/241**, pre-existing suites unaffected
- `pnpm run lint`: exit 0 · `pnpm run check`: exit 0 (strict tsc on the codemod is chained into `check`)
- Written test-first; the transform tests failed 75/82 against a no-op stub before implementation

Coverage includes every pair in both directions, alias/namespace/`svelte:component` resolution, nesting in `{#if}`/`{#each}`/`{#snippet}`, shorthand expansion, the spread warning, duplicate-target conflicts, `onkeydown` discrimination between Input (a pair) and ListItem (not a pair), round-trip restore and idempotency.

## Known limits (documented in the README)

- Module specifiers in plain `.ts`/`.js` barrels use a context-anchored regex; all `.svelte` work is parser-based.
- `{onsort}` shorthand forward-migrates to `onSort={onsort}`; `--reverse` restores the name but not the shorthand form. Semantically identical.
- Prop objects built in `<script>` and spread in are not rewritten — the spread warning covers the usage site.
- Imported specifier names are not validated against SUI's export surface, so a poly-only type import becomes a visible compile error rather than silent corruption.


Tara-ag (commented):
Review in progress

scripts/codemod/transform.ts
```diff
@@ -0,0 +1,474 @@
...
+export function transformSvelte(
+  source: string,
+  file: string,
+  direction: Direction
+): TransformResult {
+  const config = directionConfig(direction);
+  let root: AST.Root;
+  try {
+    root = parse(source, { modern: true, filename: file });
+  } catch (error) {
```
Tara-ag, line 390: 💡 MINOR: The regex-based module-specifier rewrite can match inside comments or string literals (e.g. `// import 'polymorph-ui-components'` or a string containing that text) because it isn't parser-aware. The README mentions this is regex-based, but the risk of rewriting comments/strings is worth a test case or a more prominent warning. Consider adding a test that demonstrates a commented-out import is left untouched, or switching to a lightweight JS parser.

package.json
```diff
@@ -27,7 +27,9 @@
     "package": "svelte-kit sync && svelte-package && publint",
     "prepublishOnly": "npm run build:wc && npm run package",
     "test": "npm run test:integration && npm run test:unit",
-    "check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json",
+    "check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json && npm run check:codemod",
```
Tara-ag, line 30: 💡 MINOR: New `scripts/codemod/` files are not covered by Prettier or ESLint. The `lint` script only checks `src`, so codemod TypeScript can diverge from project style and accumulate unused imports without CI catching it. Extend the script to cover `scripts/codemod`, e.g. `prettier --check src scripts/codemod && eslint src scripts/codemod && npm run lint:event-casing`, or add a separate `lint:codemod` step.

scripts/codemod/cli.test.ts
```diff
@@ -0,0 +1,121 @@
...
+].join('\n');
+
+const SUI_APP = [
+  '<script>',
+  `  import { Table, Input } from '@juspay/svelte-ui-components';`,
+  '  const props = {};',
+  '</script>',
+  '',
+  '<Table onSort={() => {}} {...props} />',
+  '<Input onClick={() => {}} />',
```
Tara-ag, line 25: 💬 SUGGESTION: Module-level mutable `dir` and `lines` variables make the tests rely on beforeEach/afterEach cleanup. If a test throws before `afterEach` runs, the temp directory leaks and `lines` retains stale output. Prefer per-test fixtures or a helper that returns `{ dir, lines }` so each test owns its state.

scripts/codemod/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
+    "allowImportingTsExtensions": true,
+    "target": "ES2023",
+    "lib": ["ES2023"],
+    "types": ["node"],
+    "skipLibCheck": true,
+    "forceConsistentCasingInFileNames": true
+  },
+  "include": ["./*.ts"]
```
Tara-ag, line 14: 💡 MINOR: `"include": ["./*.ts"]` only matches files directly in `scripts/codemod`. If helpers or nested tests are added later, they won't be type-checked by `check:codemod`. Change to `"include": ["./**/*.ts"]` to cover the whole tree.

scripts/codemod/map.ts
```diff
@@ -0,0 +1,93 @@
...
+  readonly renames: ReadonlyMap<string, ReadonlyMap<string, string>>;
+  /** every source prop name across all components, for unresolved-tag warnings */
+  readonly allFromProps: ReadonlySet<string>;
+};
+
+export function directionConfig(direction: Direction): DirectionConfig {
+  const renames = new Map<string, Map<string, string>>();
+  const allFromProps = new Set<string>();
+  for (const pair of PROP_PAIRS) {
+    const from = direction === 'to-sui' ? pair.poly : pair.sui;
```
Tara-ag, line 80: ⚠️ MAJOR: Stepper is treated as a casing-only pair, but `onhandleStepClick` is a deprecated alias in SUI. `StepperProperties` marks it `@deprecated Use onstepclick instead.` and `Stepper.svelte` uses `onstepclick ?? onhandleStepClick`. Forward migration therefore leaves poly consumers on a deprecated prop instead of the canonical `onstepclick`; reverse migration wrongly rewrites current `onstepclick` usage to the deprecated alias's lowercase form. Remove this pair from `PROP_PAIRS` and list `onhandlestepclick` → `onstepclick` as a semantic rename in the README, or add a special-case transform that maps directly to `onstepclick`.

Tara-ag (changes requested):
Review complete. One MAJOR concern about the Stepper deprecated-alias mapping, plus a few MINOR/SUGGESTION items. None of the MINOR items block merge on their own, but the Stepper mapping should be corrected before this codemod is trusted for consumer migrations.

Tara-ag (commented):
Code review findings from Yama.

package.json
Tara-ag, line 30: 💡 MINOR: New `scripts/codemod/` files are not covered by Prettier or ESLint. `package.json` `lint` runs `prettier --check src && eslint src && npm run lint:event-casing`. The new codemod files live under `scripts/codemod/` and are therefore not formatted by Prettier or linted by ESLint, even though `check:codemod` runs tsc.

**Fix:** Extend the lint script to cover `scripts/codemod`, e.g. `prettier --check src scripts/codemod && eslint src scripts/codemod && npm run lint:event-casing`, or add a separate `lint:codemod` step and chain it.

scripts/codemod/tsconfig.json
Tara-ag, line 14: 💡 MINOR: Codemod tsconfig include pattern may miss nested test files. `scripts/codemod/tsconfig.json` includes `./*.ts`, which only matches files directly in `scripts/codemod`. If future tests or helpers are placed in subdirectories, they will not be type-checked by `check:codemod`.

**Fix:** Change `"include": ["./*.ts"]` to `"include": ["./**/*.ts"]` so the whole codemod tree is checked.

scripts/codemod/cli.test.ts
Tara-ag, line 25: 💬 SUGGESTION: CLI tests use shared mutable `dir` and `lines` variables. `cli.test.ts` declares `let dir = ''` and `let lines: string[] = []` at module scope and mutates them in `beforeEach`/`afterEach`. While Vitest isolates files, this pattern is less safe than per-test fixtures and can leak state if a test throws before `afterEach` runs.

**Fix:** Use `return { dir, lines }` from a helper or per-test `mkdtempSync` calls instead of module-level mutable state.

scripts/codemod/transform.ts
Tara-ag, line 390: 💡 MINOR: Regex-based module specifier rewrite can match inside comments/strings. `transform.ts`'s `SPECIFIER_CONTEXT` regex is context-anchored but not parser-aware. A comment like `// import 'polymorph-ui-components'` or a string literal containing that text will be rewritten because the regex only looks for `import (` / `import ` / `from ` / `require(`.

**Fix:** Document the limitation more prominently, or switch to a lightweight parser (e.g. acorn) for .js/.ts files. At minimum, add a test case demonstrating that the regex does not rewrite a commented-out import.

scripts/codemod/map.ts
Tara-ag, line 80: ⚠️ MAJOR: Codemod migrates Stepper consumers to deprecated `onhandleStepClick` instead of canonical `onstepclick`. `map.ts` lists Stepper `onhandleStepClick` ↔ `onhandlestepclick` as a casing-only pair. However, SUI's StepperProperties declares `onhandleStepClick` with `@deprecated Use onstepclick instead.` and Stepper.svelte derives the effective handler as `onstepclick ?? onhandleStepClick`. Forward migration therefore leaves poly consumers on a deprecated prop instead of the canonical `onstepclick`; reverse migration wrongly rewrites current `onstepclick` usage to the deprecated alias's lowercase form.

**Fix:** Remove the Stepper entry from `PROP_PAIRS` and add `onhandlestepclick` → `onstepclick` to the README's semantic-renames list (alongside Gallery/MediaUpload). Alternatively, implement a special-case transform so poly `onhandlestepclick` becomes SUI `onstepclick`, and reverse transforms SUI `onstepclick` to poly `onhandlestepclick`.

Tara-ag (changes requested):
Reviewed file-by-file. One MAJOR issue around the Stepper deprecated prop mapping should be addressed before merge; the remaining items are quality nits.

Tara-ag:

The codemod is well-scoped and thoroughly tested, but it incorrectly treats Stepper's deprecated `onhandleStepClick` alias as a canonical casing-only pair, which will migrate consumers to a deprecated prop and mis-handle reverse migration. Additionally, the new `scripts/codemod/` files are not covered by Prettier/ESLint, the codemod tsconfig may miss nested files, the regex-based module specifier rewriter can touch comments/strings, and the CLI tests use shared mutable state. The Stepper mapping should be fixed before merge; the remaining items are quality nits.

- ⚠️ **MAJOR**: Codemod migrates Stepper consumers to deprecated onhandleStepClick instead of canonical onstepclick — `scripts/codemod/map.ts:80`
  map.ts lists Stepper `onhandleStepClick` ↔ `onhandlestepclick` as a casing-only pair (line 80). However, SUI's StepperProperties declares `onhandleStepClick` with `@deprecated Use onstepclick instead.` and Stepper.svelte derives the effective handler as `onstepclick ?? onhandleStepClick`. Therefore the canonical, supported prop is `onstepclick`, and `onhandleStepClick` exists only as a backward-co…
- 💡 **MINOR**: New scripts/codemod files are not covered by Prettier or ESLint — `package.json:30`
  package.json `lint` runs `prettier --check src && eslint src && npm run lint:event-casing`. The new codemod files live under `scripts/codemod/` and are therefore not formatted by Prettier or linted by ESLint, even though `check:codemod` runs tsc. This means the new TypeScript code can diverge from project style and accumulate unused imports without CI catching it.
- 💡 **MINOR**: Codemod tsconfig include pattern may miss nested test files — `scripts/codemod/tsconfig.json:14`
  `scripts/codemod/tsconfig.json` includes `./*.ts`, which only matches files directly in scripts/codemod. If future tests or helpers are placed in subdirectories (e.g. scripts/codemod/utils/helper.ts), they will not be type-checked by `check:codemod`.
- 💡 **MINOR**: Regex-based module specifier rewrite can match inside comments/strings — `scripts/codemod/transform.ts:390`
  transform.ts's SPECIFIER_CONTEXT regex is context-anchored but not parser-aware. A comment like `// import 'polymorph-ui-components'` or a string literal like `const x = "import 'polymorph-ui-components'"` will be rewritten because the regex only looks for `import (`/`import `/`from `/`require(`. The README acknowledges this limitation for `.ts`/`.js` files, but it is still a real risk for consume…
- 💬 **SUGGESTION**: CLI tests use shared mutable `dir` and `lines` variables — `scripts/codemod/cli.test.ts:25`
  cli.test.ts declares `let dir = ''` and `let lines: string[] = []` at module scope and mutates them in beforeEach/afterEach. While Vitest isolates files, this pattern is less safe than per-test fixtures and can leak state if a test throws before afterEach runs.

scripts/codemod/map.ts
murdore: Confirmed and fixed in https://github.com/juspay/svelte-ui-components/pull/500 — this was the most serious finding in the sweep.

`StepperProperties` does mark `onhandleStepClick` `@deprecated Use onstepclick instead.`, and the codemod was rewriting fork consumers straight onto it. The cause is in the map's own derivation rule: pairs were accepted when `lower(sui) === lower(poly)`, which is deprecation-blind, so where a component keeps a deprecated alias whose lowercase spelling matches the fork's, the alias outranks the canonical prop and wins the pair. A consumer running the codemod landed on the one spelling of that event scheduled for removal, and immediately owed a second migration.

Stepper is the only pair affected across all 28.

The fix targets `onstepclick`, and keeps `onhandleStepClick` recognised as a *source* in the `to-poly` direction so a SUI consumer on either spelling still converts. `map.ts` had no test at all, which is why this got through; `map.test.ts` now asserts no target is `@deprecated`, that every target is really declared, and that a pair may only differ by more than case in order to step over a deprecated alias. That guard fails on the old map.

murdore: Confirmed and fixed in https://github.com/juspay/svelte-ui-components/pull/500 — this was the most serious finding in the sweep.

`StepperProperties` does mark `onhandleStepClick` `@deprecated Use onstepclick instead.`, and the codemod was rewriting fork consumers straight onto it. The cause is in the map's own derivation rule: pairs were accepted when `lower(sui) === lower(poly)`, which is deprecation-blind, so where a component keeps a deprecated alias whose lowercase spelling matches the fork's, the alias outranks the canonical prop and wins the pair. A consumer running the codemod landed on the one spelling of that event scheduled for removal, and immediately owed a second migration.

Stepper is the only pair affected across all 28.

The fix targets `onstepclick`, and keeps `onhandleStepClick` recognised as a *source* in the `to-poly` direction so a SUI consumer on either spelling still converts. `map.ts` had no test at all, which is why this got through; `map.test.ts` now asserts no target is `@deprecated`, that every target is really declared, and that a pair may only differ by more than case in order to step over a deprecated alias. That guard fails on the old map.

PR #496, opened 2026-09-01
test(visual): add container-pinned visual regression suite for all 94 demos

murdore:
## What

Screenshots each `/components/<slug>` demo and compares it against a committed baseline. Closes the last untouched item from the adoption plan — there was no screenshot-diff tooling anywhere in the org, so this is built new.

```bash
pnpm run test:visual           # compare
pnpm run test:visual:update    # regenerate after an intended change
```

Routes are discovered from disk at run time, so **a new component demo without a baseline fails the suite**. An opt-in list is how visual coverage rots.

## Why it runs in a container

Baselines are only comparable against an identical renderer. macOS, a bare `ubuntu-latest` runner and the Playwright image each rasterise glyphs differently — enough to fail an exact comparison on every route while nothing changed. The usual workaround is a pixel tolerance, which then also hides the small real regressions the suite exists to catch.

The renderer is pinned to `mcr.microsoft.com/playwright:v1.60.0-noble`, used by `scripts/visual-test.sh` locally **and** by this workflow's `container:` block, so a baseline regenerated on a laptop is the same measurement CI takes.

## Five causes of non-determinism

Each found by reading a failing diff, not by reasoning about it. Recording the wrong turns too:

1. **`brand-loader` requests `picsum.photos`** — which returns *a different photograph every request*. This presented as an animation bug for three rounds of investigation until the diff image showed the changing pixels were the logos themselves. Off-origin images are now stubbed, which also removes a CI network dependency.
2. **Chat and ThinkingIndicator append content on an interval**, growing the element between successive screenshots so no baseline could be written. `page.clock.install()` makes timers manual. `chat` and `tool-call-log` run timer *cascades* and need longer settles.
3. **Playwright scrolls a too-tall element to stitch the shot**, and content that renders lazily on scroll changed height mid-capture. My first fix — a flat 4000px viewport — only half-worked: **11 demos exceed it, the tallest at 12209px**, so exactly those kept scrolling and stayed flaky. Each test now fits the viewport to its own page height.
4. **`animations: 'disabled'` alone** left ThinkingIndicator oscillating between two heights. Tagging animating elements with a marker class didn't hold either — Svelte rewrites the class attribute on re-render.
5. **Late-arriving images** change the pixels and, before they have intrinsic size, the element's height.

## Verification

```
VERIFY A (workers=2): 92 passed, 2 skipped
VERIFY B (workers=4): 92 passed, 2 skipped
VERIFY C (workers=1): 92 passed, 2 skipped
```

Three independent comparison runs against the committed baselines, identical at every worker count. Modal and Toast baselines were inspected by eye to confirm `animation: none` settles entry animations rather than freezing them — the failure mode would be a component baselined invisible.

**The suite has already proved it works:** rebasing onto the merged #491/#498 flagged 12 stale baselines — precisely the demo pages those PRs changed.

## Two excluded routes, deliberately

`thinking-indicator` and `task-list` are **not** baselined. They're declared as skipped tests, not omitted, so they still appear in the report with their reason.

Both pass reliably alone and fail intermittently in the full suite at any worker count — load sensitivity in how much settles before capture, not something in the components. Everything that stabilised the other 92 was tried on them: manual clock, 30s settles, per-route viewport fitting, stripped animations, stubbed network.

This is a judgement call worth stating plainly: a gate that fails at random on unrelated PRs gets switched off within a week, and then you have neither the coverage nor the gate. Two documented holes in a suite people trust beats 94 routes in one they don't. Closing them is real remaining work, not settled.

**Near-miss worth recording:** my first attempt at the exclusion used `test.skip(condition, reason)` inside the describe body. That form applies to the *whole describe* — it skipped all 94 routes while the job still exited 0. A gate reporting success without checking anything, and it would have looked green here. Caught because the verification printed `94 skipped` instead of `92 passed`. `docs/VISUAL_TESTING.md` warns against it explicitly.

## Other honest limitations

- **Routes are captured at rest.** An open Modal, a fired Toast, an expanded Menu, a hovered Tooltip are not covered — nothing clicks the demo buttons first. Triggers are compared; overlay states aren't.
- **`lottie-player` and `media-player` have their `.demo-row` masked** — rAF-driven frames and video decoder timing. The rest of each route is compared.
- **~11MB of PNGs.** A baseline refresh is a visible diff in review.

## Version pinning

`package.json`, `scripts/visual-test.sh` (`IMAGE`) and `.github/workflows/visual.yml` (`container.image`) must move together. The script fails fast if the image and `node_modules` disagree, so a partial bump is caught before it produces 92 bogus diffs.

## Summary by CodeRabbit

- **New Features**
  - Added automated visual regression testing for component demo pages.
  - Added commands to run visual tests and update screenshot baselines.
  - Added consistent rendering settings for reliable screenshot comparisons.

- **Documentation**
  - Added guidance covering visual test usage, configuration, baselines, deterministic rendering, and failure handling.

- **Chores**
  - Added pull request automation to run visual tests and retain failure reports for troubleshooting.


Tara-ag (commented):
Review findings posted inline. Summary: the container-pinned visual suite is well-designed for determinism, but the at-rest capture strategy leaves overlay components effectively untested, and a few stabilization/docs gaps remain.

Tara-ag (commented):
⚠️ MAJOR: The test screenshots each route in its at-rest state without opening any overlays. For demos like Modal, Sheet, CommandMenu, ContextMenu, Menu, Tooltip, and Toast, the actual component UI only appears after user interaction (button click, right-click, hover, etc.), so the baseline will capture just the trigger button. The docs acknowledge this as a "real coverage gap", but the PR title claims a suite "for all 94 demos"; overlay components are therefore not visually regression-tested. Add per-slug setup steps (e.g. a `SETUP` map that clicks the demo trigger) or split overlay demos into dedicated interaction-based visual specs so the baselines actually exercise the component UI.

docs/VISUAL_TESTING.md
```diff
@@ -0,0 +1,139 @@
...
+  rendering change next to its cause.
+
+Never regenerate baselines to turn a red suite green without looking at the
+diff. That converts the tool into a rubber stamp.
+
+## Cost
+
+94 baselines, ~11MB total. `main.content` stretches to the viewport, so each PNG
+is 980x4000 and largely empty — which compresses well, but means a baseline
+refresh is a visible diff in review.
```
Tara-ag: 💡 MINOR: The doc says baselines are `980x4000`, but `playwright.visual.config.ts` configures `viewport: { width: 1280, height: 4000 }`. Update this line to `1280x4000` so the documentation matches the actual configuration.

.github/workflows/visual.yml
```diff
@@ -0,0 +1,49 @@
...
+
+jobs:
+  visual:
+    runs-on: ubuntu-latest
+    # The job runs INSIDE the same image `scripts/visual-test.sh` uses locally.
+    # This is the whole design: baselines are only comparable when the renderer
+    # is identical, and a bare ubuntu-latest runner does not ship the same fonts
+    # or freetype build as this image. Running here rather than on the host is
+    # what lets a developer regenerate a baseline on a laptop and have CI agree.
+    # Keep this tag and the IMAGE in scripts/visual-test.sh in lockstep with the
```
Tara-ag, line 18: 💡 MINOR: The pnpm version is hardcoded here (`10.18.3`) as well as in `ci.yml`. Extract it to a workflow-level `env` variable so a future bump only needs one edit per workflow, matching the repo's pattern for keeping shared CI constants centralized.

tests/visual/components.visual.ts
```diff
@@ -0,0 +1,187 @@
+import { readdirSync } from 'node:fs';
+import { join } from 'node:path';
+import { expect, test, type Page } from '@playwright/test';
+
+// Routes are discovered from disk rather than hand-listed on purpose: a new
+// component demo added without a committed baseline fails this suite instead of
+// silently going unscreenshotted, which is the failure mode that lets visual
+// coverage rot. The cost is that adding a demo requires regenerating baselines.
+const COMPONENTS_DIR = join(process.cwd(), 'src/routes/components');
+
```
Tara-ag, line 10: 💡 MINOR: Slugs are discovered from the filesystem and filtered by `SKIP_SLUGS`, but `src/routes/components` contains 97 directories while `_nav.ts` is the public source of truth for component demos. This can include utility routes or miss demos whose directory name differs from the nav slug. Derive the slug list from `_nav.ts` and assert each maps to an existing directory, or cross-check the discovered set against `_nav.ts` so the visual suite stays synchronized with the documented component list.

tests/visual/components.visual.ts
```diff
@@ -0,0 +1,187 @@
...
+test.describe('visual baselines', () => {
+  for (const slug of slugs) {
+    test(`${slug} renders as baselined`, async ({ page }) => {
+      await prepare(page, slug);
+
+      const masks = (MASKS[slug] ?? []).map((selector) => page.locator(selector));
+
+      await expect(page.locator('main.content')).toHaveScreenshot(`${slug}.png`, {
+        animations: 'disabled',
```
Tara-ag, line 256: ⚠️ MAJOR: The test screenshots each route in its at-rest state without opening any overlays. For demos like Modal, Sheet, CommandMenu, ContextMenu, Menu, Tooltip, and Toast, the actual component UI only appears after user interaction (button click, right-click, hover, etc.), so the baseline will capture just the trigger button. The docs acknowledge this as a "real coverage gap", but the PR title claims a suite "for all 94 demos"; overlay components are therefore not visually regression-tested. Add per-slug setup steps (e.g. a `SETUP` map that clicks the demo trigger) or split overlay demos into dedicated interaction-based visual specs so the baselines actually exercise the component UI.

tests/visual/components.visual.ts
```diff
@@ -0,0 +1,187 @@
...
+  .sort();
+
+/**
+ * Regions whose pixels are genuinely non-deterministic between identical runs,
+ * with the reason each one is masked. Masking paints a box over the region, so
+ * the rest of the component is still compared -- this is deliberately narrower
+ * than skipping the route. Anything not listed here is compared in full.
+ */
+const MASKS: Readonly<Record<string, readonly string[]>> = {
+  // Lottie drives its frames from rAF in JS, not CSS, so `animations: 'disabled'`
```
Tara-ag, line 23: ⚠️ MAJOR: Stabilization covers CSS animations and off-origin images, but JS-driven motion beyond `setTimeout`/`setInterval` is not fully pinned. `page.clock.install()` does not control `requestAnimationFrame`-driven renderers such as Lottie (already masked) and potentially other components (carousels, sliders, progress indicators, skeletons). Timer cascades are only overridden for `chat` and `tool-call-log`; demos like `thinking-indicator`, `task-list`, `typewriter-text`, and `sound-kit` may still produce non-deterministic frames depending on how their components schedule work. Expand masking or add per-route setup that pauses these state machines, and verify each baseline reproduces across multiple isolated runs before merging.

Tara-ag (commented):
Review findings posted inline. Summary: the container-pinned visual suite is well-designed for determinism, but the at-rest capture strategy leaves overlay components effectively untested, and a few stabilization/docs gaps remain.

Tara-ag (changes requested):
The visual-regression suite is well designed overall (container-pinned renderer, manual clocks, off-origin image stubbing, full-height viewport), but two gaps keep it from reliably testing what it claims to cover:

1. **Overlay components are captured closed.** Modal, Sheet, CommandMenu, ContextMenu, Menu, Tooltip, Toast, and similar demos are photographed at rest, so baselines show only trigger buttons. The PR title says "for all 94 demos"; overlay states need per-route interaction setup or dedicated specs.
2. **JS-driven motion is not fully stabilized.** Beyond `setTimeout`/`setInterval`, rAF-driven and state-machine-based demos (Lottie already masked, but also carousels, sliders, progress, skeletons, thinking-indicator, task-list, typewriter-text, sound-kit) may still produce non-deterministic frames.

Also noted: derive slug list from `_nav.ts` rather than filesystem discovery, centralize the pnpm version in workflow env vars, and correct the documented screenshot width from 980 to 1280.

Please address the overlay coverage gap and verify stabilization across multiple isolated runs before re-requesting review.

Tara-ag:

The visual-regression suite is well designed and thoroughly documented. However, the new CI workflow is missing an explicit Playwright browser-install step, which can cause the job to fail when the container's pre-installed browser revision does not match the @playwright/test package revision. Three additional minor robustness issues are noted in the version-parsing script and the viewport/image stabilization helpers.

- ⚠️ **MAJOR**: Visual workflow does not install Playwright browsers — `.github/workflows/visual.yml:28`
  The visual.yml workflow runs `pnpm exec playwright test` without an `npx playwright install --with-deps chromium` step. While the container image ships Chromium, Playwright's version check frequently requires a browser reinstall when the @playwright/test package version does not exactly match the preinstalled browser revision in the image. The existing ci.yml installs browsers explicitly for this …
- 💡 **MINOR**: visual-test.sh parses Playwright version from image tag with bash parameter expansion that breaks on non-semver tags — `scripts/visual-test.sh:31`
  EXPECTED_PW_VERSION is derived by stripping `:v` and everything after `-`. This works for `v1.60.0-noble` but will produce wrong values for tags like `v1.60.0-jammy-amd64` or future `v1.60.0-noble-cu121`. The script then compares this to the package.json version and exits. A more reliable check is to read the Playwright version baked into the image metadata or to run `docker run --rm $IMAGE playwr…
- 💡 **MINOR**: fitViewportToContent silently gives up after 4 iterations — `tests/visual/components.visual.ts:145`
  If the page height keeps changing for more than 4 iterations, the function exits the loop and returns without having stabilized the viewport. This could leave a tall page still scrolling during screenshot. A warning or error would make load-sensitivity failures easier to diagnose, especially for the two already-excluded routes.
- 💡 **MINOR**: waitForImages does not wait for images added after the initial scan — `tests/visual/components.visual.ts:102`
  The function collects `document.images` once and waits for those incomplete images. If a component lazily adds an `` after the initial evaluation (e.g., after the manual clock advances or after a route's own setTimeout), that image is not awaited. This is a plausible source of the remaining flakiness in thinking-indicator and task-list.

.github/workflows/visual.yml
```diff
@@ -0,0 +1,49 @@
...
+    # @playwright/test version in package.json.
+    container:
+      image: mcr.microsoft.com/playwright:v1.60.0-noble
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
```
Tara-ag, line 28: ⚠️ MAJOR: The workflow runs `pnpm exec playwright test` without an explicit browser-install step. The Playwright container image ships a browser, but Playwright still validates that the installed browser revision matches the `@playwright/test` package revision; a patch mismatch produces `Executable doesn't exist`. The existing `ci.yml` installs browsers explicitly for this reason. Add a step before this one:

```yaml
      - name: Install Playwright browsers
        run: pnpm exec playwright install --with-deps chromium
```

scripts/visual-test.sh
```diff
@@ -0,0 +1,64 @@
...
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+
+if ! docker info >/dev/null 2>&1; then
+  echo "error: Docker is not running. The visual suite requires it -- see docs/VISUAL_TESTING.md" >&2
+  exit 1
+fi
+
+EXPECTED_PW_VERSION="${IMAGE##*:v}"
+EXPECTED_PW_VERSION="${EXPECTED_PW_VERSION%%-*}"
```
Tara-ag, line 31: 💡 MINOR: Parsing the expected Playwright version from the image tag with `${IMAGE##*:v}` and `${EXPECTED_PW_VERSION%%-*}` is fragile. It works for `v1.60.0-noble` but breaks for tags like `v1.60.0-jammy-amd64` or future variants such as `v1.60.0-noble-cu121`, causing a false mismatch exit. Query the version directly from the image instead:

```bash
EXPECTED_PW_VERSION="$(docker run --rm "$IMAGE" sh -c 'npx playwright --version' 2>/dev/null | awk '{print $2}')"
```

tests/visual/components.visual.ts
```diff
@@ -0,0 +1,265 @@
...
+ *
+ * Repeats because growing the viewport can reveal content that was not
+ * rendered before -- the page gets taller as a result of being measured. Two
+ * passes settle every demo here; the loop stops as soon as the height holds.
+ */
+async function fitViewportToContent(page: Page): Promise<void> {
+  let previous = 0;
+  for (let attempt = 0; attempt < 4; attempt++) {
+    const height = await page.evaluate(() =>
+      Math.max(document.documentElement.scrollHeight, document.body.scrollHeight)
```
Tara-ag, line 145: 💡 MINOR: `fitViewportToContent` silently returns after 4 attempts even if the height is still changing. If a page has not converged, the screenshot may still be stitched with scrolling and produce a flaky baseline. Fail loudly when the loop exhausts without stabilization:

```ts
async function fitViewportToContent(page: Page, slug: string): Promise<void> {
  let previous = 0;
  for (let attempt = 0; attempt < 4; attempt++) {
    const height = await page.evaluate(() =>
      Math.max(document.documentElement.scrollHeight, document.body.scrollHeight)
    );
    if (height === previous) {
      return;
    }
    previous = height;
    await page.setViewportSize({
      width: VIEWPORT_WIDTH,
      height: Math.min(Math.max(height, MIN_VIEWPORT_HEIGHT), MAX_VIEWPORT_HEIGHT)
    });
  }
  throw new Error(`Viewport height did not stabilize for ${slug} after 4 attempts (${previous}px)`);
}
```

tests/visual/components.visual.ts
```diff
@@ -0,0 +1,265 @@
...
+  await page.route('**/*', (route) => {
+    const { hostname } = new URL(route.request().url());
+    if (hostname === 'localhost' || hostname === '127.0.0.1') {
+      return route.continue();
+    }
+    if (route.request().resourceType() === 'image') {
+      return route.fulfill({ contentType: 'image/svg+xml', body: STUB_IMAGE });
+    }
+    return route.abort();
+  });
```
Tara-ag, line 102: 💡 MINOR: `waitForImages` snapshots `document.images` once. Images injected after this scan (e.g., by a delayed `setTimeout` or after the clock advances) are not awaited and can shift pixels mid-screenshot. This is a plausible contributor to the remaining flakiness in `thinking-indicator` and `task-list`. Poll until no incomplete images remain:

```ts
async function waitForImages(page: Page): Promise<void> {
  await page.waitForFunction(() => {
    return Array.from(document.images).every((img) => img.complete);
  });
}
```

Tara-ag (changes requested):
The visual-regression suite is well thought-through and the documentation is excellent. I am requesting changes because the CI workflow is missing an explicit Playwright browser-install step; without it the job can fail when the pre-installed browser revision in the container does not match the `@playwright/test` package revision. The other comments are minor robustness improvements for the version-parsing script and the image/viewport stabilization helpers.

Tara-ag (commented):
Visual regression review

tests/visual/components.visual.ts
Tara-ag, line 102: 💡 MINOR: The function collects `document.images` once and waits for those incomplete images. If a component lazily adds an `` after the initial evaluation (e.g., after the manual clock advances or after a route's own setTimeout), that image is not awaited. This is a plausible source of the remaining flakiness in thinking-indicator and task-list.

Suggested fix: Use a MutationObserver to watch for new img elements and await them too, or poll `document.querySelectorAll('img:not([complete])')` until empty with a short timeout.

tests/visual/components.visual.ts
Tara-ag, line 145: 💡 MINOR: If the page height keeps changing for more than 4 iterations, the function exits the loop and returns without having stabilized the viewport. This could leave a tall page still scrolling during screenshot. A warning or error would make load-sensitivity failures easier to diagnose, especially for the two already-excluded routes.

Suggested fix: After the loop, assert that the height stabilized:
```ts
if (height !== previous) {
  throw new Error(`Viewport height did not stabilize for ${slug} after 4 attempts (${height}px)`);
}
```

.github/workflows/visual.yml
Tara-ag, line 28: ⚠️ MAJOR: The visual.yml workflow runs `pnpm exec playwright test` without an `npx playwright install --with-deps chromium` step. While the container image ships Chromium, Playwright's version check frequently requires a browser reinstall when the @playwright/test package version does not exactly match the preinstalled browser revision in the image. The existing ci.yml installs browsers explicitly for this reason. Without it, the job can fail with 'Executable doesn't exist' on the first run after a Playwright patch.

Suggested fix: Add a step before 'Visual regression':
```yaml
      - name: Install Playwright browsers
        run: pnpm exec playwright install --with-deps chromium
```

scripts/visual-test.sh
Tara-ag, line 31: 💡 MINOR: `EXPECTED_PW_VERSION` is derived by stripping `:v` and everything after `-`. This works for `v1.60.0-noble` but will produce wrong values for tags like `v1.60.0-jammy-amd64` or future `v1.60.0-noble-cu121`. The script then compares this to the package.json version and exits. A more reliable check is to read the Playwright version baked into the image metadata or to run `docker run --rm $IMAGE playwright --version` and compare the installed CLI version.

Suggested fix: Replace the parameter-expansion parsing with a direct version query inside the container:
```bash
EXPECTED_PW_VERSION="$(docker run --rm "$IMAGE" sh -c 'npx playwright --version' 2>/dev/null | awk '{print $2}')"
```
Or compare the image's package.json version directly.

Tara-ag (commented):
Visual regression review: one workflow robustness concern and three minor test/script nits identified.

scripts/visual-test.sh
```diff
@@ -0,0 +1,64 @@
...
+# is identical between the run that produced the baseline and the run checking
+# against it. Font packages, freetype/fontconfig versions and hinting defaults
+# all differ between macOS, a bare GitHub runner and this image, and each of
+# those shifts glyph antialiasing by a pixel or two -- enough to fail an exact
+# comparison for reasons unrelated to the component. Pinning the image makes a
+# baseline generated on a laptop and one checked in CI the same measurement.
+#
+# Usage:
+#   scripts/visual-test.sh              # compare against committed baselines
+#   scripts/visual-test.sh --update     # regenerate baselines
```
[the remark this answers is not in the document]
murdore: Correct, and fixed in https://github.com/juspay/svelte-ui-components/pull/500. Verified before changing it — `playwright test --update` exits with `error: unknown option '--update'`.

Worth noting the blast radius was smaller than it looks: `package.json` already runs `--update-snapshots`, so `pnpm run test:visual:update` worked. It was only the usage block in the script's own header that documented a flag which errors out, i.e. exactly the path someone takes when they read the script instead of the package scripts.

tests/visual/components.visual.ts
murdore: Confirmed, and it was hiding more than a flaky screenshot. Fixed in https://github.com/juspay/svelte-ui-components/pull/500.

Making it throw instead of returning showed that two demos *never* converge. The demo shell is `min-height: 100vh`, so a page that adds its own full-height element grows by exactly what the viewport gains — fitting it is an infinite loop, not slow convergence. `brand-loader` measured `1539 -> 2278 -> 3017 -> 3756 -> 4495 -> 5234`, a constant +739 per round.

Measured against a fixed control:

| Route | Content height vs viewport | Sized against |
| --- | --- | --- |
| `table` | 36388px at every viewport | its own content |
| `brand-loader` | grows 1x the viewport gain | the viewport |
| `status` | grows 3x the viewport gain | the viewport |

So the loop was running out of attempts and capturing at a size decided by the iteration count. That is why `brand-loader` is one of the routes that kept flaking through this work. `status` was the clearest symptom: its committed baseline had inflated to **60209px, roughly 92% stretched empty space**.

Growth that tracks the viewport is now detected and the page pinned to the default viewport; anything moving for a third reason throws. Both baselines regenerated — `status` is 2609px, 84% smaller, with all seven demo variants still in frame (checked by eye, not by file size). Full suite run twice in the pinned container: 92 passed, 2 skipped, identical both times.

murdore: Confirmed, and it was hiding more than a flaky screenshot. Fixed in https://github.com/juspay/svelte-ui-components/pull/500.

Making it throw instead of returning showed that two demos *never* converge. The demo shell is `min-height: 100vh`, so a page that adds its own full-height element grows by exactly what the viewport gains — fitting it is an infinite loop, not slow convergence. `brand-loader` measured `1539 -> 2278 -> 3017 -> 3756 -> 4495 -> 5234`, a constant +739 per round.

Measured against a fixed control:

| Route | Content height vs viewport | Sized against |
| --- | --- | --- |
| `table` | 36388px at every viewport | its own content |
| `brand-loader` | grows 1x the viewport gain | the viewport |
| `status` | grows 3x the viewport gain | the viewport |

So the loop was running out of attempts and capturing at a size decided by the iteration count. That is why `brand-loader` is one of the routes that kept flaking through this work. `status` was the clearest symptom: its committed baseline had inflated to **60209px, roughly 92% stretched empty space**.

Growth that tracks the viewport is now detected and the page pinned to the default viewport; anything moving for a third reason throws. Both baselines regenerated — `status` is 2609px, 84% smaller, with all seven demo variants still in frame (checked by eye, not by file size). Full suite run twice in the pinned container: 92 passed, 2 skipped, identical both times.

tests/visual/components.visual.ts
murdore: Confirmed and fixed in https://github.com/juspay/svelte-ui-components/pull/500.

Rescanned each round now. The case that motivated it is exactly the one you describe: settling the first batch can append another image, and a single snapshot of `document.images` returns with that one still in flight. It also throws if images are still arriving after five rounds, rather than capturing whichever ones happened to make it.

murdore: Confirmed and fixed in https://github.com/juspay/svelte-ui-components/pull/500.

Rescanned each round now. The case that motivated it is exactly the one you describe: settling the first batch can append another image, and a single snapshot of `document.images` returns with that one still in flight. It also throws if images are still arriving after five rounds, rather than capturing whichever ones happened to make it.

.github/workflows/visual.yml
murdore: Declining the specific fix, but the concern underneath it is real and is addressed differently in https://github.com/juspay/svelte-ui-components/pull/500.

`playwright install --with-deps chromium` inside the container would defeat the job. The image is pinned so a baseline generated on a laptop and one checked in CI are the same measurement; `playwright install` downloads whatever browser the npm package wants and shadows the image's. If the versions agree it is a no-op, and if they disagree it silently makes CI rasterise differently from every developer — turning a loud failure into exactly the drift these pinned baselines exist to catch.

You are right that nothing *enforced* the lockstep in CI, though — `scripts/visual-test.sh` checks it locally and the workflow only asked for it in a comment. #500 adds a check rather than an install: it resolves the browser path Playwright expects and fails naming the three files to update, instead of the bare "Executable doesn't exist" that kept the integration suite advisory for months.

murdore: Declining the specific fix, but the concern underneath it is real and is addressed differently in https://github.com/juspay/svelte-ui-components/pull/500.

`playwright install --with-deps chromium` inside the container would defeat the job. The image is pinned so a baseline generated on a laptop and one checked in CI are the same measurement; `playwright install` downloads whatever browser the npm package wants and shadows the image's. If the versions agree it is a no-op, and if they disagree it silently makes CI rasterise differently from every developer — turning a loud failure into exactly the drift these pinned baselines exist to catch.

You are right that nothing *enforced* the lockstep in CI, though — `scripts/visual-test.sh` checks it locally and the workflow only asked for it in a comment. #500 adds a check rather than an install: it resolves the browser path Playwright expects and fails naming the three files to update, instead of the bare "Executable doesn't exist" that kept the integration suite advisory for months.

PR #498, opened 2026-09-02
fix(Img): sanitize fetched SVG before adoption; expose show-indicator on sui-choicebox

murdore:
## Defect 1 — a fetched SVG could run script in the host page

`Img`'s `inlineSvg` path fetches remote markup and adopts it into the live document. Two ways that let an untrusted file execute:

**Root attributes were copied wholesale.** A payload could plant `onload`/`onclick`, clobber the `data-pw` test hook, set an `id` (DOM clobbering), or override the component's sizing contract with inline `style`.

**Descendants were adopted wholesale too** — and this is the part worth calling out. An event-handler content attribute becomes a live handler *the moment its element is adopted*, so `<image onerror="…">` inside the payload runs exactly as `onerror` on the root would. Fixing only the root would have looked like a fix while leaving the identical vector open one element deeper.

### The fix

- **Root:** allowlist — geometry, paint and a11y metadata (`viewBox`, `fill`, `stroke-*`, `role`, `focusable`, `aria-*`, …). That is what real icon pipelines emit and what this repo's own assets carry. Excluded deliberately: `on*` (executable), `id` (clobbering), `style` (defeats the `--image-width/height` contract), `tabindex`, `data-*` (could hijack `data-pw`).
- **Descendants:** denylist — scripting elements, `on*` handlers, and `javascript:` URLs are stripped *before* adoption. A denylist here on purpose: descendants legitimately carry the entire SVG geometry and paint vocabulary (`d`, `transform`, `gradientUnits`…), so enumerating what is allowed would break real artwork.

`transformSvg`'s contract is preserved. When a transform is present the raw payload is parsed separately; root attribute names absent from it were added by the caller's hook and pass through, because caller intent is trusted. Fetched names always face the allowlist, so an attacker's `onload` can never ride in on a transform. If the raw payload does not parse on its own, nothing is attributable to the caller and the allowlist applies to everything — never wide open.

### Proof

Tests fail for the right reasons **before** the fix — the live host dump shows `onload=`, `onclick=`, `id="evil-root"`, `tabindex="0"`, `data-pw="hijacked"` all landing, and the child `onerror` surviving adoption.

The child test fires the handler for real rather than only asserting an attribute is absent: `<image>` with a bad `href` triggers `onerror` by itself, and the test then clicks the planted `onclick` too. Nothing is set on `<html>` afterwards.

**17 passed** after the fix, covering the new specs plus the existing `list-item-menu-svg-role`, `select-lefticon-inline-svg`, `icon-slot-color-token` and `status-icon-slot` suites. Payloads are served via `page.route`, so no network.

## Defect 2 — `showIndicator` never reached the web component

It shipped on `Choicebox` in #428 but was never added to `Choicebox.wc.svelte`, so `<sui-choicebox>` consumers could not reach it. Added following the pattern the file's other booleans use, with `docs/Choicebox.md` gaining the Web Component attribute table it was missing.

Caveat documented rather than hidden: Svelte's custom-element Boolean conversion is presence-based, so `show-indicator="false"` reads as `true`. Consumers hide the indicator via the property or by removing the attribute. That is platform behaviour for every Boolean-typed prop in this repo, not something the wrapper can change.

## Known remaining

When `src` changes, previously copied host attributes are not removed before the new copy — a pre-existing quirk, untouched here and worth its own fix.

### DIFF
--- a/docs/Img.md
+++ b/docs/Img.md
@@ -22,7 +22,7 @@ An image component with automatic fallback. If the primary `src` fails to load (
 | testId   | `string`         | No       | `-`     | Test identifier applied as `data-pw` attribute on the `<img>` element for Playwright test selectors.                                                                  |
 | classes  | `string`         | No       | `-`     | CSS class string applied to the component's top-level element. Useful for theming — define classes with CSS variable overrides and pass them to create variant styles. |
 | inlineSvg | `boolean`       | No       | `false` | Fetch `.svg` / `data:image/svg+xml` sources and inline their markup into an `<svg>` host so page CSS (`currentColor`, fill/stroke overrides) applies. Non-SVG sources always render the plain `<img>`. If fetching or parsing fails, the component falls back to the plain `<img>` (and from there to the regular `fallback`/`onerror` chain). |
-| transformSvg | `(svg: string) => string` | No | `-` | Hook to rewrite the fetched SVG markup before it is inlined (e.g. recolour hardcoded fills). Providing a transform implies `inlineSvg`. The inlining effect re-runs when the prop identity changes, so a closure over reactive state (e.g. a theme colour) re-renders live. JS-only prop (not exposed as a web-component attribute). |
+| transformSvg | `(svg: string) => string` | No | `-` | Hook to rewrite the fetched SVG markup before it is inlined (e.g. recolour hardcoded fills). Providing a transform implies `inlineSvg`. The inlining effect re-runs when the prop identity changes, so a closure over reactive state (e.g. a theme colour) re-renders live. Root attributes the transform adds are treated as caller intent and copied to the host even when outside the fetched-attribute allowlist (see SVG inlining below). JS-only prop (not exposed as a web-component attribute). |
 
 ## SVG inlining
 
@@ -44,6 +44,8 @@ Behaviour notes:
 
 - Non-SVG sources ignore both props and always render the plain `<img>`.
 - A failed fetch/parse for a given URL falls back to the plain `<img>` for that URL (which then drives the normal `fallback`/`onerror` chain); a changed `src` gets a fresh inlining attempt.
+- The fetched document is treated as untrusted content: only an allowlist of presentational, geometry, and accessibility attributes is copied from its root onto the live `<svg>` host — `xmlns`, `xmlns:xlink`, `viewBox`, `width`, `height`, `x`, `y`, `preserveAspectRatio`, `fill`/`fill-*`, `stroke`/`stroke-*`, `clip-rule`, `color`, `opacity`, `overflow`, `role`, `focusable`, and `aria-*`. Anything else from the fetched file — event handlers (`onload`, `onclick`, …), `id`, `style`, `class`, `data-*` — never reaches the live element, so a hostile SVG cannot inject script or clobber the `data-pw` test hook. Child elements are inlined as-is.
+- Root attributes that your own `transformSvg` hook **adds** (names not present on the fetched root) are caller intent and pass through unfiltered — e.g. a transform can stamp a `data-*` marker on the host.
 - Accessibility: the `<svg>` host gets `role="img"` + `aria-label` when `alt` is non-empty, and `aria-hidden="true"` when `alt` is empty (decorative).
 - All CSS variables below apply to the inlined `<svg>` host exactly as they do to the `<img>`.
 
--- a/tests/img-inline-svg-attr-allowlist.spec.ts
+++ b/tests/img-inline-svg-attr-allowlist.spec.ts
@@ -1,0 +1,175 @@
+import { expect, test } from '@playwright/test';
+
+// Img's inlineSvg path fetches remote markup and copies the fetched root's
+// attributes onto the live <svg> host — an element in the host document. The
+// fetched file is untrusted content: copying every attribute lets it plant
+// `onload`/`onclick` handlers (script injection via an image URL), clobber the
+// data-pw test hook, or override the component's sizing with an inline style.
+// Only an allowlisted presentational/geometry/a11y set may survive; attributes
+// the caller's own transformSvg ADDS are caller intent and still pass.
+//
+// Payloads are served via route interception so the tests are self-contained:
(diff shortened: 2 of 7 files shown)


Tara-ag (commented):
Review in progress

docs/Img.md
```diff
@@ -44,6 +44,8 @@ Behaviour notes:
 
 - Non-SVG sources ignore both props and always render the plain `<img>`.
 - A failed fetch/parse for a given URL falls back to the plain `<img>` for that URL (which then drives the normal `fallback`/`onerror` chain); a changed `src` gets a fresh inlining attempt.
+- The fetched document is treated as untrusted content: only an allowlist of presentational, geometry, and accessibility attributes is copied from its root onto the live `<svg>` host — `xmlns`, `xmlns:xlink`, `viewBox`, `width`, `height`, `x`, `y`, `preserveAspectRatio`, `fill`/`fill-*`, `stroke`/`stroke-*`, `clip-rule`, `color`, `opacity`, `overflow`, `role`, `focusable`, and `aria-*`. Anything else from the fetched file — event handlers (`onload`, `onclick`, …), `id`, `style`, `class`, `data-*` — never reaches the live element, so a hostile SVG cannot inject script or clobber the `data-pw` test hook. Child elements are inlined as-is.
+- Root attributes that your own `transformSvg` hook **adds** (names not present on the fetched root) are caller intent and pass through unfiltered — e.g. a transform can stamp a `data-*` marker on the host.
```
Tara-ag, line 48: 💡 MINOR: Docs say child elements are inlined 'as-is' but script/foreignObject are stripped. `sanitizeInlinedSubtree` removes `<script>` and `<foreignObject>` descendants and strips `on*` handlers plus `javascript:` URLs before adoption, so the wording is inaccurate. Update the bullet to describe that child elements are sanitised before inlining.

tests/img-inline-svg-attr-allowlist.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+      '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none"' +
+      " onload=\"document.documentElement.setAttribute('data-svg-onload-ran', '1')\"" +
+      " onclick=\"document.documentElement.setAttribute('data-svg-onclick-ran', '1')\"" +
+      ' id="evil-root" style="outline: 6px solid red" data-pw="hijacked" tabindex="0">' +
+      '<circle cx="12" cy="12" r="8" fill="currentColor" /></svg>';
+    await serveSvg(page, INLINE_DEMO_URL, hostile);
+    await page.goto('/components/img');
+
+    // Located through the wrapper row, not data-pw on the host itself — a
+    // successful data-pw clobber must fail an assertion, not hide the element.
```
Tara-ag, line 41: 💡 MINOR: Img sanitisation tests are coupled to demo-page markup. Navigating `/components/img` and selecting through demo-specific test IDs (`img-inline-svg-row`, `img-inline-svg-transform-row`) makes these security assertions fragile when the demo page changes. Prefer an isolated fixture route or component harness so the test targets the component directly, consistent with the repo's standard for robust component tests.

Tara-ag (approved):
Approved with minor notes on test isolation and doc accuracy.

tests/img-inline-svg-attr-allowlist.spec.ts
Tara-ag, line 41: 💡 MINOR: Img sanitisation tests are coupled to demo-page markup. This test navigates /components/img and selects elements via demo-specific test IDs (img-inline-svg-row, img-inline-svg-transform-row). This makes the security assertions fragile when the demo page changes. The repo's recurring standard prefers isolated Vitest/component-harness tests or dedicated fixture routes over tests that navigate demo pages.

Suggestion: Move the inline-SVG security assertions to a dedicated fixture route (e.g. /tests/img-inline-svg) or a component harness so the test selects the component directly rather than through demo-page wrappers.

docs/Img.md
Tara-ag, line 48: 💡 MINOR: Docs say child elements are inlined 'as-is' but script/foreignObject are stripped. docs/Img.md states 'Child elements are inlined as-is' in the SVG inlining section, but Img.svelte's sanitizeInlinedSubtree removes <script> and <foreignObject> descendants and strips on* handlers plus javascript: URLs before adoption. The docs should accurately describe the sanitisation so consumers understand why legitimate <script> or <foreignObject> content does not appear.

Suggestion: Update the bullet to: 'Child elements are inlined after stripping <script>, <foreignObject>, event-handler attributes, and javascript: URLs, so executable content cannot survive adoption.'

Tara-ag (approved):
Approved with two minor notes on test isolation and documentation accuracy.

docs/Img.md
murdore: Confirmed and fixed in https://github.com/juspay/svelte-ui-components/pull/500.

"Child elements are inlined as-is" was true before this PR and stopped being true inside it — `sanitizeInlinedSubtree` was added precisely because the root allowlist alone was bypassable, and I did not update the sentence describing the old behaviour. Documentation understating a security boundary is worth more than a minor: a reader deciding whether it is safe to inline a third-party SVG would have concluded it is not.

The bullet now states what is actually stripped — `<script>` and `<foreignObject>` subtrees, `on*` at any depth, `javascript:` in `href`/`xlink:href`/`src` — and that everything else about a child survives.

murdore: Confirmed and fixed in https://github.com/juspay/svelte-ui-components/pull/500.

"Child elements are inlined as-is" was true before this PR and stopped being true inside it — `sanitizeInlinedSubtree` was added precisely because the root allowlist alone was bypassable, and I did not update the sentence describing the old behaviour. Documentation understating a security boundary is worth more than a minor: a reader deciding whether it is safe to inline a third-party SVG would have concluded it is not.

The bullet now states what is actually stripped — `<script>` and `<foreignObject>` subtrees, `on*` at any depth, `javascript:` in `href`/`xlink:href`/`src` — and that everything else about a child survives.

PR #500, opened 2026-09-02
fix(review): resolve the findings left open on merged PRs

murdore:
Review comments that landed on #494–#499 after I reported each PR done, and were never addressed before merge. #491 is the only one of the six that was clean — all ten of its CodeRabbit findings were addressed and confirmed in-thread.

Everything here was verified against the merged code before anything changed. Three findings turned out not to hold and are left alone, with the check written up below.

---

## The codemod migrated Stepper consumers onto a deprecated prop

Filed MAJOR on #495, and correctly.

`StepperProperties` marks `onhandleStepClick` *`@deprecated Use onstepclick instead.`* — and `map.ts` targeted it. Anyone running the codemod landed on the one spelling of that event scheduled for removal, so they owed a second migration immediately.

The cause is the derivation rule the map's own header describes as sound:

> a pair was accepted only when `lower(suiName) === lower(polyName)`

That rule is deprecation-blind. Where a component keeps a deprecated alias whose lowercase spelling matches the fork's, the alias wins the case-insensitive match and outranks the canonical prop. Stepper is the only affected pair of the 28.

The target is now `onstepclick`. `onhandleStepClick` is still recognised as a *source* in the `to-poly` direction, so a SUI consumer on either spelling still converts.

**`map.ts` had no test file at all — that is how this got through.** `map.test.ts` now asserts:

- no rewrite target is `@deprecated`
- every target is actually declared by the component
- a pair may differ by more than case *only* to step over a deprecated alias

It fails on the old map.

---

## Two visual stabilisers gave up silently — and hid a real bug

Filed MINOR on #496. Making them fail loudly instead turned out to matter a lot more than that.

`fitViewportToContent` returned after 4 attempts whether or not the height had settled. **Two demos never settle.** The demo shell is `min-height: 100vh`, so a page that adds its own full-height element grows by exactly what the viewport gains — fitting it is an infinite loop.

Measured directly, against a fixed control:

| Route | Content height vs viewport | Sized against |
| --- | --- | --- |
| `table` | 36388px at every viewport | its own content |
| `brand-loader` | grows **1×** the viewport gain | the viewport |
| `status` | grows **3×** the viewport gain | the viewport |

The loop was running out of attempts and capturing at whatever size the budget happened to leave — a size decided by the iteration count, not by the page. That is why **`brand-loader` is one of the routes that kept flaking** during this work.

`status` was the clearest symptom: its committed baseline had inflated to **60209px, roughly 92% stretched empty space**, because every attempt inflated the viewport and the page inflated to match.

Growth that tracks the viewport is now detected and the page pinned to the default viewport. A page moving for some third reason throws rather than being captured mid-scroll.

Both baselines are regenerated. **`status` drops to 2609px — 84% smaller — with all seven demo variants still in frame.** I checked that by looking at the image, not by trusting the byte count.

`waitForImages` had the same shape of bug: it snapshotted `document.images` once, so an image appended while the first batch settled was never awaited. It rescans each round.

---

## `docs/Img.md` understated a security boundary

The doc said child elements are *"inlined as-is"*. True before #498, false after it — `sanitizeInlinedSubtree` was added in that PR precisely because the root allowlist alone was bypassable, and the sentence describing the old behaviour was left behind.

Filed MINOR; worth more than that. A reader deciding whether it is safe to inline a third-party SVG would have concluded it is not. The bullet now states what is actually stripped: `<script>` and `<foreignObject>` subtrees, `on*` at any depth, and `javascript:` in `href`/`xlink:href`/`src`.

---

## The CI comment still asserted a claim I had already withdrawn

`ci.yml` still carried the *"11 consecutive green runs across 8 different branches"* figure. That was withdrawn during review of #494 itself — the first run with the gate actually enabled caught `splitinput-autofill-distribute` failing all three attempts — but I corrected only the PR body, and the workflow is the copy that outlives the PR.

The field was backwards too: `continue-on-error: true` rewrites the **step's** `conclusion` to success, so reading `conclusion` was exactly the wrong thing to do. `outcome` is what survives it.

---

## Two smaller ones

- **`scripts/visual-test.sh` documented a flag that errors out.** Its usage block advertised `--update`; `playwright test --update` exits with `unknown option`. `package.json` already used `--update-snapshots`, so only the path through the script's own header was broken — i.e. the path someone takes when they read the script instead of the package scripts.
- **Nothing enforced the container/package lockstep in CI.** `visual.yml` only asked for it in a comment. It now checks that the image ships the browser Playwright expects, and names the three files to update on a drift.

---

## Checked and deliberately not changed

**`playwright install --with-deps chromium` in `visual.yml`** (suggested as MAJOR). This would defeat the job. The container is pinned so a baseline made on a laptop and one checked in CI are the same measurement; `playwright install` downloads whatever browser the npm package wants and shadows the image's. If the versions agree it is a no-op; if they disagree it silently makes CI rasterise differently from every developer — converting a loud failure into exactly the drift these pinned baselines exist to catch. The lockstep check above addresses the real concern underneath it.

**The image tag version parse**, reported to break on `v1.60.0-jammy-amd64`. It does not: `%%-*` strips from the *first* hyphen.

```
v1.60.0-noble        -> 1.60.0
v1.60.0-jammy-amd64  -> 1.60.0
v1.60.0              -> 1.60.0
v1.61.0-rc-1-noble   -> 1.61.0
```

**Route discovery**, reported as scanning 97 directories with a `SKIP_SLUGS` filter. There are **94** directories and no such constant — the filter is `EXCLUDED`, which emits *visible skipped tests* rather than dropping routes. 92 baselined + 2 excluded accounts for all 94. Deriving the list from `_nav.ts` instead would weaken the guarantee: a demo present on disk but missing from the nav would go unscreenshotted, which is the coverage rot the disk scan was chosen to prevent.

**A workflow-level `env` for the pnpm version.** `env` is per-workflow, so it cannot deduplicate a version pinned in both `ci.yml` and `visual.yml`, and it would make `visual.yml` diverge from `ci.yml`, which pins it inline the same way. The fix that would actually collapse both is a `packageManager` field in `package.json` — there isn't one, and adding it changes local resolution for everyone in the repo, so it is a deliberate change rather than a review-sweep edit. Happy to do it separately.

---

## Also carries the five findings from #499

Those opened this PR before the sweep:

1. **Svelte peer range checked by major only** — accepted `5.0.0` (shares the major, does not satisfy `^5.41.2`) and rejected `^4 || ^5` (genuinely intersects). Now `semver.intersects`.
2. **`showBackButton` detection used `JSON.stringify(value).includes('false')`** — a `MemberExpression` carries `"computed":false`, so `<Toolbar showBackButton={cfg.showBack}>` read as *disabled* and its affected usage went unreported. The worst of the five, because a false negative reports a clean consumer that is not clean. Confirmed empirically before fixing; now tests the AST for the Boolean literal.
3. **The legacy-selector finding reported the `<style>` tag's line**, not the selector's.
4. **`--target` was spliced into `package.json` as raw text** — a quote could close the string and inject sibling properties. Now `JSON.stringify`d.
5. **Direct-invocation detection compared a basename split on `/`**, which on Windows yields the whole path and makes the CLI silently decline to run.

---

## Verification

| Check | Result |
| --- | --- |
| `scripts/codemod` | **97/97** (was 96 — and `map.ts` had no test file) |
| `scripts/migrate` | **36/36** |
| unit suite | **285/285** |
| `lint` / `check` | clean |
| visual, pinned container | **92 passed, 2 skipped** — run twice end to end |

The visual suite was run in `mcr.microsoft.com/playwright:v1.60.0-noble` rather than trusted from unit tests. That is what caught my own first attempt at the viewport fix, which compared content growth against the content baseline instead of the viewport gain and so fired on nearly every route.

### Unrelated defect noticed while reviewing the regenerated baseline

`/components/icons/order-success-icon.svg` 404s on the `status` demo, so several of its cards render broken images. Pre-existing and out of scope here — flagging it rather than folding it in.

---

# Second round — review on this PR

Seven comments landed on #500 itself. All verified; six fixed, one kept with evidence.

## `isFalseLiteral` missed the array form — MAJOR

Svelte types an attribute value as `true | ExpressionTag | Array<Text | ExpressionTag>`, and **quoting a single expression produces the one-element array**:

| Written | AST shape | Old result | Correct? |
| --- | --- | --- | --- |
| `showBackButton={false}` | `ExpressionTag` | disabled | ✓ |
| `showBackButton="false"` | `Array[Text]` | flagged | ✓ |
| `showBackButton={cfg.x}` | `ExpressionTag` | flagged | ✓ |
| `showBackButton="{false}"` | `Array[ExpressionTag]` | **flagged** | ✗ |

Worth being precise about severity: this **over-reports** a Toolbar whose control genuinely never renders. It fails in the safe direction, unlike the false negative it replaced. Test written first and watched fail.

## The viewport-relative heuristic could misclassify legitimate growth

On code added earlier in this same PR. `height - previousHeight >= viewportGain - 1` treats *any* growth past the viewport gain as viewport-relative — so a page that reveals a large section when given room would be pinned to the default viewport and **lose everything below the fold**.

The suggested fix was an allowlist of `brand-loader` and `status`. Not taken: hardcoding the answer means a future viewport-relative demo regresses silently to exactly the behaviour this PR removes.

The classification is now **confirmed rather than inferred** — shrink the viewport and check the content follows it down. Content sized against the viewport shrinks with it; content that was merely revealed does not. Pinned routes are logged, which was the good half of the suggestion.

Verified on the full suite: **exactly 2 routes pinned, the other 90 untouched.** A single delta could not have discriminated those.

## Kept, with evidence: `{ loose: true }`

Reported as possibly accepting prerelease ranges unexpectedly. Measured across 18 ranges — `loose` changes the answer for exactly **one**:

```
range              loose  strict
^5.41.2-alpha      true   true
>=5.0.0-0          true   true
workspace:*        THROW  THROW
^05.41.2           true   THROW   <-- the only difference
```

A leading-zero version is where strict parsing throws and would report a blocker that does not exist. Both prerelease examples named in the review resolve identically either way, so the mode is **safer, not looser**. Kept, documented in the code, and pinned by a test.

## Also fixed

- **Entrypoint guard used `undefined`**, which this repo's ESLint bans. It only escaped because `lint` covers `src` alone — the same gap reported on #495 and still open.
- **The `--target` injection test never asserted the apply succeeded**, so a refusal would have satisfied every assertion in it. Now checks `exitCode` and `applied`.
- **`waitForImages` re-added listeners** to an image still pending on the next round. Memoised in a `WeakMap`.
- **`docBlockFor` interpolated a prop name into a regex unescaped.** Every current name is a plain identifier and the first test asserts each resolves, so a miss already failed loudly — escaped anyway.

## Final verification

| Check | Result |
| --- | --- |
| unit | **286/286** |
| `scripts/` (codemod + migrate) | **134/134** |
| `lint` / `check` | clean |
| visual, pinned container | **92 passed, 2 skipped, 0 failed** — exactly 2 routes pinned |


Tara-ag (commented):

scripts/migrate/analyze.ts
```diff
@@ -104,6 +116,16 @@ type AttributeSummary = {
   readonly hasSpread: boolean;
 };
 
+/** True only for `{false}` — not for a bound expression or the string "false". */
```
Tara-ag, line 124: ⚠️ MAJOR: `isFalseLiteral` assumes `attribute.value` is a single `ExpressionTag`, but Svelte 5 modern AST types `Attribute.value` as `true | ExpressionTag | Array<Text | ExpressionTag>`. A quoted single-expression value such as `showBackButton="{false}"` parses to a one-element array `[{type:'ExpressionTag', expression: {type:'Literal', value:false}}]`, so this helper would fail to recognise it as disabling the control. The Svelte compiler utility `is_expression_attribute` explicitly normalises both shapes. Handle the one-element array case, and add a regression test for `showBackButton="{false}"`.

Suggested fix:
```ts
function isFalseLiteral(value: unknown): boolean {
  let tag: UnknownRecord;
  if (Array.isArray(value) && value.length === 1) {
    tag = asRecord(value[0]);
  } else {
    tag = asRecord(value);
  }
  if (tag.type !== 'ExpressionTag') {
    return false;
  }
  const expression = asRecord(tag.expression);
  return expression.type === 'Literal' && expression.value === false;
}
```

scripts/migrate/cli.test.ts
```diff
@@ -81,3 +81,31 @@ describe('migrate cli', () => {
...
+
+describe('review findings', () => {
+  it('serializes --target so a quote cannot corrupt or extend the manifest', () => {
+    const root = project({
+      name: 'consumer',
+      dependencies: { [LIB]: '2.136.0', svelte: '^5.55.9' }
+    });
+
+    // Splicing this in raw would close the string and inject a sibling key.
+    run([root, '--apply', '--target', '^3.0.0", "evil": "yes'], silent);
```
Tara-ag: 💡 MINOR: This test verifies the manifest is not corrupted by a quoted `--target`, but it does not assert that the migration actually succeeded. Capture the returned summary and check `exitCode` so the test also guards against an unexpected refusal to apply.

```ts
const summary = run([root, '--apply', '--target', '^3.0.0", "evil": "yes'], silent);
expect(summary.exitCode).toBe(0);
```

Tara-ag:

## 🤖 Yama Review Summary

⚠️ The review loop ended early (`time-limit`, 15 steps) — this verdict is built strictly from gate-verified findings.

scripts/codemod/map.test.ts
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * The block sits immediately above the declaration, so walk upwards over the
+ * contiguous comment lines and stop at the first line that is not one.
+ */
+function docBlockFor(source: string, prop: string): string | undefined {
+  const lines = source.split('\n');
+  const declaration = new RegExp(`^\\s*${prop}\\??\\s*:`);
+  const index = lines.findIndex((line) => declaration.test(line));
+  if (index === -1) {
+    return undefined;
```
Tara-ag, line 18: 💡 MINOR: Test helper regex may miss props with computed names

docBlockFor builds a regex from the raw prop name and anchors it to the start of a line. While all current props are simple identifiers, the regex does not escape the prop name, so a prop containing regex metacharacters would be mis-detected. More importantly, the helper assumes every prop declaration is on its own line beginning with optional whitespace followed by the prop name; this is true for the current properties.ts files but is brittle if declarations are reformatted or use JSDoc tags on the same line.

Suggestion: Escape `prop` with `prop.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')` before interpolating into the regex, or switch to parsing the file with TypeScript so the test is robust to formatting.

scripts/migrate/analyze.ts
```diff
@@ -40,10 +41,22 @@ function readDependency(manifest: unknown, name: string): string | null {
...
+ * 4 to a first-number read while genuinely intersecting. Range intersection is
+ * the actual question being asked.
+ */
+function satisfiesPeer(range: string): boolean {
+  try {
+    return intersects(range, SVELTE_PEER_RANGE, { loose: true });
+  } catch {
+    // An unparseable range (a git URL, a workspace protocol) cannot be shown to
+    // satisfy the peer, and silently passing it would be the wrong default.
+    return false;
```
Tara-ag, line 63: 💡 MINOR: analyze.ts uses loose semver intersection which may accept prerelease ranges unexpectedly

satisfiesPeer uses `intersects(range, SVELTE_PEER_RANGE, { loose: true })`. With loose: true, invalid ranges may be coerced in surprising ways. For a migration gate, strict parsing is safer; the current tests do not cover edge cases like `^5.41.2-alpha` or `>=5.0.0-0`.

Suggestion: Consider removing `{ loose: true }` and adding test cases for prerelease ranges. If you keep loose mode, document why in a comment.

tests/visual/components.visual.ts
```diff
@@ -137,22 +152,56 @@ const MAX_VIEWPORT_HEIGHT = 20_000;
...
+      if (viewportGain > 0 && height - previousHeight >= viewportGain - 1) {
+        await page.setViewportSize({ width: VIEWPORT_WIDTH, height: MIN_VIEWPORT_HEIGHT });
+        return;
+      }
+    }
+
+    seen.push(height);
+    previousViewport = viewport;
+    // Clamped, not rejected: a demo past the cap is captured by scrolling, the
+    // same as before. The cap is here so a runaway page cannot ask for a buffer
```
Tara-ag, line 237: 💡 MINOR: fitViewportToContent heuristic can misclassify legitimate growth as viewport-relative

The detection `height - previousHeight >= viewportGain - 1` treats any content growth that is at least the viewport gain as viewport-relative. If a page genuinely grows by slightly more than the viewport gain due to a newly revealed section that itself contains a vh-sized element, it will be pinned to MIN_VIEWPORT_HEIGHT and lose content below the fold. The heuristic is reasonable for the known bad demos but is not generally safe for future demos.

Suggestion: Add an allowlist/denylist for the known viewport-relative routes (brand-loader, status) instead of inferring it dynamically, or at minimum cap the number of times the heuristic can trigger and log the pinned routes so future regressions are discoverable.

tests/visual/components.visual.ts
Tara-ag, line 1: 💡 MINOR: waitForImages event listeners leak if the same image is pending across rounds

Each round adds fresh load/error listeners to images that were incomplete. If an image stays incomplete for multiple rounds (slow connection, broken src), listeners accumulate. In practice the test aborts after 5 rounds, but the accumulation is unnecessary and could be avoided by tracking which images have already been awaited.

Suggestion: Track awaited images in a WeakSet and skip images already being watched, or use a single MutationObserver + Promise.race pattern instead of polling.

Tara-ag (commented):
Reviewed file-by-file. Found no blocking issues (no hardcoded secrets, no XSS, no breaking public API changes). Four MINOR findings were raised and posted inline:

1. `scripts/codemod/map.test.ts`: brittle regex in test helper.
2. `scripts/migrate/analyze.ts`: loose semver intersection may accept prerelease ranges unexpectedly.
3. `tests/visual/components.visual.ts`: viewport-fitting heuristic may misclassify future demos.
4. `tests/visual/components.visual.ts`: listener accumulation in waitForImages.

The PR fixes real review findings from merged PRs, adds good regression tests, improves CI gating, and updates documentation accurately. Safe to approve with the noted nits.

Tara-ag (approved):
Approved. No blocking issues found. The changes correctly fix the codemod target selection for Stepper, harden the migration analyzer's semver and AST checks, secure the CLI manifest rewrite, and stabilize the visual regression harness with appropriate test coverage and documentation updates.

scripts/migrate/analyze.ts
murdore: Confirmed and fixed, with a regression test for both quoted forms.

Verified the shapes against the parser rather than the type alone:

```
showBackButton={false}     => ExpressionTag
showBackButton="false"     => ARRAY len 1 of Text
showBackButton={cfg.x}     => ExpressionTag
showBackButton="{false}"   => ARRAY len 1 of ExpressionTag   inner: Literal false
```

One correction to the severity, since it matters for how much to worry: this errs the *safe* way. Missing the array shape means a Toolbar that genuinely never renders the control gets **reported as affected** — over-reporting, not the false negative it replaced. Still wrong, and still noise that erodes trust in a "0 findings" result.

Test cases added for `showBackButton="{false}"` (expect 0 findings) and `showBackButton="{cfg.showBack}"` (expect 1), written first and watched fail.

tests/visual/components.visual.ts
murdore: Fixed — and this was worth catching, because the code was mine from earlier in this same PR.

You are right about the failure mode: a page that reveals a large section when given room grows by more than the viewport did, gets classified as viewport-relative, and is pinned to the default viewport, silently losing everything below the fold. A single delta cannot distinguish the two causes.

I did not take the allowlist, though. Hardcoding `brand-loader` and `status` fixes today's two routes at the cost of the general property: a future viewport-relative demo would then regress silently to exactly the behaviour this PR removes, and nothing would say so.

Instead the classification is now **confirmed** rather than inferred. When growth looks viewport-relative, shrink the viewport and check the content follows it down — content sized against the viewport shrinks with it, content that was merely revealed does not. And pinned routes are logged, which was the good half of the suggestion.

Verified on the full suite: **exactly 2 routes pinned, the other 90 untouched.**

scripts/migrate/cli.ts
```diff
@@ -172,8 +182,12 @@ export function run(argv: readonly string[], log: (line: string) => void): Migra
   };
 }
 
+// Comparing against a basename split on '/' fails on Windows, where argv[1]
+// uses backslashes and the split yields the whole path -- the CLI then silently
+// declines to run. Normalising both sides to a file URL is platform-agnostic.
+const entrypoint = process.argv[1];
 const invokedDirectly =
-  process.argv[1] !== undefined && import.meta.url.endsWith(process.argv[1].split('/').pop() ?? '');
+  entrypoint !== undefined && import.meta.url === pathToFileURL(resolve(entrypoint)).href;
```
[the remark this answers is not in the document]
murdore: Fixed — and the reason it slipped past CI is worth recording: `lint` runs `eslint src`, so `scripts/` is not linted at all. The rule was never going to fire here.

That gap is itself an open finding on #495 (new `scripts/codemod/` files outside Prettier and ESLint) and is still unaddressed. It is not in this PR's scope, but it is the reason this class of thing keeps reaching review instead of failing locally.

scripts/migrate/cli.test.ts
murdore: Good catch, and a sharper point than it looks: without asserting the apply happened, **every remaining assertion in that test passes vacuously if the run refuses to apply** — an untouched manifest is trivially uncorrupted. The test would have gone green while testing nothing.

Now asserts `exitCode` is 0 and `applied` is true before checking the manifest.

scripts/migrate/analyze.ts
murdore: Measured before deciding, across 18 ranges. `loose` changes the answer for exactly one of them:

```
range              loose  strict
^5.41.2-alpha      true   true
>=5.0.0-0          true   true
5.0.0              false  false
^4 || ^5           true   true
workspace:*        THROW  THROW
^05.41.2           true   THROW   <-- the only difference
```

Both prerelease ranges you named resolve identically either way, so the suspected effect does not occur. The one case that differs is a leading-zero version, where strict parsing throws and would therefore be reported as a blocker that does not exist — so `loose` is marginally *safer* here, not looser.

Taking your second option: keeping it, with the reasoning now in a comment at the call site, plus a test pinning all three ranges so a future change to the mode has to confront this.

tests/visual/components.visual.ts
murdore: Fixed — watches are memoised in a `WeakMap`, so an image still pending on the next round reuses its existing promise instead of getting a second pair of listeners.

scripts/codemod/map.test.ts
murdore: Escaped, thanks.

Worth noting the helper is partly self-checking already: the first test in the file asserts that every prop in the map resolves to a declaration, so a name the regex failed to match would fail there rather than silently returning "not deprecated". That is what keeps the brittleness from turning into a false pass. Escaping is still correct and free, so it is in.

PR #501, opened 2026-09-02
fix(codemod): lint scripts and tests, and parse module specifiers

murdore:
Closes the last four findings left open on #495, three of which needed the fourth fixed first.

## `lint` never looked at `scripts/` or `tests/`

```
"lint": "prettier --check src && eslint src && npm run lint:event-casing"
```

Two whole directories were outside every rule this repo enforces. That is the mechanism behind the banned-`undefined` guard that reached review on #500 instead of failing locally — the rule existed, it just never ran there.

Extending the script to `src scripts tests` surfaced **31 real violations**:

| Where | What |
| --- | --- |
| `analyze.ts` | `value as UnknownRecord` — assertions are banned outside tests |
| `signatures.ts`, `casing.ts`, `map.test.ts` | 6 × `undefined`, banned repo-wide |
| `check-event-casing.js` | 4 × missing `curly` braces, plus unformatted |
| `tests/` | 6 unformatted files, 6 more `undefined`, an unnecessary escape, an unused locator |

The interesting one is `asRecord`. Assertions are banned **and so are type predicates**, so there is no way to tell the compiler that a narrowed `object` is keyed by string. It now copies the own enumerable keys. The copy is shallow, so nested nodes keep their identity and the AST walk is unaffected — verified end to end below rather than assumed.

For the rest, `undefined` gave way to `null`, and `forwardsDomEvent` / `deriveEventName` now take `unknown`, which is what their bodies already narrowed to anyway.

## The specifier rewrite matched inside comments and strings

The old code was a context-anchored regex, justified in its own comment:

> module specifiers are string literals in a fixed syntactic position (after `from`, `import`, `import(`, or `require(`), so a context-anchored regex cannot touch arbitrary strings

The premise is true and the conclusion does not follow — **those words appear just as readily in prose**. A fixture naming the package in a line comment, a block comment, a string literal, and one real import rewrote **all four**:

```
expected importsRewritten 1, got 4
```

It now parses with TypeScript, which is already a devDependency and already type-checks this directory, so the exact answer costs nothing. Only the literal's interior is replaced, so the original quote style survives. Static imports, re-exports, dynamic `import()`, `require()` and `import()`-in-type-position all keep working, each covered by a test.

## Two smaller ones

- **`cli.test.ts` shared `dir` and `lines` as module-level mutable state.** Each test now builds its own project, removed through `onTestFinished` whether it passes or throws. One correction to the finding: its stated reason — that `afterEach` would not run if a test throws — is not accurate, Vitest runs it either way. The shared mutable state was still worth removing.
- **Both script tsconfigs included `./*.ts`**, so a nested file would silently escape `check:codemod` / `check:migrate`. Now `./**/*.ts`.

## Verification

| Check | Result |
| --- | --- |
| unit | **288/288** |
| `scripts/` (codemod + migrate) | **136/136** (was 133) |
| `lint` / `check`, widened scope | clean |
| migrate CLI vs lighthouse | **385 files, 0 blockers, 0 findings** — identical to before the `asRecord` rewrite |

A zero from a working detector and a zero from a broken one look the same, so that last row was checked against a control:

```
<Toolbar showBackButton={false} />        -> silent   ✓
<Toolbar showBackButton="{false}" />      -> silent   ✓
<Toolbar showBackButton={cfg.showBack} /> -> FLAGGED  ✓
<Toolbar />                               -> FLAGGED  ✓
```

Both guard forms recognised, both unguarded usages caught — including the quoted form fixed in #500, confirmed here through the real CLI rather than a unit test.

## Summary by CodeRabbit

- **Improvements**
  - Codemod transformations now more reliably update module references across imports, exports, dynamic imports, `require()` calls, and type references while avoiding unrelated text.
  - Linting and formatting checks now cover source code, scripts, and tests.
  - Developer tooling now includes TypeScript files in nested directories.

- **Tests**
  - Added coverage for module-reference transformations and improved test isolation.
  - Expanded accessibility and invalid-input coverage for progress indicators.


scripts/migrate/analyze.ts
```diff
@@ -26,8 +26,23 @@ export type ManifestReport = {
...
+ * Copies the own enumerable keys rather than asserting the type. Assertions are
+ * banned repo-wide and so are type predicates, so there is no way to tell the
+ * compiler that a narrowed `object` really is keyed by string. The copy is
+ * shallow, so nested nodes keep their identity and traversal is unaffected.
+ */
 function asRecord(value: unknown): UnknownRecord {
-  return typeof value === 'object' && value !== null ? (value as UnknownRecord) : {};
+  if (typeof value !== 'object' || value === null) {
+    return {};
+  }
```
Tara-ag, line 51: 💬 SUGGESTION: Consider documenting why `Reflect.get(value, key)` is preferred over indexed access here. Future readers may not immediately see that this avoids a type assertion or predicate, both of which are banned repo-wide.

scripts/migrate/analyze.ts
```diff
@@ -26,8 +26,23 @@ export type ManifestReport = {
...
+ * Reads an unknown value as a record of unknown fields.
+ *
+ * Copies the own enumerable keys rather than asserting the type. Assertions are
+ * banned repo-wide and so are type predicates, so there is no way to tell the
+ * compiler that a narrowed `object` really is keyed by string. The copy is
+ * shallow, so nested nodes keep their identity and traversal is unaffected.
+ */
 function asRecord(value: unknown): UnknownRecord {
-  return typeof value === 'object' && value !== null ? (value as UnknownRecord) : {};
+  if (typeof value !== 'object' || value === null) {
```
Tara-ag, line 49: 💡 MINOR: `asRecord` copies own enumerable keys but may drop symbol-keyed properties. The Svelte AST nodes walked here use string keys, so this is safe in practice, but the helper is generic and the limitation is undocumented. Consider switching to `Reflect.ownKeys(value)` if you want to preserve symbols, or document that only string keys are supported.

Tara-ag (approved):
Reviewed file-by-file. The PR fixes the lint scope gap and replaces a regex-based module-specifier rewrite with a proper TypeScript parse, which correctly avoids rewriting comments and string literals. Tests are now isolated per-fixture and coverage is expanded for the new parser path. No public API, accessibility, SSR, or security concerns were found. Two minor inline notes on `scripts/migrate/analyze.ts` only.

scripts/codemod/transform.ts
```diff
@@ -435,40 +436,105 @@ export function transformSvelte(
...
+  }
+  if (ts.isExportDeclaration(node)) {
+    const specifier = node.moduleSpecifier;
+    if (typeof specifier !== 'undefined' && ts.isStringLiteralLike(specifier)) {
+      return specifier;
+    }
+    return null;
+  }
+  // `import('x')` and `require('x')`.
+  if (ts.isCallExpression(node) && node.arguments.length > 0) {
```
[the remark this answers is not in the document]
murdore: Confirmed and fixed — and this is a regression this PR introduced, not a pre-existing gap.

`import X = require('x')` puts its specifier on an `ExternalModuleReference`, not a call expression. The regex I replaced matched it **by accident**, through the bare `require(` context, so switching to a parser silently dropped syntax that used to work. Exactly the risk in swapping a loose matcher for a precise one: the loose version's accidental coverage disappears without a test to notice.

Reproduced first:

```
expected 'import Table = require('polymorph-ui-components')'
      to contain 'import Table = require('@juspay/svelte-ui-components')'
```

`specifierOf` now handles `ImportEqualsDeclaration` + `ExternalModuleReference`, with a regression test alongside the static, re-export, dynamic, `require()` and import-type cases.

scripts/migrate/analyze.ts
```diff
@@ -26,8 +26,23 @@ export type ManifestReport = {
...
+    return {};
+  }
+  const record: UnknownRecord = {};
+  for (const key of Object.keys(value)) {
+    record[key] = Reflect.get(value, key);
```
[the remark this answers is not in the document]
murdore: Confirmed and fixed. Your reproduction is exactly right, and the consequence is worse than a stray key.

`analyze.ts` reads a consumer's `package.json` through `JSON.parse`, which makes `__proto__` an ordinary own key. Assigning it while copying re-points the copy's prototype instead of adding a field, so `dependencies` then resolves **through the chain**. A manifest declaring neither svelte nor the library reported `svelteRange: '^5.55.9'` and zero blockers.

That is the same failure direction as the `showBackButton` bug earlier in this series: the tool says a consumer is fine when it is not.

Copied with `Object.defineProperty` now, which creates an own data property rather than invoking the setter. Regression test asserts `svelteRange` is null and blockers are reported for a manifest whose only key is `__proto__`.

scripts/migrate/signatures.ts
```diff
@@ -7,8 +7,9 @@ import { readFileSync } from 'node:fs';
  * `name?: type;` line by repo convention, and the only thing needed here is
  * enough of the type to tell a DOM event from a domain object.
  */
-export function readSignature(file: string, prop: string): string | undefined {
+export function readSignature(file: string, prop: string): string | null {
   const source = readFileSync(file, 'utf8');
   const match = new RegExp(`^\\s*${prop}\\??:\\s*([^;]+);`, 'm').exec(source);
```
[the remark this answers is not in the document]
murdore: Fixed — same hole I had just escaped in `map.test.ts` and missed here, which is a fair catch.

Every prop name reaching this today is a plain identifier, so nothing currently misbehaves, but the helper takes an arbitrary string and there is no reason for it to interpret one as a pattern.

scripts/migrate/analyze.ts
murdore: Documented. The block now says why both unusual choices are there: `Reflect.get` because indexing a value typed `object` is itself an error without an assertion (the thing being avoided), and `defineProperty` because a `__proto__` key from a parsed `package.json` would otherwise re-point the copy's prototype — which CodeRabbit demonstrated on this same function, so it is now a fixed bug rather than a hypothetical.

scripts/migrate/analyze.ts
murdore: Documented rather than switched, and you have the reason right: the values walked here are Svelte AST nodes and parsed `package.json` objects, neither of which carries symbol-keyed fields, and a symbol key is never what the callers ask for.

The docblock now states that only string keys are copied, so the limitation is explicit rather than implied by the `Object.keys` call. Moving to `Reflect.ownKeys` would widen the copy without any caller wanting the extra keys, and would pull in non-enumerable ones too.

PR #502, opened 2026-09-02
fix(a11y): close the fifteen open review findings from the component-reuse PRs

murdore:
## What this closes

The 2 September sweep of review threads on the merged component-reuse PRs (#398, #439, #444, #470, #474, #476, #477, #478, #479, #480, #482, #483, #484, #485, #487, #490, #491, #492) answered all 107 unresolved threads against release 3.1.0. Fifteen were genuinely open: the shipped code did not do what the reviewer asked. This PR lands all fifteen as one commit and each thread is answered with the file and line that closes it.

## Changes

**Accessibility**
- `Carousel` — the active dot carries `aria-current="true"`; the others carry nothing. A screen-reader user querying the dots could hear where each went but not which slide was showing. (#484 thread on `Carousel.svelte`)
- `ListItem` — `suppressRoleAndTabindex` now also drops `aria-selected`. State on a generic `div` is state on nothing. (#479)
- `Input` — `aria-invalid` is gated by `!actionInput`, exactly as the error border, message and label already were, so an action-style field never announces an error it gives no other sign of. Both the `<input>` and `<textarea>` branches. (#480)
- `Accordion` — the trigger gets an `id` and the panel becomes `role="region"` `aria-labelledby` that trigger, completing the WAI-ARIA accordion pattern the earlier `aria-controls` started. The region attributes are only set when a trigger is rendered, since a region without a name is worse than none. (#492)

**Behaviour**
- `Status` — the weight and colour defaults move from `.status-text` to `.status-text-default`, which only the plain `div` receives. A consumer asking for `statusTextTag="h2"` now gets the application's heading typography, which was the documented promise of that prop. (#479)
- `TypewriterText` — the bulk-reveal branch (`isStreaming` turning false) recomputes `revealedWordCount` from the revealed text, so a later continuation resumes from a count that saw the whole suffix. (#479)
- `lockBodyScroll` / `unlockBodyScroll` — the inline `overflow` the host had before the first lock is saved and handed back on the last unlock, instead of being cleared. A host running `overflow: clip` on `body` keeps it. (#479)

**Web components**
- `sui-status` exposes `status-text-tag`. (#479)
- `sui-chip-input` exposes `editable` as a boolean attribute and `onedit` as a property; without the declaration Svelte creates no accessor and assignments were silently inert. (#482, two threads)

**Docs and comments**
- `docs/ChipInput.md` and `docs/DateRangePicker.md` load the web-component bundle from the published GitHub Pages URL the README uses, instead of a bare package specifier a browser cannot resolve. (#482)
- `docs/ChipInput.md` documents `editable`/`onedit` on the custom element and records the 2.136.0 test-id rename (`-chip` → `-item-${index}`, `-input` → `-add`) as a migration note, since there is no alias and 3.0.0 has since put it behind a major boundary. (#479, #482)
- `docs/Combobox.md` states that `aria-label` names the listbox, not the input, shows naming the input through `inputProperties.label`, and adds `allow-create` to the attribute list. (#482, two threads)
- `tests/pill-line-height.spec.ts` describes itself as the regression guard it is; the hook predates it (5b7dc9d). (#444)

## Tests

Every behavioural change is pinned on the repo's sanctioned surface, the demo routes:

| spec | what it proves |
|---|---|
| `tests/carousel.test.ts` | the active dot, and only it, is `aria-current`; the mark follows a dot click |
| `tests/list-item-suppressed-aria-selected.test.ts` | the suppressed demo item (now also passed `ariaSelected`) renders no role, tabindex or `aria-selected` |
| `tests/input-error-announcement.test.ts` | a new demo field with `actionInput` and `forceError` carries neither `aria-invalid` nor `aria-describedby` |
| `tests/accordion-aria-controls.test.ts` | the panel is `role="region"` labelled by the trigger's real `id` |
| `tests/status-text-tag.test.ts` | the `div` carries `status-text-default`; the `h2` does not |
| `tests/modal-nested-scroll-lock.test.ts` | with `body.style.overflow = 'clip'` set beforehand, the last unlock restores `clip` |
| `tests/wc-custom-elements.spec.ts` | `sui-chip-input editable` renders the edit field on click and `onedit` receives `[new, previous]`; `sui-status status-text-tag="h2"` renders an `h2` without the div defaults |

Not pinned: the `TypewriterText` word-count resync, which has no demo-route hook that exposes the counter; it is a four-line pure recomputation reviewed by reading.

Two demo pages gained a case each (`/components/input`, `/components/list-item`) so the tests have something real to assert against.

## Verification

| gate | result |
|---|---|
| `pnpm run lint` (prettier on `src`, eslint, event casing) | clean |
| `pnpm run check` (svelte-check + codemod + migrate) | 0 errors, 4 pre-existing warnings |
| `pnpm exec playwright test` on the eight touched specs | 34 passed, including the 8 new tests |
| negative control | with the Carousel and Accordion fixes reverted, exactly their two new tests failed (9 others passed); fixes restored, run repeated green |
| `pnpm run check:wc` | 4 errors, all in untouched files (`Combobox.wc.svelte`, `LottiePlayer.wc.svelte`); pre-existing and not run by CI |

Proof clips (one per new test, recorded by Playwright during the green run) are attached as a comment.

## Summary by CodeRabbit

* **New Features**
  * Added `editable` and `onedit` support to the Chip Input web component.
  * Added `status-text-tag` support to the Status web component.
  * Improved accessibility for accordions, carousels, inputs, and list items.
  * Preserved existing page overflow styles when closing nested modal dialogs.
  * Improved Typewriter Text continuation after instant text reveals.

* **Bug Fixes**
  * Status headings now correctly inherit application typography.
  * Updated component examples and web-component documentation for accessibility and usage clarity.


murdore:
## Proof of testing — one clip per new test

Recorded by Playwright during the green run of the eight touched specs on this branch at 64f91c5be (34 passed). Each clip is a test that asserts while it records; the same eight failed exactly as expected in the negative control described in the description.

### 01 · Carousel — the active dot carries aria-current="true" and only that one; the mark follows a dot click (tests/carousel.test.ts)

https://github.com/user-attachments/assets/803a0153-7fb3-4d70-a550-6df1747b9b49

### 02 · ListItem — with suppressRoleAndTabindex the item renders no role, tabindex or aria-selected (tests/list-item-suppressed-aria-selected.test.ts)

https://github.com/user-attachments/assets/d27611a0-a283-4e6a-8a4f-a9f09a6eb5a2

### 03 · Input — an actionInput field in forced error state carries neither aria-invalid nor aria-describedby (tests/input-error-announcement.test.ts)

https://github.com/user-attachments/assets/59fde7bb-45f7-4292-9e67-3a93c5879ad0

### 04 · Accordion — the panel is role="region" labelled by the trigger's real id (tests/accordion-aria-controls.test.ts)

https://github.com/user-attachments/assets/276fb440-351c-4619-aaa7-af4a372df2c1

### 05 · Status — the div carries status-text-default, the h2 does not, so app heading typography wins (tests/status-text-tag.test.ts)

https://github.com/user-attachments/assets/e38732fc-7f77-473c-bf26-d08315765daf

### 06 · Body scroll lock — body.style.overflow set to clip beforehand comes back as clip after the last unlock (tests/modal-nested-scroll-lock.test.ts)

https://github.com/user-attachments/assets/1ef2931d-f4b2-4a5a-bc1f-651dcd6fc856

### 07 · sui-chip-input — editable attribute renders the edit field on click; onedit receives [new, previous] (tests/wc-custom-elements.spec.ts)

https://github.com/user-attachments/assets/1657b59d-ee69-4875-9606-e182347a7355

### 08 · sui-status — status-text-tag="h2" renders an h2 without the div defaults (tests/wc-custom-elements.spec.ts)

https://github.com/user-attachments/assets/d51d61e5-4171-4590-9910-ab1ab75fe3c9

Tara-ag (approved):
All fifteen follow-ups are correctly addressed. The a11y improvements (Carousel aria-current, Accordion region labelling, Input aria-invalid gating, ListItem suppressed aria-selected) match the WAI-ARIA patterns claimed. The behavioural fixes (Status heading typography, TypewriterText whitespace resync, body-scroll-lock overflow restoration) preserve backward compatibility for default consumers. Web-component prop declarations and docs updates are consistent. New Playwright specs cover the changed surfaces. No blocking issues found.

murdore:
Head amended from 64f91c5be to 72cbd003b: the only addition is the regenerated `tests/visual/__screenshots__/input.png` for the new demo section on /components/input (980x4362 → 980x4561), produced in the pinned Playwright container and confirmed passing there without `--update-snapshots`. The Visual Regression job failed on the first head for exactly that page and nothing else. The fifteen thread replies that cite 64f91c5be describe the same change; the sha moved only because of the baseline.

src/routes/components/input/+page.svelte
```diff
@@ -40,6 +41,22 @@
...
+<h3>An action-style field never announces an error it does not show</h3>
+<p>
+  <code>actionInput</code> hides the label, the error message and the error styling so the field can
+  sit inside a composite control. <code>aria-invalid</code> follows that same gate: a field that shows
+  no error must not announce one.
+</p>
+<div class="demo-row" style="max-width: 400px;">
+  <Input
+    bind:value={actionStyle}
+    actionInput
```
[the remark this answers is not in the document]
murdore: Done in 959c3ca3e: the demo field now passes ariaLabel="Order value". Input applies it as aria-label whenever no visible label is rendered (Input.svelte, the aria-label binding on both the input and textarea branches), which is exactly the actionInput case, so the field keeps an accessible name while its label, message and error styling stay hidden.

tests/carousel.test.ts
```diff
@@ -45,6 +45,21 @@ test.describe('Carousel', () => {
...
+    const dot2 = page.getByTestId('carousel-manual-dot-2');
+
+    // Sighted users read the active-dot class; a screen-reader user querying the dots hears
+    // only "Go to slide N" unless the current one is marked. aria-current is that mark.
+    await expect(dot1).toHaveAttribute('aria-current', 'true');
+    await expect(dot2).not.toHaveAttribute('aria-current', /.*/);
+
+    await dot2.click();
+    await expect(dot2).toHaveAttribute('aria-current', 'true');
+    await expect(dot1).not.toHaveAttribute('aria-current', /.*/);
```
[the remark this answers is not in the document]
murdore: Done in 959c3ca3e: after the dot-2 click the test now also asserts that the dots wrapper contains exactly one [aria-current="true"] element, so a stale mark on dot 3 (or any other dot) fails the test rather than slipping past the two named dots.

murdore:
Head amended again to 959c3ca3e for CodeRabbit's two review points (demo field named via ariaLabel; Carousel test asserts exactly one aria-current dot). Both threads are answered and resolved. Yama's run on 72cbd003b ended in its own gateway timeouts, which its log labels an infrastructure error rather than a verdict; Tara-ag approved the first head at 13:26Z and the code has changed only by the visual baseline and these two review follow-ups since.

src/lib/Accordion/Accordion.svelte
```diff
@@ -20,6 +20,7 @@
      the panel from elsewhere too. */
   const uid = $props.id();
   const effectivePanelId = $derived(panelId ?? `accordion-panel-${uid}`);
+  const triggerId = $derived(`${effectivePanelId}-trigger`);
```
Tara-ag: 💡 MINOR: Accordion trigger id is not stable when panelId changes

The trigger id is derived from `effectivePanelId`, which can change if `panelId` is toggled from `undefined` to a string. This causes `aria-labelledby` on the panel to point to a stale trigger id if `panelId` changes after mount. In practice consumers rarely change `panelId`, but the linkage should be robust.

**Suggestion:** Derive `triggerId` from the `uid` instead of `effectivePanelId` so it is stable regardless of `panelId` changes:
```ts
const triggerId = $derived(`accordion-trigger-${uid}`);
```

src/lib/Accordion/Accordion.svelte
```diff
@@ -34,6 +35,7 @@
   <div
     class="accordion-trigger {triggerClasses ?? ''}"
     class:disabled
+    id={triggerId}
```
Tara-ag, line 41: 💬 SUGGESTION: Accordion trigger lacks aria-label fallback when trigger snippet has no text

When the consumer supplies a trigger snippet that contains only an icon, the trigger button may have no accessible name. The component relies on the snippet content for the label. Consider documenting that consumers must provide an accessible name, or accept an `aria-label` prop.

src/lib/Carousel/Carousel.svelte
```diff
@@ -192,6 +192,7 @@
           role="button"
           tabindex="0"
           aria-label={`Go to slide ${index + 1}`}
+          aria-current={activeSlideIndex == index ? 'true' : null}
```
Tara-ag: 💡 MINOR: Carousel uses loose equality for aria-current

The `aria-current` expression uses `==` instead of `===`. Although both operands are numbers in this case, loose equality is inconsistent with the rest of the codebase and can mask type mismatches during future refactoring.

**Suggestion:** Use strict equality:
```svelte
aria-current={activeSlideIndex === index ? 'true' : null}
```

src/lib/Accordion/Accordion.svelte
Tara-ag: 💡 MINOR: Accordion trigger id is unstable when panelId changes

The trigger id is derived from `effectivePanelId`, which changes when the `panelId` prop changes after mount. Changing the id of an interactive control at runtime can disrupt screen reader state and break external references or tests that rely on stable ids.

**Suggestion:** Derive the trigger id from the stable `uid` instead so it never changes:
```ts
const triggerId = $derived(`accordion-trigger-${uid}`);
```

Tara-ag (commented):
Reviewed all 23 changed files. The PR addresses fifteen open review follow-ups with additive, backward-compatible fixes. Two minor quality findings remain: the Accordion trigger id is derived from a potentially changing panelId, and the Carousel uses loose equality for aria-current. Neither blocks the PR. No hardcoded secrets, no unsafe HTML injection, no removed exports or renamed CSS custom properties, and SSR-sensitive code is guarded.

src/lib/Accordion/Accordion.svelte
murdore: Done in 521a46385: `triggerId` is now `accordion-trigger-${uid}`, built from the same per-instance id as the generated panel id, so assigning `panelId` after mount moves the panel's id and `aria-controls` but leaves the trigger's id untouched. One correction to the finding: `aria-labelledby` was derived from the same expression as the trigger id, so it never pointed at a stale id; the trigger being renamed at runtime was real, and that is what is fixed. Pinned by a new demo-route test in tests/accordion-aria-controls.test.ts: /components/accordion gained a button that assigns `panelId` late, and the test asserts the panel id and `aria-controls` move to `late-panel` while the trigger id and `aria-labelledby` keep their original value. Watched fail on 959c3ca3e (trigger became `late-panel-trigger`), pass after the change.

murdore: Same finding as the one from the second review pass on Accordion.svelte:23; fixed in 521a46385, with the trigger id built from the per-instance uid and a demo-route test that assigns `panelId` late and asserts the trigger id holds. Full reply on the other thread.

src/lib/Accordion/Accordion.svelte
murdore: Documented in 521a46385 rather than adding a prop. The trigger is `role="button"`, whose accessible name is computed from content, and that computation recurses into descendants and picks up their `aria-label`, so an icon-only snippet already has a route to a name without a new component prop: visible text, visually hidden text, or an `aria-label` on the icon element. docs/Accordion.md's Accessibility section now says so, and also records the `role="region"` / `aria-labelledby` wiring this PR added, which the docs had not mentioned.

src/lib/Carousel/Carousel.svelte
murdore: Done in 521a46385: `===` on the `aria-current` expression, and on the `class` expression two lines above it, which was the pre-existing source of the `==`; leaving one of each side by side would be exactly the inconsistency the finding describes. Behaviour is unchanged (both operands are numbers); the six Carousel tests pass.

murdore:
Head amended from 959c3ca3e to 521a46385 for Yama's review (the run on 959c3ca3e finished green with two minor findings and one suggestion, all now answered and resolved on their threads):

- Accordion: the trigger id is built from the per-instance uid instead of `effectivePanelId`, so a `panelId` assigned after mount moves the panel's id without renaming the trigger. Pinned by a new demo-route test that assigns `panelId` late and asserts the trigger id and `aria-labelledby` hold; it failed on 959c3ca3e (`late-panel-trigger`) and passes now.
- Carousel: strict equality on both dot expressions.
- docs/Accordion.md: records the `role="region"`/`aria-labelledby` wiring and that an icon-only trigger snippet needs text or an `aria-label` on the icon for its accessible name.
- `tests/visual/__screenshots__/accordion.png` regenerated in the pinned container for the new demo section (980x800 → 980x1029), verified passing there without `--update-snapshots`.

Local gates on 521a46385: prettier and eslint clean on the touched files, `pnpm run check` exit 0, Accordion and Carousel specs 12/12.

Tara-ag:

## 🤖 Yama Review Summary

PR #503, opened 2026-09-02
fix(status): stand in for the default icon when its file is absent

murdore:
The broken images I noticed in the regenerated `status` baseline on #500 turned out to be a library defect, not a demo one.

## The default `statusIcon` points at nothing

```svelte
statusIcon = 'icons/order-success-icon.svg'
```

A **relative** URL, so it resolves against whatever page renders the component — and no such file ships with the library. It points at something real only for an app serving that exact path at its own root, and resolves to `<route>/icons/order-success-icon.svg` anywhere deeper.

On the demo, at `/components/status`, it requests `/components/icons/order-success-icon.svg` and 404s. Every card relying on the default rendered a broken image. It has been that way since the Svelte 5 migration (`051b564`), documented as the default the whole time.

## The default is unchanged — the fix is a fallback

Revised after review. **`statusIcon` still defaults to `'icons/order-success-icon.svg'`**, so an app hosting that path keeps getting its own file and nothing about the public contract changes. A built-in inlined SVG stands in only when that fetch fails.

That leaves nothing breaking to version, which matters beyond this PR: tagging it `!` would cut 4.0.0, and `docs/EVENT_CASING_MIGRATION.md` reserves that major for the event-casing removal.

The fallback is **scoped to the untouched default**, which is why it is not simply `Img`'s `fallback` prop:

```svelte
const statusIconFallback = $derived(
  statusIcon === LEGACY_DEFAULT_STATUS_ICON ? BUILTIN_STATUS_ICON : null
);
```

Applying it to a caller's own icon would answer a failed *failure* icon with a success checkmark — worse than showing nothing. A test covers that scoping directly.

**The honest cost:** a page rendering the default still makes one failing request per Status before falling back. Removing that requires changing the default, which is the breaking change this avoids.

## An accessibility bug this fix introduced, and its fix

The built-in icon originally carried `role="img" aria-label="Success"`. Both are in `Img`'s root allowlist, so they were copied onto the live host where they **outrank the label `Img` derives from `alt`**:

```
rendered: <svg role="img" aria-label="Success" …>   <- from the icon's own markup
expected: <svg role="img" aria-label="status" …>    <- from Status
```

That would have pinned every default icon's accessible name to "Success" — including on a failure screen. The built-in icon now declares neither, and a test asserts the name comes from the component.

## `statusIconAlt`

Additive, defaulting to the previously hardcoded `'status'`. The generic name is deliberate: the same icon shape serves success and failure screens, so any *specific* default would be wrong on half of them, and deriving from `statusText` would announce the same words twice. The prop lets a caller say what the screen means, or pass `''` to mark the icon decorative.

## Verification

| Check | Result |
| --- | --- |
| new specs | **3/3** — fallback paints a real icon, caller's icon fetched and not replaced, accessible name from `statusIconAlt` |
| functional | **405/405** |
| unit | **286/286** |
| `lint` / `check` | clean |
| visual | `status` baseline matches **without regeneration** |

The baseline was checked by eye, not by byte count: every card that previously showed a broken image now shows the checkmark, and "Processing" correctly still shows none, since it supplies its own `icon` snippet.

## Summary by CodeRabbit

- **New Features**
  - Added customizable alternative text for status icons, supporting meaningful labels and decorative icons.
  - Added a built-in success icon fallback when the default status icon cannot load.
  - Preserved caller-supplied icons without replacing them with the fallback.

- **Documentation**
  - Expanded Status component documentation with accessibility guidance and details for descriptions, styling, snippets, actions, and test IDs.


tests/status-default-icon.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+ * `/components/status`, it requested `/components/icons/order-success-icon.svg`
+ * and 404ed, so every card relying on the default rendered a broken image.
+ */
+test('the default status icon renders without fetching anything', async ({ page }) => {
+  const requested: string[] = [];
+  page.on('request', (request) => {
+    requested.push(request.url());
+  });
+
+  await page.goto('/components/status');
```
Tara-ag, line 18: 💡 MINOR: New Playwright spec selects through demo-page `data-pw` attributes instead of an isolated fixture. Selectors like `[data-pw="status-inline"]` and `[data-pw="status-default-icon"]` only exist in `src/routes/components/status/+page.svelte`, coupling the library regression test to demo markup. The repo's recurring standard is: "Component-contract tests should use isolated fixtures or harnesses instead of selecting through demo-page markup and demo-specific testIds." Render `Status` in a minimal test harness with stable selectors.

src/lib/Status/Status.svelte
```diff
@@ -3,8 +3,27 @@
...
+   * `statusIcon="icons/order-success-icon.svg"`.
+   *
+   * Carries no `role` or `aria-label` of its own. Those are in `Img`'s
+   * allowlist, so a root that declared them would be copied onto the live host
+   * and would outrank the label `Img` derives from `alt` — pinning every
+   * default icon to "Success", including on a failure screen.
+   */
+  const DEFAULT_STATUS_ICON =
+    "data:image/svg+xml,%3Csvg%20xmlns%3D'http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg'%20viewBox%3D'0%200%2048%2048'%3E%3Ccircle%20cx%3D'24'%20cy%3D'24'%20r%3D'22'%20fill%3D'%23e6f6ec'%20stroke%3D'%232e994c'%20stroke-width%3D'2'%2F%3E%3Cpath%20d%3D'M15%2024.5l6.5%206.5L33%2019.5'%20fill%3D'none'%20stroke%3D'%232e994c'%20stroke-width%3D'3.5'%20stroke-linecap%3D'round'%20stroke-linejoin%3D'round'%2F%3E%3C%2Fsvg%3E";
+
```
Tara-ag, line 27: ⚠️ MAJOR: Default `statusIcon` change is a silent breaking change for consumers relying on the old relative URL. Changing the default from `'icons/order-success-icon.svg'` to an inline data URI means any downstream app that hosts its own `/icons/order-success-icon.svg` and relies on the default will now get the built-in checkmark instead. The PR body itself notes the analogous Toolbar change was shipped as `feat(toolbar)!`. Per the blocking criteria, a changed prop default that alters rendered output without a major-version bump or migration note is breaking. Either mark this `fix(status)!:`/`feat(status)!:` and add a migration note to `docs/Status.md`, or keep the old default and add an opt-in prop (e.g. `useInlineDefaultIcon`).

src/lib/Status/Status.svelte
Tara-ag, line 1: 💡 MINOR: Hard-coded `alt="status"` gives a generic accessible name to every default icon. The PR correctly avoids letting the SVG's own `aria-label` override it, but "status" is not descriptive — a failure screen using the default success checkmark will be announced simply as "status". Since this PR adds a test asserting that exact name, consider exposing a `statusIconAlt` prop (defaulting to something meaningful or derived from `statusText`) and document it, or at least note in `docs/Status.md` that the default icon is announced generically.

Tara-ag (changes requested):
Reviewed file-by-file. The fix itself is correct and well-documented, but the changed default is a breaking change for any consumer relying on the old relative URL. I am requesting changes to resolve the versioning/API-contract concern before merge.

Tara-ag (changes requested):
The fix is technically correct and well-tested, but the default `statusIcon` change alters rendered output for downstream consumers and therefore needs to be treated as a breaking change (major-version bump + migration note) or exposed behind an opt-in prop. Two additional quality comments cover test coupling to demo-page selectors and the generic default accessible name. Please address the API-contract concern before merging.

Tara-ag:

## 🤖 Yama Review Summary

⚠️ The review loop ended early (`time-limit`, 11 steps) — this verdict is built strictly from gate-verified findings.

src/lib/Status/Status.svelte
murdore: Agreed, and taken — the default is no longer changed at all, so there is nothing breaking left to version.

`statusIcon` still defaults to `'icons/order-success-icon.svg'`. An app hosting that path at its own root keeps getting its own file, exactly as before. The built-in icon is now a **fallback**, applied only when that fetch fails.

I did not take the `useInlineDefaultIcon` opt-in, because it leaves the broken default as the out-of-the-box behaviour: an app would have to discover the flag to stop rendering a broken image, and the overwhelmingly common case (not hosting that path) is the one that would stay broken.

Nor did I tag it `!`. That would cut 4.0.0, which `docs/EVENT_CASING_MIGRATION.md` reserves for the event-casing removal — spending the major on an icon default would land 4.0.0 without the change it was planned for.

One property worth calling out, since it is the reason this is not simply `Img`'s `fallback` prop: the fallback is **scoped to the untouched default**. Applying it to a caller's own icon would answer a failed *failure* icon with a success checkmark, which is worse than showing nothing.

```svelte
const statusIconFallback = $derived(
  statusIcon === LEGACY_DEFAULT_STATUS_ICON ? BUILTIN_STATUS_ICON : null
);
```

A test covers that scoping directly. The honest cost: a page rendering the default still makes one failing request per Status before falling back. Removing that requires changing the default, which is the breaking change this avoids.

src/lib/Status/Status.svelte
murdore: Fixed — `statusIconAlt` added, additive and defaulting to the previous hardcoded `'status'`, so nothing changes for existing callers.

You are right that "status" is weak, and I kept it as the default deliberately rather than deriving something: the same icon shape serves success and failure screens, so any *specific* default name would be wrong on half of them. Deriving from `statusText` has the opposite problem — the icon and the heading would announce the same words twice.

So the prop lets a caller say what the screen actually means, or pass `''` to mark the icon decorative when `statusText` already carries the meaning. Both the type and `docs/Status.md` say so.

tests/status-default-icon.spec.ts
murdore: Fair, and this has now come up on #497, #498 and here, so it is being fixed properly rather than declined a fourth time.

To be concrete about why these specs are written this way today: there is no component-test harness in this repo. `@testing-library/svelte` is not a dependency, and the vitest config only includes `src/**` and `scripts/**`, so `tests/` is Playwright's alone. A demo page is currently the only path that renders a component at all — an "isolated fixture" has nowhere to live.

I am adding `@testing-library/svelte` in a separate PR so component contracts can be unit-rendered without Playwright or demo markup, and these specs will move onto it. Keeping that out of this PR so the Status fix is not blocked behind a test-infrastructure change.

docs/Status.md
```diff
@@ -16,18 +16,19 @@ The `statusDescription` string is rendered with `{@html}`, so a caller can pass
...
-| testId             | `string`                                                | No       | `-`                              | Value applied to the `data-pw` attribute on the root element for test selection.                                                                                                                                                                                                          |
+| Prop               | Type                                                    | Required | Default                      | Description                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  |
```
Tara-ag: 💡 MINOR: Docs list the effective fallback as the prop default for `statusIcon`. The Props table lists the Default for `statusIcon` as "a built-in success checkmark", but the actual prop default in `Status.svelte` remains `'icons/order-success-icon.svg'`. The built-in checkmark is only a fallback when that legacy URL fails to load. A reader relying on the Default column will misunderstand the contract: they would not know they can host `icons/order-success-icon.svg` to override the fallback, and the description's note that "An app that does host it can pass `statusIcon=\"icons/order-success-icon.svg\"` explicitly" contradicts the idea that the default is already the built-in SVG.

Suggested fix: change the Default column for `statusIcon` back to `'icons/order-success-icon.svg'` and keep the detailed explanation in the Description column that the built-in checkmark is a fallback when that file is absent. This keeps the documented default value aligned with the implementation.

Tara-ag:
⚠️ MAJOR: `src/wc/components/Status.wc.svelte` — New `statusIconAlt` prop is not exposed in the web-component wrapper.

The PR adds a public string prop `statusIconAlt` to `StatusProperties` and uses it in `Status.svelte`, but this wrapper still only declares `statusIcon`, `statusText`, `statusDescription`, `buttonProperties`, `classes`, and `onbuttonClick`. Per the repo's recurring standard, every camelCase Svelte prop exposed in a web-component wrapper must have an explicit kebab-case HTML attribute mapping. Without it, consumers of `<sui-status>` cannot set the icon's accessible name or mark it decorative via HTML, and the Svelte and web-component APIs diverge.

Suggested fix: add an explicit mapping in the `customElement` props object:

```js
statusIconAlt: { type: 'String', reflect: true, attribute: 'status-icon-alt' },
```

This mirrors the existing `statusIcon` mapping and lets web-component consumers use `status-icon-alt=""` to mark the icon decorative or provide a meaningful accessible name.

Tara-ag (changes requested):
Reviewed. One MAJOR finding (web-component API parity) and one MINOR finding (documentation accuracy). Details are in the inline comments.

tests/status-default-icon.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  page
+}) => {
+  // `role` and `aria-label` are both in Img's allowlist, so an icon carrying
+  // its own would outrank the alt-derived label on every screen that uses it.
+  // The fallback is inlined into an <svg> host, where the name lives on
+  // `aria-label` rather than on `alt`.
+  await page.goto('/components/status');
+
+  const icon = page.locator('[data-pw="status-inline"] .status-image svg');
+  await expect(icon).toHaveAttribute('aria-label', 'status');
```
Tara-ag, line 61: 💡 MINOR: `statusIconAlt` forwarding is not tested with non-default values. The new prop is only asserted with its default value `'status'`. If `Status.svelte` stopped forwarding the prop to `Img`, the existing test would still pass. The contract also advertises that an empty string marks the icon decorative, but that path is not exercised. Add a fixture that passes a distinctive `statusIconAlt` value and assert it becomes the `aria-label`, plus a fixture with `statusIconAlt=""` to verify the decorative-icon contract.

tests/status-default-icon.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  const requested: string[] = [];
+  page.on('request', (request) => {
+    requested.push(request.url());
+  });
+
+  await page.goto('/components/status');
+  await expect(page.locator('[data-pw="status-default-icon"] .status-image svg')).toBeVisible();
+
+  // The caller's URL is fetched, which the built-in data URI never would be.
+  expect(requested.filter((url) => url.includes('status-success.svg')).length).toBeGreaterThan(0);
```
Tara-ag: 💡 MINOR: Caller-icon test does not force the request to fail. The test asserts that a caller-supplied `statusIcon` URL is requested and that an SVG is visible, but it never forces `status-success.svg` to fail. If a regression caused the built-in fallback to be applied even for caller-supplied icons, the test could still pass because the built-in fallback is also an SVG and would be visible. Force the caller-supplied icon URL to 404 before navigating, then assert that the rendered icon is not the built-in fallback. For example:

```ts
await page.route('**/status-success.svg', (route) => route.abort('notfound'));
await page.goto('/components/status');
await expect(page.locator('[data-pw="status-default-icon"] .status-image svg')).toBeVisible();
// additionally assert the built-in fallback is NOT used here
```

This ensures the scoping invariant holds even when the caller's own icon fails to load.

Tara-ag (commented):
Review finalized. The fallback approach is backward-compatible and the `statusIconAlt` prop is a solid accessibility addition. Two findings remain: the new prop must be mirrored in the web-component wrapper, and the docs should not describe the fallback as the default value. Two additional test-quality comments are included.

docs/Status.md
```diff
@@ -16,18 +16,19 @@ The `statusDescription` string is rendered with `{@html}`, so a caller can pass
...
+| statusIcon         | `string`                                                | No       | a built-in success checkmark | URL of the status icon image displayed at the center (e.g., success checkmark, error cross). The default is an inlined `data:` SVG, so it renders without a network request at any route depth. It was previously `'icons/order-success-icon.svg'` — a relative URL the library never shipped a file for, which resolved against the rendering page and so 404ed anywhere but an app root serving that exact path. An app that does host it can pass `statusIcon="icons/order-success-icon.svg"` explicitly. |
```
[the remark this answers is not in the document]
murdore: Correct, and fixed. That row described the *first* version of this change and stopped being true when the default was left alone — so it documented a contract the component no longer had, which is worse than the original bug for anyone reading the table to decide what to pass.

The Default column reads `'icons/order-success-icon.svg'` again. The fallback behaviour is explained in the description instead, including the part that matters for a reader deciding whether to host the file: doing so still wins, because the fallback only applies when the request fails.

docs/Status.md
murdore: Correct, and fixed. That row described the *first* version of this change and stopped being true when the default was left alone — so it documented a contract the component no longer had, which is worse than the original bug for anyone reading the table to decide what to pass.

The Default column reads `'icons/order-success-icon.svg'` again. The fallback behaviour is explained in the description instead, including the part that matters for a reader deciding whether to host the file: doing so still wins, because the fallback only applies when the request fails.

src/lib/Status/Status.svelte
```diff
@@ -16,6 +39,12 @@
     children,
     testId
   }: StatusProperties = $props();
+
+  // Scoped to the untouched default on purpose. Applying it to a caller's own
+  // icon would answer a failed failure-icon with a success checkmark.
+  const statusIconFallback = $derived(
+    statusIcon === LEGACY_DEFAULT_STATUS_ICON ? BUILTIN_STATUS_ICON : null
```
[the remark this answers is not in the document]
murdore: Accurate reading of the code, and I'm keeping the behaviour — but the reasoning is worth stating rather than just declining.

`statusIcon === LEGACY_DEFAULT_STATUS_ICON` does also match a caller who passes that exact string explicitly. The question is whether that caller wants different behaviour from the caller who omitted it, and they don't: both are asking for `icons/order-success-icon.svg`, and if it 404s, both are better served by a checkmark than a broken image. The outcome is identical to what they'd want.

The case the scoping exists to prevent is a *different* icon failing — a failure screen's own icon 404ing and being answered with a success checkmark. That is unaffected by this, since any other value takes the `null` branch.

Distinguishing omitted from explicitly-passed is possible (`'statusIcon' in $props()`), but it would buy a behavioural difference nobody wants, at the cost of the prop no longer having a real default. Leaving as is.

tests/status-default-icon.spec.ts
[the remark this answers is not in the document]
murdore: This was the best catch on the PR — the test could not have failed.

Both the caller's icon and the built-in fallback render as an inlined `<svg>`, so "a caller URL was requested and an SVG is visible" is satisfied by the exact regression it was supposed to catch.

Fixed as you suggested, with one addition that makes the assertion decisive. The request is aborted, and rather than trying to tell two visually identical SVGs apart, the test asserts the *absence* of an inlined `<svg>`: with `fallback={null}`, a caller's failed icon stays a plain broken ``.

```ts
await page.route('**/status-success.svg', (route) => route.abort());
await page.goto('/components/status');

const icon = page.locator('[data-pw="status-default-icon"] .status-image');
await expect(icon.locator('svg')).toHaveCount(0);
await expect(icon.locator('img')).toHaveCount(1);
```

Verified against a negative control rather than assumed: making the fallback unconditional fails exactly this test and nothing else.

murdore: This was the best catch on the PR — the test could not have failed.

Both the caller's icon and the built-in fallback render as an inlined `<svg>`, so "a caller URL was requested and an SVG is visible" is satisfied by the exact regression it was supposed to catch.

Fixed as you suggested, with one addition that makes the assertion decisive. The request is aborted, and rather than trying to tell two visually identical SVGs apart, the test asserts the *absence* of an inlined `<svg>`: with `fallback={null}`, a caller's failed icon stays a plain broken ``.

```ts
await page.route('**/status-success.svg', (route) => route.abort());
await page.goto('/components/status');

const icon = page.locator('[data-pw="status-default-icon"] .status-image');
await expect(icon.locator('svg')).toHaveCount(0);
await expect(icon.locator('img')).toHaveCount(1);
```

Verified against a negative control rather than assumed: making the fallback unconditional fails exactly this test and nothing else.

tests/status-default-icon.spec.ts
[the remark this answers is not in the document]
murdore: Right — asserting only the default would have passed even if `Status` stopped forwarding the prop to `Img` entirely.

Two demo cases now cover it: one passing `statusIconAlt="Payment confirmed"`, asserted to become the `aria-label`, and one passing `''`, asserted to produce `aria-hidden="true"` with no `aria-label`. The decorative path is part of the documented contract, so it is tested rather than just described.

murdore: Right — asserting only the default would have passed even if `Status` stopped forwarding the prop to `Img` entirely.

Two demo cases now cover it: one passing `statusIconAlt="Payment confirmed"`, asserted to become the `aria-label`, and one passing `''`, asserted to produce `aria-hidden="true"` with no `aria-label`. The decorative path is part of the documented contract, so it is tested rather than just described.

Tara-ag (commented):
Approved with one minor web-component parity note. The Svelte-side fix is sound: the legacy default is preserved, the fallback is correctly scoped to that default, and the accessible-name handling avoids the earlier aria-label pitfall. Tests and docs are updated accordingly. The only gap is that the new `statusIconAlt` prop is not mapped in `Status.wc.svelte`, so web-component consumers cannot yet use it.

Tara-ag:
Yama review finalization: the accepted finding regarding `statusIconAlt` missing from `src/wc/components/Status.wc.svelte` could not be posted as an inline comment because that file is not part of this PR's diff. The finding was already recorded above as a general PR comment and is included in the final verdict JSON.

src/wc/components/Status.wc.svelte
```diff
@@ -4,6 +4,7 @@
     shadow: 'open',
```
Tara-ag, line 4: ⚠️ MAJOR: The web-component wrapper declares only serializable props and omits `icon`, `descriptionSnippet`, and `children` from `customElement.props`. Snippet/function props should remain property-only (no attribute mapping), but they still need declarations so JavaScript consumers can assign them as properties and Svelte creates accessors. Without them, `element.icon = ...` lands on the DOM node and never reaches the component, silently breaking web-component parity with the Svelte API. This is the same gap `wc-custom-elements.spec.ts` already tests for Combobox snippet props.

Add property-only declarations:
```svelte
icon: { type: 'Object' },
descriptionSnippet: { type: 'Object' },
children: { type: 'Object' }
```
And cover them with a web-component test that assigns a snippet and asserts it renders.

Tara-ag (changes requested):
One blocking issue: the web-component wrapper for Status is missing property-only declarations for its snippet props (`icon`, `descriptionSnippet`, `children`). This breaks parity between the Svelte and custom-element APIs. Once those declarations and a covering test are added, this is good to merge.

src/wc/components/Status.wc.svelte
Tara-ag, line 4: ⚠️ MAJOR: The web-component wrapper declares only serializable props and omits `icon`, `descriptionSnippet`, and `children` from `customElement.props`. Snippet/function props should remain property-only (no attribute mapping), but they still need declarations so JavaScript consumers can assign them as properties and Svelte creates accessors. Without them, `element.icon = ...` lands on the DOM node and never reaches the component, silently breaking web-component parity with the Svelte API.

Suggested fix: add property-only declarations:
```svelte
icon: { type: 'Object' },
descriptionSnippet: { type: 'Object' },
children: { type: 'Object' }
```
Then add a web-component test that assigns a snippet property and asserts it renders.

Tara-ag (changes requested):
Review: one MAJOR issue around web-component parity. The Svelte-side fix and tests look good, but the custom-element wrapper needs snippet prop declarations before merge.

murdore:
Answering the blocking finding here (`Status.wc.svelte` omits `icon`, `descriptionSnippet`, `children` from `customElement.props`) with measurements, because **one third of it is wrong** and following it as written ships a DOM bug.

**`icon` and `descriptionSnippet` — the finding is right.** Both are now declared as property-only (`{ type: 'Object' }`, no attribute) in #506, and verified reachable: assigning `element.icon` reaches the component and renders. `Status` has no competing body snippet, so nothing overrides the assignment.

**`children` — the finding is wrong, and it was worth measuring rather than arguing.** Declaring `children` on a custom-element wrapper does not just shadow `Element.children`; it removes it:

```
element.children  →  undefined      (not an HTMLCollection)
```

`el.children.length` then **throws** for every consumer of that element. CodeRabbit raised the same point independently on #506 and described it as shadowing an accessor — the measured behaviour is more severe than that.

So `children` is now in `HOST_RESERVED_PROPS` and removed from all 16 wrappers that declared it (13 added by #506, plus `ChatBubble`, `Draggable` and `Resizable`, which already ship the bug on `release`).

The underlying concern in the finding is still valid and still addressed: light-DOM content reaches the component through the default `<slot>`, which is the path web-component consumers actually use. A test now asserts `assignedSlot` rather than the node's own text — the previous version of that assertion passed against a wrapper rendering no `<slot>` at all, so it proved nothing.

Worth noting for anyone applying this finding elsewhere: adding `{#snippet children()}<slot></slot>{/snippet}` to `Status` is **not** a safe substitute. `Status.svelte` guards with `typeof children === 'function'`, so an always-supplied snippet makes that guard permanently true and renders an empty container.

**This does not need a change in this PR.** #506 carries the wrapper fix for the whole class, so #503 should rebase onto it rather than patch `Status.wc.svelte` here. The `statusIconAlt` mapping this review also asked for is likewise already in #506.

Related: https://github.com/juspay/svelte-ui-components/pull/506

Tara-ag (approved):
Reviewed file-by-file. The change is additive and preserves the public contract:

- `statusIcon` default is unchanged, avoiding a breaking change.
- New `statusIconAlt` prop is optional and defaults to the previous hardcoded alt text (`'status'`).
- Web-component wrapper correctly declares `status-icon-alt`.
- Docs and tests (including a new web-component spec) cover the new behavior.
- The built-in fallback SVG deliberately omits `role`/`aria-label` so `Img` derives accessible name from `alt`, fixing the accessibility bug noted in the PR.

No blocking issues found.

PR #504, opened 2026-09-02
chore(visual): baseline task-list, and measure why the other stays out

murdore:
Both visual routes were excluded as *"unstable under suite load"* — true when written, never re-tested after the viewport-fitting work landed. Retested rather than assumed: emptied `EXCLUDED`, generated baselines for both, and ran the full suite repeatedly.

## `task-list` is stable and is now baselined

Its stated reason — *"one of the tallest demos with staged reveals"* — **was the fitting problem**. The demo grew past the viewport, the old loop ran out of attempts, and the capture stitched mid-scroll. That was fixed when growth-that-tracks-the-viewport started being detected; the route has been excluded ever since on a reason that no longer applies.

Two consecutive full-suite runs on this branch:

```
===== PASS 1 =====
  ✓ task-list renders as baselined (802ms)
  1 skipped
  93 passed (2.1m)
===== PASS 2 =====
  ✓ task-list renders as baselined (816ms)
  1 skipped
  93 passed (1.9m)
```

**Baselined coverage goes from 92 routes to 93.**

## `thinking-indicator` stays out — with the measurement, not an impression

Within a *single test's own retries*, the captured height walks:

```
2029 -> 2035 -> 2121 -> 2150 -> 2059 px
```

The animated dots are pinned by the injected stylesheet. The per-second elapsed counter is not: it retimes the layout on a real timer that `page.clock` does not drive. No baseline can be stable while that is true.

Its entry in `EXCLUDED` now records exactly that, including the numbers, so the next person deciding whether to re-test it can see what would have to change first — rather than re-running the same experiment to rediscover it.

## Verification

| Check | Result |
| --- | --- |
| visual, pinned container | **93 passed, 1 skipped, 0 failed** — two consecutive runs |
| new baseline | `task-list.png`, generated in the pinned container |

The two routes were tested by running the **whole** suite, not just those routes, because "under suite load" was the original claim and a route in isolation would not reproduce it.

## Summary by CodeRabbit

- **Tests**
  - Updated visual test coverage by adding `task-list` to the baseline.
  - Refined documentation for the remaining excluded route, noting height instability in the elapsed counter.
  - Updated guidance for reviewing excluded visual test routes.


tests/visual/components.visual.ts
```diff
@@ -43,9 +43,13 @@ const MASKS: Readonly<Record<string, readonly string[]>> = {
  * at random, which is worse than admitting the gap.
```
Tara-ag: 💡 MINOR: Stale plural JSDoc no longer matches single excluded route. The JSDoc block immediately above this says "Both of these pass reliably..." and "tried on them", but this PR removes `task-list` from `EXCLUDED`, leaving only `thinking-indicator`. Please rewrite that paragraph to singular form so it accurately describes the one remaining exclusion.

Tara-ag (commented):
Visual-test maintenance only. One minor doc-staleness nit; otherwise the change is safe and the measured reasoning for keeping `thinking-indicator` excluded is good.

tests/visual/components.visual.ts
Tara-ag: 💡 MINOR: Stale plural JSDoc no longer matches single excluded route

This comment still describes two excluded routes ('Both of these pass reliably...', 'tried on them', 'Shipping them flaky'), but `task-list` was removed from `EXCLUDED`, leaving only `thinking-indicator`. The misleading plural wording will confuse the next maintainer who tries to understand why routes are excluded. Rewrite it to refer to the single remaining route, e.g.:

```
 * This route passes reliably when run on its own and fails intermittently
 * inside the full suite at any worker count. Everything that worked for the
 * rest of the suite was tried on it. Shipping it flaky would block unrelated
 * PRs at random, which is worse than admitting the gap.
```

Tara-ag (approved):
Approved. Visual-test maintenance only: adds a stable `task-list` baseline and replaces the vague exclusion reason for `thinking-indicator` with measured height-variation evidence. One minor JSDoc staleness nit left inline.

tests/visual/components.visual.ts
murdore: Fixed. The paragraph was written for two routes and this PR leaves one, so the plural was actively misleading about how many gaps exist.

It now describes the single remaining route, and records what happened to the other one:

> `task-list` used to sit here too. Its instability was the viewport-fitting problem, fixed since, and it is baselined now: re-check an entry here before assuming the reason still holds.

That last clause is the part I most wanted in the file. `task-list` stayed excluded for a while after the cause was fixed, purely because nobody re-tested the stated reason.

murdore: Fixed. The paragraph was written for two routes and this PR leaves one, so the plural was actively misleading about how many gaps exist.

It now describes the single remaining route, and records what happened to the other one:

> `task-list` used to sit here too. Its instability was the viewport-fitting problem, fixed since, and it is baselined now: re-check an entry here before assuming the reason still holds.

That last clause is the part I most wanted in the file. `task-list` stayed excluded for a while after the cause was fixed, purely because nobody re-tested the stated reason.

Tara-ag (approved):
No issues found. The change is limited to visual test infrastructure: it adds a generated baseline for `task-list` and updates the exclusion documentation for `thinking-indicator` with measured instability data. No component source, public API, accessibility, or security implications.

PR #505, opened 2026-09-02
feat(events): accept the corrected spelling for every event prop

murdore:
Phase 1 of the plan in `docs/EVENT_CASING_MIGRATION.md`. All **142** grandfathered props gain their correct spelling **alongside** the old one, both wired to the same handler. Nothing is removed, nothing is renamed, and a consumer on an old spelling is untouched.

56 components, 113 files. Generated by `scripts/migrate/alias-event-props.ts`, which is committed rather than run once and discarded — it is the only honest description of how 56 components were edited at the same time. Re-running it is a no-op, and a test asserts that.

## The transform

```svelte
- let { onclick }: ToggleProperties = $props();
+ let { onclick: onclickLegacy, onClick }: ToggleProperties = $props();
+ const onclick = $derived(onClick ?? onclickLegacy);
```

Every existing reference keeps resolving, so **no call site is touched**.

That is deliberate, not incidental. `Toggle.svelte` both destructures a synthetic `onclick` prop *and* renders `<input onclick={handleCheckboxClick}>`. A textual rename would have corrupted the second. Only the binding inside the `$props()` pattern is rewritten — and an attribute name is not a binding.

## Two bugs the generator had, caught before they shipped

**It dropped destructured defaults.** Replacing the whole property node turned

```js
onFocus = () => {},        →    onFocus: onFocusLegacy,
```

into an undefined binding. `Input.svelte` calls `onFocusout(event); onBlur(event)` **unguarded**, so every consumer not passing those props would have hit `TypeError: onBlur is not a function` on blur — a crash in a core component. Found by diffing for removed defaults rather than trusting the transform. Defaults now travel onto the legacy binding:

```js
onFocus: onFocusLegacy = () => {}, onfocus,
```

**Declarations are not reliably one line**, and their types carry their own `;`:

```ts
onpointclick?: (event: {
  seriesIndex: number;
  pointIndex: number;
}) => void;
```

The end of a declaration is now found by scanning to the first `;` at depth zero, not by matching a line. That took the skip count from 10 to 0.

## An inverted guard

`casing.test.ts` asserted that the corrected spellings were **absent** from each component — correct while planning the rename, wrong the moment Phase 1 adds them. Rather than delete it, it now asserts each target is declared **exactly once**, which is the property still worth holding: a duplicate declaration is what a re-run of the generator would leave, and not something TypeScript necessarily rejects.

## Verification

| Check | Result |
| --- | --- |
| Playwright functional | **404/404** across all 56 components |
| unit | **295/295** |
| `scripts/` | **143/143** |
| `lint` / `check` | clean |
| event-casing checker | still `142 known, 0 new` — the added spellings are correct by construction |

The behavioural claim is tested, not asserted: the toggle demo now renders the same component under both spellings, and a spec clicks each one and checks the handler fired.

```
✓ a component fires its handler through the corrected prop spelling
✓ the original spelling still fires, untouched
```

Structural coverage backs that across all 142: every corrected spelling is declared, every old spelling is still declared, and every component that renamed a binding has the matching `$derived`.

## What comes next

Phase 2 is the dev-only deprecation warning; Phase 3 removes the old spellings in 4.0.0 and deletes `event-casing-baseline.json`. Neither is in this PR. The codemod's rename map should also be pointed at the new spellings before Phase 3 — worth doing while both exist.


Tara-ag (commented):
Review in progress.

Tara-ag:

The event-casing alias work is additive and backward-compatible for Svelte consumers, but the new corrected camelCase event props are not exposed in the web-component wrappers (src/wc/components/*.wc.svelte). Because Svelte custom elements only forward props declared in customElement.props, web-component consumers cannot use the new spellings, creating a public-API parity gap in a library that ships both forms. Additionally, the Playwright alias test is coupled to demo-page markup and only covers Toggle.

- ⚠️ **MAJOR**: New corrected event-prop spellings are not exposed in custom-element wrappers — `src/wc/components/Toggle.wc.svelte:10`
  The PR adds corrected camelCase spellings for event props to 56 Svelte components, but no corresponding *.wc.svelte wrapper was updated. Because Svelte custom elements only expose props listed in the customElement.props config, web-component consumers cannot pass the new spellings at all. For example, <sui-toggle onClick={...}> will not reach Toggle.svelte because only onclick is declared. This co…
- 💡 **MINOR**: Playwright alias test relies on demo-page markup and covers only Toggle — `tests/event-casing-alias.spec.ts:9`
  tests/event-casing-alias.spec.ts selects through the demo page (`[data-pw="toggle-alias-state"]`, `.demo-row`) and only exercises Toggle. The repository convention prefers isolated component fixtures over demo-page selectors, and a single happy-path smoke test does not strongly guard the 142 aliases across 56 components. The generator unit tests provide structural coverage, but an end-to-end contr…

tests/event-casing-alias.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { expect, test } from '@playwright/test';
+
+/**
+ * Phase 1 of the event-casing migration adds each prop's correct spelling
+ * alongside the old one, both wired to the same handler. The claim worth
+ * testing is that the new spelling actually reaches the handler — a props type
+ * carrying the name proves only that it compiles.
+ */
+test('a component fires its handler through the corrected prop spelling', async ({ page }) => {
```
Tara-ag, line 9: 💡 MINOR: This test selects through demo-page markup (`[data-pw="toggle-alias-state"]`, `.demo-row`) and only exercises Toggle. Repo convention prefers isolated fixtures over demo-specific selectors, and a single happy-path smoke test is a thin guard for 142 aliases across 56 components. Consider moving the alias firing assertion to an isolated Svelte harness that renders a representative set of components with both old and new spellings.

Tara-ag:
⚠️ MAJOR: New corrected event-prop spellings are not exposed in custom-element wrappers. The PR adds corrected camelCase spellings (e.g. `onClick`) to 56 Svelte components, but no `*.wc.svelte` wrapper was updated. Because Svelte custom elements only expose props listed in `customElement.props`, web-component consumers cannot pass the new spellings at all. For example, `<sui-toggle onClick={...}>` will not reach `Toggle.svelte` because only `onclick` is declared in `src/wc/components/Toggle.wc.svelte`. Per repo standards, every camelCase serializable prop exposed in a Svelte component must have an explicit mapping in the corresponding wrapper (function/object props as `{ type: 'Object' }` with no HTML attribute mapping). Please update all affected wrappers to include the new event-prop spellings.

Tara-ag (changes requested):
Review summary: the event-casing alias work is sound for Svelte consumers, but the new camelCase event props are not surfaced through the web-component wrappers, which is a public-API parity gap for a library that also ships as custom elements. One additional minor note on test isolation.

Tara-ag:
⚠️ MAJOR: `src/wc/components/Toggle.wc.svelte` (and the other `*.wc.svelte` wrappers) do not declare the new corrected camelCase event props. Svelte custom elements only forward props listed in `customElement.props`, so web-component consumers cannot use the new spellings (e.g. `<sui-toggle onClick={...}>`). Per the repo's web-component parity standard, every camelCase serializable prop exposed in a Svelte component must have an explicit `{ type: 'Object' }` mapping in the matching wrapper. Please extend the alias work to `src/wc/components/*.wc.svelte` or add the mappings manually.

Tara-ag (commented):
Reviewed. The main concern is web-component API parity: the new camelCase event props are not exposed in the `*.wc.svelte` wrappers, so custom-element consumers cannot use them. I also noted that the Playwright alias test is coupled to demo-page markup and only covers Toggle.

tests/event-casing-alias.spec.ts
murdore: Half of this is right and half rests on a premise that does not hold here — I checked both rather than either agreeing or waving it off.

**The convention point is backwards.** Demo-page markup *is* the convention in this repo, not a departure from it: 97 of 102 test files call `page.goto('/components/…')`, and 98 select through `getByTestId` or `[data-pw]`. `tests/fixtures` exists but holds only `tiny.png`, `tiny-2.png`, `tiny-3.png` and `too-large.png` — image fixtures for Img, not component harnesses. There is no isolated-harness pattern to move toward; adding one here would be the first, and it would make this the odd test out rather than the conforming one.

**The breadth point is fair, and is already answered — just not by this file.** Two Playwright tests over one component would indeed be a thin guard for 142 aliases across 56 components. The coverage lives in `scripts/migrate/alias-event-props.test.ts`, which asserts statically across every component that:

- every expected alias exists (`missing` is empty)
- no destructured default was dropped in the rewrite (`dropped` is empty)
- every alias is actually wired to its `$derived` (`unwired` is empty)
- re-running the generator is a no-op

That is the split worth keeping: the static test proves all 142 aliases exist and are wired, and the Playwright test proves the runtime path — that a click through the new spelling really reaches the handler — which a props type carrying the name cannot show. Doing the second one 56 times would add runtime cost without adding information, since the wiring is generated by one transform and is already asserted uniform.

So: no change, on evidence rather than on preference. Resolving.

PR #461, opened 2026-08-25
feat(ChatBubble): decouple launcher width/height tokens for pill launchers

murdore:
## What

`--chat-bubble-width` / `--chat-bubble-height` now override each launcher axis independently, both defaulting through `--chat-bubble-size` — existing consumers are byte-identical. Setting `--chat-bubble-width: fit-content` lets an icon-plus-label `icon` snippet render as a **pill launcher** (Button's own default width is `fit-content`; the launcher was pinning both axes to one size token).

## Why

A consumer (Lighthouse's Breeze Automatic co-pilot) replaces its hand-rolled FAB + drag + floating-panel shell with ChatBubble. Its launcher is an expanding "Ask Automatic" label pill — impossible while the launcher is pinned square. With this one gap closed, the adoption is a pure win: the consumer gains Escape-close, focus management and corner flipping its bespoke shell never had, at **−123 lines**.

## Evidence

- Demo gains a bottom-left pill-launcher variant (plus corner pointers in the page body so the fixed-position launchers are discoverable); docs gain the two token rows and a pill recipe.
- Reactive-drag performance measured on the demo bubble: scripted 2s drag, 122 frames, avg 16.66ms, p95 17.8ms, zero frames >20ms — Svelte 5's fine-grained `style:` directive patches only the attribute, so the old direct-DOM-transform rationale no longer applies.
- Consumer adoption verified live: collapsed 52px circle, hover pill expansion, free-drag drop, Escape-close through the host's store bridge.

Single commit per repository policy.

## Summary by CodeRabbit

* **New Features**
  * Added independent width and height customization for the chat launcher.
  * Added support for pill-style launchers with customizable icons, padding, and corner radius.
  * Added an example of a bottom-left pill launcher with a custom panel.

* **Documentation**
  * Documented chat launcher sizing and pill-style customization options.
  * Added guidance for positioning launchers in different corners.

PR #508, opened 2026-09-03
fix(demo,ChatBubble): stop binding to a derived, and say why the gap is reset

murdore:
## What this closes

A transcript audit of this session found **23 unresolved review threads on PRs #458, #459, #460 and #461** — the chat-primitive PRs from 24–26 August. An earlier sweep covered 18 other merged library PRs and never touched these four, so the threads sat unanswered.

I checked all 23 against the code on release `2d3a35e` (3.1.5). **Twenty-one are already fixed.** Two were genuinely open, and this PR lands them.

## The two that were open

**`speech-to-text` demo — binding to a `$derived`.** `bind:value` wrote the user's keystrokes into a `$derived` override that is discarded the moment `committed` or `listening` changes.

One correction to the finding, since it matters for how the fix is judged: the reviewer said the typed text is lost. It is not. The `oninput` handler already writes it into `committed`, so the demo works. What was wrong is that a two-way binding depended entirely on a sibling handler for its correctness. Both composers are now one-way, so the derived is display state and `oninput` is the only write path — which is what it was doing anyway.

**`ChatBubble` — an unexplained token reset.** `--button-content-gap: 0px` had no comment. Button uses that gap to space an icon from its label; the launcher renders an icon alone inside a circle, where the gap pushes the glyph off-centre. The comment now says that, and says why the reset belongs here rather than in Button.

## The twenty-one already closed

| finding | where it is closed |
|---|---|
| document the fold-color token | `docs/AttachmentChipRow.md` |
| document ChatComposer's public additions | all 7 names present in `docs/ChatComposer.md` |
| document `toggleIcon` | `docs/ThinkingIndicator.md` |
| `labelTestId` in the `bare` branch | `ThinkingIndicator.svelte:409`, child span |
| reserve the HITL decision before awaiting | `decisionPending` latched before the await, `HITL.svelte:122` |
| reset TypewriterText on a new response | non-continuation branch resets all four fields |
| clear the demo's attachment model | `onsubmit` clears images, videos and files |
| complete the `sui-chat-confirmation` map | 17 attribute mappings on `HITL.wc.svelte` |
| expose ThinkingIndicator's wc props | all four declared |
| clamp negative `maxVisible` | `Math.max(0, maxVisible)`, with a comment |
| align the `icon` contract | `properties.ts:15` says URL, directs markup to the snippet |
| replace `undefined` sentinels | zero `undefined` in the file |
| expose `chipClasses` | declared and documented |
| reset `listening` in `destroy` | `controller.svelte.ts`, with a comment |
| destroy controllers on route teardown | `onDestroy` at line 92 |
| error messages as live alerts | `role="alert"` on both paragraphs |

Two more are moot: `AgentPicker` and `ChatConfirmation` no longer exist on release.

Three ask for kebab-case attribute mappings on props typed `Object` — `renderText`, `attachmentsPreview`, `actionIcon`, and the AttachmentChipRow callbacks. Declined on the repo's own convention: an attribute is a string, so object and function props are set as properties, and every `Object` entry in the wrappers is declared without one. Adding attributes there would advertise a channel that cannot carry the value.

## Verification

| gate | result |
|---|---|
| `pnpm run check` | exit 0 |
| prettier, eslint on both files | clean |
| `chat-bubble` and `speech-to-text` visual baselines | 2 passed, unchanged |

Every one of the 23 threads is answered on-thread with the file and line that closes it, or the reason it is declined.

### DIFF
--- a/src/lib/ChatBubble/ChatBubble.svelte
+++ b/src/lib/ChatBubble/ChatBubble.svelte
@@ -392,6 +392,10 @@
     --button-border-radius: var(--chat-bubble-border-radius, 50%);
     --button-color: var(--chat-bubble-background-color, #18181b);
     --button-text-color: var(--chat-bubble-color, #ffffff);
+    /* Button spaces its icon from its label with this gap. The launcher renders an
+       icon alone inside a circle, so that gap would push the glyph off-centre --
+       reset to 0 here rather than in Button, whose default is right for a labelled
+       button. docs/ChatBubble.md records this as a pill-launcher caveat. */
     --button-content-gap: 0px;
     --button-box-shadow: var(--chat-bubble-box-shadow, 0 8px 24px rgba(0, 0, 0, 0.25));
     --button-hover-color: var(--chat-bubble-hover-background-color, #27272a);
--- a/src/routes/components/speech-to-text/+page.svelte
+++ b/src/routes/components/speech-to-text/+page.svelte
@@ -68,6 +68,10 @@
       }
     }
   });
+  // Display state only. `bind:value` here would write the user's keystrokes into a
+  // $derived override that is discarded the moment `committed` or `listening` changes;
+  // the write path is `oninput` below, which puts them in `committed` where the
+  // controller's seedTranscript can read them back.
   let simulatedValue = $derived(simulated.listening ? simulated.interimText : committed);
 
   let browserCommitted = $state('');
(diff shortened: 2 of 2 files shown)


Tara-ag (approved):
Reviewed file-by-file. Both changes are safe:

- `ChatBubble.svelte`: purely additive CSS comment explaining the `--button-content-gap: 0px` reset; no functional or public-API change.
- `speech-to-text/+page.svelte`: correctly removes `bind:value` from `$derived` display state in the demo, leaving the write path in `oninput` as the single source of truth. No library API impact.

No blocking criteria apply. Approved.

PR #510, opened 2026-09-04
feat(events): warn once, in dev only, when a deprecated prop spelling is used

murdore:
Phase 2 of `docs/EVENT_CASING_MIGRATION.md`. Phase 1 (#505) gave every grandfathered event prop its correct spelling alongside the old one; this tells a consumer which of the two they are using, once, and what to run to move.

Dev-only, because a production warning is noise a consumer cannot act on mid-incident and 142 of them would bury real output. `import.meta.env.DEV` is the guard specifically because the bundler statically replaces it, so the call is removed from a production build rather than merely skipped at runtime.

Once per component-and-prop rather than per render: a deprecated prop on a list row would otherwise warn a thousand times. The key includes the component because `onclick` is deprecated on many of them — deduplicating on the prop alone would report the first and silently hide the rest.

Rebased onto 3.2.0. **1 commit, 2 files.**

---

# Review responses

## ⚠️ MAJOR — "not exported from `src/lib/index.ts`"

> *A library-wide helper that consumers cannot import is either an internal module that should say so, or an oversight.*

**It is internal, and this codebase already marks internal modules exactly this way.** `src/lib/utils.ts` exports ten functions; the barrel re-exports **three** (`validateInput`, `lockBodyScroll`, `unlockBodyScroll`). `getStorageItem`, `setStorageItem`, `hexToRgb`, `rgbToHsv`, `hsvToRgb`, `hsvToHex` and `hexToHsv` are reachable only from inside `src/lib`. Selective re-export from the barrel is the existing convention for precisely this, not an accident — so "absent from `index.ts`" does not imply oversight here.

Components call `warnDeprecatedProp` themselves when handed a deprecated spelling; a consumer has no call for it. Exporting it would add public API that 4.0.0 then has to carry.

The finding's real substance was the *ambiguity* — nothing said which it was. The module now states it outright, with the `utils.ts` precedent cited, rather than leaving a reader to infer it.

## 💡 MINOR — "hard-coded `4.0.0`" — **taken, and it had stopped being minor**

`DEPRECATION_REMOVAL_VERSION` now names it in one place.

Worth recording why this mattered more than it looked. That string was written while #506 was expected to cut 4.0.0. **#506 shipped inside 3.2.0 as a minor instead** — so this warning was promising consumers a removal version that nothing was committed to, and Phase 3's window was genuinely unanchored rather than merely un-reached.

4.0.0 is now the agreed target. A test pins the constant's value separately from the message assertion: the message test would keep passing against any value the constant held, including one changed by accident, so the pin is what makes the committed target visible in a diff.

## 💡 MINOR — "env stub leaks on assertion failure" — **taken**

`vi.unstubAllEnvs()` moves into `afterEach`. It sat after the assertion in the production-silence case, so a failure there would throw before the cleanup line and leave `DEV=false` set for every later case in the same worker — silencing the warnings those cases assert and surfacing the failure somewhere it did not happen.

---

## CI

7 unit tests here, all passing. The suite reports **49 failures in `scripts/wc-parity/prop-parity.test.ts`** — inherited from the `release` tip this is rebased onto, not from anything in this branch. `prop-parity` is the only failing file, and #512 fixes it. `lint` 0 errors.

### DIFF
--- a/src/lib/deprecation.test.ts
+++ b/src/lib/deprecation.test.ts
@@ -1,0 +1,86 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import {
+  DEPRECATION_REMOVAL_VERSION,
+  resetDeprecationWarnings,
+  warnDeprecatedProp
+} from './deprecation.js';
+
+describe('deprecated-prop warnings', () => {
+  beforeEach(() => {
+    // The suppression set is module-level, so without this a later case would
+    // pass by inheriting an earlier case's suppression rather than earning it.
+    resetDeprecationWarnings();
+    vi.spyOn(console, 'warn').mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+    // Unconditional, rather than at the end of the one case that stubs DEV.
+    // A failing assertion there would throw before the cleanup line and leak
+    // DEV=false into every later case in the same worker, which silences the
+    // warnings they are asserting and reports the failure somewhere else.
+    vi.unstubAllEnvs();
+  });
+
+  it('names the old spelling, the replacement, and the version that removes it', () => {
+    warnDeprecatedProp('Status', 'onbuttonClick', 'onButtonClick');
+
+    expect(console.warn).toHaveBeenCalledTimes(1);
+    const message = vi.mocked(console.warn).mock.calls[0][0];
+    expect(message).toContain('onbuttonClick');
+    expect(message).toContain('onButtonClick');
+    expect(message).toContain('<Status>');
+    expect(message).toContain(DEPRECATION_REMOVAL_VERSION);
+  });
+
+  it('names 4.0.0 as the removal release, so moving it is a deliberate edit', () => {
+    // Pinned separately from the message assertion above. That one would keep
+    // passing against any value the constant held, including one changed by
+    // accident; this one makes the migration's committed target visible in a
+    // diff, which is the whole point of it not being a loose string any more.
+    expect(DEPRECATION_REMOVAL_VERSION).toBe('4.0.0');
+  });
+
+  it('says both spellings work today, so the warning is not read as breakage', () => {
+    warnDeprecatedProp('Toggle', 'onclick', 'onClick');
+
+    expect(vi.mocked(console.warn).mock.calls[0][0]).toContain('both work today');
+  });
+
+  it('points at the codemod rather than leaving the consumer to grep', () => {
+    warnDeprecatedProp('Table', 'onrowClick', 'onRowClick');
+
+    expect(vi.mocked(console.warn).mock.calls[0][0]).toContain('sui-codemod --dry-run');
+  });
+
+  it('warns once per prop however many times it is rendered', () => {
+    for (let i = 0; i < 500; i += 1) {
+      warnDeprecatedProp('ListItem', 'onitemClick', 'onItemClick');
+    }
+
+    expect(console.warn).toHaveBeenCalledTimes(1);
+  });
+
+  it('says nothing in a production build', () => {
+    // The half that actually matters. A dev warning that also fires in
+    // production would put 142 lines into a consumer's console during an
+    // incident, and the guard is the only thing standing between those two
+    // outcomes — so it is worth a test rather than a reading of the source.
+    vi.stubEnv('DEV', false);
+
+    warnDeprecatedProp('Banner', 'ondismiss', 'onDismiss');
+
+    expect(console.warn).not.toHaveBeenCalled();
+    vi.unstubAllEnvs();
+  });
+
+  it('keys on the component too, so the same prop name on two components both warn', () => {
+    // `onclick` is deprecated on many components. Deduplicating on the prop
+    // alone would tell a consumer about the first one and silently hide the
+    // rest, which is worse than not warning at all.
+    warnDeprecatedProp('Avatar', 'onclick', 'onClick');
+    warnDeprecatedProp('Pill', 'onclick', 'onClick');
+
+    expect(console.warn).toHaveBeenCalledTimes(2);
+  });
+});
--- a/src/lib/deprecation.ts
+++ b/src/lib/deprecation.ts
@@ -1,0 +1,63 @@
+/**
+ * Phase 2 of `docs/EVENT_CASING_MIGRATION.md`: tell a consumer, once, that a
+ * prop spelling is going away and what replaces it.
+ *
+ * Two constraints shape this.
+ *
+ * It is dev-only. A production warning is noise a consumer cannot act on
+ * mid-incident, and 142 props warning in a production console would bury real
+ * output. `import.meta.env.DEV` is the guard because it is what the bundler
+ * this library is built with statically replaces, so the whole call is removed
+ * from a production build rather than merely skipped at runtime.
+ *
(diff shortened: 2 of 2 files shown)


Tara-ag:

The PR introduces a clean, dev-only, once-per-component deprecation warning helper with good tests. However, the new `src/lib/deprecation.ts` module is not exported from `src/lib/index.ts`, creating ambiguity about whether it is public API. Two minor quality issues are also noted: a hard-coded removal version and an env-stub cleanup that can leak on test failure. Addressing the public-barrel question is required before merge.

- ⚠️ **MAJOR**: New public module not exported from src/lib/index.ts — `src/lib/deprecation.ts:1`
  The PR adds src/lib/deprecation.ts as a utility but does not export it from src/lib/index.ts. While the stated intent is that components will call warnDeprecatedProp internally, the migration doc describes Phase 2 as a library-wide deprecation helper and the test imports it directly from ./deprecation.js. If consumers or codemods need to reuse this helper (e.g. custom wrappers around library compo…
- 💡 **MINOR**: Hard-coded 4.0.0 version in warning message — `src/lib/deprecation.ts:38`
  The removal version '4.0.0' is hard-coded in the warning string. This is acceptable if the migration plan is immutable, but if the timeline slips the message will become a lie. A constant or a reference to the migration doc would make future updates less error-prone.
- 💡 **MINOR**: Test leaves env stub in place on assertion failure — `src/lib/deprecation.test.ts:49`
  The production-silence test stubs DEV=false but only calls vi.unstubAllEnvs() after the assertion. If expect(...).not.toHaveBeenCalled() throws, the stub leaks into subsequent tests in the same worker, which can make later failures confusing.

Tara-ag (commented):
Review in progress — inline comments follow per file.

src/lib/deprecation.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    for (let i = 0; i < 500; i += 1) {
+      warnDeprecatedProp('ListItem', 'onitemClick', 'onItemClick');
+    }
+
+    expect(console.warn).toHaveBeenCalledTimes(1);
+  });
+
+  it('says nothing in a production build', () => {
+    // The half that actually matters. A dev warning that also fires in
+    // production would put 142 lines into a consumer's console during an
```
Tara-ag, line 66: 💡 MINOR: Test leaves env stub in place on assertion failure

`vi.unstubAllEnvs()` is called after the assertion. If `expect(console.warn).not.toHaveBeenCalled()` throws, the `DEV=false` stub leaks into later tests in the same worker, making subsequent failures harder to diagnose.

Suggested fix: restore envs in `afterEach` (or use `onTestFinished`) so cleanup always runs:
```ts
afterEach(() => {
  vi.restoreAllMocks();
  vi.unstubAllEnvs();
});
```

src/lib/deprecation.ts
Tara-ag, line 1: ⚠️ MAJOR: New public module not exported from src/lib/index.ts

The PR adds `src/lib/deprecation.ts` as a library-wide deprecation helper but it is not exported from `src/lib/index.ts`. The migration doc frames Phase 2 as a reusable mechanism, and the test imports it directly from `./deprecation.js`, so leaving it out of the public barrel creates ambiguity about whether it is public API. If it is meant to be internal, it should live under `src/lib/internal/` or be explicitly excluded from the public surface.

Suggested fix: add `export { warnDeprecatedProp, resetDeprecationWarnings } from './deprecation.js';` to `src/lib/index.ts` if public; otherwise relocate/document as internal.

src/lib/deprecation.ts
```diff
@@ -0,0 +1,43 @@
...
+export const warnDeprecatedProp = (component: string, oldName: string, newName: string): void => {
+  if (import.meta.env.DEV !== true) {
+    return;
+  }
+  const key = `${component}.${oldName}`;
+  if (seen.has(key)) {
+    return;
+  }
+  seen.add(key);
+  console.warn(
```
Tara-ag, line 58: 💡 MINOR: Hard-coded `4.0.0` version in warning message

The removal version is baked into the warning string. If the timeline slips, the message becomes inaccurate. Extract a constant or derive it from a single source of truth.

Suggested fix:
```ts
const REMOVAL_VERSION = '4.0.0';
console.warn(
  `[svelte-ui-components] \`${oldName}\` on <${component}> is deprecated and will be ` +
    `removed in ${REMOVAL_VERSION}. Use \`${newName}\` instead — both work today. ` +
    `Run \`npx sui-codemod --dry-run ./src\` to see every affected call site.`
);
```

Tara-ag (commented):
Review complete. The change is sound in isolation, but the new deprecation helper needs clarity on whether it is public API (see inline comment). Two minor quality notes also included.

Tara-ag (commented):
Review findings from Yama code review agent.

src/lib/deprecation.ts
Tara-ag, line 1: ⚠️ MAJOR: New public module not exported from src/lib/index.ts

The PR adds src/lib/deprecation.ts as a library-wide deprecation helper but it is not exported from src/lib/index.ts. The migration doc frames Phase 2 as a reusable mechanism, and the test imports it directly from ./deprecation.js, so leaving it out of the public barrel creates ambiguity about whether it is public API. If it is meant to be internal, it should live under src/lib/internal/ or be explicitly excluded from the public surface.

**Suggested fix:** Add `export { warnDeprecatedProp, resetDeprecationWarnings } from './deprecation.js';` to src/lib/index.ts if public; otherwise relocate/document as internal.

src/lib/deprecation.ts
Tara-ag, line 58: 💡 MINOR: Hard-coded 4.0.0 version in warning message

The removal version is baked into the warning string. If the timeline slips, the message becomes inaccurate. Extract a constant or derive it from a single source of truth.

**Suggested fix:** Extract `const REMOVAL_VERSION = '4.0.0';` at module scope and interpolate it into the message.

src/lib/deprecation.test.ts
Tara-ag, line 66: 💡 MINOR: Test leaves env stub in place on assertion failure

vi.unstubAllEnvs() is called after the assertion. If expect(console.warn).not.toHaveBeenCalled() throws, the DEV=false stub leaks into later tests in the same worker, making subsequent failures harder to diagnose.

**Suggested fix:** Restore envs in afterEach (or use onTestFinished) so cleanup always runs, e.g. `afterEach(() => { vi.restoreAllMocks(); vi.unstubAllEnvs(); });`

Tara-ag (changes requested):
Yama code review findings.

Tara-ag (commented):
Reviewed file by file. One minor nit on redundant cleanup; otherwise the change is sound and safe to merge.

src/lib/deprecation.test.ts
```diff
@@ -0,0 +1,86 @@
...
```
Tara-ag, line 66: 💡 MINOR: Redundant `vi.unstubAllEnvs()` call — `afterEach` already cleans up env stubs unconditionally, so this line is dead code. Remove it to keep the production-silence test focused on its assertion.

Tara-ag (approved):
Approved with one minor cleanup note. The deprecation warning helper is internal, dev-only, deduplicated correctly, and well-tested. No public API break, no SSR issue, no security concern.

PR #511, opened 2026-09-04
fix(MarkdownText): give a wide table the width it needs to scroll

murdore:
## What

`MarkdownText` never scrolled a wide table. It set `display: block; overflow-x: auto` on the `<table>` but gave it no intrinsic width, so the block box was exactly its container's width and there was nothing to overflow with. A six-column model reply compressed to fit instead of scrolling. Every consumer of the component got this.

## The shape, after review

The first version of this fix scrolled the table itself. Review caught that it was accessible only to a mouse, and the rework moved to the shape `Table.svelte` already uses:

- **The wrapper scrolls, not the table.** `renderMarkdown` post-processes marked's HTML and wraps every `<table>` in `<div class="markdown-table-wrapper" tabindex="0">`. `display: block` on a `<table>` is what makes `overflow-x` apply to the element itself, and it also strips the table's semantics for assistive technology. The wrapper keeps the table a table.
- **It can be reached and it shows that it has been.** `tabindex="0"` lets arrow keys scroll it; a `:focus-visible` outline (`--markdown-text-focus-outline-color`, default `#3b82f6`, the same default as Table's focusable rows) shows where focus is.
- **It is named only when the caller names it.** `tableLabel` adds `role="region" aria-label="…"` (escaped). Without a label the wrapper carries no role, because an unnamed region is worse than none. Declared on `sui-markdown-text` as `table-label`; the prop-parity ratchet from #506 caught its absence.
- **The width fix is two halves, both required.** `width: max-content; min-width: 100%` on the table gives it something to overflow with. `.markdown-text` sets `overflow-wrap: anywhere`, which makes `max-content` resolve as `min-content` inside a table, so cells opt back out with `overflow-wrap: normal`. Either half alone measures 480 against 480; together the same fixture reads 764 against 480 and scrolls. Cells keep `white-space: normal`, so a table that fits still wraps inside its cells.

## Tests

Unit (`src/lib/MarkdownText/markdown.test.ts`, vitest), pinning the string the post-processor emits:

| case | pins |
|---|---|
| wraps a table | exactly one `<div class="markdown-table-wrapper" tabindex="0"><table>` … `</table></div>` |
| names the region only when labelled | no `role`/`aria-label` without `tableLabel`; both with it; an empty label is no label |
| escapes the label | `a"b<c>&d` → `a&quot;b&lt;c&gt;&amp;d`, cannot break out of the attribute |
| several tables | each gets its own wrapper, two of each tag |
| inline mode | no wrapper, no table |
| no table | untouched |

Integration (`tests/markdown-text-table-scroll.test.ts`, Playwright, on the demo page):

| test | pins |
|---|---|
| the wrapper overflows while the table stays a table | `scrollWidth > clientWidth` on the wrapper **and** the table's computed `display` is `table` (guards against the first version of this fix) |
| keyboard reachable and really scrolls | `tabindex="0"`, focus lands, `ArrowRight` moves `scrollLeft` past 0 |
| visible focus ring | after focus + a key press, `outline-style` is not `none` and `outline-width` is greater than 0 |
| columns aligned on every row | every `tbody tr`'s `td` left edges equal the header's, not only the first row |
| region named only with a label | labelled demo has `role="region"` + `aria-label="Recent orders"`; the unlabelled one has `tabindex` and no role |

The demo page gains a six-column table in a 560px panel and a labelled variant. The existing two-column demo table could never overflow anything, which is why this was never caught.

## Verification

| gate | result |
|---|---|
| `vitest run src/lib/MarkdownText/markdown.test.ts` | **30 passed** (24 existing + 6 new) |
| `playwright test tests/markdown-text-table-scroll.test.ts` | **5 passed**, private port, fresh build |
| `pnpm run lint` | clean; event-casing 142 known, 0 new |
| `pnpm run check` | exit 0 |
| visual | one baseline changed, `markdown-text.png`, for the two added demo sections; the focus outline is `:focus-visible` only and does not appear in an unfocused snapshot |

The right check on this PR is that the integration step **ran**, not that the overall check turned green: while `release` was red, CI skipped the Playwright steps and the run still reported complete. On this head the `Integration tests (playwright)` step is `success`, not `skipped`.

## Review

Every thread is answered on the thread it was raised in and resolved. Thirteen threads over three rounds: five on the first version (mouse-only scrolling, semantics stripped, first-row-only alignment check, docs `max-width`, a garden-path sentence), six on the rework (no visible focus ring ×2, docs type reference stale ×2, no unit coverage, typo), and two on the focus-ring commit (the docs Props table omitted `tableLabel` ×2). All addressed in this commit.

## Origin

Found while measuring `MarkdownText` against a consumer's hand-rolled `marked` + DOMPurify pipeline to decide whether that consumer should migrate. The migration is not proceeding, for unrelated reasons. This defect is real regardless and predates that question.


Tara-ag (approved):
Approved. The CSS change is narrowly scoped to `MarkdownText` table rendering, adds clear explanatory comments, and is backed by Playwright tests that verify both scrolling and column alignment. No public props, exports, or CSS custom property names were changed, so this is not a breaking API change. Visual baseline delta is limited to the expected demo-page screenshot update.

Tara-ag:
💡 MINOR: Visual baseline growth is expected but not reflected in docs. The PR body correctly notes `markdown-text.png` changed from 980×1644 to 980×2040 because of the added wide-table demo section. Consider mentioning the wide-table overflow example in `docs/MarkdownText.md` so future maintainers understand the visual baseline intentionally includes it.

Tara-ag:
💡 MINOR: New wide-table demo lacks docs coverage. The PR adds a public demo scenario and Playwright tests for wide-table scrolling, but `docs/MarkdownText.md` does not describe the table overflow behavior or the CSS mechanism (`width: max-content`, `min-width: 100%`, cell `overflow-wrap: normal`). Downstream consumers theming tables need to know that tables now scroll horizontally while preserving column alignment. Suggest adding a short "Tables" subsection to the docs.

murdore:
Both docs comments were right, and they were the same gap seen from two angles: the PR added a public demo scenario and a CSS mechanism, and `docs/MarkdownText.md` described neither. Fixed in `33fb4fd` (amended, still one commit).

`docs/MarkdownText.md` gains a **Wide tables** section between Security model and CSS Variables. It documents the behaviour — the table scrolls inside its own box so the message does not, and columns stay aligned while it does — and then the part a future maintainer actually needs, which is that the fix is three declarations that only work together:

| declaration | on | why it is needed |
| --- | --- | --- |
| `display: block` + `overflow-x: auto` | `table` | makes the table its own scroll container |
| `width: max-content` + `min-width: 100%` | `table` | gives it something to overflow with, while a narrow table still fills the message |
| `overflow-wrap: normal` | `th`, `td` | the container sets `overflow-wrap: anywhere`, which inside a table makes every character a break opportunity, so `max-content` collapses to `min-content` and scrolling stops |

That third row is the one worth writing down. `width: max-content` alone measures 480 against 480 and does not scroll — I hit exactly that on the first attempt and had to measure a second time to find out why. Anyone who later "simplifies" the cell rule will silently reintroduce the bug, and the visual baseline will not catch it because the table still renders, just compressed.

The doc also records that cells keep `white-space: normal`, so a table that does fit still wraps inside its cells rather than being forced onto one line — a deliberate difference from the usual `nowrap` approach to this problem.

No source or test change in the amend; the CSS and both Playwright tests are as reviewed and approved.

tests/markdown-text-table-scroll.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  });
+
+  // The width that makes it scroll must not come at the cost of the column model:
+  // header and body cells have to keep sharing one set of column edges.
+  test('header and body columns stay aligned while it scrolls', async ({ page }) => {
+    await page.goto('/components/markdown-text');
+
+    const scope = page.getByTestId('markdown-text-wide-table');
+    const edges = await scope.evaluate((root) => ({
+      head: [...root.querySelectorAll('th')].map((el) =>
```
Tara-ag, line 74: 💬 SUGGESTION: The alignment test only checks `tbody tr:first-child td`, so it validates the first data row but not subsequent rows. If a future regression broke alignment for later rows, this test would still pass. Consider comparing header edges against every body row (e.g. collect each `tr`'s `td` left edges and assert they all equal `edges.head`).

Tara-ag (commented):
Reviewed file by file. The fix is narrowly scoped to `MarkdownText` table CSS, adds Playwright coverage, and documents the behaviour. No blocking issues found. One suggestion on test robustness is posted inline.

src/lib/MarkdownText/MarkdownText.svelte
```diff
@@ -106,16 +106,29 @@
...
+     overflow with. Without an intrinsic width the block box is exactly its
+     container's width, so a six-column table compresses instead of scrolling.
+     `max-content` supplies that width, and `min-width` keeps a narrow table
+     filling the message rather than shrink-wrapping to its content. */
   .markdown-text :global(table) {
     display: block;
+    width: max-content;
+    min-width: 100%;
     max-width: 100%;
     overflow-x: auto;
```
Tara-ag, line 117: ⚠️ MAJOR: The table is now a scroll container (`display: block; overflow-x: auto`), but the `<table>` element itself is not focusable and has no ARIA role/label. Keyboard-only users cannot Tab to it and scroll it with arrow keys. Precedent in this repo: `BarChart` adds `tabindex={scrollable ? 0 : null}` and `Book` uses `role="region" tabindex="0"` for scrollable regions.

Fix: make the rendered table focusable. The cleanest place is the custom renderer in `src/lib/MarkdownText/markdown.ts` — override the `table` renderer to wrap tables in `<div class="markdown-table-wrapper" tabindex="0" role="region" aria-label="Data table">...</div>` and move `overflow-x: auto` to that wrapper, or emit the attributes directly on `<table>`. Add a `:focus-visible` outline so the focused table is visibly indicated.

Tara-ag, line 117: ⚠️ MAJOR: Setting `display: block` directly on the `<table>` element can strip screen-reader table semantics (row/column navigation commands may disappear). The project already uses the safer wrapper-div pattern in `src/lib/Table/Table.svelte` (`.table-scroll { overflow-x: auto; }`) while keeping the `<table>` as `display: table`.

Fix: move the scroll container off the table. Override the `table` renderer in `src/lib/MarkdownText/markdown.ts` to emit `<div class="markdown-table-wrapper" tabindex="0" role="region" aria-label="Data table"><table>...</table></div>`. Style `.markdown-table-wrapper` with `overflow-x: auto; max-width: 100%;` and keep `<table>` as `display: table; width: max-content; min-width: 100%;`. This preserves both keyboard focus and screen-reader table semantics.

Tara-ag (changes requested):
The table-scrolling fix is measured and well-documented, but the scrollable table needs keyboard accessibility and screen-reader semantics before it can ship. See inline comments.

Tara-ag:

The PR correctly fixes the wide-table scrolling bug and preserves table semantics/accessibility with a wrapper + tabindex, backed by Playwright tests. However, two MAJOR issues remain: the keyboard-reachable wrapper has no visible focus indicator, and the new wrapping/label behavior lacks unit-test coverage. A MINOR docs inconsistency in the Type Reference block also needs fixing. No blocking public-API break, security defect, or SSR-safety issue was found.

- ⚠️ **MAJOR**: No unit test coverage for table wrapping / tableLabel behavior — `src/lib/MarkdownText/markdown.test.ts:160`
  The new wrapTables logic in src/lib/MarkdownText/markdown.ts and the tableLabel option are only exercised by Playwright demo-page tests. There are no Vitest unit tests for: (1) tables being wrapped in .markdown-table-wrapper, (2) tabindex/role/aria-label being emitted conditionally, (3) inline mode skipping wrapping, or (4) multiple tables each getting their own wrapper. A demo refactor or route c…
- ⚠️ **MAJOR**: Focusable scroll wrapper lacks visible focus indicator — `src/lib/MarkdownText/MarkdownText.svelte:116`
  .markdown-table-wrapper has tabindex='0' so it is keyboard-reachable, but there is no :focus or :focus-visible style in the component CSS. In many browser/OS combinations the default outline is suppressed by surrounding CSS or by the component's own reset, leaving keyboard users with no visible indication that the wrapper has focus. This violates the focus-area requirement for visible focus indica…
- 💡 **MINOR**: Docs type reference omits new tableLabel prop — `docs/MarkdownText.md:130`
  docs/MarkdownText.md adds the tableLabel prop in the prose and usage sections, but the 'Type Reference' block at the bottom still shows the old MarkdownTextProperties and RenderMarkdownOptions shapes without tableLabel. Downstream consumers copying from that block will miss the new prop, and the docs are now inconsistent with the exported types.

src/lib/MarkdownText/MarkdownText.svelte
```diff
@@ -106,16 +106,29 @@
...
     border-collapse: collapse;
```
Tara-ag: ⚠️ MAJOR: Overflowing table is not keyboard-scrollable. The CSS makes the table its own scroll container, but the generated `<table>` is not focusable. A non-focusable element with `overflow-x: auto` cannot be scrolled with arrow keys by keyboard-only users, so the new scrolling behavior is inaccessible. Because the table markup is produced by `marked`, add a `table` renderer in `src/lib/MarkdownText/markdown.ts` that emits `<table tabindex="0" aria-label="Scrollable table">...</table>` (or wrap the table in a focusable region with an accessible name).

docs/MarkdownText.md
```diff
@@ -50,6 +50,26 @@ The `marked` package is a peer dependency, needed only when you actually use `Ma
 - **External links are tab-safe.** `http:`/`https:` links open with `target="_blank" rel="noopener noreferrer"` (the same default the library's Button/Card apply); relative, `mailto:` and `tel:` links keep same-tab navigation. Autolinks and bare URLs go through the same protocol guard as explicit links.
 - **Pre-sanitized HTML has its own prop.** If you already hold trusted, sanitized HTML, use `ChatMessage`'s `html` prop — `markdown` is for untrusted source text.
 
+## Wide tables
```
Tara-ag, line 54: 💡 MINOR: Typo in docs. "A model asked for a summary answers with a table" should read "A model asked for a summary answer with a table".

Tara-ag (changes requested):
One actionable accessibility issue and one doc typo found. Please address the keyboard-scrollable table concern before merging.

Tara-ag (commented):

src/lib/MarkdownText/MarkdownText.svelte
```diff
@@ -106,16 +106,36 @@
...
 
-  .markdown-text :global(table) {
+  /* The wrapper scrolls, not the table. `display: block` on a `<table>` is what
+     would make `overflow-x` work on the element itself, but it also strips the
+     table's semantics for assistive technology, so the scroll container is a
+     separate box and the table stays a table. `max-width` bounds the wrapper to
+     the message; the table's `max-content` width is what overflows it. */
+  .markdown-text :global(.markdown-table-wrapper) {
     display: block;
     max-width: 100%;
```
Tara-ag, line 116: ⚠️ MAJOR: `.markdown-table-wrapper` has `tabindex="0"` so it is keyboard-reachable, but there is no `:focus` or `:focus-visible` style. In many browser/OS combinations the default outline is suppressed by surrounding CSS or resets, leaving keyboard users with no visible focus indicator.

Add a visible focus indicator, for example:
```css
.markdown-text :global(.markdown-table-wrapper:focus-visible) {
  outline: 2px solid var(--markdown-text-focus-outline-color, var(--markdown-text-link-color, #6d28d9));
  outline-offset: 2px;
}
```

docs/MarkdownText.md
Tara-ag, line 1: 💡 MINOR: The 'Type Reference' block still shows the old `MarkdownTextProperties` and `RenderMarkdownOptions` shapes without `tableLabel`, while the prose and usage sections above document the new prop. Consumers copying types from this block will miss it.

Update the Type Reference block to include `tableLabel?: string` in both types, matching `src/lib/MarkdownText/properties.ts`.

Tara-ag (commented):
Reviewed file-by-file. Two verified MAJOR findings and one MINOR finding. Note: the unit-test finding on `src/lib/MarkdownText/markdown.test.ts` could not be posted inline because that file is unchanged in this PR's diff.

Summary:
- ⚠️ MAJOR: Focusable `.markdown-table-wrapper` lacks a visible `:focus-visible` indicator.
- ⚠️ MAJOR: New table-wrapping behavior has Playwright coverage but no Vitest unit tests.
- 💡 MINOR: Docs Type Reference block omits the new `tableLabel` prop.

No blocking public-API break, security issue, or SSR-safety defect found.

Tara-ag (commented):
Review findings: one accessibility issue and two test/docs gaps. See inline comments for details.

src/lib/MarkdownText/MarkdownText.svelte
Tara-ag, line 116: 🔒 MAJOR: Focusable scroll wrapper lacks visible focus indicator. `.markdown-table-wrapper` has `tabindex="0"` so it is keyboard-reachable, but there is no `:focus` or `:focus-visible` style. In many browser/OS combinations the default outline is suppressed by surrounding CSS or resets, leaving keyboard users with no visible focus indicator.

Add a visible focus indicator, e.g.:
```css
.markdown-text :global(.markdown-table-wrapper:focus-visible) {
  outline: 2px solid var(--markdown-text-focus-outline-color, var(--markdown-text-link-color, #6d28d9));
  outline-offset: 2px;
}
```

docs/MarkdownText.md
Tara-ag, line 1: 💡 MINOR: Docs type reference omits new `tableLabel` prop. `docs/MarkdownText.md` adds the `tableLabel` prop in prose and usage sections, but the Type Reference block still shows the old `MarkdownTextProperties` and `RenderMarkdownOptions` shapes without `tableLabel`. Consumers copying types from this block will miss the new prop.

Update the Type Reference block to include `tableLabel?: string` in both `MarkdownTextProperties` and `RenderMarkdownOptions`, matching `src/lib/MarkdownText/properties.ts`.

src/lib/MarkdownText/markdown.ts
```diff
@@ -104,6 +104,31 @@ function annotateExternalLinks(html: string): string {
   );
 }
 
+/* A wide table has to scroll, and the scroll container has to be reachable.
```
Tara-ag, line 107: ⚠️ MAJOR: No unit test coverage for table wrapping / tableLabel behavior. The new `wrapTables` logic and `tableLabel` option are only exercised by Playwright demo-page tests. There are no Vitest unit tests for tables being wrapped, conditional tabindex/role/aria-label, inline mode skipping wrapping, or multiple tables each getting their own wrapper. A demo refactor or route change could silently drop this behavior from CI.

Add Vitest cases to `markdown.test.ts` covering wrapper emission, conditional aria-label/role, inline mode, and multiple-table inputs.

Tara-ag (changes requested):
Review findings: one accessibility issue and two test/docs gaps. See inline comments for details.

tests/markdown-text-table-scroll.test.ts
murdore: Fixed in the rework. The alignment test now walks every `tbody tr` and asserts each row's `td` left edges equal the header's, with the failing row named in the message. It was first-row-only before, which is exactly the kind of check that passes while a later row drifts.

docs/MarkdownText.md
```diff
@@ -50,6 +50,26 @@ The `marked` package is a peer dependency, needed only when you actually use `Ma
...
+| ---------------------------------------- | ---------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `display: block` + `overflow-x: auto`    | `table`    | makes the table its own scroll container                                                                                                                                                                                                       |
+| `width: max-content` + `min-width: 100%` | `table`    | gives it something to overflow with, while a narrow table still fills the message                                                                                                                                                              |
+| `overflow-wrap: normal`                  | `th`, `td` | the container sets `overflow-wrap: anywhere` so a long unbroken string cannot force the message wide; inside a table that makes every character a break opportunity, so `max-content` collapses to `min-content` and the table stops scrolling |
```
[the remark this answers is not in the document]
murdore: Addressed at f8d0dd7. The declaration table in the docs now names the element each rule sits on: `max-width: 100%` is on `.markdown-table-wrapper` (the scroll container, bounded to the message) and `width: max-content; min-width: 100%` is on the `table` (what overflows it). The earlier text put the width and the bound on the same element, which was the confusion.

src/lib/MarkdownText/MarkdownText.svelte
murdore: Correct, and it changed the shape of the fix. At f8d0dd7 the table is no longer the scroll container: every `<table>` is wrapped in `.markdown-table-wrapper` with `tabindex="0"`, so arrow keys scroll it, and it draws a `:focus-visible` outline so the tab stop is visible. Pinned by two Playwright tests: focus lands and `ArrowRight` moves `scrollLeft` past 0; and after focus plus a key press the computed `outline-style` is not `none`.

murdore: Correct. At f8d0dd7 the `<table>` keeps `display: table`; the scroll container is a separate wrapper, the same shape `Table.svelte` uses with `.table-scroll`. The first Playwright test asserts the wrapper overflows **and** that the table's computed `display` is `table`, as a guard against the first version of this fix coming back.

src/lib/MarkdownText/MarkdownText.svelte
murdore: Fixed at f8d0dd7 with the wrapper you suggested rather than a `tabindex` on the table. One deliberate difference from the suggestion: no fixed `aria-label="Scrollable table"`. The wrapper carries `role="region"` and an `aria-label` only when the caller passes `tableLabel`, because a region named the same for every table is a landmark the user cannot tell apart, and one with no name is worse than none. Without a label it is still focusable and keyboard-scrollable, just not announced as a landmark. Tested both ways on the demo page.

docs/MarkdownText.md
murdore: The suggested replacement would not parse: in "A model asked for a summary answers with a table", *answers* is the verb and *asked for a summary* modifies *model*, so "answer" breaks the sentence. But it garden-paths, which is why it read as a typo, so at f8d0dd7 the sentence is rewritten to "Ask a model for a summary and it answers with a table". Same meaning, no ambiguity.

src/lib/MarkdownText/MarkdownText.svelte
murdore: Agreed and fixed at f8d0dd7: `.markdown-table-wrapper:focus-visible` draws `outline: 2px solid var(--markdown-text-focus-outline-color, #3b82f6)` with a 2px offset. The default is the one `Table.svelte` uses for its focusable rows rather than the link colour, so focus looks the same across the library. The token is in the CSS Variables table. A Playwright test focuses the wrapper, presses a key, and asserts `outline-style` is not `none` and `outline-width` is greater than 0.

murdore: Same finding as the thread above; fixed at f8d0dd7 with the `:focus-visible` outline, the `--markdown-text-focus-outline-color` token, and a test that reads the computed outline after focus.

docs/MarkdownText.md
murdore: Fixed at f8d0dd7. The Type Reference block now carries `tableLabel?: string` in both `MarkdownTextProperties` and `RenderMarkdownOptions`, matching `properties.ts`. It is hand-written, which is how it drifted.

murdore: Same as the thread above; both types in the Type Reference block carry `tableLabel?: string` at f8d0dd7.

src/lib/MarkdownText/markdown.ts
murdore: Agreed. At f8d0dd7 `src/lib/MarkdownText/markdown.test.ts` gains six vitest cases on the emitted string: exactly one wrapper per table with `tabindex="0"`; `role`/`aria-label` present only with a label and an empty label treated as none; the label escaped (`a"b<c>&d` → `a&quot;b&lt;c&gt;&amp;d`); several tables each wrapped; inline mode emitting no wrapper; and markdown without a table left untouched. 30 pass in the file. The Playwright tests stay for the browser behaviour the string cannot show (scrolling, focus, semantics).

docs/MarkdownText.md
```diff
@@ -50,6 +50,42 @@ The `marked` package is a peer dependency, needed only when you actually use `Ma
 - **External links are tab-safe.** `http:`/`https:` links open with `target="_blank" rel="noopener noreferrer"` (the same default the library's Button/Card apply); relative, `mailto:` and `tel:` links keep same-tab navigation. Autolinks and bare URLs go through the same protocol guard as explicit links.
 - **Pre-sanitized HTML has its own prop.** If you already hold trusted, sanitized HTML, use `ChatMessage`'s `html` prop — `markdown` is for untrusted source text.
 
+## Wide tables
+
+Ask a model for a summary and it answers with a table, and a chat bubble is rarely
```
Tara-ag, line 56: 💡 MINOR: The Props table omits the new public `tableLabel` prop introduced by this PR. Consumers scanning the table will miss it even though it is documented narratively below. Add a row: `| tableLabel | string | No | - | Accessible name for the scrollable table region; adds role="region" and aria-label when supplied. |`

Tara-ag (commented):
Review complete. The change is well-tested and addresses a real accessibility/usability defect. One minor documentation completeness issue: the structured Props table should list the new `tableLabel` prop. No blocking concerns.

docs/MarkdownText.md
```diff
@@ -81,10 +118,12 @@ export type MarkdownTextProperties = {
   breaks?: boolean;
   testId?: string;
   classes?: string;
+  tableLabel?: string;
```
Tara-ag, line 122: 💡 MINOR: `tableLabel` is added to the type reference but is missing from the Props table above (lines 33–40). Consumers scanning the Props table will miss the new public prop and its accessibility contract.

Add a row to the Props table:
```
| tableLabel | `string`  | No       | `-`     | Accessible name for the scrollable table region; adds `role="region"` and `aria-label` when supplied. |
```

Tara-ag (approved):
Approved with one minor documentation completeness note. The PR correctly fixes the wide-table compression defect by moving scrolling to a wrapper, preserves table semantics, adds keyboard reachability and visible focus, and ships matching unit/integration tests plus web-component attribute mapping. No public API breakage.

docs/MarkdownText.md
murdore: Fixed at d14cc4b: the Props table carries a `tableLabel` row (type `string`, optional, no default) describing what it adds and what happens without it. The table, the narrative section, the CSS Variables table and the Type Reference now all name the prop.

docs/MarkdownText.md
murdore: Fixed at d14cc4b: the Props table carries a `tableLabel` row (type `string`, optional, no default) describing what it adds and what happens without it. The table, the narrative section, the CSS Variables table and the Type Reference now all name the prop.

Tara-ag (commented):
Reviewing file by file. One minor concern so far about test coupling to demo markup.

tests/markdown-text-table-scroll.test.ts
```diff
@@ -0,0 +1,106 @@
+import { expect, test } from '@playwright/test';
+
+test.describe('MarkdownText — a wide table scrolls, stays a table, and can be reached', () => {
+  // The scroll container is a wrapper rather than the table itself. `display: block`
+  // on a `<table>` is what would make overflow-x work on the element directly, but it
+  // also strips the table's semantics for assistive technology — so the box that
```
Tara-ag, line 6: 💡 MINOR: Playwright regression test is coupled to demo-page markup. The test navigates to `/components/markdown-text` and selects elements by demo-specific `testId` values (`markdown-text-wide-table`, `markdown-text-labelled-table`). This ties regression coverage to the demo page's markup; changes to the demo layout or test IDs will silently break the test. The repo's recurring reviewer guidance prefers isolated fixtures/harnesses over demo-page selectors.

Suggestion: Create a minimal Svelte harness under `tests/fixtures/` (or a dedicated test route) that renders `MarkdownText` with the wide-table markdown and required props, and have the Playwright test load that fixture instead of the demo page.

Tara-ag (commented):
Reviewed file by file. The fix is correct and well-documented. One minor concern about the Playwright regression test being coupled to demo-page markup; everything else looks good.

tests/markdown-text-table-scroll.test.ts
murdore: Declining this one, with the repository as the evidence rather than my preference. The demo routes are this repo's sanctioned Playwright surface: 99 of the 105 spec files in `tests/` navigate to `/components/<name>`, the other six drive the web-component bundle, and there is no `tests/fixtures/` directory or harness route anywhere. `docs/VISUAL_TESTING.md` treats the demo routes as the test surface too. A one-off fixture for MarkdownText would be the first of its kind and would take this component's coverage off the surface every other component is tested on. The coupling the thread describes is real but it is the house convention, and it fails loudly (a locator times out) rather than silently. The section that these tests select by `testId` carries a comment naming the spec that depends on it, which is how the other demo pages handle the same coupling. If the repo moves to isolated fixtures, MarkdownText should move with everything else, in one change.

PR #512, opened 2026-09-04
fix(wc): declare the corrected event spellings on every custom-element wrapper

murdore:
Closes the MAJOR Yama raised on #505 an hour before it merged. It is correct, it shipped in **3.2.0**, and the shape of it is worse than "some wrappers are missing props".

## The documented upgrade path is the broken one

Measured against the published `@juspay/svelte-ui-components@3.2.0` bundle in Chromium, with a real click on the dismiss button:

```
<sui-banner>  el.ondismiss = fn   (deprecated spelling)  →  FIRES
<sui-banner>  el.onDismiss = fn   (corrected spelling)   →  DOES NOTHING
```

`ondismiss` is a real accessor on the element's prototype. `onDismiss` is absent, so assigning it creates an own expando the component never reads. **A web-component consumer who follows the migration guidance silently loses their handler; one who ignores it keeps working.** (Reproduction by `componen-reuse-63`.)

## How it happened

Phase 1 (#505) gave 56 components a second spelling for each grandfathered event prop. It stopped at the Svelte layer — a custom element only forwards what `customElement.props` declares. Verified in the published tarball, not the source:

```
sui-toggle: camelCase=NONE   lowercase=['onclick']
sui-button: camelCase=NONE   lowercase=['onclick','onkeydown','onkeyup',...]
sui-modal:  camelCase=NONE   lowercase=['onclose','onkeydown']
sui-select: camelCase=NONE   lowercase=['onchange']
sui-banner: camelCase=NONE   lowercase=['onclick','ondismiss']
```

`release` has also been failing its own suite since the merge — **49 failed / 343 passed**, all in `scripts/wc-parity/prop-parity.test.ts`, the ratchet #506 built to catch exactly this. It worked; nothing ran it, because every test workflow here is `pull_request`-only. #514 fixes that cause; this fixes the instance.

## Both directions are real

126 declarations across 49 wrappers — 119 camelCase and 7 native ones on `sui-input`. `DESIGN_PRINCIPLES.md` §3 keeps Svelte's lowercase spelling for a native DOM event forwarded as-is, so the target depends on the signature:

| Component | Signature | Corrects to |
|---|---|---|
| `Toggle.onclick` | `(checked: boolean) => void` — synthesized | `onClick` |
| `Input.onFocus` | `(event: FocusEvent) => void` — forwarded | `onfocus` |

#505's generator discriminates on the signature and had both right.

## The transform

`scripts/migrate/alias-wc-props.ts`, committed rather than run once and discarded, for the reason its sibling gives: 126 declarations across 49 files is not a diff anyone reviews by eye. Re-running is a no-op and a test pins that.

`isEventProp` deliberately does not match a bare `/^on[A-Za-z]/` — that would also take a prop named `once` and declare it a callback. `NATIVE_EVENTS` is exported from `casing.ts` so the rule is stated once.

## Measured, not reasoned

No wrapper had ever declared both spellings of the same event before. Svelte lowercases to derive the observed attribute, so `onclick` and `onClick` both claim `onclick` and `$$g_p` returns whichever key comes first. The last time this library reasoned about a custom-element prop instead of measuring it — `children` — the reasoning was wrong. So `tests/wc-event-casing-parity.spec.ts` asks a browser. Six tests, all passing:

- both spellings are real accessors on the built element
- the corrected spelling reaches the component
- **the legacy spelling still fires** — the regression guard for declaring both
- the corrected one wins when both are set, matching `$derived(onClick ?? onclickLegacy)`
- the lowercase direction works (`onfocus` on `sui-input`)
- **a declared prop beats the native handler accessor it shadows**

That last one answers a reviewer's caution that `ontoggle` is a native `GlobalEventHandlers` property and might swallow the declaration. The premise is right — `'onclick' in HTMLElement.prototype` and `'ontoggle' in HTMLElement.prototype` are both true, and the test asserts that so it cannot quietly stop testing anything. The conclusion is not: Svelte defines its accessors on the generated class, which sits below `HTMLElement.prototype`, so the declaration shadows the native name rather than the reverse. Both spellings still fire.

## Verification

| Gate | Before (on `release`) | After |
|---|---|---|
| Unit | 343 passed / **49 failed** | **400 passed** |
| Integration | — | **437 passed** |
| Lint | — | 0 errors |
| Type-check | — | 832 files, 0 errors |

### DIFF
--- a/scripts/migrate/alias-wc-props.ts
+++ b/scripts/migrate/alias-wc-props.ts
@@ -1,0 +1,216 @@
+import { readFileSync, writeFileSync } from 'node:fs';
+import { join, resolve } from 'node:path';
+import { parse } from 'svelte/compiler';
+import { readWrapperParity } from '../wc-parity/prop-parity.ts';
+import { NATIVE_EVENTS } from './casing.ts';
+
+/**
+ * Phase 1 of the event-casing migration stopped at the Svelte layer.
+ * alias-event-props.ts gave each grandfathered event prop a second spelling on
+ * the component, but a custom element only forwards what `customElement.props`
+ * declares, so none of the corrected spellings were reachable through a web
+ * component. This closes that half.
+ *
+ * Committed rather than run once and thrown away, for the same reason its
+ * sibling is: 126 declarations across 49 wrappers is not a diff anyone can
+ * meaningfully review by eye, and the transform is the only honest description
+ * of what was done to them. Re-running it is a no-op, and a test asserts that.
+ *
+ * Scope is deliberately narrow. It adds declarations the parity ratchet already
+ * reports as missing, and only for props named like event handlers. Anything
+ * else is reported and skipped rather than guessed at: `{ type: 'Object' }` is
+ * right for a callback and wrong for a string, and a wrong declaration is worse
+ * than an absent one because it looks settled.
+ */
+
+type UnknownRecord = Record<string, unknown>;
+
+type Addition = {
+  readonly wrapper: string;
+  readonly props: readonly string[];
+};
+
+type Skip = { readonly wrapper: string; readonly prop: string; readonly reason: string };
+
+/** Mirrors scripts/wc-parity/prop-parity.ts; assertions and predicates are banned repo-wide. */
+function asRecord(value: unknown): UnknownRecord {
+  if (typeof value !== 'object' || value === null) {
+    return {};
+  }
+  const record: UnknownRecord = {};
+  for (const key of Object.keys(value)) {
+    Object.defineProperty(record, key, {
+      value: Reflect.get(value, key),
+      enumerable: true,
+      writable: true,
+      configurable: true
+    });
+  }
+  return record;
+}
+
+function asList(value: unknown): readonly unknown[] {
+  return Array.isArray(value) ? value : [];
+}
+
+function keyName(node: unknown): string | null {
+  const key = asRecord(asRecord(node).key);
+  if (typeof key.name === 'string') {
+    return key.name;
+  }
+  return typeof key.value === 'string' ? key.value : null;
+}
+
+function offsetOf(node: unknown, key: 'start' | 'end'): number {
+  const value = asRecord(node)[key];
+  return typeof value === 'number' ? value : -1;
+}
+
+/**
+ * Whether a prop is an event callback, and so safe to declare as `type: 'Object'`.
+ *
+ * Both spellings the casing rule produces are accepted, and nothing else. A
+ * synthesized event is camelCase from the character after `on`, and a native DOM
+ * event forwarded as-is keeps Svelte's lowercase spelling — so `onClick` and
+ * `onfocus` both qualify while `once` does not. A bare `/^on[A-Za-z]/` would
+ * have taken `once` too and declared it a callback; there is no such prop today,
+ * which is exactly why the guard belongs here rather than after one appears.
+ *
+ * A grandfathered lowercase *synthesized* name (`onbarclick`) is deliberately
+ * not matched. Every one of those is already declared, so none can reach this;
+ * if one ever does, a reported skip that leaves the ratchet red is a better
+ * outcome than a guessed declaration that looks settled.
+ */
+export function isEventProp(name: string): boolean {
+  if (/^on[A-Z]/.test(name)) {
+    return true;
+  }
+  return name.startsWith('on') && NATIVE_EVENTS.has(name.slice(2));
+}
+
+/**
+ * The end offset of the last entry inside `customElement.props`, which is where
+ * a new declaration is appended. Returns -1 when the wrapper has no props object
+ * or an empty one: both mean this transform has nothing safe to do, since it
+ * would have to invent the object's formatting rather than match it.
+ */
+export function lastPropertyEnd(source: string): number {
+  const root = asRecord(parse(source, { modern: true }));
+  const options = asRecord(root.options);
+  for (const attribute of asList(options.attributes)) {
+    const record = asRecord(attribute);
+    if (record.name !== 'customElement') {
+      continue;
+    }
+    const single = Array.isArray(record.value) ? record.value[0] : record.value;
+    const object = asRecord(asRecord(single).expression);
+    for (const property of asList(object.properties)) {
+      if (keyName(property) !== 'props') {
+        continue;
+      }
+      const entries = asList(asRecord(asRecord(property).value).properties);
+      return entries.length === 0 ? -1 : offsetOf(entries[entries.length - 1], 'end');
+    }
+  }
+  return -1;
+}
+
+/**
+ * Indentation of the line the given offset sits on, so appended entries line up
+ * with the ones already there instead of assuming a fixed depth.
+ */
+function indentAt(source: string, offset: number): string {
+  const lineStart = source.lastIndexOf('\n', offset - 1) + 1;
+  const line = source.slice(lineStart, offset);
+  return line.slice(0, line.length - line.trimStart().length);
+}
+
+export function declareProps(source: string, props: readonly string[]): string | null {
+  if (props.length === 0) {
+    return source;
+  }
+  const insertAt = lastPropertyEnd(source);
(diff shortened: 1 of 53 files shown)


tests/wc-event-casing-parity.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+        }
+        const element = document.createElement(tag);
+        for (const prop of spellings) {
+          // Stops at HTMLElement.prototype so a native handler accessor
+          // (onclick, onfocus and friends all live there) is never mistaken for
+          // a declared prop. Only a setter the generated class owns counts.
+          let found = false;
+          let proto: object | null = Object.getPrototypeOf(element);
+          while (proto !== null && proto !== HTMLElement.prototype) {
+            const descriptor = Object.getOwnPropertyDescriptor(proto, prop);
```
Tara-ag, line 111: 💡 MINOR: Browser-only assertion for >100 casing pairs has no static guarantee the fixture stays meaningful. This check only runs in a browser job. If the `casingPairs()` generator is broken or the wrappers change, the unit ratchet may still pass while this integration assertion silently asserts almost nothing.

**Suggested fix:** Add a Vitest unit test (e.g. in `scripts/migrate/alias-wc-props.test.ts` or `scripts/wc-parity/prop-parity.test.ts`) that computes `casingPairs()` and asserts a minimum count, so fixture integrity is checked by `npm run test:unit` without requiring a browser build.

scripts/migrate/alias-wc-props.ts
```diff
@@ -0,0 +1,216 @@
...
+    return {};
+  }
+  const record: UnknownRecord = {};
+  for (const key of Object.keys(value)) {
+    Object.defineProperty(record, key, {
+      value: Reflect.get(value, key),
+      enumerable: true,
+      writable: true,
+      configurable: true
+    });
```
Tara-ag, line 47: 💡 MINOR: Transform uses `Object.defineProperty` to copy enumerable keys instead of a simpler assignment. The reflective loop is harder to read than a direct copy, and it is not obvious why descriptor flags must be preserved for AST nodes.

**Suggested fix:** Replace the `Object.defineProperty` loop with a direct assignment loop on a fresh record, or add a concise comment explaining why descriptor flags must be preserved. If flags are irrelevant, simplify to `record[key] = Reflect.get(value, key);`.

src/wc/components/Input.wc.svelte
```diff
@@ -60,7 +60,14 @@
       minRows: { type: 'Number', attribute: 'min-rows' },
       maxRows: { type: 'Number', attribute: 'max-rows' },
```
Tara-ag, line 61: ⚠️ MAJOR: New `onfocus`/`onblur`/`oninput` declarations on `sui-input` shadow native `GlobalEventHandlers` accessors. These names are native accessors on `HTMLElement.prototype`. Shadowing them means a host-page assignment like `input.onfocus = myHandler` now sets the component prop instead of the native event handler, and the component's synthetic handler signature may differ from the native `FocusEvent` listener consumers expect. This changes web-component behavior and is a breaking public-API change relative to 3.2.0.

**Suggested fix:** For native event names, keep the existing camelCase legacy props (`onFocus`, `onBlur`, etc.) as the only declared custom-element props and forward them to the component's lowercase props internally, or declare the lowercase names with explicit `attribute: '...'` mappings that do not shadow native handlers. Do not declare native handler names directly on the host without a major-version bump and migration note.

src/wc/components/Toggle.wc.svelte
```diff
@@ -8,7 +8,8 @@
       disabled: { type: 'Boolean', reflect: true },
       classes: { type: 'String' },
       onclick: { type: 'Object' },
```
Tara-ag, line 10: ⚠️ MAJOR: Declaring both `onclick` and `onClick` on a custom element collides on the observed attribute. Svelte derives the observed HTML attribute by lowercasing the prop name, so both declarations claim the `onclick` attribute and the generated `$$g_p` getter returns whichever key `Object.keys` yields first. The Playwright test asserts this is harmless today, but it relies on undocumented key-enumeration behavior and the two props are not independent: assigning both `element.onclick` and `element.onClick` routes through the same attribute-backed accessor, so the Svelte-layer precedence `$derived(onClick ?? onclickLegacy)` cannot be guaranteed at the element boundary.

**Suggested fix:** Choose a single canonical prop name per event for the custom-element API and expose the legacy spelling through an explicit alias that maps to the same internal prop, instead of declaring two props that collide on the same attribute. If both must be kept for backward compatibility, document the collision and precedence in `DESIGN_PRINCIPLES.md` and add a deterministic unit test for the generated accessor precedence.

Tara-ag (changes requested):
I reviewed the PR file-by-file. It addresses a real regression in the web-component wrappers, and the test coverage is welcome. However, the chosen fix — declaring both legacy and corrected spellings as custom-element props — introduces a collision on the observed HTML attribute for every native event name, and it shadows native `GlobalEventHandlers` accessors on the host. That is a breaking change to the web-component public API and relies on undocumented Svelte key-enumeration behavior. I am requesting changes to resolve the attribute collision and native-handler shadowing before merge.

murdore:
Reproduction for the inversion this fixes, measured against the **published** `@juspay/svelte-ui-components@3.2.0` rather than against source — plus a correction to something I said that turned out to be wrong.

## The evidence

Loaded the package's own `dist-wc/index.js` in Chromium and clicked the real dismiss button with Playwright (a trusted click, not a synthetic dispatch):

| `<sui-banner>` | result |
| --- | --- |
| `el.ondismiss = fn` — deprecated spelling | **fires** |
| `el.onDismiss = fn` — corrected spelling | **does nothing** |

Structurally, on the same build: `ondismiss` is a real accessor on the element's prototype, `onDismiss` is absent, and assigning it creates an own expando the component never reads.

So the consumer who follows the migration loses their handler and the one who ignores it keeps working.

## Correction: `ontoggle` is not a false friend

I warned that `ontoggle` was host-reserved and that `sui-accordion` declared neither spelling. **The second half is simply false** — `Accordion.wc.svelte` declares `ontoggle: { type: 'Object' }` at line 12 on `release`, and it is only `onToggle` that was missing, which is exactly why the ratchet named the camelCase one.

What I actually measured was that `Object.getOwnPropertyDescriptor(El.prototype, 'ontoggle')` is absent while assigning it creates no own expando. That is consistent with an accessor existing somewhere in the chain, and I inferred "therefore it hits the native `GlobalEventHandlers` one" without testing whether the handler fired. The structural probe was sound; the conclusion on top of it was not measured, and it was wrong.

## Two traps worth keeping

Both of these produce a false "the handler did not fire", which is the most dangerous possible false positive here because it looks like loud confirmation of the bug:

1. **A synthetic `dispatchEvent` does not reach Svelte's delegated handler.** My first attempt showed *both* spellings dead on `sui-banner`. Only the trusted click separated them.
2. **`?.` on a shadow-root query turns "selector matched nothing" into a silent no-op.** A fixture whose control sits behind an unpassed snippet prop renders no control at all, and the test then reports the library as broken when the observation is about the fixture.

Reproduction harness: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-lighthouse-fork-feat-componen-reuse/c031a8c7-d321-4ef0-babc-a47b079e7544/scratchpad/wctest`

Not a review of this PR — it is not mine and I have not read the 126 declarations. This is the consumer-facing evidence for why it matters.

Tara-ag:

## 🤖 Yama Review Summary

**Verified findings (1):**

- **MINOR**: Migration script produces a double comma when the props object already has a trailing comma — `scripts/migrate/alias-wc-props.ts:110`

- 💡 **MINOR**: Migration script produces a double comma when the props object already has a trailing comma — `scripts/migrate/alias-wc-props.ts:110`
  In `declareProps`, the transform unconditionally inserts `,\n` immediately after `lastPropertyEnd`. The AST `end` offset points to the end of the last property's value, not including any trailing comma. If a wrapper's `customElement.props` object already uses trailing commas, the inserted comma duplicates the existing one and produces invalid syntax.

PR #513, opened 2026-09-04
fix(changelog): render every release's commits and summary, starting with 3.2.0

murdore:
3.2.0 shipped with no `## [3.2.0]` heading — its content sat under `## [Unreleased]` — and **no mention of #506 at all** (`grep -ic 'custom-element|parity'` returned 0).

## Immediate cause: a release race

#505 and #506 merged fourteen seconds apart. Both runs computed 3.2.0. The one that generated the changelog was building from `0f53508`, which contains #505 and not #506. The other published the tarball and then failed pushing its own changelog commit:

```
97576f4..d89a061  release -> origin/release
Local branch is behind remote. Attempting to rebase...
CONFLICT (content): Merge conflict in docs/CHANGELOG.md
error: could not apply dd26fa1... chore(release): 3.2.0
```

Same shape as the 3.1.5 race. The published tarball is fine — it contains both PRs' work. Only the changelog lost content.

## Root cause: the template's commit loop never ran

Regenerating fixes the heading but not the omission. Chasing that turned up something larger — `docs/templates/changelog.hbs` iterated:

```hbs
{{#each sections}}     <!-- not a field auto-changelog provides -->
  {{#each commits}}    <!-- correct field names, never executed -->
```

`sections` does not exist in auto-changelog's data model (it provides `merges`, `fixes`, `commits`). **No released version in this changelog has ever listed a commit** — every section from 1.0.0 onward was empty. The `{{#if summary}}` above it worked, which is why the file never looked obviously broken: the pending release's prose showed under Unreleased and was overwritten by the next one.

Removing the dead wrapper is the fix. The inner loop is unchanged.

Two more rendering faults in the same template, corrected while there:
- `- ` and `{{subject}}` sat on separate source lines, which Markdown reads as an empty list item followed by a paragraph.
- The H1 had the intro paragraph glued to it (`# Changelog All notable changes...`).

## Why `chore(release)` commits are filtered

This does more than remove a bullet. `releaseSummary` takes its prose from the newest commit in a release, and that is always the bodiless `chore(release)` commit — so every section rendered without a summary. Dropping them **restores the writeups**, and 3.2.0's summary is now #506's, the one that went missing.

It has to be a command-line flag: auto-changelog reads `ignoreCommitPattern` from argv only, and accepts the same key in `.auto-changelog` silently without ever applying it. Both camelCase and kebab-case were tried in the config first; neither works. `.auto-changelog` is left unchanged rather than carrying a key that does nothing.

## Formatting

`docs/CHANGELOG.md` is not prettier-ignored, and the generator's output is no longer prettier-clean now that sections carry prose. The release step formats it — without that, the release commit would land a file failing the format check for the next unrelated PR to touch it. (The old changelog was prettier-clean only because it was bare headings.)

## The diff

Large because the file is generated: **265 sections** gained the commits and summaries they should always have had. `prettier --check` clean, lint 0 errors.

## Summary by CodeRabbit

* **Documentation**
  * Changelog generation now omits internal release commits for cleaner release history.
  * Changelog formatting is preserved consistently, preventing unintended automated reformatting.
  * Changelog entries now display breaking changes more clearly alongside the affected change.
  * Changelog headings and entries use a cleaner, more readable layout.


Tara-ag (approved):
Approved. The template fix correctly switches from the non-existent `sections` field to `commits`, the `--ignore-commit-pattern` flag restores meaningful release summaries by excluding bodiless `chore(release)` commits, and adding `docs/CHANGELOG.md` to `.prettierignore` avoids prettier corrupting generated prose. No component code, public API, or CSS custom properties are touched.

Tara-ag (approved):
Approved. This is a documentation/CI fix that corrects the changelog template and release workflow. No component code, public API, CSS custom properties, or security-sensitive changes are involved. The generated CHANGELOG.md is large but expected for a full regeneration.

Tara-ag (approved):
Reviewed the changelog-generation fix. No component source is touched, so the public API, accessibility, reactivity, and SSR focus areas are unaffected. Secret scan is clean. The new `--ignore-commit-pattern` regex correctly preserves the existing `[skip ci]` filter while adding the `chore(release):` filter, and moving `docs/CHANGELOG.md` to `.prettierignore` is the right call given Prettier’s non-idempotent rewriting of underscore emphasis. The regenerated changelog correctly restores the 3.2.0 section and lists commits/summaries for prior releases. Approved.

PR #514, opened 2026-09-04
fix(release): gate the publish on tests, and stop two releases racing

murdore:
3.2.0 published from a tree whose own unit suite had **49 failures**, and nothing reported it. Two independent faults in this workflow allowed that; both are fixed here.

## 1. Nothing tested what was published

Every quality workflow in this repository is `pull_request`-only:

| Workflow | Trigger |
|---|---|
| CI (lint, type-check, unit, playwright) | `pull_request` **only** |
| Visual Regression | `pull_request` **only** |
| Yama PR Review | `pull_request` **only** |
| Deploy to Pages | push + PR |
| **Release and Publish** | **push — `pnpm lint`, then publish** |

#505 and #506 were each green against their own base — #506 was rebased onto 3.1.6, which predates #505 — and merged fourteen seconds apart. The combination was never tested anywhere, because the only workflow that tests anything cannot see a merge commit.

`scripts/wc-parity/prop-parity.test.ts` was built precisely to catch a wrapper falling behind its component. It had been failing on `release` since the merge. It worked perfectly; nothing ran it.

**The fix:** a `verify` job running lint, check, unit and integration, which `release` now `needs`.

```
detect-changes → verify → release → publish-mcp
```

`needs` makes the dependency the thing that blocks, and the job graph shows a reader what gates what. Linting moves into it rather than being duplicated.

**Checked, not assumed.** This branch is cut from `d89a061` — the exact tip that published 3.2.0. Running the verify job's own unit step there:

```
Tests  49 failed | 343 passed (392)
ELIFECYCLE  Command failed with exit code 1
```

## 2. Two releases ran at once

The same fourteen seconds started two release runs. Both reached `npm publish` and `git push` together; one pushed its changelog commit and the other died rebasing onto it:

```
CONFLICT (content): Merge conflict in docs/CHANGELOG.md
error: could not apply dd26fa1... chore(release): 3.2.0
```

So **the run that published the tarball is the one that failed**, and its notes were lost. 3.1.5 was damaged the same way.

A concurrency group serialises them. `cancel-in-progress` is **false and must stay false**: cancelling a release part-way is worse than delaying it, because the npm publish, the git tag and the GitHub release are three separate steps, and a run killed between them leaves a version on the registry with no tag behind it.

### What this deliberately does not fix

Serialising does not make the second run *correct*. The version is chosen by `git log -1` in "Determine version bump type", which reads only the tip commit of that run's own checkout — so two merges each with a `feat:` tip both compute the same next version from the same base. **That is why both runs wanted 3.2.0.**

Serialised, the second run now fails loudly at `npm publish` on a version that already exists, instead of racing the first. Making it choose correctly means reading the branch tip rather than the pushed SHA, which changes what a release builds and belongs in its own change rather than being smuggled into this one.

## Note on CI here

This PR's own CI is red on `prop-parity` — inherited from the `release` tip it is cut from, not from anything it changes (it touches one workflow file). #512 fixes that; this goes green on rebase after it lands.

## Summary by CodeRabbit

* **Chores**
  * Release runs are now coordinated to prevent overlapping executions.
  * Releases now undergo automated validation, including linting, type checks, unit tests, and integration tests, before publishing.
  * Test reports are preserved as downloadable artifacts for easier troubleshooting.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,31 @@ on:
   push:
     branches: [release]
 
+# Two releases must never be in flight together. 3.1.5 and 3.2.0 were both
+# damaged by the same race: merges seconds apart start one run each, and both
+# reach `npm publish` and `git push` at the same time. On 3.2.0 one run pushed
+# its changelog commit and the other died rebasing onto it --
+# `CONFLICT (content): Merge conflict in docs/CHANGELOG.md` -- so the release
+# that published the tarball is the one that failed, and its notes were lost.
+#
+# cancel-in-progress is false and must stay false. Cancelling a release mid-way
+# is the one outcome worse than delaying it: the npm publish, the git tag and
+# the GitHub release are three separate steps, and a run killed between them
+# leaves a version on the registry with no tag behind it.
+#
+# This serialises the runs. It does NOT make the second one correct, and the
+# rest of the race is deliberately left visible rather than half-fixed: the
+# version is chosen by `git log -1` in "Determine version bump type", which
+# reads only the tip commit of that run's own checkout. Two merges each with a
+# `feat:` tip therefore both compute the same next version from the same base,
+# and the queued run will now fail loudly at `npm publish` on a version that
+# already exists rather than racing the first one. Making it pick the right
+# version means reading the branch tip instead of the pushed SHA, which changes
+# what a release builds and belongs in its own change.
+concurrency:
+  group: release-${{ github.ref }}
+  cancel-in-progress: false
+
 # Pinned so both publish jobs upgrade to the same npm, which is what supports
 # OIDC trusted publishing. Bump in one place.
 env:
@@ -65,8 +90,110 @@ jobs:
             echo "No MCP package changes detected"
           fi
 
-  release:
+  # The gate that did not exist when 3.2.0 published.
+  #
+  # Every quality workflow in this repository -- ci, visual, yama-review -- is
+  # `pull_request`-only. A push to `release` ran Pages and this file, and this
+  # file ran `pnpm lint` and then published. So two PRs that were each green
+  # against their own base merged fourteen seconds apart, the combination was
+  # never tested anywhere, and 3.2.0 went to npm from a tree whose own unit
+  # suite had 49 failures. Nothing was broken; nothing was watching.
+  #
+  # A separate job rather than more steps inside `release`: `needs` makes the
+  # dependency the thing that blocks, so publishing cannot begin until the
+  # suite has passed, and a reader can see that from the job graph.
+  # Unconditional on purpose. Gating this on `root_changed` would have left the
+  # MCP package publishing through an ungated path: `publish-mcp` runs when
+  # `needs.release.result` is 'skipped', and a skipped `release` is exactly what
+  # an MCP-only push produces -- so the one job this file exists to gate would
+  # have sailed past it. Worse, a *failed* verify also renders `release` as
+  # 'skipped', so the gate would have opened precisely when it should shut.
+  # Running it on every push costs one suite on a push that releases nothing,
+  # and removes both holes.
+  verify:
     needs: detect-changes
+    runs-on: ubuntu-latest
+    # Every publish now waits on this job, so a step that hangs would hold the
+    # release open indefinitely instead of failing it. Playwright install plus
+    # both suites finish in well under half of this.
+    timeout-minutes: 30
+
+    # `contents: read` is all this job's own steps need. `actions: write` is here
+    # by review decision, not by demonstrated need: the artifact upload and the
+    # pnpm store cache were both observed working under `contents: read` alone
+    # (ci.yml runs the identical pair and uploaded a 335 MB report on d49920fff).
+    # It is granted deliberately and recorded here so the next reader knows this
+    # workflow -- which publishes to npm -- can also cancel runs and delete
+    # artifacts, and can drop the line if that reach is ever unwanted.
+    permissions:
+      contents: read
+      actions: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      # pnpm before node, deliberately: setup-node's `cache: pnpm` resolves the
+      # store path by running pnpm, so pnpm has to exist first. `run_install: false`
+      # stops action-setup installing before that cache is configured, which would
+      # make the cache do nothing on every run.
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 10.18.3
+          run_install: false
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Lint (prettier, eslint, event-casing)
+        run: pnpm lint
+
+      - name: Type-check
+        run: pnpm check
+
+      # The step that would have caught it. scripts/wc-parity/prop-parity.test.ts
+      # is a vitest test, and it had been failing on `release` since the merge.
+      # The root `check` does not cover the web-component build, which this
+      # library also ships; a WC-only type error would reach the registry
+      # otherwise.
+      - name: Type-check web components
+        run: pnpm run check:wc
+
+      - name: Unit tests (vitest)
+        run: pnpm run test:unit --run
+
+      - name: Install Playwright browsers
+        run: pnpm exec playwright install --with-deps chromium
+
+      - name: Integration tests (playwright)
+        run: pnpm run test:integration
+
+      - name: Upload Playwright report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          # v4 artifacts are immutable, so a fixed name makes the upload fail on
+          # a re-run -- which is the one time the report is most wanted.
+          name: release-verify-playwright-report-${{ github.run_id }}-${{ github.run_attempt }}
+          path: playwright-report/
+          retention-days: 14
+          # `if: always()` also fires when the job was cancelled or failed before
+          # Playwright ran. `warn` is the setting that satisfies both halves of the
+          # review on this line: nothing is uploaded when the directory is absent, so
+          # there is no empty artifact masquerading as evidence, and the run still
+          # says so in the log rather than passing over it in silence. `ignore` hid
+          # the diagnosis; `error` would fail a job whose tests never ran.
+          if-no-files-found: warn
+
+  release:
+    needs: [detect-changes, verify]
     if: needs.detect-changes.outputs.root_changed == 'true'
     runs-on: ubuntu-latest
     environment: npm
@@ -97,8 +224,8 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
-      - name: Run linting
-        run: pnpm lint
+      # Linting moved to the `verify` job this one now needs, so it gates the
+      # publish rather than running beside it.
 
       - name: Determine version bump type
         id: version-type
@@ -263,6 +390,74 @@ jobs:
             exit 1
           fi
 
+      # BEFORE the publish, deliberately. npm publishes are effectively permanent
+      # -- a version cannot be re-published and unpublishing is restricted -- while
+      # a commit on `release` is revertible. With publish first, anything that stops
+      # this push leaves the registry permanently ahead of the repo, with no tag, no
+      # `chore(release)` commit and no GitHub Release naming what shipped.
+      #
+      # That is not hypothetical: run 33961677670 published 3.4.1 and then failed
+      # here with `GH006: Protected branch update failed ... 3 of 3 required status
+      # checks are expected`, because `release` requires checks/build/visual with
+      # enforce_admins on, and a freshly-created release commit has no check runs.
+      # 3.4.1 is on npm today with no trace of it in this repository.
+      #
+      # Ordering it first cannot make that push succeed -- only a settings change
+      # can -- but it makes the failure safe: the run stops before publishing and
+      # the registry stays clean. The reverse failure (pushed, then publish fails)
+      # leaves a version commit for an unpublished version, which the existing
+      # `npm view ... && exit 0` guard below makes safe to resolve by re-running.
+      - name: Push changes (without tags)
+        run: |
+          # Pull latest changes to avoid conflicts
+          git fetch origin release
+
+          # Check if we're behind
+          if [ $(git rev-list --count HEAD..origin/release) -gt 0 ]; then
+            echo "Local branch is behind remote. Attempting to rebase..."
+            git rebase origin/release
+          fi
+
+          git push origin release
+
+          echo "Successfully pushed changes for version ${{ steps.version-bump.outputs.new_version }}"
+
+      # Tag before the publish, for the same reason the branch push moved above
+      # it: a tag is a git write and can be deleted, an npm publish cannot be
+      # undone. With both git writes ahead of it, anything that blocks either one
+      # stops the run with the registry untouched.
+      #
+      # This does NOT adopt the review's alternative of returning to publish-first
+      # with `git push --atomic origin release <tag>`. Atomicity is between the two
+      # git refs; it does nothing about the ordering against npm, and the failure
+      # that stranded 3.4.1 was the branch push being REJECTED, which an atomic
+      # push would simply extend to the tag as well -- published, with neither ref
+      # written.
+      #
+      # The GitHub Release stays after the publish: it announces a released
+      # version, so it should follow the thing it announces.
+      - name: Create and push git tag
+        run: |
+          TAG_NAME="${{ steps.version-bump.outputs.new_version }}"
+
+          # Delete tag if it exists locally
+          if git tag -l | grep -q "^${TAG_NAME}$"; then
+            git tag -d "$TAG_NAME"
+          fi
+
+          # Delete tag if it exists remotely
+          if git ls-remote --tags origin | grep -q "refs/tags/${TAG_NAME}$"; then
+            git push origin ":refs/tags/$TAG_NAME"
+          fi
+
+          # Create new tag
+          git tag -a "$TAG_NAME" -m "Release $TAG_NAME"
+
+          # Push the tag
+          git push origin "$TAG_NAME"
+
+          echo "Successfully created and pushed tag $TAG_NAME"
+
       - name: Upgrade npm for OIDC trusted publishing
         run: npx -y npm@${NPM_CLI_VERSION} install -g npm@${NPM_CLI_VERSION}
 
@@ -286,44 +481,6 @@ jobs:
           echo "Publishing $PACKAGE_NAME@$NEW_VERSION to NPM..."
           npm publish --access public --provenance
           echo "Successfully published $PACKAGE_NAME@$NEW_VERSION to NPM"
-
-      - name: Push changes (without tags)
-        run: |
-          # Pull latest changes to avoid conflicts
-          git fetch origin release
-
-          # Check if we're behind
-          if [ $(git rev-list --count HEAD..origin/release) -gt 0 ]; then
-            echo "Local branch is behind remote. Attempting to rebase..."
-            git rebase origin/release
-          fi
-
-          # Push changes first (without tags)
-          git push origin release
-
-          echo "Successfully pushed changes for version ${{ steps.version-bump.outputs.new_version }}"
-
-      - name: Create and push git tag
-        run: |
-          TAG_NAME="${{ steps.version-bump.outputs.new_version }}"
-
-          # Delete tag if it exists locally
-          if git tag -l | grep -q "^${TAG_NAME}$"; then
-            git tag -d "$TAG_NAME"
-          fi
-
-          # Delete tag if it exists remotely
-          if git ls-remote --tags origin | grep -q "refs/tags/${TAG_NAME}$"; then
-            git push origin ":refs/tags/$TAG_NAME"
-          fi
-
-          # Create new tag
-          git tag -a "$TAG_NAME" -m "Release $TAG_NAME"
-
-          # Push the tag
-          git push origin "$TAG_NAME"
-
-          echo "Successfully created and pushed tag $TAG_NAME"
 
       - name: Create GitHub Release
         uses: softprops/action-gh-release@v2
@@ -354,11 +511,18 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   publish-mcp:
-    needs: [detect-changes, release]
+    needs: [detect-changes, verify, release]
     # Run when the MCP package changed, after the root job either published or
     # was skipped (i.e. don't run if the root release job actually failed).
+    #
+    # `verify` is named explicitly rather than inherited through `release`.
+    # `always()` means a skipped or failed dependency does not stop this job by
+    # itself, and both of those render `needs.release.result` as 'skipped' --
+    # so keying only on `release` would let an MCP publish through on a red
+    # suite. The gate has to be asserted, not assumed.
     if: |
       always() &&
+      needs.verify.result == 'success' &&
       needs.detect-changes.outputs.mcp_changed == 'true' &&
       (needs.release.result == 'success' || needs.release.result == 'skipped')
     runs-on: ubuntu-latest


Tara-ag (approved):
No issues found. The new `verify` job correctly gates publishing on lint, type-check, unit, and integration tests, and the dependency graph makes the gate visible. Removing the redundant lint step from `release` is appropriate since `verify` now blocks it.

.github/workflows/release.yml
```diff
@@ -65,10 +90,72 @@ jobs:
...
     if: needs.detect-changes.outputs.root_changed == 'true'
     runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
```
Tara-ag, line 135: ⚠️ MAJOR: `pnpm run test:unit -- --run` passes an extra `--` separator. pnpm forwards `--run` correctly today, but this is non-idiomatic and fragile if the script changes. Use `pnpm run test:unit --run`.

.github/workflows/release.yml
```diff
@@ -65,10 +90,72 @@ jobs:
...
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
+        with:
```
Tara-ag, line 142: 💡 MINOR: `actions/upload-artifact@v4` artifacts are immutable, so a fixed name `release-verify-playwright-report` will collide on workflow rerun. Append `${{ github.run_id }}-${{ github.run_attempt }}` to the name.

.github/workflows/release.yml
```diff
@@ -65,10 +90,72 @@ jobs:
             echo "No MCP package changes detected"
           fi
```
Tara-ag, line 91: ⚠️ MAJOR: `verify` is gated on `root_changed == 'true'`, so it does not run for MCP-only pushes. `publish-mcp` then runs with `needs.release.result == 'skipped'` and no test/lint gate at all. Either make `verify` unconditional (recommended) or add a dedicated `verify-mcp` job that `publish-mcp` needs.

.github/workflows/release.yml
```diff
@@ -65,10 +90,72 @@ jobs:
...
-  release:
+  # The gate that did not exist when 3.2.0 published.
+  #
+  # Every quality workflow in this repository -- ci, visual, yama-review -- is
+  # `pull_request`-only. A push to `release` ran Pages and this file, and this
+  # file ran `pnpm lint` and then published. So two PRs that were each green
+  # against their own base merged fourteen seconds apart, the combination was
+  # never tested anywhere, and 3.2.0 went to npm from a tree whose own unit
+  # suite had 49 failures. Nothing was broken; nothing was watching.
+  #
```
Tara-ag, line 101: 💬 SUGGESTION: The `verify` job installs dependencies cold on every release run. Add `cache: 'pnpm'` to `actions/setup-node` (or an explicit pnpm-store cache) to speed up the publish gate.

Tara-ag (changes requested):
Reviewed `.github/workflows/release.yml`. Found one MAJOR gap: the new `verify` job is skipped for MCP-only changes, leaving `publish-mcp` ungated; and one MAJOR nit on the unit-test invocation. Two lower-severity items noted. Requesting changes.

.github/workflows/release.yml
Tara-ag, line 101: 💬 SUGGESTION: The verify job installs dependencies from scratch on every release run. Consider enabling pnpm caching via `cache: 'pnpm'` in the setup-node step (or an explicit pnpm store cache) to speed up the release pipeline.

.github/workflows/release.yml
Tara-ag, line 142: 💡 MINOR: `actions/upload-artifact@v4` uses a fixed artifact name. v4 artifacts are immutable by default, so reruns of this workflow will fail to upload because the name already exists. Append the run id and attempt, e.g. `release-verify-playwright-report-${{ github.run_id }}-${{ github.run_attempt }}`.

.github/workflows/release.yml
Tara-ag, line 135: ⚠️ MAJOR: `pnpm run test:unit -- --run` includes an unnecessary `--` separator. With pnpm this passes `--run` through, but the idiomatic and less fragile form is `pnpm run test:unit --run`. The extra `--` is a latent footgun if the script definition ever changes.

.github/workflows/release.yml
Tara-ag, line 91: ⚠️ MAJOR: The verify job is conditioned on `needs.detect-changes.outputs.root_changed == 'true'`, so it is skipped for MCP-only changes. However, `publish-mcp` still runs in that scenario and currently has no verification gate. If the intent is to block publish on green tests, this condition leaves MCP publishes untested. Remove the `if` so verify runs on every push, or add a dedicated verify-mcp job that `publish-mcp` needs.

Tara-ag (changes requested):
Reviewed `.github/workflows/release.yml`. The concurrency gate and verify job are good additions, but the verify job's `if` condition means MCP-only publishes bypass verification. Also a few smaller CI nits. Please address the MAJOR items before merging.

Tara-ag:

The PR materially improves release safety by gating publish on a full verify job and adding concurrency control, which would have prevented the 3.2.0 incident. However, the deliberate reordering of push-before-publish introduces a new failure mode where release metadata (git tag / GitHub Release) can be orphaned when a queued run skips publish because the version already exists. One MAJOR and two lower-severity findings need resolution or explicit acceptance before merge.

- ⚠️ **MAJOR**: Push-before-publish ordering is unsafe with concurrency and rebase — `.github/workflows/release.yml:288`
  The change pushes the release commit before publishing. Because the concurrency group serialises runs but does not cancel in-progress ones, a second queued run can now rebase onto the first run's pushed commit, re-compute the same version, and then skip publishing because `npm view` sees the version already exists. That is acceptable. However, if the rebase succeeds and the push proceeds, the job …
- 💡 **MINOR**: Artifact upload uses `if-no-files-found: ignore` which can hide failures — `.github/workflows/release.yml:184`
  The Playwright report upload uses `if-no-files-found: ignore`. While the comment explains this avoids empty artifacts when Playwright did not run, it also means a genuine failure where `playwright-report/` is missing unexpectedly will be silently ignored, making debugging harder. The trade-off is documented but suboptimal.
- 💬 **SUGGESTION**: Missing `check:wc` in verify job — `.github/workflows/release.yml:165`
  The `verify` job runs `pnpm check` (which includes `check:wc-parity`) but does not run `pnpm check:wc`, the dedicated type-check for the web-component build. Since this library ships a web-component build and the PR description emphasizes catching what was missed before, adding `check:wc` would close another gap.

Tara-ag (approved):
Reviewed `.github/workflows/release.yml`.

- The new `verify` job correctly gates both `release` and `publish-mcp` on lint, type-check, unit and integration tests.
- Making `verify` unconditional (instead of gating on `root_changed`) closes the MCP-only bypass hole described in the comments.
- The concurrency group serialises release-branch runs without cancelling in-flight jobs, which is the right call given the publish/tag/release sequence.
- `publish-mcp` now asserts `needs.verify.result == 'success'` explicitly rather than inheriting it through `release`, preventing a skipped/failed verify from opening the gate.

No blocking issues found. The change is safe to merge.

.github/workflows/release.yml
```diff
@@ -65,8 +90,79 @@ jobs:
...
+  # A separate job rather than more steps inside `release`: `needs` makes the
+  # dependency the thing that blocks, so publishing cannot begin until the
+  # suite has passed, and a reader can see that from the job graph.
+  # Unconditional on purpose. Gating this on `root_changed` would have left the
+  # MCP package publishing through an ungated path: `publish-mcp` runs when
+  # `needs.release.result` is 'skipped', and a skipped `release` is exactly what
+  # an MCP-only push produces -- so the one job this file exists to gate would
+  # have sailed past it. Worse, a *failed* verify also renders `release` as
+  # 'skipped', so the gate would have opened precisely when it should shut.
```
Tara-ag, line 110: 💬 SUGGESTION: Consider adding `timeout-minutes: 30` to the `verify` job. It runs lint, type-check, unit tests, browser install, and Playwright integration tests unconditionally; a hung test or browser install could otherwise consume runner minutes indefinitely.

.github/workflows/release.yml
```diff
@@ -65,8 +90,79 @@ jobs:
...
+
+      - name: Lint (prettier, eslint, event-casing)
+        run: pnpm lint
+
+      - name: Type-check
+        run: pnpm check
+
+      # The step that would have caught it. scripts/wc-parity/prop-parity.test.ts
+      # is a vitest test, and it had been failing on `release` since the merge.
+      - name: Unit tests (vitest)
```
Tara-ag, line 169: 💡 MINOR: `if: always()` here uploads the Playwright report even when the preceding browser-install or integration-test step failed or was skipped, which can produce empty/misleading artifacts on red runs. Prefer `if: failure() || success()` and add `if-no-files-found: ignore` so the artifact is only uploaded when it actually exists.

.github/workflows/release.yml
```diff
@@ -65,8 +90,79 @@ jobs:
...
+  # an MCP-only push produces -- so the one job this file exists to gate would
+  # have sailed past it. Worse, a *failed* verify also renders `release` as
+  # 'skipped', so the gate would have opened precisely when it should shut.
+  # Running it on every push costs one suite on a push that releases nothing,
+  # and removes both holes.
+  verify:
     needs: detect-changes
+    runs-on: ubuntu-latest
+
+    permissions:
```
Tara-ag, line 128: ⚠️ MAJOR: The new `verify` job installs pnpm and runs `pnpm install` on every push to `release`, but neither `actions/setup-node` nor `pnpm/action-setup` is configured with a dependency cache. Because this job now runs unconditionally on every release push, uncached installs add wall-clock time and registry-flakiness risk. Add `cache: 'pnpm'` to `setup-node` (and `run_install: false` to `pnpm/action-setup`) so the lockfile is cached between runs.

Tara-ag (commented):
The workflow change correctly addresses the two release faults: it serialises release runs and gates publishing on a `verify` job that runs the full quality suite. I have one MAJOR concern about missing dependency caching in the new unconditional `verify` job, plus a MINOR artifact-hygiene note and a SUGGESTION for job timeouts. None of these are blocking correctness issues, so the PR is safe to merge once the team weighs the caching feedback.

.github/workflows/release.yml
Tara-ag, line 169: 💡 MINOR: The Upload Playwright report step uses `if: always()`, which will execute even if the preceding Install Playwright browsers or Integration tests steps failed or were skipped. A missing or partial `playwright-report/` directory can produce empty or misleading artifacts on a red run.

Suggestion: Change the condition and add an existence guard:
```yaml
      - name: Upload Playwright report
        if: failure() || success()
        uses: actions/upload-artifact@v4
        with:
          name: release-verify-playwright-report-${{ github.run_id }}-${{ github.run_attempt }}
          path: playwright-report/
          retention-days: 14
          if-no-files-found: ignore
```

.github/workflows/release.yml
Tara-ag, line 110: 💬 SUGGESTION: The verify job runs lint, type-check, unit tests, Playwright browser install, and integration tests unconditionally on every release push. Without a job-level timeout, a hung Playwright test or browser install can consume runner minutes indefinitely.

Suggestion: Add `timeout-minutes: 30` to the verify job (and consider the same for release).

.github/workflows/release.yml
Tara-ag, line 128: ⚠️ MAJOR: The new verify job installs pnpm and runs pnpm install on every push to release, but neither actions/setup-node nor pnpm/action-setup is configured with a dependency cache. For a workflow that now runs unconditionally on every release push, this adds minutes of install latency and unnecessary registry load.

Suggestion: Add cache to the setup steps:
```yaml
      - name: Setup Node.js
        uses: actions/setup-node@v4
        with:
          node-version: '22'
          cache: 'pnpm'
      - name: Install pnpm
        uses: pnpm/action-setup@v4
        with:
          version: 10.18.3
          run_install: false
```

murdore:
## The three post-approval findings are applied

All three were in the `verify` job; all three are now in the file.

- **No dependency cache.** `pnpm/action-setup` now runs **before** `setup-node`, with `run_install: false`, and `setup-node` carries `cache: 'pnpm'`. The order is load-bearing: `cache: pnpm` resolves the store path by invoking pnpm, so pnpm must exist first, and letting action-setup install before the cache is configured would make the cache do nothing on every run.
- **No job timeout.** `timeout-minutes: 30`. Every publish waits on this job now, so a hung step would hold the release open instead of failing it.
- **`if: always()` upload with no existence guard.** Added `if-no-files-found: ignore`. `always()` also fires when the job was cancelled or failed before Playwright ran, and upload-artifact's default is to warn and still create the artifact — a report entry that looks like evidence and contains nothing.

Scoped to `verify` deliberately: the same setup block appears in `release` and `publish-mcp`, and those were not part of the finding, so they are untouched.

The three findings that earned the earlier approval — the MCP-bypass gap, the `--run` separator, and the artifact-name collision — remain fixed. Rebased onto `release` at 3.4.0, single commit.

Tara-ag (commented):
Review in progress.

.github/workflows/release.yml
```diff
@@ -65,8 +90,93 @@ jobs:
             echo "No MCP package changes detected"
           fi
 
-  release:
+  # The gate that did not exist when 3.2.0 published.
+  #
+  # Every quality workflow in this repository -- ci, visual, yama-review -- is
+  # `pull_request`-only. A push to `release` ran Pages and this file, and this
+  # file ran `pnpm lint` and then published. So two PRs that were each green
+  # against their own base merged fourteen seconds apart, the combination was
```
Tara-ag, line 98: ⚠️ MAJOR: The `verify` job explicitly sets `permissions: contents: read`, which strips the default `actions: write` permission. `actions/setup-node` with `cache: 'pnpm'` writes to the Actions cache, and `actions/upload-artifact@v4` uploads the Playwright report; both require `actions: write`. Without it, the artifact upload step will fail on every release run, blocking the pipeline.

Suggested fix: add `actions: write` to the job permissions:
```yaml
permissions:
  contents: read
  actions: write
```

Tara-ag (changes requested):
The PR adds a necessary concurrency gate and a verify job to prevent untested releases, which is the right direction. However, the new `verify` job declares overly restrictive permissions (`contents: read` only) that will break `actions/setup-node` cache writes and `actions/upload-artifact` uploads, blocking the release pipeline. One actionable MAJOR fix is required.

.github/workflows/release.yml
murdore: Not applying this one — the finding is refuted on both halves, and `actions: write` on a workflow that publishes to npm is a real privilege increase (it permits cancelling and re-running workflows and deleting artifacts), so it should not go in without a reason.

**Artifact upload does not require `actions: write`.** Measured on this exact commit:

| | |
|---|---|
| workflow | `.github/workflows/ci.yml`, which declares `permissions: contents: read` |
| run | 33954275809, branch `fix/release-test-gate`, commit `d49920fff` — the commit under review |
| conclusion | **success** |
| artifact | `playwright-report`, **351,109,070 bytes**, not expired |

`actions/upload-artifact@v4` uploaded 335 MB under the identical permission line, on the same commit, today. `visual.yml` does the same. Artifacts authenticate with the runner's own token, not with `GITHUB_TOKEN`'s `permissions:` scopes.

**Cache save does not require it either, and cannot block the pipeline.** Per `actions/cache`'s own documentation, cache access is governed by the run's cache credential rather than a `permissions:` block — which is why a fork PR can restore but not save. And when the cache token is read-only, *the save step does not fail the job*: it emits one warning (`Failed to save: ... cache write denied`) and the step still succeeds. So even if the premise held, the consequence — "the artifact upload step will fail on every release run, blocking the pipeline" — would not follow.

One difference worth naming: `ci.yml` sets `contents: read` at workflow level and `verify` sets it at job level. Both replace the default token scopes for the job identically, so the comparison holds.

Happy to be shown otherwise with a failing run, but on the evidence available `contents: read` is correct and stays.

.github/workflows/release.yml
```diff
@@ -65,8 +90,93 @@ jobs:
...
+        uses: pnpm/action-setup@v4
+        with:
+          version: 10.18.3
+          run_install: false
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+          cache: 'pnpm'
```
Tara-ag, line 150: 💡 MINOR: Node.js version `22` is repeated across `verify`, `release`, and `publish-mcp`. Move it to the workflow `env` block to keep runtime versions consistent.

```yaml
env:
  NODE_VERSION: '22'
```

Then use `node-version: ${{ env.NODE_VERSION }}` in each `setup-node` step.

.github/workflows/release.yml
```diff
@@ -65,8 +90,93 @@ jobs:
...
+      # pnpm before node, deliberately: setup-node's `cache: pnpm` resolves the
+      # store path by running pnpm, so pnpm has to exist first. `run_install: false`
+      # stops action-setup installing before that cache is configured, which would
+      # make the cache do nothing on every run.
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 10.18.3
+          run_install: false
+
```
Tara-ag, line 145: 💡 MINOR: pnpm version `10.18.3` is hard-coded in both the `verify` and `release` jobs. Centralize it in the workflow `env` block so bumps happen in one place and cannot drift between jobs.

```yaml
env:
  NPM_CLI_VERSION: '11.19.0'
  PNPM_VERSION: '10.18.3'
```

Then reference it as `version: ${{ env.PNPM_VERSION }}`.

.github/workflows/release.yml
```diff
@@ -65,8 +90,93 @@ jobs:
...
+      - name: Lint (prettier, eslint, event-casing)
+        run: pnpm lint
+
+      - name: Type-check
+        run: pnpm check
+
+      # The step that would have caught it. scripts/wc-parity/prop-parity.test.ts
+      # is a vitest test, and it had been failing on `release` since the merge.
+      - name: Unit tests (vitest)
```
Tara-ag, line 169: 💬 SUGGESTION: Only the Playwright report is uploaded. Since the PR's purpose is to catch unit-test failures before publish, consider also persisting the Vitest results/logs as an artifact so failures on `release` can be diagnosed without re-running the full suite.

```yaml
- name: Upload unit-test report
  if: always()
  uses: actions/upload-artifact@v4
  with:
    name: release-verify-unit-report-${{ github.run_id }}-${{ github.run_attempt }}
    path: |
      test-results/
      coverage/
    if-no-files-found: ignore
```

Tara-ag (commented):
The workflow change correctly addresses both stated problems: a `verify` job now gates release/publish-mcp, and a concurrency group serializes release runs without allowing cancellation mid-flight. I noted three maintainability/observability items inline; none are blocking. The PR is safe to merge once CI is green.

.github/workflows/release.yml
murdore: Applied in `4d91c80` — `actions: write` is now on the `verify` job.

To be transparent about why, since I argued against it above: the evidence still says neither step needs it. `ci.yml` declares the identical `permissions: contents: read`, and its run on `d49920fff` — the exact commit this review was submitted against — succeeded and uploaded a 351,109,070-byte `playwright-report` artifact. `actions/cache` documents that cache access is governed by the run's own cache credential rather than a `permissions:` block, and that a read-only token makes the save step warn rather than fail the job.

It is applied as a deliberate decision by the repo owner, not because the premise was confirmed. That distinction is recorded in a comment beside the grant, along with the fact that this workflow publishes to npm and can now also cancel runs and delete artifacts — so a later reader can drop the line if that reach is unwanted.

Separately, this push carries a real fix for a failure in the same area. Run 33961677670 published 3.4.1 and then failed at `Push changes` with `GH006: Protected branch update failed for refs/heads/release — 3 of 3 required status checks are expected`. `release` requires checks/build/visual with `enforce_admins` on, so a release commit created seconds earlier can never satisfy them. The tag and GitHub Release steps were skipped, and **3.4.1 is on npm right now with no tag, no `chore(release)` commit and no release notes** (its tarball is byte-identical to 3.4.0 apart from the version string). The push now runs before the publish, so that failure stops the run with the registry untouched instead of stranding an untraceable version. Ordering cannot make the push itself succeed — that needs a branch-protection change.

Tara-ag (commented):

.github/workflows/release.yml
```diff
@@ -65,8 +90,101 @@ jobs:
...
+        run: pnpm run test:unit --run
+
```
Tara-ag, line 171: 💬 SUGGESTION: The verify job runs `pnpm check` but not `pnpm check:wc`, the dedicated type-check for the web-component build. Since this library ships a WC build and the goal is to catch gaps that let 3.2.0 ship broken, adding a `check:wc` step here closes another hole.

.github/workflows/release.yml
```diff
@@ -65,8 +90,101 @@ jobs:
...
+        with:
+          # v4 artifacts are immutable, so a fixed name makes the upload fail on
+          # a re-run -- which is the one time the report is most wanted.
+          name: release-verify-playwright-report-${{ github.run_id }}-${{ github.run_attempt }}
+          path: playwright-report/
+          retention-days: 14
+          # `if: always()` also fires when the job was cancelled or failed before
+          # Playwright ran, and upload-artifact's default is to warn and still create
+          # the artifact -- a report entry that looks like evidence and holds nothing.
+          if-no-files-found: ignore
```
Tara-ag: 💡 MINOR: `if-no-files-found: ignore` will silently swallow a missing `playwright-report/` directory. That makes genuine failures (e.g., Playwright misconfigured or the report path changed) harder to diagnose. Prefer conditioning the upload on test outcome and using `warn` or `error` so an unexpectedly missing report is surfaced.

.github/workflows/release.yml
```diff
@@ -263,6 +381,38 @@ jobs:
...
+      # Ordering it first cannot make that push succeed -- only a settings change
+      # can -- but it makes the failure safe: the run stops before publishing and
+      # the registry stays clean. The reverse failure (pushed, then publish fails)
+      # leaves a version commit for an unpublished version, which the existing
+      # `npm view ... && exit 0` guard below makes safe to resolve by re-running.
+      - name: Push changes (without tags)
+        run: |
+          # Pull latest changes to avoid conflicts
+          git fetch origin release
+
```
Tara-ag, line 414: ⚠️ MAJOR: Push-before-publish ordering is unsafe with concurrency and rebase. Serialising runs prevents the 3.2.0-style git race, but a queued second run can now rebase onto the first run's pushed commit, compute the same version, skip publish because `npm view` sees it already exists, and then proceed to create a tag/GitHub Release for a version whose tarball was published by the *first* run. Worse, if a rerun sees the version on npm, it exits 0 before recreating a missing tag or release, orphaning release metadata. Consider creating the tag and GitHub Release immediately after this push step and before publish, or keeping publish-first and using `git push --atomic origin release <tag>`.

Tara-ag (commented):
Reviewed `.github/workflows/release.yml`. The PR materially improves release safety by gating publish on a full verify job and adding concurrency control. I have one MAJOR concern about the new push-before-publish ordering creating orphaned release metadata, plus two smaller notes on artifact visibility and the missing `check:wc` step. None of these are hard blockers, but the MAJOR item should be addressed or explicitly accepted before merge.

.github/workflows/release.yml
Tara-ag, line 171: 💬 SUGGESTION: Missing `check:wc` in verify job

The verify job runs `pnpm check` but not `pnpm check:wc`, the dedicated type-check for the web-component build. Since this library ships a WC build and the goal is to close gaps that let 3.2.0 ship broken, adding `check:wc` would catch another failure mode.

**Fix:** Add a step `- name: Type-check web components
        run: pnpm check:wc` after the existing type-check step.

.github/workflows/release.yml
Tara-ag: 💡 MINOR: Artifact upload uses `if-no-files-found: ignore` which can hide failures

The Playwright report upload uses `if-no-files-found: ignore`, which silently swallows a missing `playwright-report/` directory and makes genuine misconfigurations (e.g., Playwright output path changed) harder to diagnose.

**Fix:** Condition the upload on test outcome and use `warn` or `error` so an unexpectedly missing report is surfaced.

.github/workflows/release.yml
```diff
@@ -263,6 +381,38 @@ jobs:
             exit 1
           fi
 
+      # BEFORE the publish, deliberately. npm publishes are effectively permanent
+      # -- a version cannot be re-published and unpublishing is restricted -- while
+      # a commit on `release` is revertible. With publish first, anything that stops
+      # this push leaves the registry permanently ahead of the repo, with no tag, no
+      # `chore(release)` commit and no GitHub Release naming what shipped.
```
Tara-ag, line 397: ⚠️ MAJOR: Push-before-publish ordering is unsafe with concurrency and rebase

Serialising runs prevents the 3.2.0-style git race, but a queued second run can rebase onto the first run's pushed commit, compute the same version, skip publish because `npm view` sees it already exists, and then proceed to create a tag/GitHub Release for a version whose tarball was published by the first run. A rerun that sees the version on npm exits 0 before recreating a missing tag or release, orphaning release metadata.

**Fix:** Create the tag and GitHub Release immediately after the push step and before publish, or keep publish-first and use `git push --atomic origin release <tag>`.

Tara-ag (changes requested):
Yama review: 3 findings accepted (1 MAJOR, 1 MINOR, 1 SUGGESTION). See inline comments.

.github/workflows/release.yml
murdore: Half of this is right and is applied in `ed868f1`; half rests on a mechanism that does not hold. Taking them separately.

**Applied — tag before publish.** The order is now push branch → push tag → publish → GitHub Release. This is the same reasoning that moved the branch push: both git writes are revertible, the npm write is not, so with both refs ahead of it anything that blocks either one stops the run with the registry untouched. The GitHub Release stays after the publish because it announces a released version.

**Not adopted — publish-first with `git push --atomic origin release <tag>`.** Atomicity holds between the two git refs; it says nothing about ordering against npm. The failure that stranded 3.4.1 was the branch push being **rejected** by branch protection (`GH006 ... 3 of 3 required status checks are expected`), and an atomic push would have extended that rejection to the tag as well — still published, now with neither ref written. That is strictly worse than what is here.

**Incorrect — "a rerun that sees the version on npm exits 0 before recreating a missing tag or release, orphaning release metadata."** That `exit 0` is inside a `run:` step. It ends the step successfully; the job continues to the next step. So the tag and GitHub Release steps still execute on a rerun, which is precisely the repair path for a missing tag rather than a way of orphaning one. Worth correcting because the conclusion inverts the actual behaviour.

**On the queued-second-run scenario:** `concurrency: group: release-${{ github.ref }}` with `cancel-in-progress: false` serialises runs, so the second starts after the first finishes. If the first completed, its tag exists and the second computes the next version. If the first failed before tagging, the second computing the same version and creating the tag is the recovery, and it tags a commit whose `package.json` carries that exact version — not orphaned metadata.

Also applied from this review: `check:wc` added to `verify` (the root `check` does not cover the web-component build this library ships), and the Playwright upload moved from `if-no-files-found: ignore` to `warn`. Worth noting `ignore` was itself added earlier in this PR at this reviewer's request; `warn` satisfies both readings — no empty artifact posing as evidence, and the absence is still stated in the log.

Tara-ag (commented):
Review in progress.

.github/workflows/release.yml
```diff
@@ -65,8 +90,110 @@ jobs:
...
+    runs-on: ubuntu-latest
+    # Every publish now waits on this job, so a step that hangs would hold the
+    # release open indefinitely instead of failing it. Playwright install plus
+    # both suites finish in well under half of this.
+    timeout-minutes: 30
+
+    # `contents: read` is all this job's own steps need. `actions: write` is here
+    # by review decision, not by demonstrated need: the artifact upload and the
+    # pnpm store cache were both observed working under `contents: read` alone
+    # (ci.yml runs the identical pair and uploaded a 335 MB report on d49920fff).
```
Tara-ag, line 124: 💡 MINOR: The `verify` job grants `actions: write` without a concrete need. The inline comment acknowledges this is "by review decision, not by demonstrated need" and even notes it lets this publishing workflow cancel runs and delete artifacts. A test-only job should run with least privilege; drop this permission unless a specific step requires it.

Tara-ag (approved):
This PR correctly addresses the two release-process faults it identifies: it adds a `verify` job that gates publishing on lint/type-check/unit/integration, and it serialises release runs with a concurrency group that keeps `cancel-in-progress: false`. The reordering of git writes before npm publish is also the safer failure mode.

One minor note: the `verify` job grants `actions: write` without a demonstrated need (left inline). That is worth tightening, but it does not block the fix.

Approving.

PR #517, opened 2026-09-04
test(parity,wc): close every open review finding across the parity harness and wrappers

murdore:
One PR for every finding still outstanding after 3.2.3, rather than four. **Supersedes #509** — it carries that harness, because the findings cannot be fixed without the code they are about.

## 1. The parity harness asserted almost nothing (#509)

Three of its assertions were decorative:

| Assertion | Problem |
|---|---|
| `consoleErrors` | collected, written to `results.json`, attached — **never asserted** |
| `expect(Array.isArray(bothBlank)).toBe(true)` | true of every array; row count could fall to zero without failing |
| only `suiOnlyBroken` gated | while the description claimed *"asymmetric failures 0 in either direction"* |

`consoleErrors` is now asserted **first**, deliberately: `<svelte:boundary>` catches a render failure and turns it into a visible cell, which is the result this harness wants, but an error thrown from a timer, a promise or an event handler escapes it. A component that throws asynchronously can still leave a plausible DOM behind, so the rendering checks below would pass while the page was on fire.

The one-directional gate mattered because the claim rested on a number nobody could reproduce from the test. Both directions are asserted now.

**Measured after the change:**

```
components compared : 68
sui rendered        : 68        poly rendered      : 68
sui-only broken     : 0         poly-only broken   : 0
both blank          : 0         console errors     : 0
rows differing on computed style : 15 of 68
```

The description's claim is now something the suite enforces rather than something it was trusted on.

`Cell.svelte` also loses its `name` prop — it built a `data-pw` no selector used.

**On `$bindable`:** the finding is correct that `{...props}` does not establish a binding, and that is the right shape here. This harness asks whether both libraries render the same fixture — a question about first paint. Nothing types into a component or reads a value back, so a two-way binding would have nothing to carry. Documented in `Cell.svelte` rather than left to be rediscovered, and Input, Checkbox, Toggle and Select render on both sides, which the suite now asserts.

## 2. Both #512 findings hold up — and neither needs the code changed

**Native handler shadowing.** Declaring `onfocus`/`oninput`/`onclick` really does shadow the native `GlobalEventHandlers` accessor. What matters is whether that costs a consumer the DOM, and it does not — the declaration replaces one property, not the event system. A test now proves `addEventListener('click')` and the declared prop **both fire on the same element**, because *"use addEventListener instead"* is the entire mitigation and was until now an assertion rather than a fact.

**Attribute collision.** `onclick` and `onClick` do both claim the `onclick` attribute — confirmed, `observedAttributes` contains it twice, and the test asserts that so the premise cannot quietly stop being true. The finding calls this reliance on undocumented key-enumeration order. It is, *for the attribute path*, which is inert because a function cannot be written as an HTML attribute. Property assignment is per-key and unaffected, which is why precedence holds in the existing tests. Pinned anyway: the order is stable but incidental, and the generator appends corrected spellings after legacy ones, so a reordering would silently change which key the attribute resolves to and nothing else would notice.

## 3. One correction to DESIGN_PRINCIPLES (#516)

The pressed-state section claimed a missing declaration is *"the one gap in the contract a consumer cannot work around"*. True through the custom elements, where the shadow boundary ends it. Not true importing the Svelte components, where styles are scoped rather than encapsulated and a consumer **can** write the missing rule — at the cost of reaching past the token API into class names that are not part of it. Both cases are now stated, since the difference decides whether a consumer has no way out or only an unsupported one.

## Deliberately not here

The release workflow's version bump reads `git log -1` — the tip commit only — so two merges each with a `feat:` tip compute the same version. That is why 3.2.0 was computed twice. It belongs with #514's release-pipeline work, and putting it here would collide with that PR over the same file.

## Verification

**443 integration · 407 unit · lint 0 errors · check 978 files 0 errors**

## Summary by CodeRabbit

- **New Features**
  - Added a visual comparison page for reviewing shared SUI and polymorph UI components side by side.
  - Displays component render results, styling differences, and isolated rendering errors for easier comparison.

- **Tests**
  - Added automated visual-parity checks covering shared components and expected styling differences.
  - Added browser tests validating custom-element event handling and event-name casing behavior.

- **Documentation**
  - Clarified theming limitations and updated formatting in the design principles documentation.

### DIFF
--- a/tests/wc-event-casing-parity.spec.ts
+++ b/tests/wc-event-casing-parity.spec.ts
@@ -261,3 +261,85 @@ test.describe('a declared prop whose name is also a native handler', () => {
     ]);
   });
 });
+
+test.describe('declaring a native handler name does not take the DOM path with it', () => {
+  test('addEventListener still works on an element that declares the same name', async ({
+    page
+  }) => {
+    await loadBundle(page);
+
+    // Review on #512 held that declaring `onclick`/`onfocus`/`oninput` shadows
+    // the native GlobalEventHandlers accessor, and so changes web-component
+    // behaviour. The shadowing is real and measured in the test above. What
+    // matters for a consumer is whether it costs them the DOM, and it does not:
+    // the declaration replaces one property, not the event system. This asserts
+    // the escape hatch actually exists rather than assuming it does, because
+    // "use addEventListener instead" is the whole mitigation.
+    const seen = await page.evaluate(async () => {
+      const order: string[] = [];
+
+      const element = document.createElement('sui-toggle');
+      element.addEventListener('click', () => order.push('addEventListener'));
+      Reflect.set(element, 'onClick', () => order.push('declared-prop'));
+      document.body.append(element);
+      await new Promise((resolve) => requestAnimationFrame(resolve));
+      element.shadowRoot?.querySelector('input')?.click();
+
+      return order;
+    });
+
+    // Both run. The listener is not displaced by the prop, and the prop is not
+    // displaced by the listener.
+    expect(seen, 'declaring the prop cost the element its DOM event').toContain('addEventListener');
+    expect(seen).toContain('declared-prop');
+  });
+
+  test('the shared observed attribute resolves to the legacy spelling', async ({ page }) => {
+    await loadBundle(page);
+
+    // The other half of that review: `onclick` and `onClick` both lowercase to
+    // the `onclick` attribute, so Svelte's `$$g_p` returns whichever key
+    // `Object.keys` yields first, and the finding called that undocumented
+    // key-enumeration behaviour to rely on. It is -- for the attribute path,
+    // which is inert here because a function cannot be written as an HTML
+    // attribute. Property assignment is unaffected: each declared key gets its
+    // own accessor, which is why precedence holds in the tests above.
+    //
+    // Pinned by driving the attribute rather than by counting it. An earlier
+    // version asserted only that `onclick` appears twice in observedAttributes,
+    // while the comment claimed it would catch a reordering -- which it would
+    // not have, since both entries are the same string. Review caught the gap.
+    // Setting the attribute and reading both properties back shows which key
+    // actually won, so a reordering of the declarations fails here.
+    const resolved = await page.evaluate(async () => {
+      const constructor = customElements.get('sui-toggle');
+      const observed =
+        typeof constructor === 'function' ? Reflect.get(constructor, 'observedAttributes') : [];
+      const list: string[] = Array.isArray(observed) ? observed : [];
+
+      const element = document.createElement('sui-toggle');
+      document.body.append(element);
+      await new Promise((resolve) => requestAnimationFrame(resolve));
+      // `type: 'Object'` runs the attribute through JSON.parse, so a parseable
+      // marker lands on whichever prop the attribute resolved to.
+      element.setAttribute('onclick', '{"marker":true}');
+      await new Promise((resolve) => requestAnimationFrame(resolve));
+
+      return {
+        onclickCount: list.filter((name) => name === 'onclick').length,
+        legacyGotIt: JSON.stringify(Reflect.get(element, 'onclick')),
+        correctedGotIt: JSON.stringify(Reflect.get(element, 'onClick'))
+      };
+    });
+
+    // Both declarations really do claim the same attribute -- the premise the
+    // finding rests on, confirmed rather than taken on trust.
+    expect(resolved.onclickCount, 'the two spellings no longer share an attribute').toBe(2);
+
+    // And the legacy key is the one it resolves to, because it is declared
+    // first. This is the assertion a reordering would break.
+    expect(resolved.legacyGotIt, 'the attribute stopped resolving to the legacy spelling').toBe(
+      '{"marker":true}'
+    );
+  });
+});
--- a/DESIGN_PRINCIPLES.md
+++ b/DESIGN_PRINCIPLES.md
@@ -20,9 +20,17 @@ as a bug, the same as an unhandled `null`.
 **A token can only reach a state the component actually paints.** The cascade cannot add a
 declaration that does not exist: a `.toggle-button:hover` rule with no `:active` beside it
 leaves nothing for `--theme-switcher-bg-pressed` to land in, so no consumer can supply a
-pressed state however carefully they arrange their own stylesheet. It is the one gap in the
-contract a consumer cannot work around, and it is invisible from the source — the component
-reads as fully tokenised, because every rule it *does* have is.
+pressed state through the token contract however carefully they arrange their own
+stylesheet. It is invisible from the source, too — the component reads as fully tokenised,
+because every rule it _does_ have is.
+
+How far that goes depends on which build a consumer is on, and the distinction is worth
+keeping straight. Through the custom elements the shadow boundary is the end of it: an
+outside stylesheet cannot reach a rule that does not exist, and cannot add one. Importing
+the Svelte components directly, styles are scoped rather than encapsulated, so a consumer
+can still write the missing rule themselves — at the cost of reaching past the token API
+into class names that are not part of it. Neither case makes the missing declaration
+acceptable; the first leaves no way out at all, and the second only an unsupported one.
 
 Measured on an embedded Shopify Admin surface (42 routes, 573 elements, five interaction
 states each, states forced through CDP rather than inferred): every divergence from the host
@@ -56,7 +64,7 @@ hook for Playwright.
   `onSelectionChange`. This matches the wider Svelte-component ecosystem's convention for
   invented callback props.
 
-This is deliberately *not* polymorph's rule (lowercase for everything, including
+This is deliberately _not_ polymorph's rule (lowercase for everything, including
 synthesized events). That's a real, coherent choice, but it trades away the native/synthetic
 distinction that Svelte 5 itself draws, and it isn't obviously better — just different.
 What actually needed fixing here wasn't the rule, it was that this codebase had never
@@ -81,6 +89,6 @@ otherwise it stays a plain element.
 ## 5. Documentation is built for machines, not just humans
 
 This library ships an MCP server (`mcp/`) exposing `list_components`-style tools over
-`docs/*.md` — components are increasingly composed and themed *through* AI assistants, not
+`docs/*.md` — components are increasingly composed and themed _through_ AI assistants, not
 just read about by humans. `docs/_index.json` is the source of truth; keep it current when
 a component is added, the same way you'd keep an export current.


Tara-ag (commented):
Review in progress — posting findings file by file.

src/routes/parity/+page.svelte
```diff
@@ -0,0 +1,89 @@
...
+
+  // This repo bans both type assertions and type predicates, so the narrowing
+  // has to happen at the type boundary instead of at the use site.
+  // `Object.entries` on a plain object yields `any` values, which assign to the
+  // declared return type without an `as` — and the runtime filter below is then
+  // an ordinary boolean check rather than a predicate.
+  const toBag = (module: object): Record<string, Renderable> =>
+    Object.fromEntries(Object.entries(module));
+
+  const isRenderable = (value: Renderable): boolean => typeof value === 'function';
```
Tara-ag: ⚠️ MAJOR: `toBag` declares a return type of `Record<string, Renderable>` but returns `Object.fromEntries(Object.entries(module))`, which is effectively `Record<string, unknown>`. This is a silent cast around the repo-banned type assertion. Because `pair.Sui` is already typed as `Renderable`, the `isRenderable` predicate on line 21 is typed `(value: Renderable) => boolean` and cannot actually narrow anything. A non-function export would satisfy the prop type and be passed into `<Cell component={...}>`, where Svelte would try to instantiate it.

Make `isRenderable` a real type guard and filter inside `toBag`:

```ts
const isRenderable = (value: unknown): value is Renderable => typeof value === 'function';

const toBag = (module: object): Record<string, Renderable> =>
  Object.fromEntries(Object.entries(module).filter(([, v]) => isRenderable(v)));
```

Then drop the redundant `.filter((pair) => isRenderable(pair.Sui) && isRenderable(pair.Poly))` on line 34.

tests/visual-parity.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  const styleDiffs: Array<{
+    component: string;
+    diff: string[];
+    suiStyle: Record<string, string> | null;
+    polyStyle: Record<string, string> | null;
+  }> = [];
+
+  for (let i = 0; i < count; i += 1) {
+    const row = rows.nth(i);
```
Tara-ag: 💡 MINOR: `styleDiffs` is collected and written to `results.json`, but never asserted. The PR description notes 15 of 68 components differ on computed style and treats that as acceptable, but the test does not enforce any bound. A future regression that made 67 components differ would still pass.

Either assert a ceiling (e.g. `expect(styleDiffs.filter(d => d.diff.length).length).toBeLessThanOrEqual(15)`) with a comment explaining the baseline, or add an explicit comment that the array is diagnostic-only and intentionally unasserted.

tests/wc-event-casing-parity.spec.ts
```diff
@@ -261,3 +261,64 @@ test.describe('a declared prop whose name is also a native handler', () => {
...
+    // the `onclick` attribute, so Svelte's `$$g_p` returns whichever key
+    // `Object.keys` yields first, and the finding called that undocumented
+    // key-enumeration behaviour to rely on. It is -- for the attribute path,
+    // which is inert here because a function cannot be written as an HTML
+    // attribute. Property assignment is unaffected: each declared key gets its
+    // own accessor, which is why precedence holds in the tests above.
+    //
+    // Pinned anyway. The order is stable but incidental, and the generator
+    // appends corrected spellings after the legacy ones. If a future edit
+    // reorders them, the attribute would start resolving to the other key, and
```
Tara-ag: 💡 MINOR: This test asserts that `'onclick'` appears twice in `observedAttributes`, confirming the attribute collision. However, the comment warns that the generator appends corrected spellings after legacy ones and that a reordering would change which key the attribute resolves to. The count assertion does not catch that reordering.

Assert the order, not just the count:

```ts
const onclickIndices = list
  .map((name, i) => (name === 'onclick' ? i : -1))
  .filter((i) => i !== -1);
expect(onclickIndices, 'legacy onclick should precede corrected onClick in observedAttributes')
  .toEqual([legacyIndex, correctedIndex]);
```

If the exact indices are unstable, derive the expected order from the wrapper source so the test actually guards the hazard it describes.

tests/visual-parity.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+  const consoleErrors: string[] = [];
+  page.on('pageerror', (e) => consoleErrors.push(e.message));
+
+  await page.goto('/parity');
+  await page.waitForSelector('[data-pw="parity-row"]', { timeout: 20_000 });
+  // Components with mount-time timers or transitions need a beat before capture.
+  await page.waitForTimeout(1500);
+
+  const rows = page.locator('[data-pw="parity-row"]');
+  const count = await rows.count();
```
Tara-ag: ⚠️ MAJOR: The `pageerror` listener registered here is never removed, and the test writes screenshots/results into a fixed `parity-report/` directory. Listeners accumulate across tests in the same worker, and parallel workers can overwrite each other's output. Writing to the project root also leaves stale files that can mask regressions in later runs.

Make the test hermetic: write under `testInfo.outputDir` and detach the listener on finish:

```ts
const onError = (e: Error) => consoleErrors.push(e.message);
page.on('pageerror', onError);
testInfo.onTestFinished(() => page.off('pageerror', onError));
```

Use `testInfo.outputDir` (or `test.use({ outputDir: 'parity-report' })`) for screenshots and `results.json` so workers cannot collide.

Tara-ag (commented):
Review complete. Four findings: one MAJOR type-safety issue in the parity harness, one MAJOR test-hermeticity issue, and two MINOR test-design/clarity issues. None are blocking security or public-API breaks.

Tara-ag:

The PR improves the visual parity harness and adds web-component event-casing regression tests. No public API, accessibility, SSR, or security regressions were found. The accepted findings are all quality issues: one MAJOR non-hermetic test setup (leaked listener and fixed output directory) and two MINOR items (misleading type contract in toBag, incomplete Playwright attachment). These should be addressed before merge.

- ⚠️ **MAJOR**: Non-hermetic visual parity test writes to project root and leaks pageerror listener — `tests/visual-parity.spec.ts:8`
  The test registers a pageerror listener and writes screenshots/results.json into a fixed parity-report/ directory under the project root, but never removes the listener or cleans the directory. Leftover files from prior runs can mask failures or pollute CI artifacts, and the unremoved listener accumulates across tests in the same worker.
- 💡 **MINOR**: Parity page relies on implicit any widening to type module exports as components — `src/routes/parity/+page.svelte:11`
  `toBag` declares a return type of `Record<string, Renderable>` but simply returns `Object.fromEntries(Object.entries(module))`. Because the parameter is typed `object`, `Object.entries` yields `[string, any][]`, so TypeScript silently accepts the annotation without proving values are Svelte components. The runtime guard catches this, but the compile-time contract is misleading.
- 💡 **MINOR**: Attached parity-results omits styleDiffs and consoleErrors written to results.json — `tests/visual-parity.spec.ts:145`
  The file artifact written to `parity-report/results.json` contains `results`, `styleDiffs`, and `consoleErrors`, but the Playwright attachment `parity-results` only contains `results`. Anyone downloading the CI attachment misses the style-diff and console-error context that the on-disk file preserves.

tests/visual-parity.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+  expect(consoleErrors, 'an error escaped <svelte:boundary> during the parity run').toEqual([]);
+
+  // The assertion that matters: SUI must not fail where polymorph succeeds.
+  expect(
+    suiOnlyBroken.map((r) => `${r.component}: ${r.suiError || 'rendered nothing'}`),
+    'SUI failed to render a component that polymorph rendered from the same fixture'
+  ).toEqual([]);
+
+  // And the other direction, which was described as holding but never checked.
+  // The PR that added this harness reported "asymmetric failures 0 in either
```
Tara-ag: 💡 MINOR: The `parity-results` attachment only contains `results`, but `parity-report/results.json` also includes `styleDiffs` and `consoleErrors`. Attach the same combined object so CI downloads preserve the full context written to disk.

tests/visual-parity.spec.ts
```diff
@@ -0,0 +1,197 @@
+import { expect, test } from '@playwright/test';
+import { mkdirSync, writeFileSync } from 'node:fs';
+
+/**
+ * Captures the SUI and polymorph rendering of every shared component, side by
+ * side, from one fixture per component.
+ *
+ * The comparison this produces is structural rather than pixel-exact, and
```
Tara-ag: ⚠️ MAJOR: Non-hermetic test setup: `page.on('pageerror', ...)` is registered but never removed, and screenshots/results are written to a fixed `parity-report/` directory under the project root without cleanup. Leftover files can mask regressions or pollute CI artifacts, and the leaked listener accumulates across tests in the same worker. Write artifacts to `testInfo.outputDir` (Playwright cleans it per test) and remove the listener with `page.off('pageerror', handler)` before exit.

src/routes/parity/+page.svelte
```diff
@@ -0,0 +1,103 @@
...
+  import type { Component } from 'svelte';
+  import * as sui from '$lib/index.js';
+  import * as poly from 'polymorph-ui-components';
+  import { FIXTURES, NON_COMPONENTS } from './fixtures.js';
+  import Cell from './Cell.svelte';
+
+  type Renderable = Component<Record<string, unknown>>;
+
+  // Review called this return type a silent cast, and that is exactly what it
+  // is: a barrel exports components, constants and helpers alike, so claiming
```
Tara-ag: 💡 MINOR: `toBag` declares `Record<string, Renderable>` but returns `Object.fromEntries(Object.entries(module))`. Because `module: object` makes `Object.entries` yield `[string, any][]`, TypeScript silently accepts the annotation; the runtime guard saves it, but the compile-time contract is misleading. Consider returning `Record<string, unknown>` from `toBag` and doing the component narrowing explicitly where pairs are built, or document why the widening is unavoidable.

Tara-ag (changes requested):

The PR materially strengthens the parity harness and pins web-component event-casing behavior with focused tests. All findings are test-harness quality issues; no public API, accessibility, SSR, or security regressions were identified.

### Findings
- **⚠️ MAJOR** `tests/visual-parity.spec.ts:8` — Non-hermetic test writes to a fixed `parity-report/` directory and leaks the `pageerror` listener. Clean up artifacts and unregister the listener.
- **💡 MINOR** `src/routes/parity/+page.svelte:11` — `toBag` silently widens module exports to `Record<string, Renderable>` via `Object.entries` on `object`. Localize/document the cast or narrow explicitly.
- **💡 MINOR** `tests/visual-parity.spec.ts:145` — `parity-results` attachment omits `styleDiffs` and `consoleErrors` that are written to `results.json`.

Please address the MAJOR issue before merge; the MINOR items are recommended.

murdore:
## All open findings addressed — by removing what they were about

Every finding still outstanding on this PR was in the polymorph parity harness:

| finding | disposition |
|---|---|
| MAJOR — leaked `pageerror` listener + fixed `parity-report/` output dir (`tests/visual-parity.spec.ts`) | **removed with the file** |
| MINOR (was MAJOR) — `toBag` widening in `src/routes/parity/+page.svelte` | **removed with the file** |
| MINOR — `parity-results` attachment omits `styleDiffs`/`consoleErrors` | **removed with the file** |
| MINOR — `styleDiffs` collected but unasserted | **removed with the file** (had already been fixed to assert the exact 15-component list) |
| MINOR — `onclick` `observedAttributes` count cannot prove ordering | **fixed** — the test now sets the attribute at runtime and reads both properties back, so which key it resolves to is asserted, not described |

`polymorph-ui-components@0.7.0` was a devDependency added only to render the two libraries side by side. It is not wanted, so the dependency, `src/routes/parity/*`, `tests/visual-parity.spec.ts` and the `parity-report/` ignore rule are all gone. Lockfile regenerated: 0 hits. #509, which carried the same harness, is closed and its branch restored to its pre-rebase tip.

The Pill and Toggle fixes that were also on this branch shipped in 3.4.0 via #518, so they have dropped out of the diff.

**What is left is two files**: `tests/wc-event-casing-parity.spec.ts` and the `DESIGN_PRINCIPLES.md` encapsulation correction. Rebased onto `release` at 3.4.0, single commit, message rewritten to describe only that.

Tara-ag (approved):
Approved. The remaining diff is limited to two files: a documentation correction in DESIGN_PRINCIPLES.md and additional web-component event-casing regression tests. No public API, accessibility, SSR, security, or reactivity issues were found. The previously reported parity-harness findings have been removed along with the harness itself, as explained in the commit message and PR comment.

PR #519, opened 2026-09-05
fix(migration): ship a runnable codemod, warn on deprecated spellings, and close the audit gaps

murdore:
One commit closing every gap the completion audit found still open in this repository, rebased onto `release` (3.4.0 + #517 + #514). #513 already merged as 3.3.2; #514 merged on its own, so its release-workflow changes are no longer part of this diff.

## Consumer migration — did not work as shipped

| Defect | Fix | Proof |
|---|---|---|
| `npx sui-codemod` exited 1 on the real 3.3.1 tarball: the bin pointed at a `.ts` file and Node 24 refuses type-stripping under `node_modules` | Codemod compiled to `dist-codemod/` on `build`/`prepublishOnly` (`rewriteRelativeImportExtensions`, output dir cleaned first); bin → `dist-codemod/bin.js`; a launcher resolves `svelte/compiler` first and fails with one line instead of a module stack | `npm pack` → extracted under a consumer's `node_modules` → `--help` exit 0 via the npm bin symlink; without `svelte`: one-line message, exit 1 |
| The codemod carried a second, cross-library direction (`--reverse`, a two-way prop map, import-specifier rewriting) that targeted spellings 4.0.0 removes, and the repo referenced the other library throughout | The codemod now has one job: move a consumer of **this** package off the 142 deprecated event-prop spellings (`legacy-pairs.ts`, derived from the baseline via `deriveEventName`, drift pinned by test) and report `children` assignments on the three custom elements that stop declaring it. The other direction, the parity route, its Playwright spec and the devDependency are deleted. `git grep -i polymorph` outside the generated changelog: **0 hits** | packed CLI dry run on `<Toggle onclick>`, `<Input onBlur>`, `<ListItem oncenterTextClick>` + a `.children =` in a `.ts` file: 3 renames, 1 warning, exit 0; idempotent on a second run |
| The "never rewrite onto a deprecated prop" test passed vacuously because no prop carried `@deprecated` | `legacy-pairs.test.ts` checks every legacy spelling **is** tagged and every target is **not** | 142 per-pair transform tests + the tag checks, all green |

## Deprecation — a helper nobody called

- All 140 `$derived(corrected ?? legacy)` aliases (56 components) now go through `resolveDeprecatedProp`, which warns once per prop in dev when only the legacy spelling is supplied. Generated, not hand-edited (`alias-event-props.ts --wire-warnings`).
- All 142 legacy spellings carry `@deprecated Use \`x\` instead; both work until 4.0.0.` in `properties.ts` (generated by `mark-deprecated.ts`, pinned by test).
- Input's legacy bindings drop their `= () => {}` defaults (a default made "legacy supplied" indistinguishable from "nothing supplied", so the warning would have fired for everyone) and the five call sites are guarded instead.
- **Render proof:** `@testing-library/svelte` + jsdom are wired into vitest; `Toggle.svelte.test.ts` renders with `onclick` → one warning naming `onClick`; with `onClick` → silent; the handler fires in both cases.
- `postinstall.mjs` now names both 4.0.0 removals, not just `children`.

## Smaller ledger items

- **Step.svelte** read only the legacy `onclick` (its props live in `Stepper/properties.ts`, so the alias generator never reached it); once `Stepper` handed it the corrected `onClick`, every step click was dropped. It now resolves both spellings like every other component; `Stepper.svelte.test.ts` clicks through both.
- **Img:** a `src` change now removes the root attributes the previous file supplied (they used to linger and blend two icons), restoring the component's own `role`/`aria-label`/`testId` where the file had overwritten them. Regression test drives `sui-img` through two intercepted SVGs.
- **docs/_index.json** gains AttachmentChipRow, HITL, KeyValue, TypewriterText; `src/lib/docs-index.test.ts` keeps the index and `docs/` in lockstep both ways (the MCP server only sees what is indexed). Negative control: removing an entry fails the test.
- **ROADMAP.md** no longer claims those docs need writing.

## Gates (run locally on this exact commit)

- `pnpm lint` clean (142 known / 0 new casing violations), `pnpm check` 0 errors across all four tsconfigs, `pnpm run test:unit --run` **517/517** (33 files), `vite build` exit 0 with the parity route gone.
- Required checks on `release` are now enforced (`checks`, `build`, `visual`; strict up-to-date), so this PR's own CI is the merge gate.

## Deliberately not in this commit

- The 4.0.0 removal itself — a major release, sequenced after these warnings and the working codemod have shipped in a 3.x.
- ostia/helm `CLAUDE.md` drift and Lighthouse's 3.2.0 pin — other repositories.
- ThinkingIndicator's excluded visual route and open-overlay screenshots — separate, documented coverage holes.

Review-tool note: this touches ~190 files, which is over CodeRabbit's 100-file cap (the same cap that left #505 unreviewed). The commit body itemises every change so a human review has a map.


Tara-ag:

Review could not be performed: the time budget expired before any files or the PR diff were examined. No claims about correctness, accessibility, reactivity, SSR safety, tests/docs, or security could be verified. The PR must be re-reviewed with adequate time to inspect each changed file before merging.

murdore:
## Recordings for `f19108c`

All recorded on this exact commit. Every Playwright test in this repo already records video + trace (`playwright.config.ts`), so the three test clips below are the raw test-run recordings; the two demos were recorded from a headed browser and a real CLI session respectively.

### 1. Deprecation warning — real browser, dev build

Page video from a headed Chromium session against `vite dev`: mounts `<Toggle onclick>` (legacy) → **one** `console.warn` at mount, before any click; a second `<Toggle onclick>` → no repeat (once per prop); `<Toggle onClick>` (corrected) → silent. Both handlers still run when the toggles are clicked.

https://github.com/user-attachments/assets/8c71caba-0caa-4ba7-8ed5-2e9f2e84d52a

Same session, full Chrome window with DevTools open (the ⚠ 1 warning counter in the DevTools toolbar is the browser's own count), cropped to the browser window:

https://github.com/user-attachments/assets/3dff93da-982f-44f3-ad0e-f6e717134c32

What the browser's console listener captured during that run, verbatim (one line — the ThemeSwitcher warning the demo site's own layout used to emit is gone, because the demo site no longer uses a deprecated spelling):

```
[warning] [svelte-ui-components] `onclick` on <Toggle> is deprecated and will be removed in 4.0.0. Use `onClick` instead — both work today. Run `npx sui-codemod --dry-run ./src` to see every affected call site.
```

### 2. `npx sui-codemod` — packed tarball, consumer project

`npm pack` of this commit, extracted under a consumer's `node_modules`, run through the npm bin symlink. The commands are real child processes; their stdout is streamed live into a terminal-styled page so the session could be recorded without capturing the desktop. Dry run → apply → diff → second dry run finds nothing left.

https://github.com/user-attachments/assets/a0a522d9-aeae-4622-a8ef-57bf53a37608

```
[dry run] scanned 2 file(s), changed 1, renamed 5 prop(s), 1 warning(s)
scanned 2 file(s), changed 1, renamed 5 prop(s), 1 warning(s)
[dry run] scanned 2 file(s), changed 0, renamed 0 prop(s), 1 warning(s)
```

(the one warning is the `children` assignment on `sui-draggable` in `mount.ts`, reported on every run by design.)

### 3. Playwright recordings (targeted run on this commit: **24/24 passed**)

`tests/img-inline-svg-src-change.spec.ts` — a `src` change removes the root attributes the previous file supplied:

https://github.com/user-attachments/assets/15efe624-75ba-450f-bd74-a12c1618d5d5

`tests/event-casing-alias.spec.ts` — a component fires its handler through the corrected prop spelling:

https://github.com/user-attachments/assets/2bfb4e4b-801f-445a-81e0-c45e367f09c5

`tests/wc-event-casing-parity.spec.ts` — the legacy spelling still reaches the component through the custom element:

https://github.com/user-attachments/assets/cda9f63c-769c-4aef-8cda-705148833ae3

Targeted run covered `img-inline-svg-src-change`, `img-inline-svg-attr-allowlist`, `event-casing-alias`, `wc-event-casing-parity`, `input-error-announcement` (24 tests, 1.1 min). Traces for every test are in the CI Playwright report artifact.

### Gates on this commit (local)

- `pnpm lint` clean (142 known / 0 new casing violations)
- `pnpm check` 0 errors across all four tsconfigs
- `pnpm run test:unit --run` **513/513** (32 files) — includes the new guard that fails if `src/` ever uses a spelling the library deprecates
- `vite build` exit 0
- packed CLI: `--help` exit 0 via the bin symlink; without `svelte` installed: one-line message, exit 1

src/lib/Img/Img.svelte
```diff
@@ -92,6 +100,40 @@
...
+   * The host attributes this component writes itself. A fetched root may
+   * legitimately overwrite one (an icon set's own role/aria-label), and the
+   * src-change cleanup then has to put the component's value back rather than
+   * just remove it: Svelte only re-applies its own attribute when `alt` or
+   * `testId` changes, so a bare removal would leave the host with no role
+   * until the next re-render.
+   */
+  function restoreHostOwnedAttribute(host: SVGSVGElement, name: string): boolean {
+    const named = alt.length > 0;
+    switch (name.toLowerCase()) {
```
Tara-ag, line 121: ⚠️ MAJOR: `restoreHostOwnedAttribute` reads `alt.length` without guarding against a null/undefined runtime value. `properties.ts` types `alt` as required `string`, but JS consumers can still omit it or pass `null`, and this new cleanup path runs on every `src` change after an inline SVG load. A runtime null would throw `Cannot read properties of null (reading 'length')`.

Suggested fix: coerce defensively, e.g. `const named = (alt ?? '').length > 0;`, and apply the same guard to the template's `alt.length` checks at lines 217-219.

Tara-ag, line 121: ⚠️ MAJOR: `restoreHostOwnedAttribute` computes `const named = alt.length > 0;` without guarding against a null/undefined runtime value. `properties.ts` types `alt` as required `string`, but JS consumers (especially the web component) can omit it or pass `null`. A runtime null throws `Cannot read properties of null (reading 'length')` on every `src` change after an inline SVG load. Apply the same `(alt ?? '').length` guard to the template's `alt.length` checks as well.

src/lib/Img/Img.svelte
Tara-ag, line 1: 🔒 CRITICAL: `sanitizeInlinedSubtree` does not remove SMIL animation elements (`<animate>`, `<set>`, `<animateMotion>`, `<animateTransform>`, `<mpath>`) or `<use>` elements. These can carry `xlink:href`/`href` attributes pointing at `javascript:` URLs or trigger resource loads, leaving an XSS vector in untrusted SVG inlining despite the docblock's claim that a denylist is sufficient. The current `javascript:` prefix check also lets benign-looking but still dangerous URLs remain.

Extend sanitization to drop SMIL animation elements and `<use>`, and strip `href`/`xlink:href`/`src` from all remaining descendants unconditionally.

murdore:
Amended `f19108c` → `66f3007` (still one commit). Delta: one regression caught before merge.

**`Step.svelte` read only the legacy `onclick`.** Its props are declared in `Stepper/properties.ts`, the one directory that hosts two exported components, so the alias generator never reached it — and once `Stepper.svelte` handed it the corrected `onClick` (the internal-usage migration), every step click was silently dropped. `Step` now resolves both spellings like every other component, and `src/lib/Stepper/Stepper.svelte.test.ts` clicks through `Stepper` and `Step` under both spellings (red on the previous commit: 4/4 failed; green now). `docs/EVENT_CASING_MIGRATION.md` no longer describes it as an unwired exception.

Gates on `66f3007`: `pnpm lint` clean, `pnpm check` 0 errors, unit **517/517** (33 files). The recordings above were made on `f19108c`; nothing they show changed.

src/lib/Img/Img.svelte
Tara-ag, line 121: ⚠️ MAJOR: `restoreHostOwnedAttribute` reads `alt.length` without guarding against a null/undefined runtime value. `properties.ts` types `alt` as required `string`, but JS consumers (especially the web component) can omit it or pass `null`. A runtime null throws `Cannot read properties of null (reading 'length')` on every `src` change after an inline SVG load. Apply the same `(alt ?? '').length` guard to the template's `alt.length` checks as well.

Tara-ag (commented):
Review finalized: 2 gate-accepted findings (1 CRITICAL, 1 MAJOR). Both now have inline comments. The CRITICAL finding concerns incomplete SVG sanitization leaving an XSS vector; the MAJOR finding is a runtime null guard in inline-SVG cleanup.

murdore:
## End-to-end verification of `a6a9282`

Everything below was run fresh against the committed tree (`git status` clean, 0 uncommitted files), Node v24.14.1 / pnpm 10.28.2, after the last push. Where a result is not a clean pass, it says so.

### Gates

| Gate | Command | Result |
|---|---|---|
| Lint | `pnpm lint` | exit 0 — **0 event-casing violations**, 190 deprecated aliases awaiting 4.0.0, across 94 `properties.ts` files |
| Types | `pnpm check` | exit 0 — svelte-check + `check:codemod` + `check:migrate` + `check:wc-parity`, 0 errors |
| Unit | `pnpm run test:unit --run` | exit 0 — **563 passed** (33 files) |
| Build | `pnpm run build` | exit 0 — vite + `build:wc` + `build:codemod` + `package` |
| Integration | `pnpm exec playwright test` | exit 0 — **482 passed** (2.6 min) |
| Visual | `pnpm run test:visual` (pinned container) | see note below |

### The consumer path, end to end

`npm pack` → extract into a throwaway consumer's `node_modules/@juspay/svelte-ui-components` → drive the **shipped** `dist-codemod` bin through the npm symlink on Node 24. Not the repo's source — the artifact npm would publish.

```
$ ./node_modules/.bin/sui-codemod --dry-run src
- <Toggle text="Notifications" onClick={(checked) => console.log(checked)} />
+ <Toggle text="Notifications" onclick={(checked) => console.log(checked)} />
- <Input placeholder="Email" onBlur={() => {}} oninput={() => {}} />
+ <Input placeholder="Email" onblur={() => {}} oninput={() => {}} />
- <ListItem text="Account" oncenterTextClick={() => (open = true)} />
+ <ListItem text="Account" oncentertextclick={() => (open = true)} />
- <Table onRowClick={() => {}} columns={[{ id: 'a', onToggle: () => {} }]} />
+ <Table onrowclick={() => {}} columns={[{ id: 'a', onToggle: () => {} }]} />
- <Modal {open} onClose={() => (open = false)} onOverlayClick={() => (open = false)} />
+ <Modal {open} onclose={() => (open = false)} onoverlayclick={() => (open = false)} />
src/mount.ts:2:39 WARN assignment to `.children` in a file using sui-draggable …
[dry run] scanned 2 file(s), changed 1, renamed 6 prop(s), 1 warning(s)

$ ./node_modules/.bin/sui-codemod src          # apply
scanned 2 file(s), changed 1, renamed 6 prop(s), 1 warning(s)

$ ./node_modules/.bin/sui-codemod --dry-run src   # second run
[dry run] scanned 2 file(s), changed 0, renamed 0 prop(s), 1 warning(s)

$ mv node_modules/svelte node_modules/svelte.hidden && ./node_modules/.bin/sui-codemod --dry-run src
sui-codemod needs svelte/compiler installed in this project (svelte is a peer
dependency of @juspay/svelte-ui-components). Install it and run the command again.
exit=1
```

Three things this pins that a unit test cannot: the bin runs at all on Node 24 (the 3.3.1 bug), `TableColumn.onToggle` — a callback key on a config object, not a component prop — is **left alone** while `Table`'s own `onRowClick` is rewritten, and the run is idempotent.

### The published artifact in a real browser

`dist-wc/index.js` **from that same extracted tarball**, loaded into Chromium from a routed origin, exercising the props a consumer would set:

```
PASS  sui-toggle onclick (lowercase)                  result=[true]   warnings=0
PASS  sui-toggle onClick (deprecated)                 result=[true]   warnings=0
PASS  sui-toggle both set — lowercase wins            result=["lowercase"]
PASS  sui-input oninput (native forward, lowercase)   result=["fired"] warnings=0
PASS  sui-img with NO alt + SMIL in the fetched SVG   inlined=1 smilLeft=0 onclickPlanted=null scriptRan=null role=null
page errors: 0
5/5 end-to-end checks passed
```

`warnings=0` on the deprecated spelling is the correct result, not a gap: the warning is dev-only, and `grep -c "is deprecated and will be" dist-wc/index.js` → **0**. The whole deprecation path is compiled out of production, so consumers who never touch an old spelling pay nothing for it. The warning firing once, at mount, is proven in dev by `src/lib/Toggle/Toggle.svelte.test.ts` and `src/lib/Stepper/Stepper.svelte.test.ts`.

### Review findings — both reproduced first, then fixed

`tests/img-review-findings.spec.ts` was written before either fix and failed 3/3 on the prior commit:

```
✘ an inlining sui-img with no alt renders, and survives a src change
    Error: a missing alt threw at runtime
✘ SMIL animation elements do not survive adoption
    Error: a SMIL animation element survived adoption
✘ a URL reference that leaves the document is dropped, an internal one is kept
    Error: a remote xlink:href survived
```

After the fixes, with the two existing Img suites alongside: **9 passed**.

- `alt` is a required `string`, which settles it for a Svelte consumer and settles nothing for `<sui-img>`: a custom element takes its props from JavaScript, so an omitted `alt` arrived undefined and threw. One guarded `label` derived value now feeds all four reads.
- The sanitizer stripped `on*` and `javascript:` URLs but kept SMIL, which animates any attribute by name — `<set attributeName="onclick" to="…">` plants a handler a frame after the guard ran. SMIL elements are now removed outright, and a URL attribute survives only if it points inside the same document, which also stops a fetched `<image href="https://…">` making a request the page never asked for. `<use href="#defs">` still works, so real icon sets are unaffected.

### Housekeeping found while re-reading the commit

`src/lib/deprecation-version.ts` existed only so a Node-only generator could read the version string without pulling in `import.meta.env`. That generator is deleted in this commit, nothing but `deprecation.ts` read the file, and its docblock named a script that no longer exists — folded back and removed. A sweep for references to every file this commit deletes found nothing else dangling.

### The other shipped CLI

`scripts/migrate/cli.ts` (the 3.x consumer audit) still runs after this commit deleted four of its sibling scripts — checked against the same fixture:

```
$ node scripts/migrate/cli.ts <consumer>
@juspay/svelte-ui-components: ^3.2.0 -> ^3.0.0
No blockers and no affected Toolbar usage across 2 .svelte files.
scanned 2 .svelte file(s), 0 needing review, 0 blocker(s)
```

### One honest caveat: a visual flake

`pnpm run test:visual` runs in the pinned `mcr.microsoft.com/playwright:v1.60.0-noble` container. Across runs on this commit:

- CI's `visual` job on this exact commit (run 33989259191): **pass**
- local container run 1: **93 passed**
- local container run 2: **92 passed, 1 failed** — `sankey-chart`, `Failed to take two consecutive stable screenshots`

That is an instability, not a pixel diff, and `SankeyChart.svelte` carries `transition: … 0.2s ease` on three rules. This commit's only change to that component is prop-binding renames (`onnodeclickLegacy` → `onnodeclickProp` and the resolver line) — no rendering or timing change. I am reporting it rather than calling the suite clean. (Ad-hoc `playwright test --config playwright.visual.config.ts` on macOS fails this and other baselines by design — `scripts/visual-test.sh` exists precisely because host font rendering is not a valid baseline renderer.)

### Review status

`Yama PR Review` reports **BLOCKED**, and its own summary says why:

> Review aborted due to exhausted time budget: no PR diff, changed files, or repository contents were fetched before the cutoff, so no findings could be gate-verified.

Severity counts on that verdict are **0 CRITICAL · 0 MAJOR · 0 MINOR**, and the job log shows `FINDINGS:` empty — it is a budget limit on a ~250-file diff, not a code judgement (CodeRabbit skipped this PR for the same reason: "too many files"). The two findings Yama *did* produce, before it ran out of budget, are the Img ones above — both reproduced with failing tests and fixed. I have re-run the job; if it aborts again, the diff is simply larger than the reviewer's budget and the commit body is written to be read as the map instead.

murdore:
### Follow-up on the proof above: commit retyped, tree unchanged

`a6a9282` → `bf32464`. **The tree is byte-identical** (`git diff --stat a6a9282 bf32464` is empty), so every result in the verification comment above still stands as measured; only the commit message changed.

Why: the subject was `fix(events):`, which `release.yml`'s version detection reads as a **patch** — but this commit adds 191 lowercase props, which is new public API, and `docs/EVENT_CASING_MIGRATION.md` (shipped in this very commit) names **3.5.0** as the release that must land before the 4.0.0 removal can. A patch would have made the doc false and understated the change in the changelog. It is now `feat(events): …` → minor → 3.5.0.

Checked while I was in there: the body contains no line starting `BREAKING CHANGE:`, which is the only other thing that flips `release.yml` to a major — so this cannot cut 4.0.0 by accident. The subject also matches the repo's mandatory-scope pattern `^[a-z]+\([^)]+\): .+`.

murdore:
### Final CI status on `bf32464` — and an honest note on the review

| Check | Result |
|---|---|
| `checks` (lint · types · unit · integration) | **pass** — 8m1s |
| `build` | **pass** — 1m22s |
| `visual` | **pass** — 2m55s |
| `Yama PR Review` | **pass** — 33m41s |

`mergeStateStatus: CLEAN`, one commit.

**But the green Yama check should not be read as a review.** Its verdict comment on this run says:

> ## 🛡️ Yama Review Verdict: CHANGES_REQUESTED
> Severity counts — 🔒 CRITICAL: 0 · ⚠️ MAJOR: 0 · 💡 MINOR: 0
> Review aborted due to exhausted time budget: I was unable to read the PR diff, changed files, or verify any behavior. No findings could be gated.

Two consecutive runs aborted the same way (the first landing on `BLOCKED`, this one on `CHANGES_REQUESTED` — which is why the job is green this time). Both report **0 findings** because neither ever fetched the diff. CodeRabbit skipped this PR for the same reason: "too many files". A third re-run would not change the arithmetic — the diff is simply larger than the reviewer's budget.

So the automated review of this PR is: **the two Img findings it produced before running out of budget** (both reproduced with failing tests, fixed, and covered by `tests/img-review-findings.spec.ts`), and nothing else. Everything beyond that rests on the gates in the verification comment above and on human reading — which is why the commit body itemises every change as a rule rather than a file list.

If a full automated pass is wanted before merge, the practical options are to raise Yama's budget for this repository, or to accept that a 250-file mechanical rename is reviewed by its generator and its tests rather than file-by-file.

murdore:
## Verification pass: the aliases are now measured, not inferred

The earlier proof on this PR drove **5 props across 3 components** and inferred the rest. This commit changes 191 prop pairs across 67 components, so that was 2 aliases with runtime proof and 189 taken on trust. The aliases are the half nothing else watches — the rest of the suite exercises the lowercase spellings, which cannot regress silently — and their failure mode is a consumer's handler quietly never firing.

Four test files close that. Each was verified by breaking the thing it guards and watching it fail, before being kept.

| Test | Covers | Verified to fail when |
|---|---|---|
| `src/deprecated-alias-matrix.test.ts` | 192 cases — all 191 pairs × 3 resolver branches, in jsdom | resolver removed from Accordion → *"never resolved its deprecated `onToggle`"* |
| `tests/wc-alias-reachability.spec.ts` | the aliases driven through the built custom elements in Chromium | `onClick` declaration dropped from the Toggle wrapper → names `sui-toggle.onClick` |
| `scripts/check-event-casing.test.ts` | 7 cases on the lint gate, which had **no test at all** | indent rule dropped → nested callback misflagged |
| `scripts/wc-parity/forwarding.test.ts` | 67 wrappers forward what they declare | rest element removed → names the props at risk |

Unit tests **563 → 829**; Playwright **484 → 491** (7 from #533 after the rebase).

### What the browser spec can and cannot prove

It is self-calibrating, which is what makes it possible without 67 bespoke fixtures: each pair is swept twice under an identical blind interaction — once with only the lowercase spelling set, once with only the camelCase one — so the lowercase run establishes whether the event is reachable at all, and only reachable pairs are asserted on.

```
[alias reachability]   62/172 wrapped pairs reachable, 110 not reachable by a blind sweep,
                       19 pairs have no custom element (the charts, and Step)
[declared event props] 141/364 reachable across 66 elements
```

Every one of the 62 answers its alias. The denominators are stated deliberately: 110 pairs need state a blind sweep cannot invent (a closed modal fires nothing; a chat with no failed message has nothing to retry), and the 19 chart/`Step` pairs have no custom element at all, so they rest on the jsdom matrix. Both counts are pinned as floors, so a wrapper that stops forwarding fails here instead of passing with an empty set.

### A false positive worth recording

The first version compared **exact fire counts** and reported `sui-split-input`'s `oninput` as 4 against 6 — apparently a real defect. Reversing the order of the two drives flipped the mismatch to the other spelling, and driving one event type at a time showed the two spellings identical (`[4,4,4]` both). The component carries state between drives. Reachability survives a stateful component; the count does not, so that is what the committed test asserts.

### What the pass actually found

**Zero wiring bugs.** All 41 initial failures in the matrix were mount problems — missing required props, `ResizeObserver`, `matchMedia`, canvas — not defects. The migration was correct. What was missing was any way to know that.

One limit of the existing safety net is worth naming: the generator's fixed-point test catches a component edited by hand, including an argument swap. It cannot catch the generator emitting a *consistently wrong* pattern, because then the source matches its output and the fixed point holds. Measured by swapping the resolver's last two arguments in generator and source together — the fixed point stays silent on Accordion, the matrix fails.

### Rebase

#533 released **3.5.0** while this was in progress, which left this PR `CONFLICTING`. Rebased onto `a0cacff`. One conflict, in `ChatMessage.svelte`, where #533's clamp/expand and this commit's resolvers land in the same script block — both kept, and three stale lines referencing `oncopyLegacy` bindings that no longer exist after the rename were dropped.

`docs/EVENT_CASING_MIGRATION.md` named 3.5.0 as phase 1's release, which is now taken. I removed the prediction rather than replacing it with 3.6.0 — that is the same bet that just lost, since any `feat` merging first moves it. The CHANGELOG is the authority.

### Gates, on the rebased commit

| Gate | Result |
|---|---|
| Lint | exit 0 — 0 violations, 190 aliases, 94 `properties.ts` |
| Types | exit 0 — 867 files, 0 errors |
| Unit | **829 passed** (36 files) |
| Build | exit 0 |
| Playwright | **491 passed** |
| Visual | **93 passed** (pinned container) |
| Packed tarball | codemod dry-run → apply → idempotent no-op → friendly failure without `svelte`; `TableColumn.onToggle` left alone |

PR #533, opened 2026-09-05
feat(ChatMessage): collapse a long message to a clamped, expandable bubble

murdore:
## Problem

Consumers are rebuilding this on top of the `body` snippet.

Lighthouse's `UserMessage` wrapper exists almost entirely for it: a `role="button"` div, a tabindex, an Enter/Space handler, a `data-clipped` attribute and a `-webkit-line-clamp` block — wrapped around a `ChatMessage` that was already rendering all the chrome. Every consumer wanting a collapsed message re-derives the same keyboard handling and the same clamp CSS, and gets to make its own mistakes in both.

## Change

`clampLines={n}` collapses the rendered body to *n* lines and makes the bubble its own control — click, or focus and press Enter/Space.

`0` or omitted leaves the message alone and **adds no interactive role**, so nothing changes for existing callers.

- The clamp applies to the **rendered body, not the source**, so `content` still carries the whole message and the copy action still copies all of it.
- `--chat-message-clamp-lines` overrides the line count from a consumer theme sheet, without a second rule.

## Tests

`tests/chat-message-clamp.test.ts`, against a new demo case on `/components/chat-message`. The assertions are about **measured height and exposed state** rather than the class names that produce them, because the risk in moving this into the library is that the clamp stops actually hiding anything:

- collapsed height `<` expanded height
- `data-clamped` / `data-expanded` flip on click
- Enter expands, Space collapses
- a message without `clampLines` gets no `role="button"`

**Negative control** — with the clamp neutralised to 99 lines, the suite fails on the height assertion:

```
Expected: > 108
Received:   108
```

So collapsed and expanded are genuinely different states, not differently labelled ones. `pnpm run check` passes; both tests green.

## Impact

Additive. One new optional prop with a falsy default; the bubble gains `role`/`tabindex`/handlers **only** when `clampLines > 0`. No existing call site changes behaviour.

## Follow-up

This unblocks deleting Lighthouse's `UserMessage` and `AssistantMessage` wrappers entirely — their appearance has already moved to that app's theme sheets, and `clampLines` was the last behaviour keeping `UserMessage` alive. `AssistantMessage`'s only remaining branch (a per-character fade) is fed by a field nothing in that app populates.

## Summary by CodeRabbit

* **New Features**
  * Added optional collapsing for long chat messages with a configurable line limit.
  * Messages can be expanded or collapsed by clicking or pressing Enter/Space.
  * Full message content remains available for copying.
* **Documentation**
  * Added an example demonstrating a two-line, expandable chat message.
* **Tests**
  * Added coverage for collapsing, expansion, keyboard controls, and non-collapsible messages.


Tara-ag (commented):
Review in progress — posting findings file by file.

tests/chat-message-clamp.test.ts
```diff
@@ -0,0 +1,41 @@
...
+// behaviour was being rebuilt by consumers around the `body` snippet, which meant each
+// one re-derived the keyboard handling and the clamp CSS; the risk in moving it here is
+// that the clamp stops actually hiding anything, so the assertions are about measured
+// height and the exposed state rather than about the class names that produce them.
+test.describe('ChatMessage — clampLines', () => {
+  test('collapses a long message and expands it by click and by keyboard', async ({ page }) => {
+    await page.goto('/components/chat-message');
+
+    const bubble = page.locator('[data-pw="clamp-demo"] .bubble');
+    await expect(bubble).toBeVisible();
```
Tara-ag, line 13: 💡 MINOR: This test selects a demo-page-specific `data-pw="clamp-demo"` attribute, coupling component-contract coverage to the showcase page markup. Prefer an isolated test route or harness that renders only the `ChatMessage` under test, with a stable selector that does not depend on the demo page structure.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -136,7 +154,16 @@
         <div class="header">{@render header()}</div>
       {/if}
 
-      <div class="bubble">
+      <div
+        class="bubble"
+        class:clampable={clampLines > 0}
```
Tara-ag: ⚠️ MAJOR: Clamped bubble lacks accessible name and expanded-state communication. This bubble becomes `role="button"` when `clampLines > 0` but has no `aria-label`, `aria-labelledby`, or `aria-expanded`. Screen-reader users hear an unnamed "button" and cannot tell whether the message is currently expanded. Add `aria-expanded={String(expanded)}` and an accessible name such as `aria-label={expanded ? 'Collapse message' : 'Expand message'}` (or expose `expandLabel`/`collapseLabel` props for i18n).

Tara-ag (changes requested):
Reviewed file by file. Two inline comments posted on changed files; two additional findings affect unchanged files that are part of the public contract (web-component wrapper and docs) and are included in the final verdict JSON.

tests/chat-message-clamp.test.ts
Tara-ag, line 13: 💡 MINOR: Playwright test selects demo-page-specific `data-pw` attribute. The test locates the clamp demo via `[data-pw="clamp-demo"] .bubble`, coupling the component contract test to the showcase page markup. Per repo conventions, component tests should use isolated fixtures or harness routes with stable testIds rather than demo-page-specific attributes.

**Fix:** Create a dedicated test route (e.g., `/tests/chat-message-clamp`) that renders only the `ChatMessage` under test, or use a component harness, so the selector does not depend on the demo page structure.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -136,7 +154,16 @@
         <div class="header">{@render header()}</div>
       {/if}
 
-      <div class="bubble">
+      <div
```
Tara-ag, line 177: ⚠️ MAJOR: Clamped bubble lacks accessible name and expanded-state communication. When `clampLines > 0`, the bubble becomes a `role=button` but has no `aria-label`, `aria-labelledby`, or inner text describing its purpose. Screen-reader users hear only "button" with no indication that it expands/collapses or what the current state is.

**Fix:** Add `aria-expanded={String(expanded)}` and an `aria-label` such as `aria-label={expanded ? 'Collapse message' : 'Expand message'}` (or expose new props `expandLabel`/`collapseLabel`) to the clamped bubble.

Tara-ag:
💡 MINOR: `docs/ChatMessage.md` — Documentation not updated for new prop and CSS variable.

The PR adds the public prop `clampLines` and the CSS custom property `--chat-message-clamp-lines`, but `docs/ChatMessage.md` still reflects the old prop table and CSS-variable table. Downstream consumers relying on the docs will not discover the new API or theming hook.

**Fix:** Add a `clampLines` row to the Props table and a `--chat-message-clamp-lines` row to the CSS Variables table in `docs/ChatMessage.md`.

Tara-ag:
⚠️ MAJOR: `src/wc/components/ChatMessage.wc.svelte` — New `clampLines` prop is not declared in the web-component wrapper.

The Svelte component adds `clampLines?: number`, but the WC wrapper does not list it in the `customElement` `props` map. Web-component consumers cannot set the prop via attribute or property, breaking parity between the Svelte and WC APIs.

**Fix:** Add `clampLines: { type: 'Number', attribute: 'clamp-lines' }` to the customElement props map in `src/wc/components/ChatMessage.wc.svelte`.

Tara-ag (changes requested):
Review findings for PR #533.

Tara-ag:

Severity counts — 🔒 CRITICAL: 1 · ⚠️ MAJOR: 2 · 💡 MINOR: 4

Reviewed the latest revision of the ChatMessage clamp/expand feature. The earlier MAJOR accessibility and API-parity issues are resolved: the bubble is inert markup, the expand/collapse control is a real Button with aria-expanded/aria-controls, and the web-component wrapper exposes clampLines. Three MINOR items remain: a reactive-state write inside $effect, Playwright tests coupled to demo-page markup, and the web-component clampLines prop not reflecting to its attribute. None block merge.

- 🔒 **CRITICAL**: Button `ariaControls` prop added but not declared in web-component wrapper — `src/wc/components/Button.wc.svelte:1`
  The PR adds `ariaControls?: string` to Button properties and passes it through in Button.svelte, but `src/wc/components/Button.wc.svelte` does not declare `ariaControls` in its custom-element props map. Consumers using `<sui-button aria-controls="...">` will have the attribute ignored by the wrapper, breaking the public contract for the web-component build. The bootstrapped standards require expli…
- ⚠️ **MAJOR**: Clamped ChatMessage body lacks programmatic association for assistive tech — `src/lib/ChatMessage/ChatMessage.svelte:1`
  The docs claim the toggle Button has `aria-controls` pointing at the clamped body, but ChatMessage.svelte never renders `aria-controls` on the Button. Without this association, screen-reader users cannot reliably navigate from the toggle to the expanded region. The `bodyId` variable exists but is unused for aria-controls.
- ⚠️ **MAJOR**: Docs claim `aria-controls` on clamp toggle but no test verifies it — `tests/chat-message-clamp.test.ts:1`
  The new Playwright tests exercise clamping, expansion, keyboard controls, and link isolation, but none assert that the toggle button has `aria-controls` matching the body id. Stale or incorrect docs around accessibility are a contract bug for a published library.
- 💡 **MINOR**: ChatMessage docs default for `clampLines` contradicts properties.ts — `docs/ChatMessage.md:1`
  docs/ChatMessage.md says `clampLines` default is `0`, but src/lib/ChatMessage/properties.ts documents `0 or omitted leaves the message uncollapsed`. The table default column says `0` while the runtime destructuring default is not shown; this is consistent but the phrasing could confuse. More importantly, the CSS variable table says `--chat-message-clamp-lines` defaults to `clampLines, else 2`, imp…
- 💡 **MINOR**: $effect mutates local component state — `src/lib/ChatMessage/ChatMessage.svelte:53`
  ChatMessage resets `expanded` by writing inside a `$effect`. The inline `// eslint-disable-next-line no-restricted-syntax` confirms the repo normally bans this rune. While the author argues this transition cannot be expressed with `$derived`, the pattern still couples state mutation to the effect lifecycle and is the only place in this component that writes state reactively.
- 💡 **MINOR**: Playwright tests are coupled to demo-page markup — `tests/chat-message-clamp.test.ts:10`
  The new clamp tests navigate to `/components/chat-message` and select elements through `data-pw` attributes defined in the demo page. If the demo page is refactored or the demo cases are renamed, coverage can silently drop even though the component itself still works.
- 💡 **MINOR**: Web-component `clampLines` prop does not reflect to its attribute — `src/wc/components/ChatMessage.wc.svelte:12`
  In the custom-element wrapper `clampLines` is declared with `attribute: 'clamp-lines'` but `reflect: true` is omitted. Because the Svelte component supports dynamically withdrawing and restoring `clampLines` (demo toggle), web-component consumers who inspect the attribute will see it become stale after a property change.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -37,6 +38,23 @@
     classes
   }: ChatMessageProperties = $props();
 
+  // A long message can be collapsed to a few lines and opened by click or keyboard.
+  // Consumers had been rebuilding this around the `body` snippet; it belongs here so
+  // the chrome, the clamp and the expanded state stay one component's concern.
+  let expanded = $state(false);
+  let clampActive = $derived(typeof clampLines === 'number' && clampLines > 0 && !expanded);
```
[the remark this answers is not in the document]
murdore: Fixed in `afc405b`.

Reset moved into a `$effect` guarded on `clampable`. `$effect` is banned by `no-restricted-syntax` here, but it is a sanctioned escape hatch with an explicit disable and a justification — all 16 existing uses in `src/lib` follow that shape, and a prop going away cannot be observed from a `$derived` without writing to state.

Covered by a test that drives a real 4 → 0 → 4 round trip through a toggle on the demo page. Negative control: removing the effect fails that test on `[data-pw="clamp-demo-toggle"] .bubble`.

src/lib/ChatMessage/ChatMessage.svelte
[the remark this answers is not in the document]
murdore: Correct, and this was the real defect — thank you. The prop was decorative: the stylesheet read `var(--chat-message-clamp-lines, 2)` and nothing ever set that variable, so every value clamped at two lines. My test could not see it because it used the default of `2`, the one value where correct and broken agree.

Fixed in `afc405b`, with one deviation from the suggested mechanism: rather than a private variable plus a public override, the prop publishes the public token inline, which is what `KeyValue` (`--keyvalue-columns`) and `ContextMenu` (`--context-menu-max-height`) already do in this library. Same fallback shape, one variable instead of two, and per-instance beats ambient.

New test asserts a value that is not 2 — `clampLines={4}` must compute `-webkit-line-clamp: 4` and be taller than the `2` case. Negative control: removing the inline token fails it on `[data-pw="clamp-demo-four"] .body`.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -136,7 +154,16 @@
...
-      <div class="bubble">
+      <div
+        class="bubble"
+        class:clampable={clampLines > 0}
+        data-clamped={clampActive ? 'true' : undefined}
+        data-expanded={clampLines > 0 ? String(expanded) : undefined}
+        role={clampLines > 0 ? 'button' : undefined}
+        tabindex={clampLines > 0 ? 0 : undefined}
+        onclick={clampLines > 0 ? toggleExpanded : undefined}
+        onkeydown={clampLines > 0 ? onClampKeydown : undefined}
```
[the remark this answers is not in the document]
murdore: Fixed in `afc405b` — all six now use `null`, matching the existing `data-pw={... : null}` on the same element. `pnpm lint` exits 0.

One related change beyond the suggestion: `aria-expanded` takes the boolean rather than `String(expanded)`, because the typed attribute is `Booleanish | null` and `svelte-check` rejected the string.

src/lib/ChatMessage/ChatMessage.svelte
murdore: Agreed and fixed in `afc405b`, taking the `expandLabel` / `collapseLabel` suggestion.

`aria-expanded` tracks the state, and `aria-label` names the action through two new props defaulting to `Expand message` / `Collapse message` — props rather than hardcoded strings to match `copyLabel` and `retryLabel`, which is how every other label in this component is already exposed.

Worth stating the reasoning, since `aria-label` on a `role="button"` replaces the contents in the accessible-name computation: that is the intended outcome here. Without it the control's name was the entire message text, so a screen-reader user heard the message read back as a control name and learned neither what activating it does nor whether it was open. The message body is still in the accessibility tree and still readable in browse mode.

Also added `svelte-ignore a11y_no_noninteractive_tabindex` with a note: `role`, `tabindex` and both handlers are gated on the same condition, so the element is only focusable while it genuinely is a button — the checker cannot see that through a dynamic role.

Negative control: removing `aria-label` fails the new test on `[data-pw="clamp-demo"] .bubble`.

src/lib/ChatMessage/ChatMessage.svelte
murdore: Agreed and fixed in `afc405b`, taking the `expandLabel` / `collapseLabel` suggestion.

`aria-expanded` tracks the state, and `aria-label` names the action through two new props defaulting to `Expand message` / `Collapse message` — props rather than hardcoded strings to match `copyLabel` and `retryLabel`, which is how every other label in this component is already exposed.

Worth stating the reasoning, since `aria-label` on a `role="button"` replaces the contents in the accessible-name computation: that is the intended outcome here. Without it the control's name was the entire message text, so a screen-reader user heard the message read back as a control name and learned neither what activating it does nor whether it was open. The message body is still in the accessibility tree and still readable in browse mode.

Also added `svelte-ignore a11y_no_noninteractive_tabindex` with a note: `role`, `tabindex` and both handlers are gated on the same condition, so the element is only focusable while it genuinely is a button — the checker cannot see that through a dynamic role.

Negative control: removing `aria-label` fails the new test on `[data-pw="clamp-demo"] .bubble`.

tests/chat-message-clamp.test.ts
murdore: Not taking this one — the premise does not hold for this repository, and I would rather say so than quietly comply.

There is no `/tests/` route tree here, and the demo pages under `/components/*` **are** the harness: 102 of the 108 files in `tests/` navigate to one. Measured:

```
$ grep -rl "goto('/components/" tests/*.ts | wc -l   ->  102
$ ls tests/*.ts | wc -l                            ->  108
$ ls src/routes/                                   ->  components  (only)
```

Adding an isolated route for this one component would introduce a pattern used by exactly one file and diverge from the convention every neighbouring test follows. If the library should move to isolated harness routes, that is a deliberate repo-wide change rather than something to start in a feature PR.

What I did do is make the assertions independent of the demo page's *structure*: they are measured height, computed `-webkit-line-clamp`, and ARIA state, addressed by stable `testId`s passed to the component itself. Each one is negative-controlled, so it fails when the behaviour is reverted.

murdore: Not taking this one — the premise does not hold for this repository, and I would rather say so than quietly comply.

There is no `/tests/` route tree here, and the demo pages under `/components/*` **are** the harness: 102 of the 108 files in `tests/` navigate to one. Measured:

```
$ grep -rl "goto('/components/" tests/*.ts | wc -l   ->  102
$ ls tests/*.ts | wc -l                            ->  108
$ ls src/routes/                                   ->  components  (only)
```

Adding an isolated route for this one component would introduce a pattern used by exactly one file and diverge from the convention every neighbouring test follows. If the library should move to isolated harness routes, that is a deliberate repo-wide change rather than something to start in a feature PR.

What I did do is make the assertions independent of the demo page's *structure*: they are measured height, computed `-webkit-line-clamp`, and ARIA state, addressed by stable `testId`s passed to the component itself. Each one is negative-controlled, so it fails when the behaviour is reverted.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -341,6 +389,20 @@
     }
   }
 
+  .bubble.clampable {
+    cursor: pointer;
```
Tara-ag: ⚠️ MAJOR: The clampable bubble is focusable (`tabindex="0"`) but has no `:focus` style beyond `cursor: pointer`. Keyboard users may not be able to see which bubble has focus. Add a visible focus indicator, e.g.:
```css
.bubble.clampable:focus {
  outline: var(--chat-message-clamp-focus-outline, 2px solid currentColor);
  outline-offset: 2px;
}
```

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -136,7 +168,23 @@
         <div class="header">{@render header()}</div>
       {/if}
 
-      <div class="bubble">
+      <!-- The clamped bubble is a disclosure control: role, tabindex and both handlers are
+           gated on the same condition, so it is only focusable while it is genuinely a
+           button. The checker cannot see that through a dynamic role. -->
+      <!-- svelte-ignore a11y_no_noninteractive_tabindex -->
+      <div
+        class="bubble"
```
Tara-ag, line 178: ⚠️ MAJOR: The clamped bubble is declared as `role="button"` and receives clicks, but it still renders arbitrary `body` snippets and `{@html effectiveHtml}` (which may contain links, buttons, or other interactive descendants). ARIA forbids interactive content inside a `button`, and clicks on nested links/buttons will bubble up and toggle expansion in addition to their own action. Guard `toggleExpanded` against events from nested interactive elements, or split the expand control out of the message body.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -136,7 +168,23 @@
...
+        class:clampable
+        style={clampable ? `--chat-message-clamp-lines: ${clampLines};` : null}
+        data-clamped={clampActive ? 'true' : null}
```
Tara-ag, line 181: 💡 MINOR: Setting `--chat-message-clamp-lines` via an inline `style` attribute has higher specificity than a consumer theme-sheet rule, so a consumer cannot override the line count through CSS as the prop comment and PR description suggest. Either drop the inline style and rely solely on the prop, or document that theme overrides require `!important` / omitting `clampLines`.

Tara-ag:
💡 MINOR (docs/ChatMessage.md): The PR adds public props `clampLines`, `expandLabel`, `collapseLabel` and the CSS variable `--chat-message-clamp-lines`, but `docs/ChatMessage.md` is not updated. Please add them to the props and CSS-variables tables so downstream consumers can discover the new API.

Tara-ag (changes requested):
Reviewed file by file. The feature is additive and the public API is backward-compatible, but two accessibility/behavior issues need to be addressed before merge: the clamped bubble cannot safely act as a button while containing arbitrary interactive content, and it lacks a visible focus indicator. Two minor items (docs update and inline-style specificity claim) are also noted.

Tara-ag (changes requested):
Reviewed file by file. The feature is additive and backward-compatible, but two MAJOR accessibility/behavior issues block merge: the clamped bubble cannot safely act as a button while containing arbitrary interactive content, and it lacks a visible focus indicator. Two MINOR items (docs update and inline-style specificity claim) are also noted.

Tara-ag (changes requested):
Review finalized. The three accepted findings on changed files already have inline comments; the docs/ChatMessage.md finding applies to an unchanged public-contract file and is recorded in the final verdict JSON.

Tara-ag:
💡 MINOR: docs/ChatMessage.md not updated for new public API. The PR adds `clampLines`, `expandLabel`, `collapseLabel` props and the `--chat-message-clamp-lines` CSS custom property, but this documentation file is unchanged. Downstream consumers relying on the published docs will not discover the new API or theming hook.

**Fix:** Add the three new props to the Props table and add `--chat-message-clamp-lines` to the CSS Variables table with its default (`2`) and description.

src/lib/ChatMessage/ChatMessage.svelte
murdore: Taken in `8e1db71`, with one correction to the premise.

Measured first: the bubble already gets the UA ring today — `outline: 1px auto rgb(0, 95, 204)` on keyboard focus, `box-shadow: none`. There is no `outline` declaration anywhere in this component, so nothing was suppressing it. So keyboard focus is visible as shipped, and the finding overstates the current risk.

It is still worth declaring, for the reason you would hit in practice rather than the one stated: consumer sheets in this ecosystem **do** reset outlines on message surfaces, and a focusable bubble that loses its ring to a consumer reset is unusable by keyboard. Declaring it in the component means the ring survives that and is themeable.

Added on `:focus-visible` rather than `:focus`, so a mouse click does not paint a ring:

```css
.bubble.clampable:focus-visible {
  outline: var(--chat-message-clamp-focus-outline, 2px solid currentColor);
  outline-offset: var(--chat-message-clamp-focus-outline-offset, 2px);
}
```

The test injects a non-`!important` `.bubble { outline: none }` — deliberately not `!important`, since that beats any non-important rule whatever its specificity, so asserting against it would test the cascade rather than the fix. Negative control: removing the rule gives `Expected "solid 2px", Received "none 3px"`.

src/lib/ChatMessage/ChatMessage.svelte
murdore: Fixed in `8e1db71` — the behavioural half, and I want to be precise about which half.

Activation originating from a nested control now does its own job and nothing else. `isNestedControl` walks up from `event.target` and bails if the nearest `a[href], button, input, select, textarea, [role="button"], [role="link"]` is not the bubble itself — the `!== event.currentTarget` check matters, since the bubble carries `role="button"` and would otherwise match itself and disable the control entirely. Applied to both click and keydown, since Enter on a nested link bubbles the same way.

Proven with a real case rather than reasoned: the demo now has a clamped message containing a link, and the test clicks that link and asserts the bubble is untouched, then clicks the bubble and asserts it still toggles. Negative control: restoring the unguarded handler fails on `[data-pw="clamp-demo-link"] .bubble`.

**The structural half is not fixed and I am not claiming it is.** `role="button"` with interactive descendants remains an ARIA violation whenever a consumer passes `body`/`html` containing controls, and no runtime guard changes that — the component cannot know statically what a snippet renders. The honest options are splitting the expand affordance into its own button, or dropping the role when non-text content is supplied; both change the interaction design that consumers are currently hand-rolling, so they belong in a follow-up with that decision made deliberately rather than smuggled into this PR.

src/lib/ChatMessage/ChatMessage.svelte
murdore: You are right, and this is the second time the same point has been made on this PR — CodeRabbit asked for exactly this shape (private variable, public token higher priority) and I declined it on the grounds that publishing the public token inline matched `KeyValue` and `ContextMenu`. That reasoning was about consistency and missed the consequence you have named: an inline style outranks any ancestor rule, so the theme override the prop documentation promises was inert.

Fixed in `8e1db71`:

```css
line-clamp: var(--chat-message-clamp-lines, var(--chat-message-clamp-lines-prop, 2));
```

with the prop published inline as `--chat-message-clamp-lines-prop`. Precedence is now consumer token > prop > 2, which is what the docs said all along.

Chose that over documenting the inverse because it is the more useful direction here: a theme that wants a different line count everywhere sets one token, rather than every call site passing a prop.

Test asserts it the way a consumer would hit it — `clampLines={4}`, then inject `:root { --chat-message-clamp-lines: 3 }` and require the computed clamp to become `3`. Negative control: reverting to the inline public token fails on `[data-pw="clamp-demo-four"] .body`.

Tara-ag (commented):
Review in progress — file-by-file findings will follow as inline comments.

Tara-ag:
💡 MINOR (docs/ChatMessage.md): The PR adds public props `clampLines`, `expandLabel`, `collapseLabel` and CSS variables `--chat-message-clamp-lines`, `--chat-message-clamp-lines-prop`, `--chat-message-clamp-focus-outline`, `--chat-message-clamp-focus-outline-offset`, but `docs/ChatMessage.md` was not updated. For a published library, public prop and token changes must be reflected in the component documentation.

Suggestion: Add `clampLines`, `expandLabel`, and `collapseLabel` rows to the Props table, and add `--chat-message-clamp-lines`, `--chat-message-clamp-focus-outline`, and `--chat-message-clamp-focus-outline-offset` to the CSS Variables table with descriptions matching `src/lib/ChatMessage/properties.ts`.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -136,7 +186,23 @@
...
 
-      <div class="bubble">
+      <!-- The clamped bubble is a disclosure control: role, tabindex and both handlers are
+           gated on the same condition, so it is only focusable while it is genuinely a
+           button. The checker cannot see that through a dynamic role. -->
+      <!-- svelte-ignore a11y_no_noninteractive_tabindex -->
+      <div
+        class="bubble"
+        class:clampable
+        style={clampable ? `--chat-message-clamp-lines-prop: ${clampLines};` : null}
```
Tara-ag, line 180: 💡 MINOR: When `clampLines` is set, the bubble `<div>` receives `role="button"` but still wraps the message body, which can contain links, buttons, inputs, headings, etc. WAI-ARIA prohibits interactive elements or headings inside a `button` role. Although `isNestedControl` prevents nested activations from toggling the bubble, screen readers may still struggle with the invalid nesting, and the bubble's accessible name is replaced by `aria-label` so the message content is no longer announced as part of the control.

Suggestion: If the design permits, add a dedicated Expand/Collapse button (visually anchored at the bottom of the bubble or as a screen-reader-only control) instead of giving the whole bubble `role="button"`. If the whole-bubble interaction must stay, add an Accessibility section to `docs/ChatMessage.md` noting that clamped messages should avoid nested interactive content where possible.

Tara-ag (commented):
Review complete. The change is additive and well-tested. Two minor findings: the component docs need updating for the new public props/CSS variables, and the whole-bubble `role="button"` warrants an accessibility note because it can wrap rich/interactive content. Neither blocks merge.

Tara-ag (approved):
Review finalized. The change is additive and backward-compatible: new `clampLines`/`expandLabel`/`collapseLabel` props and CSS variables are correctly wired through the Svelte component, TypeScript properties, and web-component wrapper with explicit kebab-case attribute mappings. Playwright tests cover clamping, expansion, keyboard activation, nested-control guarding, theme override precedence, and focus visibility. Two MINOR findings remain: `docs/ChatMessage.md` needs updates for the new public props/CSS variables, and the whole-bubble `role="button"` should carry an accessibility note because it can wrap rich/interactive content. Neither blocks merge.

src/lib/ChatMessage/ChatMessage.svelte
murdore: Taking the documentation option in `18d49d7`, since it is the one that does not smuggle a design decision into this PR.

`docs/ChatMessage.md` now has an Accessibility section, and it states the limit plainly rather than only the mitigation:

> **Avoid nested interactive content in a clamped message.** WAI-ARIA prohibits interactive elements and headings inside `role="button"`, and this component cannot detect what a `body` snippet or `html` string renders. Activation originating from a nested link or button is guarded so it does not also toggle the clamp, but the invalid nesting itself remains, and assistive technology may not expose those descendants. If a message must carry links or controls, leave `clampLines` unset for it.

It also records the name trade-off you identified, and why it was chosen rather than fallen into: `aria-label` replaces name-from-content deliberately, because otherwise the control's accessible name is the entire message text, which says nothing about what activating it does. The body stays in the accessibility tree and reads normally in browse mode.

On the dedicated control: agreed that a separate Expand/Collapse button is the unambiguously correct shape, and I am not arguing otherwise. It is a change to the interaction design that three consumers are currently hand-rolling as whole-bubble click, so it wants its own decision and its own before/after, not a late addition here. The documented caveat makes the current limit visible in the meantime, which is what stops it being an invisible one.

The section also documents `clampLines`, `expandLabel` / `collapseLabel` and the three CSS variables, none of which had reached the docs. The docs-accuracy suite passes (33).

Tara-ag (commented):
Review comments incoming as line-level notes.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -37,6 +40,53 @@
...
+  };
+  // A message body can carry links and buttons. Those have their own job, and the
+  // bubble wrapping them must not turn every such activation into a second,
+  // unasked-for expand or collapse.
+  const isNestedControl = (event: Event) => {
+    const target = event.target;
+    if (!(target instanceof Element)) {
+      return false;
+    }
+    const nested = target.closest(
```
Tara-ag: 💡 MINOR: The nested-control guard misses `[contenteditable]`, `[tabindex]:not([tabindex='-1'])`, `<area href>`, custom elements, and JS-driven click targets. Activating those can still bubble to the bubble and toggle the clamp. Broaden the selector and consider honoring `event.defaultPrevented` in `onClampClick`/`onClampKeydown` so nested handlers that call `preventDefault` do not also expand/collapse.

tests/chat-message-clamp.test.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    await link.click();
+    // The link did its own job; the bubble must be exactly as it was.
+    await expect(bubble).toHaveAttribute('data-clamped', 'true');
+    await expect(bubble).toHaveAttribute('aria-expanded', 'false');
+
+    // The bubble itself still toggles, so the guard has not disabled the control.
+    await bubble.click({ position: { x: 5, y: 5 } });
+    await expect(bubble).toHaveAttribute('aria-expanded', 'true');
+  });
```
Tara-ag: ⚠️ MAJOR: This test is fragile. Programmatic `bubble.focus()` after a single `Tab` may not match `:focus-visible` and may land on an earlier focusable element; `getComputedStyle` outline serialization also varies by browser. Assert that the outline is visible (`outlineStyle !== 'none'` and `outlineWidth > 0`) after keyboard-navigating to the bubble, rather than expecting an exact `'solid 2px'` string.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -136,7 +186,23 @@
         <div class="header">{@render header()}</div>
       {/if}
 
-      <div class="bubble">
+      <!-- The clamped bubble is a disclosure control: role, tabindex and both handlers are
+           gated on the same condition, so it is only focusable while it is genuinely a
+           button. The checker cannot see that through a dynamic role. -->
+      <!-- svelte-ignore a11y_no_noninteractive_tabindex -->
```
Tara-ag: ⚠️ MAJOR: The bubble becomes `role="button"` while its `.body` child can render paragraphs, lists, headings, and other structural markup from `content`/`html`/`markdown`/`body`. WAI-ARIA prohibits buttons from containing structural descendants; screen readers may flatten or misrepresent the message. The docs warn about nested interactive content, but this also affects the common non-interactive markup produced by `html`/`markdown`. Consider separating the disclosure control from the message body, or clearly documenting that `clampLines` must not be used with structured markup.

Tara-ag (commented):
Reviewed file by file. The feature is additive and backward-compatible, but there are accessibility and test-quality concerns that should be addressed before merge.

Tara-ag (commented):
Review in progress — posting findings for the accepted issues.

src/lib/ChatMessage/ChatMessage.svelte
murdore: Both taken in `62c1456`.

The selector now keys on `[tabindex]:not([tabindex="-1"])` as well as a tag list, which is the part that generalises: custom elements and anything a consumer made focusable carry one, so it no longer depends on enumerating tags I happened to think of. Added `area[href]`, `label`, `summary`, `audio[controls]`, `video[controls]`, `[contenteditable]:not([contenteditable="false"])`, and the `checkbox` / `menuitem` / `tab` / `switch` / `textbox` roles alongside `button` / `link`.

`defaultPrevented` is the better of the two suggestions and I would not have got there — a nested handler that called `preventDefault` has already claimed the event, and that covers JS-driven targets that match no selector at all. Both handlers now go through:

```ts
const shouldToggleFrom = (event: Event) =>
  clampable && !event.defaultPrevented && !isNestedControl(event);
```

Note the `nested !== event.currentTarget` check is load-bearing now in a way it was not before: the bubble carries `tabindex="0"` and `role="button"`, so it matches two of its own selectors and without that check the control would disable itself entirely.

tests/chat-message-clamp.test.ts
murdore: Hardened in `62c1456`, though the measurement disagrees with part of the diagnosis.

`:focus-visible` does match here — the test now asserts that directly rather than inferring it:

```ts
expect(ring.focusVisible).toBe(true);   // element.matches(':focus-visible')
expect(ring.style).toBe('solid');
expect(ring.width).toBeGreaterThan(0);
```

It passes, so the Tab-then-`focus()` sequence does produce a `:focus-visible` match. Asserting it explicitly is the real improvement: a future failure now says which half broke instead of only that the outline was wrong.

On landing elsewhere — `bubble.focus()` targets the bubble directly, and Playwright fails the call if the locator does not resolve, so the Tab is only there to make the subsequent focus read as keyboard-driven.

Kept the width as a parsed number rather than a string, which removes the serialization concern. Did **not** loosen the style to `!== 'none'`: the UA ring is `outline-style: auto` and ours is `solid`, so `solid` is precisely what distinguishes the component's rule from the browser default. A looser assertion would pass on the UA ring too, which would make the test unable to fail for the reason it exists. This project runs one browser project (`functional`, Chromium), so there is no cross-browser variance to absorb.

tests/chat-message-clamp.test.ts
```diff
@@ -0,0 +1,155 @@
...
```
Tara-ag: ⚠️ MAJOR: This focus-ring test is fragile. A single `Tab` may land on an earlier focusable element (e.g., the markdown link in the first demo), and `bubble.focus()` is programmatic focus that does not reliably match `:focus-visible`. Drive focus to the bubble with repeated `Tab` presses and assert `:focus-visible` before reading the outline, or test the CSS rule presence directly.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -136,7 +195,23 @@
         <div class="header">{@render header()}</div>
```
Tara-ag, line 170: ⚠️ MAJOR: `role="button"` on `.bubble` wraps structural content (`<p>`, `<ul>`, `<h1>`–`<h6>`) rendered by `effectiveHtml`/`content`/`body`, which violates WAI-ARIA (button descendants must be phrasing content). Screen readers will flatten or misannounce the message. Consider making the expand control a separate element rather than putting the button role on the message container, or provide a prop to disable the button role for rich-content messages.

Tara-ag (changes requested):
Two MAJOR issues need addressing before merge:

1. **Accessibility**: placing `role="button"` on `.bubble` violates ARIA because the bubble's `.body` can contain block-level structural elements (`<p>`, `<ul>`, headings, etc.) from `content`/`html`/`markdown`/`body`. Button descendants must be phrasing content; structural descendants will be flattened or misannounced by screen readers. The docs acknowledge the problem for nested interactive content but not for structural content, which is equally invalid.

2. **Test fragility**: the focus-ring test uses a single `Tab` followed by programmatic `bubble.focus()`, then asserts `:focus-visible` matching and a solid outline. Programmatic focus does not reliably match `:focus-visible`, and the lone `Tab` may land on an earlier focusable element (e.g., the inner markdown link), making the assertion flaky across browsers and demo-order changes.

Both are actionable and have inline comments with suggested fixes.

Tara-ag (commented):
Reviewed file-by-file. Two MAJOR findings remain:

1. **Accessibility** (`src/lib/ChatMessage/ChatMessage.svelte:195`): `role="button"` on the bubble container violates WAI-ARIA when the message body carries structural elements (`<p>`, `<ul>`, headings, etc.) from `content`/`html`/`markdown`/`body`. The docs now warn about nested interactive content but not structural content, which is equally prohibited. The nested-control guard and `aria-label` improvements are good, but the invalid nesting itself remains.

2. **Test fragility** (`tests/chat-message-clamp.test.ts:118`): The focus-ring test relies on a single `Tab` followed by programmatic `bubble.focus()`, which may land on an earlier focusable element and does not reliably match `:focus-visible`. It should drive focus to the bubble with repeated Tab presses and assert `:focus-visible` before reading outline, or test the CSS rule directly.

Both have inline comments with concrete fixes. Once addressed, this is good to merge.

src/lib/ChatMessage/ChatMessage.svelte
murdore: You were right, and the fix in `78d6085` goes further than the documentation option I proposed on the previous thread — I was wrong to stop there.

What changed my mind is that this hits the common case, not an edge one. `markdown` is a first-class prop on this component, and the consumer this feature was extracted from renders `{@html renderMarkdown(message.text)}` through the `body` snippet. So "avoid structured markup with `clampLines`" would have forbidden the exact usage the feature exists for. Documenting that would have been writing a warning instead of a fix.

**The bubble is no longer a button.** The disclosure control is this library's own `Button`, rendered below the clamped body:

```svelte
{#if clampable}
  <div class="clamp-toggle">
    <Button
      variant="ghost"
      size="sm"
      onclick={toggleExpanded}
      ariaExpanded={expanded}
      text={expanded ? collapseLabel : expandLabel}
    />
  </div>
{/if}
```

`Button` already exposed `ariaExpanded`, so the control carries name, state, focus order and native keyboard activation without this component hand-rolling any of it — and `ChatMessage` was already importing `Button` for its copy/retry/feedback actions, so this is the same primitive those use.

Consequences, stated rather than implied:

- Presentational-children flattening is gone entirely. Paragraphs, lists and headings from `markdown` / `html` / `body` keep their semantics.
- Both earlier findings on this axis are resolved by construction rather than by guard — there is no `role="button"` for interactive or structural descendants to be invalid inside.
- The bubble keeps its click handler as a pointer convenience, with no role and no tabindex.
- The nested-control guard became load-bearing rather than defensive: the Button's own click reaches the bubble handler too, so without the guard every toggle fires twice and nets zero. Negative control proves it — dropping the guard gives `Expected "true", Received "false"`.
- The `:focus-visible` rule from the previous round is deleted. The bubble is not focusable any more, so it was dead code; `Button` owns focus styling.

New assertions, each negative-controlled: the bubble is not `role="button"` and not focusable; the Button is, is named, and reports `aria-expanded`; Enter and Space on it work; clicking it toggles once. Restoring `role="button"` on the bubble fails with `Expected not "button", Received "button"`.

Visual baseline regenerated — the toggle is a visible control now, so the demo page legitimately changed. 93 visual, 485 functional, 413 unit, 33 docs-accuracy, lint 0, check 0.

tests/chat-message-clamp.test.ts
murdore: Obsolete as of `78d6085` — that test no longer exists.

The bubble is not focusable any more: the disclosure control is a library `Button`, so the component-declared focus ring was dead code and both the rule and the test asserting it are deleted. `Button` owns its focus styling and is covered by its own suite.

For the record on the earlier version, since the same concern was raised twice: the test had been hardened to assert `element.matches(':focus-visible')` directly, and it passed — so the Tab-then-focus sequence did produce a `:focus-visible` match. The point is moot now either way.

src/lib/ChatMessage/ChatMessage.svelte
murdore: Fixed in `78d6085`, which landed after this run started — `role="button"` is gone from `.bubble` entirely.

The disclosure control is now this library's `Button`, rendered below the clamped body with `ariaExpanded`, so there is no button role for structural content to be flattened inside. The bubble keeps only a click handler as a pointer convenience, with no role and no tabindex.

Pinned by a test that asserts the absence rather than trusting it: reinstating `role="button"` on the bubble fails with `Expected not "button", Received "button"`.

Detail in the reply on the earlier thread for the same finding.

Tara-ag (commented):

src/wc/components/ChatMessage.wc.svelte
```diff
@@ -9,6 +9,7 @@
       markdown: { type: 'String' },
       body: { type: 'Object' },
       streaming: { type: 'Boolean', reflect: true },
+      clampLines: { type: 'Number', attribute: 'clamp-lines' },
```
Tara-ag, line 12: 💡 MINOR: `clampLines` is declared as a `Number` prop with a kebab-case attribute but `reflect` is not enabled. Because the prop can be withdrawn and restored by a parent (see the demo toggle), consider whether web-component consumers need the attribute to stay in sync. If so, add `reflect: true`; otherwise document the one-way behavior in the Web Component section of `docs/ChatMessage.md`.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -136,7 +194,21 @@
...
```
Tara-ag, line 170: ⚠️ MAJOR: The clamped bubble is clickable but has no keyboard equivalent or visible focus indicator. The `onclick` handler plus `cursor: pointer` makes this a pointer-only control, while the deliberate avoidance of `role="button"` (to preserve child semantics) means it is unreachable by keyboard and receives no focus ring. Either remove the bubble click affordance entirely and rely on the real `Button` below, or add `tabindex="0"`, an `onkeydown` handler for Enter/Space, and a `:focus-visible` rule that consumes the documented `--chat-message-clamp-focus-outline` / `--chat-message-clamp-focus-outline-offset` tokens.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -37,6 +40,61 @@
...
+  // Consumers had been rebuilding this around the `body` snippet; it belongs here so
+  // the chrome, the clamp and the expanded state stay one component's concern.
+  let expanded = $state(false);
+  let clampable = $derived(typeof clampLines === 'number' && clampLines > 0);
+  let clampActive = $derived(clampable && !expanded);
+  // Turning clamping off must not leave the message stuck open: without this, a parent
+  // that sets clampLines to 0 and later restores it would get an already-expanded bubble
+  // instead of the clamped default. A prop going away cannot be observed from a $derived
+  // without writing to state, so this is the sanctioned escape hatch rather than a
+  // reactive convenience.
```
Tara-ag, line 53: 💡 MINOR: Writing to `expanded` inside a `$effect` is a reactive anti-pattern. Consider deriving the effective expanded state instead, e.g. `let effectiveExpanded = $derived(clampable && expanded);`, and use that for `data-expanded` and the toggle button label. That removes the need for the `eslint-disable-next-line no-restricted-syntax` comment and the state mutation in an effect.

tests/chat-message-clamp.test.ts
```diff
@@ -0,0 +1,140 @@
+import { expect, test } from '@playwright/test';
+
+// `clampLines` collapses the rendered body and makes the bubble its own control. The
+// behaviour was being rebuilt by consumers around the `body` snippet, which meant each
+// one re-derived the keyboard handling and the clamp CSS; the risk in moving it here is
+// that the clamp stops actually hiding anything, so the assertions are about measured
+// height and the exposed state rather than about the class names that produce them.
+test.describe('ChatMessage — clampLines', () => {
+  test('collapses a long message and expands it by click and by keyboard', async ({ page }) => {
+    await page.goto('/components/chat-message');
```
Tara-ag, line 10: 💡 MINOR: The test navigates to the demo page and selects via `data-pw` hooks defined there, coupling the test to demo markup. If the demo page is refactored, coverage can silently drop. Prefer an isolated component harness or a dedicated `/tests/*` route for behavioral assertions, keeping the demo page for visual/manual verification.

Tara-ag (changes requested):
Review complete. One MAJOR accessibility concern and three MINOR suggestions are noted inline. The MAJOR issue should be addressed before merge: the clamped bubble is currently a pointer-only control with no keyboard equivalent or visible focus indicator, despite the documented focus-outline CSS variables.

Tara-ag (changes requested):
The clamp/expand feature is well-scoped and the implementation mostly avoids common pitfalls (copy text preserved, nested controls guarded, CSS variable override supported). However, there are two MAJOR issues that need to be addressed before merge:

1. **Keyboard accessibility of the clickable bubble**: the `.bubble` div has `cursor: pointer` and an `onclick` handler when clampable, but no `tabindex`, `role`, or keyboard activation. This leaves keyboard-only users unable to use the convenience affordance. Either make it fully focusable/keyboard-activatable or remove the bubble-click behavior.

2. **State mutation inside `$effect`**: resetting `expanded` when `clampable` becomes false is done via an effect write. This is a reactive anti-pattern in Svelte 5 and can cause timing/stale-state issues. Derive or reset this more explicitly.

Two MINOR items are also noted: the expand/collapse `Button` should use `aria-controls` to relate to the body, and the Playwright tests should not depend on demo-page `data-pw` selectors.

Decision: **CHANGES_REQUESTED**.

Tara-ag (changes requested):
Review complete. All accepted findings have been posted as inline comments. The PR introduces a useful clamp/expand feature but has outstanding MAJOR accessibility and reactivity concerns that should be addressed before merge.

murdore:
## Addressing the BLOCKED verdict — `b89f72b`

Three MAJORs, which are two distinct issues.

### The bubble was a pointer-only control — fixed

Findings at `:194` and `:197` are the same one, and they are right. The previous revision let a click anywhere on the bubble toggle, as a convenience layered on top of the Button. That is worse than it looks: a click target that is not focusable is a control a keyboard user cannot reach, and it was the reason a nested-activation guard existed at all — without it the Button’s own click reached the bubble handler too and toggled twice.

Both are gone. The `Button` is now the only control, for every input method; the bubble is inert markup with no role, no handler and no `cursor: pointer`. That also deleted `isNestedControl` / `shouldToggleFrom` and both `svelte-ignore` suppressions — roughly 20 lines of machinery that only existed to make a non-standard affordance behave.

Negative control: restoring `role="button"` on the bubble fails with `Expected not "button", Received "button"`.

### `$effect` writing state — keeping it, and here is the proof rather than an opinion

The suggested alternative is to derive the effective state from `clampable` and `expanded` together. **That cannot express this requirement**, and it is worth showing why rather than asserting it.

The requirement is that expansion does not survive `clampLines` going `4 → 0 → 4`. Any `$derived` is a pure function of current state. At the first `4` and the second `4` every input is identical — same `clampLines`, same `clampable`, same stored toggle state. A pure function must return the same answer to both. Keying on the value does not help:

```js
let expandedFor = $state(0);
let expanded = $derived(clampable && expandedFor === clampLines);
// 4 -> 0 -> 4: expandedFor is 4, clampLines is 4 again, so this is true. Unchanged.
```

Distinguishing them requires observing the *transition*, which requires writing state, which is what an effect is. There is no formulation without one.

It is also the sanctioned pattern here rather than a novelty: `$effect` is banned by `no-restricted-syntax` and used **16 times** in `src/lib` with an explicit disable and a justification — `Input`, `Toast`, `Menu`, `Select`, `MediaPlayer`, `LottiePlayer`, `TypewriterText` and six charts. `Menu`’s comment calls it “the sanctioned reactive escape hatch.” This one follows that form.

On the flash concern: the effect writes `expanded = false` only when `clampable` is already false, and while `clampable` is false the clamp is off and no control renders, so nothing observable depends on `expanded` in that state. Covered by a test that drives a real round trip through a toggle.

If the project would rather drop the reset than carry the effect, that is a reasonable call and a one-line deletion — but it reintroduces the behaviour CodeRabbit opened this PR’s review by flagging, so it needs to be a deliberate choice rather than a silent one.

### Minors

- **`aria-controls`** — added. `Button` gains an `ariaControls` prop next to the `ariaLabel` / `ariaExpanded` it already had, so this is reusable rather than a one-off; the id comes from `$props.id()`, as `Accordion` and `Input` do. Test resolves the attribute and asserts it lands on the clamped body. Negative control: removing it gives `Received null`.
- **`reflect` on `clamp-lines`** — not changed. The finding itself notes it is consistent with the file; `Gauge` reflects `value` and not `max`, so there is no convention to follow, and property→attribute reflection is a separate decision for the wrapper as a whole.
- **Demo-page selectors** — declined, with the measurement on the original thread: 102 of 108 test files here drive `/components/*`, and there is no `/tests/` route tree. The demo route is this library’s harness. Note the bootstrap step that derives “repo standards” timed out this run (`litellm generate operation timed out after 5m`), so that standard was not read from the repository.
- **`prefers-reduced-motion`** — not added. There is no transition on the clamp; a guard for a hypothetical future one would be a rule with nothing to govern.

### Verification

lint 0 · check 0 errors · 413 unit · 485 functional · 93 visual · 33 docs-accuracy · wc prop-parity 90

docs/ChatMessage.md
```diff
@@ -65,58 +67,70 @@ type ChatRole = ChatParty | 'user' | 'assistant' | 'system' | (string & {});
...
+| Variable                                       | Default                   | CSS Property   | Description                                             |
+| ---------------------------------------------- | ------------------------- | -------------- | ------------------------------------------------------- |
+| `--chat-message-max-width`                     | `82%`                     | max-width      | Max width of the message.                               |
+| `--chat-message-margin`                        | `0`                       | margin         | Margin around the message.                              |
+| `--chat-message-gap`                           | `10px`                    | gap            | Gap between avatar and bubble.                          |
+| `--chat-message-content-gap`                   | `6px`                     | gap            | Gap between header, bubble, attachments.                |
+| `--chat-message-header-font-size`              | `0.75rem`                 | font-size      | Header row font size.                                   |
```
Tara-ag: 💡 MINOR: The CSS variables table lists `--chat-message-clamp-focus-outline` and `--chat-message-clamp-focus-outline-offset`, but no rule in ChatMessage.svelte consumes them. Since the bubble is intentionally not focusable, these tokens cannot take effect and will mislead themers. Either wire them to the real `.clamp-toggle :global(button):focus-visible` rule or remove the rows.

src/lib/ChatMessage/properties.ts
```diff
@@ -33,6 +33,16 @@ export type OptionalChatMessageProperties = {
    * action still has something to copy.
    */
   body?: Snippet | null;
+  /**
```
Tara-ag, line 36: 💡 MINOR: The `clampLines` JSDoc still says "Clicking the bubble does the same", but the implementation deliberately makes the bubble inert (no `role`, no `onclick`) and the docs explicitly state "The bubble itself is deliberately not a button." Remove the stale sentence so the JSDoc matches the actual interaction model.

docs/ChatMessage.md
```diff
@@ -65,58 +67,70 @@ type ChatRole = ChatParty | 'user' | 'assistant' | 'system' | (string & {});
...
-| `--chat-message-action-hover-background-color` | `#f4f4f5`                 | background    | Action button hover background.                         |
+| Variable                                       | Default                   | CSS Property   | Description                                             |
+| ---------------------------------------------- | ------------------------- | -------------- | ------------------------------------------------------- |
+| `--chat-message-max-width`                     | `82%`                     | max-width      | Max width of the message.                               |
+| `--chat-message-margin`                        | `0`                       | margin         | Margin around the message.                              |
+| `--chat-message-gap`                           | `10px`                    | gap            | Gap between avatar and bubble.                          |
+| `--chat-message-content-gap`                   | `6px`                     | gap            | Gap between header, bubble, attachments.                |
```
Tara-ag: 💡 MINOR: Documented clamp focus CSS variables are not implemented. ChatMessage.svelte never consumes `--chat-message-clamp-focus-outline` or `--chat-message-clamp-focus-outline-offset`, and the bubble is intentionally not focusable. Either implement these tokens on the toggle button's `:focus-visible` state or remove these rows so the docs stay synchronized with the implementation.

Tara-ag (approved):
Approved with one minor documentation sync note: two CSS custom properties are documented but not implemented in the component styles.

Tara-ag (commented):

src/wc/components/ChatMessage.wc.svelte
Tara-ag, line 12: 💡 MINOR: `clampLines` web-component prop does not reflect its attribute. It is declared with `attribute: 'clamp-lines'` but `reflect: true` is omitted. Since the component supports dynamically withdrawing/restoring `clampLines`, web-component consumers may expect the attribute to stay in sync with the property. Add `reflect: true`, or document the one-way behavior in the Web Component section of `docs/ChatMessage.md`.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -37,6 +40,37 @@
...
```
Tara-ag, line 53: 💡 MINOR: `$effect` used to reset `expanded` state violates the repo's `no-restricted-syntax` ban on `$effect`. The inline `eslint-disable-next-line` confirms the rule is being bypassed. Derive the effective state instead, e.g. `let effectiveExpanded = $derived(clampable && expanded);`, and use that for `data-expanded` and the toggle button label, avoiding the reactive anti-pattern and the disable comment.

src/lib/ChatMessage/properties.ts
Tara-ag, line 36: 💡 MINOR: Stale JSDoc says clicking the bubble toggles expansion. The current implementation deliberately makes the bubble inert (no `role`, no `tabindex`, no `onclick`), and `docs/ChatMessage.md` explicitly states "The bubble itself is deliberately not a button." Remove the sentence "Clicking the bubble does the same." so the JSDoc matches the actual interaction model where only the `Button` toggles.

Tara-ag (commented):
Review complete. The feature is sound and safe to merge; three MINOR doc/reactivity/web-component parity issues noted inline.

murdore:
Note for anyone reading the pinned verdict comment above: **it is stale.** It was last written at `02:10:46Z` by the run against `78d6085`.

The run against `b89f72b` decided `APPROVED` and the check is green:

```
Decision: APPROVED
Yama decision: APPROVED (action outcome: success)
Yama review passed (decision=APPROVED).
```

The workflow only writes the summary comment on a non-approving verdict — the `Post verdict summary comment` step reports `outcome=skipped` on this run — so the BLOCKED text stays on the page after the findings behind it are fixed. Worth knowing before treating that comment as current state; the check status is the live signal.

docs/ChatMessage.md
murdore: Fixed in `7926884`. Genuinely my mistake, and an instructive one: when I deleted the focus rule I updated the docs with a `str.replace` that had no assertion on it, the anchor had already shifted, and it silently replaced nothing. The rows stayed while the CSS they described was gone.

The table now lists only `--chat-message-clamp-lines` and `--chat-message-clamp-toggle-margin-top`, which are the two the component actually reads, and I checked it mechanically rather than by eye:

```
$ for v in $(grep -o "`--chat-message-clamp[a-z-]*`" docs/ChatMessage.md | tr -d "`" | sort -u); do
    grep -q -- "$v" src/lib/ChatMessage/ChatMessage.svelte && echo "OK $v" || echo "DANGLING $v"; done
  OK   --chat-message-clamp-lines
  OK   --chat-message-clamp-toggle-margin-top
```

docs/ChatMessage.md
murdore: Fixed in `7926884`. Genuinely my mistake, and an instructive one: when I deleted the focus rule I updated the docs with a `str.replace` that had no assertion on it, the anchor had already shifted, and it silently replaced nothing. The rows stayed while the CSS they described was gone.

The table now lists only `--chat-message-clamp-lines` and `--chat-message-clamp-toggle-margin-top`, which are the two the component actually reads, and I checked it mechanically rather than by eye:

```
$ for v in $(grep -o "`--chat-message-clamp[a-z-]*`" docs/ChatMessage.md | tr -d "`" | sort -u); do
    grep -q -- "$v" src/lib/ChatMessage/ChatMessage.svelte && echo "OK $v" || echo "DANGLING $v"; done
  OK   --chat-message-clamp-lines
  OK   --chat-message-clamp-toggle-margin-top
```

src/lib/ChatMessage/properties.ts
murdore: Fixed in `7926884` — the sentence now reads "that button is the only control".

Same root cause as the CSS-variable drift on the other thread: I removed the whole-bubble click and updated `docs/ChatMessage.md`, but the JSDoc in `properties.ts` says the same thing in different words and I only corrected one of the two. Verified by grep across both files that no "clicking the bubble" claim survives anywhere.

murdore: Fixed in `7926884` — the sentence now reads "that button is the only control".

Same root cause as the CSS-variable drift on the other thread: I removed the whole-bubble click and updated `docs/ChatMessage.md`, but the JSDoc in `properties.ts` says the same thing in different words and I only corrected one of the two. Verified by grep across both files that no "clicking the bubble" claim survives anywhere.

src/lib/ChatMessage/ChatMessage.svelte
murdore: Answered in detail in the [verdict response comment](https://github.com/juspay/svelte-ui-components/pull/533#issuecomment-5556329582) — keeping the effect, with the reason rather than a preference.

Short version: the suggested alternative cannot express the requirement. Expansion must not survive `clampLines` going `4 -> 0 -> 4`, and at the first `4` and the second `4` every input to a `$derived` is identical, so a pure function must answer both the same way. Keying on the value does not help either:

```js
let expandedFor = $state(0);
let expanded = $derived(clampable && expandedFor === clampLines);
// 4 -> 0 -> 4: expandedFor is 4, clampLines is 4 again. Still expanded.
```

Distinguishing them means observing the transition, which means writing state. It is also the pattern this library already sanctions 16 times in `src/lib` with the same disable and a justification — `Menu` calls it "the sanctioned reactive escape hatch".

If the project would rather drop the reset than carry the effect, that is a one-line deletion and a fair call — but it restores the behaviour this review opened by flagging, so it should be deliberate.

src/lib/ChatMessage/ChatMessage.svelte
murdore: Answered in detail in the [verdict response comment](https://github.com/juspay/svelte-ui-components/pull/533#issuecomment-5556329582) — keeping the effect, with the reason rather than a preference.

Short version: the suggested alternative cannot express the requirement. Expansion must not survive `clampLines` going `4 -> 0 -> 4`, and at the first `4` and the second `4` every input to a `$derived` is identical, so a pure function must answer both the same way. Keying on the value does not help either:

```js
let expandedFor = $state(0);
let expanded = $derived(clampable && expandedFor === clampLines);
// 4 -> 0 -> 4: expandedFor is 4, clampLines is 4 again. Still expanded.
```

Distinguishing them means observing the transition, which means writing state. It is also the pattern this library already sanctions 16 times in `src/lib` with the same disable and a justification — `Menu` calls it "the sanctioned reactive escape hatch".

If the project would rather drop the reset than carry the effect, that is a one-line deletion and a fair call — but it restores the behaviour this review opened by flagging, so it should be deliberate.

src/wc/components/ChatMessage.wc.svelte
murdore: Not changing this, and the finding half-agrees: it notes the declaration is consistent with the rest of the file.

There is no convention to follow here — `Gauge` reflects `value` and not `max`, and most numeric wrapper props do not reflect. Property-to-attribute reflection changes what a wrapper writes back to the DOM for every consumer, so turning it on for one prop in one component is the kind of local inconsistency that makes the next person guess. If the wrappers should reflect numeric props, that is worth doing across the file in its own change.

Nothing here depends on it: the demo drives `clampLines` as a property and the round-trip test passes, since Svelte tracks the property whether or not it is mirrored to an attribute.

murdore: Not changing this, and the finding half-agrees: it notes the declaration is consistent with the rest of the file.

There is no convention to follow here — `Gauge` reflects `value` and not `max`, and most numeric wrapper props do not reflect. Property-to-attribute reflection changes what a wrapper writes back to the DOM for every consumer, so turning it on for one prop in one component is the kind of local inconsistency that makes the next person guess. If the wrappers should reflect numeric props, that is worth doing across the file in its own change.

Nothing here depends on it: the demo drives `clampLines` as a property and the round-trip test passes, since Svelte tracks the property whether or not it is mirrored to an attribute.

src/lib/ChatMessage/ChatMessage.svelte
murdore: Already fixed, before this comment was written — see `b89f72b`.

The bubble has no click handler, no role, no tabindex and no `cursor: pointer`. The `Button` is the only control, which is exactly what this finding asks for. The `svelte-ignore` suppressions that made the pointer-only version possible are deleted along with it.

Pinned by a test asserting the absence: reinstating `role="button"` on the bubble fails with `Expected not "button", Received "button"`.

tests/chat-message-clamp.test.ts
murdore: Same answer as the earlier threads on this, with the measurement rather than an opinion:

```
$ grep -rl "goto('/components/" tests/*.ts | wc -l   ->  102
$ ls tests/*.ts | wc -l                            ->  108
$ ls src/routes/                                   ->  components  (only)
```

There is no `/tests/` route tree in this repository, and the demo pages under `/components/*` are the harness. An isolated route for this one component would be a pattern used by exactly one file.

Worth noting for this run specifically: the step that derives "repo standards" timed out (`litellm generate operation timed out after 5m`), so the convention cited here was not read from the repository.

The assertions themselves are independent of demo *structure* — measured height, computed `-webkit-line-clamp`, ARIA state — reached through `testId`s passed to the component, and each is negative-controlled.

murdore: Same answer as the earlier threads on this, with the measurement rather than an opinion:

```
$ grep -rl "goto('/components/" tests/*.ts | wc -l   ->  102
$ ls tests/*.ts | wc -l                            ->  108
$ ls src/routes/                                   ->  components  (only)
```

There is no `/tests/` route tree in this repository, and the demo pages under `/components/*` are the harness. An isolated route for this one component would be a pattern used by exactly one file.

Worth noting for this run specifically: the step that derives "repo standards" timed out (`litellm generate operation timed out after 5m`), so the convention cited here was not read from the repository.

The assertions themselves are independent of demo *structure* — measured height, computed `-webkit-line-clamp`, ARIA state — reached through `testId`s passed to the component, and each is negative-controlled.

tests/chat-message-clamp.test.ts
murdore: Same answer as the earlier threads on this, with the measurement rather than an opinion:

```
$ grep -rl "goto('/components/" tests/*.ts | wc -l   ->  102
$ ls tests/*.ts | wc -l                            ->  108
$ ls src/routes/                                   ->  components  (only)
```

There is no `/tests/` route tree in this repository, and the demo pages under `/components/*` are the harness. An isolated route for this one component would be a pattern used by exactly one file.

Worth noting for this run specifically: the step that derives "repo standards" timed out (`litellm generate operation timed out after 5m`), so the convention cited here was not read from the repository.

The assertions themselves are independent of demo *structure* — measured height, computed `-webkit-line-clamp`, ARIA state — reached through `testId`s passed to the component, and each is negative-controlled.

Tara-ag (approved):
Reviewed file-by-file. The new `clampLines` feature is additive, the web-component wrappers map the new props (including `aria-controls` on Button), accessibility is handled via a real disclosure Button with `aria-expanded`/`aria-controls`, and the Playwright tests cover clamping, expansion, keyboard activation, link isolation, and the CSS-token override. No blocking issues found.

Tara-ag (changes requested):
Correcting my earlier approval: file-by-file review found a CRITICAL web-component prop parity gap and a MAJOR accessibility/docs mismatch that need to be addressed before merge.

Tara-ag (approved):
Correction to my earlier review: upon re-reading the actual branch contents, the concerns I raised were unfounded. `src/wc/components/Button.wc.svelte` already declares `ariaControls` with `attribute: 'aria-controls'`, `ChatMessage.svelte` already wires `ariaControls={bodyId}` to the toggle Button and assigns `id={bodyId}` to the body, and `tests/chat-message-clamp.test.ts` already asserts the `aria-controls`/`id` relationship. The implementation is sound: the bubble stays a non-interactive container, the toggle is a real Button with proper disclosure semantics, clamping is covered by Playwright tests including keyboard/link/CSS-override paths, and the web-component wrappers map the new props. Approving.

Tara-ag (commented):
Reviewed the latest revision. The MAJOR accessibility and API-parity issues from earlier rounds are resolved: the bubble is inert, the toggle is a real Button with `aria-expanded`/`aria-controls`, and the web-component wrapper exposes `clampLines`. Three MINOR items remain; see inline comments.

src/wc/components/ChatMessage.wc.svelte
Tara-ag, line 12: 💡 MINOR: Web-component `clampLines` prop does not reflect to its attribute. In the custom-element wrapper `clampLines` is declared with `attribute: 'clamp-lines'` but `reflect: true` is omitted. Because the Svelte component supports dynamically withdrawing and restoring `clampLines` (demo toggle), web-component consumers who inspect the attribute will see it become stale after a property change.

Suggestion: Add `reflect: true` to keep the attribute in sync with the property, or document the one-way behavior in the Web Component section of `docs/ChatMessage.md`.

src/lib/ChatMessage/ChatMessage.svelte
Tara-ag, line 53: 💡 MINOR: `$effect` mutates local component state. The inline `// eslint-disable-next-line no-restricted-syntax` confirms the repo normally bans this rune. While the author argues this transition cannot be expressed with `$derived`, the pattern still couples state mutation to the effect lifecycle and is the only place in this component that writes state reactively.

Suggestion: Add a short code comment explaining *why* `$derived` cannot satisfy the `4 -> 0 -> 4` reset requirement, or extract the reset into an explicit helper whose side-effect nature is obvious. If the project prefers to drop the reset, delete the effect and document that expansion survives clamping being withdrawn.

tests/chat-message-clamp.test.ts
```diff
@@ -0,0 +1,143 @@
...
```
Tara-ag, line 10: 💡 MINOR: Playwright tests are coupled to demo-page markup. The new clamp tests navigate to `/components/chat-message` and select elements through `data-pw` attributes defined in the demo page. If the demo page is refactored or the demo cases are renamed, coverage can silently drop even though the component itself still works.

Suggestion: Render the component in an isolated test harness or dedicated `/tests/*` route, and pass stable `testId` props to the component itself rather than selecting through demo-page wrappers.

Tara-ag (commented):
Reviewed the latest revision. The earlier MAJOR accessibility and API-parity issues are resolved: the bubble is inert markup, the expand/collapse control is a real Button with `aria-expanded`/`aria-controls`, and the web-component wrapper exposes `clampLines`. Three MINOR items remain; see inline comments.

murdore:
## The BLOCKED verdict on `7926884` is reviewing content that is not on this branch

The CRITICAL and both MAJORs describe code that does not match the commit they were run against. Verified against GitHub's own copy of `7926884` rather than my working tree:

```
$ gh api -H "Accept: application/vnd.github.raw" \
    repos/juspay/svelte-ui-components/contents/src/wc/components/Button.wc.svelte?ref=7926884 \
  | grep -n ariaControls
16:      ariaControls: { type: 'String', attribute: 'aria-controls' },

$ ... contents/src/lib/ChatMessage/ChatMessage.svelte?ref=7926884 | grep -n ariaControls
200:              ariaControls={bodyId}

$ ... contents/tests/chat-message-clamp.test.ts?ref=7926884 | grep -n aria-controls
96:    const controls = await toggle.getAttribute('aria-controls');
```

Point by point:

- 🔒 **CRITICAL — "`ariaControls` not declared in the web-component wrapper."** It is declared, at `Button.wc.svelte:16`. This is also the one finding the repository checks mechanically: `scripts/wc-parity/prop-parity.test.ts` fails when a wrapper omits a prop its component declares — it is what caught my earlier omission of `clampLines` / `expandLabel` / `collapseLabel`, failing the `checks` job with `ChatMessage.wc.svelte is missing: clampLines, expandLabel, collapseLabel`. It passes on this commit (90 tests), and `checks` is green on this exact sha.
- ⚠️ **MAJOR — "ChatMessage.svelte never renders `aria-controls`; `bodyId` is unused."** `bodyId` is used four times: on each of the three `.body` branches and as `ariaControls` on the Button at `:200`.
- ⚠️ **MAJOR — "no test asserts `aria-controls`."** `tests/chat-message-clamp.test.ts:96` reads the attribute and asserts it resolves to the clamped body. Negative control: removing `ariaControls={bodyId}` fails it with `Received null`.

The summary text contradicts its own findings, which is the tell — it opens by stating that "the expand/collapse control is a real Button with aria-expanded/**aria-controls**" and that "None block merge", then lists a CRITICAL saying `aria-controls` is missing.

Worth noting the previous run logged `[NEUROLINK:ERROR] Generate failed for litellm: litellm generate operation timed out after 5m` in its repo-standards step, and this verdict cites "bootstrapped standards" as its authority for the CRITICAL.

I am re-running the review job rather than changing code to satisfy findings that describe code that is not there. If it reproduces, the specific claims above are falsifiable with the three commands at the top — please check them before treating the verdict as a code judgement.

### Genuinely outstanding, and answered on their threads

- `$effect` for the reset — kept, with a proof that the `$derived` alternative cannot express the requirement (the `4 → 0 → 4` transition is invisible to a pure function of current props). 16 sanctioned precedents in `src/lib`.
- Demo-page selectors — declined; 102 of 108 test files here drive `/components/*` and there is no `/tests/` route tree.
- `reflect` on `clamp-lines` — declined; no convention to follow (`Gauge` reflects `value`, not `max`), and it is a wrapper-wide decision.

All three are MINOR and none is a defect in the code.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -136,14 +170,38 @@
...
         {#if hasBody}
-          <div class="body">{@render body?.()}</div>
+          <div class="body" id={bodyId}>{@render body?.()}</div>
         {:else if hasHtml}
           <!-- eslint-disable-next-line svelte/no-at-html-tags -->
-          <div class="body">{@html effectiveHtml}</div>
+          <div class="body" id={bodyId}>{@html effectiveHtml}</div>
         {:else if content.length > 0}
-          <div class="body text">{content}</div>
+          <div class="body text" id={bodyId}>{content}</div>
```
Tara-ag, line 190: 💡 MINOR: Expand/collapse toggle can reference a missing body element. When `clampable` is true but the message has no `body`, `html`, or `content`, the clamp toggle Button still renders with `ariaControls={bodyId}` pointing to an element that is never mounted. This creates a dangling `aria-controls` reference and a control that toggles nothing.

Suggestion: Render the toggle only when there is actual body content to clamp. Introduce a derived such as `let hasBodyContent = $derived(hasBody || hasHtml || content.length > 0);` and change the condition to `{#if clampable && hasBodyContent}`.

Tara-ag, line 190: 💬 SUGGESTION: Toggle renders even when content does not overflow. The expand/collapse Button is rendered whenever `clampable` is true, even if the rendered body is shorter than `clampLines`. Activating it then only swaps the label with no visual change, which is confusing for users.

Suggestion: Measure the body element after mount (e.g. `bodyEl.scrollHeight > bodyEl.clientHeight`) and only render or show the toggle when the content actually overflows. Alternatively, expose a prop such as `showExpandToggle` so callers can suppress the button for short messages.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -136,14 +170,38 @@
         <div class="header">{@render header()}</div>
       {/if}
 
-      <div class="bubble">
+      <!-- The bubble deliberately carries no role and no handler of its own. A button has
+           presentational children, so making the bubble one would flatten the paragraphs and
+           lists a message body renders; and a click target that is not focusable is a control
+           keyboard users cannot reach. The Button below is the whole control. -->
+      <div
```
Tara-ag, line 177: 💬 SUGGESTION: Non-integer clampLines values produce invalid CSS. `clampLines` accepts any positive number, including floats like `2.5` or `Infinity`. These values are interpolated directly into `--chat-message-clamp-lines-prop`, which `line-clamp` may ignore or handle inconsistently.

Suggestion: Normalize the prop: treat only finite integers as clampable (`Number.isFinite(clampLines) && clampLines > 0`) and use `Math.floor(clampLines)` in the inline style, e.g. `--chat-message-clamp-lines-prop: ${Math.floor(clampLines)};`.

Tara-ag (commented):
Reviewed file-by-file. No blocking issues found. The change is a clean, additive feature: new optional props with sensible defaults, no public API breakage, proper accessibility via a real `Button` disclosure control, SSR-safe ID generation, and comprehensive tests/docs. Left three non-blocking notes on the toggle rendering edge cases.

murdore:
Re-run result: **the BLOCKED verdict did not reproduce on the same commit.**

```
job 101420159132  attempt=2  sha=7926884  conclusion=success
  success  Yama Review
  success  Enforce verdict          <- exits 1 on BLOCKED; it did not
```

Identical content, opposite verdict. That is the strongest available evidence that the CRITICAL and the two MAJORs were not grounded in the tree — the three commands in my previous comment already showed `ariaControls` present in the wrapper, wired in `ChatMessage.svelte:200` and asserted at `chat-message-clamp.test.ts:96` on that exact sha.

Recording it because a re-run that flips a verdict is easy to read as "the retry fixed it". Nothing was fixed; nothing changed. `7926884` is the same commit both times.

The pinned verdict comment above may still show stale text — the workflow only rewrites it on a non-approving verdict, so an approval leaves the previous BLOCKED summary in place. The check status is the live signal.

All checks green, `CLEAN / MERGEABLE`.

PR #534, opened 2026-09-06
feat(pill): add a semantic tone prop, and stop assuming every pill is clickable

### LINKED ISSUE #520: Pill: A semantic tone/variant prop on Pill (accent/ok/warn/danger/muted), plus letter-spacing/tex…
## Need
A semantic tone/variant prop on Pill (accent/ok/warn/danger/muted), plus letter-spacing/text-transform and data-attribute passthrough

## Closest today
Pill (classes-only styling hook, cursor defaults to pointer even when non-interactive)

## Proposal
Add `tone?: 'accent'|'ok'|'warn'|'danger'|'muted'` mapping to --pill-* background/color pairs, a `--pill-cursor` default of 'default' when no onclick is supplied, and `--pill-letter-spacing`/`--pill-text-transform` CSS vars.

## Where it bites
Found while migrating TARA (a Svelte 5 / SvelteKit 2 consumer pinned to 3.3.0) onto the library. Without this the following consumer files can only migrate part-way and keep hand-rolled markup for the rest: `+page.svelte`, `SessionHeaderBar.svelte`, `TalkStatusPill.svelte`, `Terminal.svelte`.

Related but distinct from #469 (interactive variant) and #464 (dark-theme recipe): this asks for a semantic tone prop on the existing non-interactive Pill.

murdore:
Pill had no way to express a semantic status (accent/ok/warn/danger/muted)
without a consumer hand-rolling its own tone-* class per call site and
repeating the same background/color pair everywhere a status chip was
needed. TARA's own web app does exactly this today: apps/web's
TalkStatusPill.svelte and every Pill status badge in the settings page pass
classes="tara-ui-pill-status-badge tara-ui-pill-status-badge-{accent,ok,warn,
danger,muted}", and packages/ui/src/styles/components.css spells out, in its
own comment, "Pill has no tone prop and ... no letter-spacing or
text-transform variable" as the reason those five classes and their
uppercase/tracking rules exist. This closes that gap at the source. Filed as
#520.

## `tone` prop

`tone?: 'accent' | 'ok' | 'warn' | 'danger' | 'muted'` (`PillTone`, exported
from `src/lib/Pill/properties.ts`). Unset by default — no class added, no CSS
variable read, rendering byte-for-byte unchanged from before this prop
existed. The mapping from tone to CSS class is a pure, unit-tested function
(`pillToneClass` in `src/lib/Pill/pillTone.ts`, covered by
`src/lib/Pill/pill-tone.test.ts`) rather than five inline `class:` directives,
since this repo's vitest suite runs no browser and cannot assert on rendered
DOM.

Each tone sets an internal `--_pill-tone-background` / `--_pill-tone-color`
pair through `--pill-tone-{tone}-background` / `--pill-tone-{tone}-color`
(each with a built-in default, documented in `docs/Pill.md`), read by
`--pill-background` / `--pill-color` as a fallback layer — the same
intermediate-variable precedence Button's `variant` already uses relative to
`--button-color`. An explicit `--pill-background` / `--pill-color`, set
directly or via `classes`, always wins over the tone default, so a one-off
recolor never has to fork the tone.

## `--pill-cursor` no longer implies clickability

The issue named this directly: `--pill-cursor` defaulted to `pointer`
unconditionally, so a purely informational pill (no `onclick`) looked
clickable. The default is now `default`, with `pointer` restored only once
the pill is actually interactive (`role="button"`, set exactly when `onclick`
is supplied). An explicit `--pill-cursor` still wins in both cases.

This changes the rendered cursor for an already-shipped, non-interactive Pill
that relied on the old implicit `pointer` default without setting
`--pill-cursor` itself. It is deliberately released as a minor rather than a
major: the effect is a cursor affordance on pills that never had a working
click handler, and a consumer that wants the old look sets
`--pill-cursor: pointer`. docs/Pill.md documents the change and that
one-line opt-back-in.

## `--pill-letter-spacing` / `--pill-text-transform`

Two new CSS variables (`normal` / `none` defaults, matching every other
already-`normal`/`none`-defaulted Pill property) so a status-chip theme can
set `text-transform: uppercase` without a bespoke class — the exact rule
`.tara-ui-pill-status-badge` above has to apply itself today.

## Also

- `src/wc/components/Pill.wc.svelte`: `tone` added to the custom element's
  declared props (`check:wc-parity`).
- `docs/Pill.md`: new prop row, CSS variable rows, a "Theming with tone"
  section alongside the existing classes-based theming section, and a
  breaking-change callout for the `--pill-cursor` default.
- `src/routes/components/pill/+page.svelte`: a `pill-tone-fixtures` demo
  section exercising every tone, the tone/override precedence, the cursor
  default, and letter-spacing/text-transform.
- `tests/visual/__screenshots__/pill.png`: regenerated for the added demo
  section (980x1092, up from 980x800).

data-attribute passthrough, also named in the issue's "Need" section, is not
part of the "Proposal" this issue actually asked for and is left for a
separate issue.

Closes #520

---

### Verification

**Spec runs on this branch** — `src/lib/Pill/pill-tone.test.ts`, 3 vitest specs, part of a green suite. The Playwright visual baseline for the pill route was regenerated for the added demo section.

**Fail-before** — source reverted to `origin/release` with only the spec kept: **fails** (`pillTone` does not exist there).

`pnpm test:unit`, `pnpm check`, `pnpm lint` and `pnpm build` each exit 0 on this branch; exit codes captured directly, never through a pipe.

### On the `--pill-cursor` default, and the review that flagged it

The review asked for this to be marked `feat(pill)!:` with a `BREAKING CHANGE:` footer, citing Toolbar's 3.0.0 precedent. That is a fair reading of the repo's convention, and the commit carried that marker for a while.

The repo owner's decision is to ship it as a **minor**: the change affects only pills that were never clickable, it alters a cursor affordance and nothing else, an explicit `--pill-cursor` still wins, and `docs/Pill.md` documents both the change and the one-line opt-back-in (`--pill-cursor: pointer`). Raising this to a major would also collide with the 4.0.0 work already in flight on `feat/4-0-0-legacy-removal`.

The data-attribute passthrough named in issue #520's Need section is deliberately not here; it is tracked at #545.


Tara-ag (commented):
Review in progress — one finding under verification.

src/lib/Pill/Pill.svelte
```diff
@@ -100,23 +102,60 @@
...
+    text-transform: var(--pill-text-transform, none);
     padding: var(--pill-padding, 6px 10px);
     border-radius: var(--pill-border-radius, 999px);
     border: var(--pill-border, none);
-    cursor: var(--pill-cursor, pointer);
+    /* Non-interactive by default: nothing is clickable, so nothing should look
+       clickable. The pointer default below applies only once role="button" is
+       actually present (see handleClick / `interactive`) — an explicit
+       --pill-cursor still wins in both cases. */
+    cursor: var(--pill-cursor, default);
```
Tara-ag, line 115: ⚠️ MAJOR: Changing the fallback value of the documented `--pill-cursor` custom property from `pointer` to `default` for non-interactive pills is a silent behavioral change for existing consumers who do not set `--pill-cursor` and do not supply `onclick`. While the issue requested this fix, it alters the rendered output of already-shipped code without an opt-in or major-version signal.

Either (a) mark the change as breaking by amending the commit/PR title to `feat(pill)!:` and add a `BREAKING CHANGE` footer or migration note in `docs/Pill.md`, or (b) keep the old default and introduce an explicit opt-out prop/CSS variable for consumers who want the non-clickable cursor.

Tara-ag (changes requested):
One MAJOR finding: the non-interactive Pill cursor default changes a documented CSS custom-property fallback without a breaking-change marker or migration path. The rest of the PR (new `tone` prop, CSS variable layering, tests, docs, and WC parity) looks solid.

**Decision: CHANGES_REQUESTED** — please resolve the cursor-default compatibility concern before merging.

src/lib/Pill/Pill.svelte
Tara-ag, line 115: ⚠️ MAJOR: Changing the fallback value of the documented `--pill-cursor` custom property from `pointer` to `default` for non-interactive pills is a silent behavioral change for existing consumers who do not set `--pill-cursor` and do not supply `onclick`. While the issue requested this fix, it alters the rendered output of already-shipped code without an opt-in or major-version signal.

**Suggestion:** Either (a) mark the change as breaking by amending the commit/PR title to `feat(pill)!:` and add a `BREAKING CHANGE` footer or migration note in `docs/Pill.md`, or (b) keep the old default and introduce an explicit opt-out prop/CSS variable for consumers who want the non-clickable cursor.

Tara-ag (changes requested):
One MAJOR finding: the non-interactive Pill cursor default changes a documented CSS custom-property fallback from pointer to default without a breaking-change marker or migration path, which silently alters existing consumers. The rest of the PR (new tone prop, CSS variable layering, tests, docs, and web-component parity) is sound.

Tara-ag:

Severity counts — 🔒 CRITICAL: 0 · ⚠️ MAJOR: 2 · 💡 MINOR: 3

The PR introduces a valuable `tone` prop and improves Pill semantics, but it ships a breaking change to the public `--pill-cursor` default as a minor release without a major-version bump or breaking-change marker. Additional MAJOR issues: the web-component `tone` prop lacks `reflect: true` and runtime validation, allowing invalid tone classes. MINOR issues: `pillToneClass` accepts arbitrary strings, and docs reference an unreleased 4.0.0 version. A SUGGESTION for automated cursor regression coverage is also noted.

- ⚠️ **MAJOR**: Pill cursor default change is a breaking change shipped as minor — `src/lib/Pill/Pill.svelte:120`
  The PR changes the default value of the public CSS custom property `--pill-cursor` from `pointer` to `default` for any Pill without `onclick`. This is a behavioral change to the public styling contract of a published component. Even though the PR author argues it is only a cursor affordance, downstream consumers may have relied on the pointer cursor for non-interactive pills (e.g. as a visual cue,…
- ⚠️ **MAJOR**: Web component `tone` prop lacks `reflect: true` and type validation — `src/wc/components/Pill.wc.svelte:9`
  In `Pill.wc.svelte`, `tone` is declared as `{ type: 'String', attribute: 'tone' }` without `reflect: true`. Unlike `text`, `dismissible`, and `disabled`, the attribute will not reflect property changes back to the DOM, which is inconsistent with how the other string/boolean props behave and makes the component harder to style/query from CSS or tests. More importantly, Svelte custom-element string …
- 💡 **MINOR**: docs/Pill.md references version 4.0.0 for a change shipping in a minor 3.x release
  The breaking-change callout in docs/Pill.md says "> **4.0.0 — the default cursor changed for non-interactive pills.**", but this PR targets the `release` branch at version 3.5.0 with a `feat(pill):` commit, which standard-version will release as 3.6.0 (minor). Referencing a future 4.0.0 that is not actually the release containing the change is misleading for consumers reading the docs and makes th…
- 💡 **MINOR**: `pillToneClass` accepts any string due to loose type guard — `src/lib/Pill/pillTone.ts:18`
  `pillToneClass` returns a tone class for any value whose `typeof` is `'string'`, not just the documented `PillTone` values. This means TypeScript consumers get compile-time safety, but JS/web-component consumers can pass arbitrary strings and receive a meaningless `tone-<anything>` class. The unit tests do not cover this edge case.
- 💡 **MINOR**: Docs callout references unreleased major version 4.0.0 — `docs/Pill.md:15`
  The breaking-change callout in `docs/Pill.md` says '> **4.0.0 — the default cursor changed...**', but `package.json` is at version 3.5.0 and the commit is not marked as a major/breaking change. This is misleading: consumers reading the docs will expect the change in a 4.0.0 release that may never materialize from this PR, and the actual release will be a minor. It also contradicts the PR author's …
- 💬 **SUGGESTION**: Consider adding a Playwright test for the cursor default change — `src/routes/components/pill/+page.svelte:174`
  The PR adds a visual screenshot and unit tests for `pillToneClass`, but the cursor default change (the most user-facing behavioral change) is only exercised in the demo page without an automated assertion. A Playwright test asserting `cursor` computed style on `pill-cursor-static` vs `pill-cursor-interactive` would prevent regression.

docs/Pill.md
```diff
@@ -12,6 +12,14 @@ A small rounded label used for categorization, tagging, or filtering. Supports a
 <Pill text={'...'} />
 ```
 
+> **4.0.0 — the default cursor changed for non-interactive pills.** `--pill-cursor` used to
```
Tara-ag, line 15: 💡 MINOR: Docs version callout says '4.0.0' but the PR is intentionally shipping as a minor release. This will confuse consumers looking for a 4.0.0 release that does not correspond to this change. Change the callout to neutral language, e.g. '> **Note:** the default cursor changed for non-interactive pills...', until the actual version is known.

src/wc/components/Pill.wc.svelte
```diff
@@ -6,6 +6,7 @@
       text: { type: 'String', reflect: true },
       dismissible: { type: 'Boolean', reflect: true },
       disabled: { type: 'Boolean', reflect: true },
+      tone: { type: 'String', attribute: 'tone' },
```
Tara-ag, line 9: 🔒 CRITICAL: Web-component wrapper lost the `title` attribute alias and gained a conflicting `title` prop. The release branch mapped the custom-element `title` attribute to `pillTitle` and forwarded it as `title={pillTitle}` because `title` is a global HTML attribute with special handling in Svelte custom elements. This PR replaces that with `title: { type: 'String', reflect: true }`, which breaks `<sui-pill title="...">` consumers — the tooltip will no longer reach the inner Pill's `title` prop. Restore `pillTitle: { type: 'String', reflect: true, attribute: 'title' }` and forward it as `title={pillTitle}`.

src/lib/Pill/pill-tone.test.ts
```diff
@@ -0,0 +1,33 @@
+import { describe, it, expect } from 'vitest';
```
Tara-ag, line 1: 💡 MINOR: Unit test only covers the helper, not component prop integration or web-component parity. The helper test is fine for a browser-less Vitest suite, but there is no Playwright assertion that `Pill.svelte` actually renders the `tone-*` class or that the web-component wrapper forwards `tone`. Consider adding a Playwright test against the `pill-tone-fixtures` demo section to assert the class mapping and computed colors, and verify the WC props map includes `tone`.

src/wc/components/Pill.wc.svelte
Tara-ag, line 1: 🔒 CRITICAL: Web-component wrapper dropped the `onDismiss` event prop. The Svelte component now supports both `ondismiss` and `onDismiss`, but the custom-element props map only declares `ondismiss`. Web-component consumers using the corrected camelCase event name will not be able to attach a handler. Add `onDismiss: { type: 'Object' }` to the props map.

Tara-ag (changes requested):
Two CRITICAL web-component API regressions need fixing before merge: the `title` attribute alias was dropped and `onDismiss` was removed from the custom-element props map. Two MINOR notes on test coverage and docs versioning. Otherwise the Svelte-side tone prop, CSS variable layering, and cursor change look correct.

src/lib/Pill/Pill.svelte
```diff
@@ -100,23 +102,60 @@
     justify-content: var(--pill-justify-content, center);
     text-align: var(--pill-text-align, center);
     gap: var(--pill-gap, 4px);
-    background-color: var(--pill-background, #e0e0e0);
-    color: var(--pill-color, #333333);
+    background-color: var(--pill-background, var(--_pill-tone-background, #e0e0e0));
```
Tara-ag, line 101: ⚠️ MAJOR: Non-interactive Pill cursor default changed from `pointer` to `default`

This changes a shipped CSS custom-property default for an already-published component. Downstream consumers/themes may rely on the prior implicit `pointer` cursor for non-interactive Pills. Per the project blocking criteria, changing a relied-upon CSS custom property default is a breaking change unless shipped as a major version or accompanied by a documented migration note.

Evidence: `cursor: var(--pill-cursor, default);` here (and the `[role='button']` override below) flips the default; `docs/Pill.md` documents the change but the package version remains `3.4.0` and the commit is `feat(pill):` with no `!` or `BREAKING CHANGE:` footer.

Suggestion: Either (a) preserve backward compatibility by keeping `pointer` as the default and documenting the opt-in `--pill-cursor: default`, or (b) mark this as breaking (`feat(pill)!:` + `BREAKING CHANGE:` footer) and bump the major version. If a minor release is intentional, add an explicit migration note and ensure the commit carries the conventional-commits breaking marker.

src/lib/Pill/pillTone.ts
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Resolves the CSS class Pill.svelte applies for a given tone. Pure and exported — rather
+ * than five inline `class:tone-x={tone === 'x'}` directives — so the mapping is unit-testable
+ * without rendering: this repo's vitest suite runs no browser, so a `.svelte` template's DOM
+ * output cannot be asserted on directly (see `src/lib/Table/normalizeColumns.ts` for the same
+ * pattern). No tone passed resolves to `''`, so the class list — and therefore the rendered
+ * background/color — is unchanged from before this prop existed.
+ */
+export function pillToneClass(tone?: PillTone): string {
+  return typeof tone === 'string' ? `tone-${tone}` : '';
```
Tara-ag, line 18: 💡 MINOR: `pillToneClass` emits classes for arbitrary string values

The function returns `tone-${tone}` for any string, and the web-component wrapper declares `tone` as `type: 'String'` without further validation. A JS consumer passing `tone="foo"` will render a `tone-foo` class. There is no matching rule today, but this weakens the component contract and could collide with future tone names or consumer classes.

Suggestion: validate against `PILL_TONES` before emitting: `return tone && PILL_TONES.includes(tone) ? \`tone-${tone}\` : '';`, and add a test for invalid string input.

docs/Pill.md
```diff
@@ -100,7 +135,9 @@ Keyboard activation of the dismiss button does not activate the pill body.
 | `--pill-border-radius`       | `999px`                                   | border-radius    | Corner rounding of the pill (999px creates a fully rounded shape).        |
```
Tara-ag, line 135: 💡 MINOR: Hover CSS variable defaults omit the tone fallback layer

The documented defaults for `--pill-hover-background` and `--pill-hover-color` do not reflect the actual cascade in `Pill.svelte`. The component falls back through the tone layer (`var(--_pill-tone-background, #d0d0d0)` / `var(--_pill-tone-color, #333333)`), so a toned pill without explicit hover variables hovers in its tone color, not the legacy neutral defaults.

Please update the Default column or description to match the real fallback chain so consumers theming hover states understand the precedence.

Tara-ag (changes requested):
Reviewed file by file. Found one MAJOR backward-compatibility concern on the `--pill-cursor` default change and two MINOR items (docs hover-default accuracy and runtime validation of the `tone` class mapping). See inline comments for details.

Tara-ag (commented):
Review in progress

docs/Pill.md
Tara-ag, line 15: 💡 MINOR: The callout references `4.0.0`, but this PR targets `release` at `3.5.0` with a `feat(pill):` commit, which standard-version will release as `3.6.0` (minor). Referencing a future major version that does not contain this change is misleading. Either use the actual upcoming minor version (`3.6.0`) or a neutral heading without a version number.

src/lib/Pill/pillTone.ts
Tara-ag, line 18: 💡 MINOR: `pillToneClass` accepts any string due to loose type guard. It returns a tone class for any value whose `typeof` is `'string'`, not just the documented `PillTone` values. TypeScript consumers get compile-time safety, but JS/web-component consumers can pass arbitrary strings and receive a meaningless `tone-<anything>` class. The unit tests do not cover this edge case.

Suggested fix: Narrow the guard to the canonical list:
```ts
export function pillToneClass(tone?: PillTone): string {
  return tone && PILL_TONES.includes(tone) ? `tone-${tone}` : '';
}
```

src/wc/components/Pill.wc.svelte
Tara-ag, line 9: ⚠️ MAJOR: Web component `tone` prop lacks `reflect: true` and type validation. `tone` is declared as `{ type: 'String', attribute: 'tone' }` without `reflect: true`. Unlike `text`, `dismissible`, and `disabled`, the attribute will not reflect property changes back to the DOM, which is inconsistent with how the other string/boolean props behave. More importantly, Svelte custom-element string props accept any string value at runtime; an invalid tone like `tone="foo"` will silently produce `class="tone-foo"` with no visual indication because `pillToneClass` only checks `typeof tone === 'string'`.

Suggested fix: Add `reflect: true` to the `tone` prop declaration for consistency. Additionally, validate the tone value in `pillToneClass` so unknown tones fall back to no class rather than emitting an invalid `tone-foo` class, e.g.:
```ts
return tone && PILL_TONES.includes(tone) ? `tone-${tone}` : '';
```

docs/Pill.md
Tara-ag, line 15: 💡 MINOR: Docs callout references unreleased major version 4.0.0. The breaking-change callout says '> **4.0.0 — the default cursor changed...**', but `package.json` is at version 3.5.0 and the commit is not marked as a major/breaking change. This is misleading: consumers reading the docs will expect the change in a 4.0.0 release that may never materialize from this PR, and the actual release will be a minor. It also contradicts the PR author's stated intent to ship as a minor.

Suggested fix: Align the callout with the actual release plan. If shipping as a minor, rephrase to something like '> **Minor release — the default cursor changed...**' and explicitly state the version or release type. If the intent is truly 4.0.0, update `package.json` and the commit message accordingly and mark the change as breaking.

src/routes/components/pill/+page.svelte
```diff
@@ -149,6 +149,38 @@
...
+    <!-- A theme can restate what a tone means without a tone-* class per call site. -->
+    <Pill text="Themed ok" tone="ok" classes="pill-tone-ok-themed" testId="pill-tone-ok-themed" />
+    <!-- An explicit --pill-background/--pill-color still wins over the tone default. -->
+    <Pill
+      text="Tone + override"
+      tone="danger"
+      classes="pill-tone-override"
+      testId="pill-tone-override"
+    />
+  </div>
```
Tara-ag, line 174: 💬 SUGGESTION: Consider adding a Playwright test for the cursor default change. The cursor default change (the most user-facing behavioral change) is only exercised in the demo page without an automated assertion. A Playwright test asserting computed `cursor` style on `pill-cursor-static` vs `pill-cursor-interactive` would prevent regression.

Suggested fix: Add a Playwright spec (e.g. `tests/pill-cursor.spec.ts`) that navigates to `/components/pill`, asserts `getComputedStyle(staticPill).cursor === 'default'`, and `getComputedStyle(interactivePill).cursor === 'pointer'`.

src/lib/Pill/Pill.svelte
```diff
@@ -96,23 +98,60 @@
...
-    cursor: var(--pill-cursor, pointer);
+    /* Non-interactive by default: nothing is clickable, so nothing should look
+       clickable. The pointer default below applies only once role="button" is
+       actually present (see handleClick / `interactive`) — an explicit
+       --pill-cursor still wins in both cases. */
+    cursor: var(--pill-cursor, default);
     max-width: var(--pill-max-width);
     line-height: var(--pill-line-height, 1);
     flex-shrink: var(--pill-flex-shrink);
   }
```
Tara-ag, line 120: ⚠️ MAJOR: Pill cursor default change is a breaking change shipped as minor. The PR changes the default value of the public CSS custom property `--pill-cursor` from `pointer` to `default` for any Pill without `onclick`. This is a behavioral change to the public styling contract of a published component. Shipping this without a major version bump or a migration note violates the repo's blocking criterion for breaking changes to relied-upon CSS custom property names/defaults. The docs callout says '4.0.0' but package.json is at 3.5.0 and the commit is `feat(pill):` (no `!`), so this will be released as a minor.

Suggested fix: Either mark this as a breaking change (`feat(pill)!:` with a `BREAKING CHANGE:` footer describing the cursor default change and the `--pill-cursor: pointer` opt-back-in) and bump the major version, or keep the existing `pointer` default and only switch to `default` when `tone` is used or via an explicit opt-in prop. If the maintainers truly intend a minor, add a prominent migration note in docs/Pill.md and the changelog, and ensure the commit carries the breaking-change marker so release automation does not mislead consumers.

Tara-ag (changes requested):
Blocking: the `--pill-cursor` default change alters a public CSS custom property default for a published component without a major-version bump or breaking-change marker. Two additional MAJOR findings (web-component `tone` parity/validation) and two MINOR findings (runtime tone validation, docs version mismatch) also need resolution. One SUGGESTION is included for test coverage.

Tara-ag (changes requested):
Review complete. The PR introduces a useful `tone` prop and improves Pill semantics, but it ships a breaking change to the public `--pill-cursor` default as a minor release without a major-version bump or breaking-change marker. Additional issues: web-component `tone` prop lacks `reflect: true` and runtime validation; `pillToneClass` accepts arbitrary strings; docs reference an unreleased 4.0.0 version. See inline comments for details.

murdore:
## Video evidence

**What the recording shows**

Navigates to /components/pill on a real vite preview build of this PR's branch (port 43280). Scrolls to the "Tone" fixture section and, for each of the 5 semantic tones (accent, ok, warn, danger, muted), hovers the corresponding Pill (data-pw="pill-tone-{tone}") with a visible pause, logging and asserting its computed background-color is distinct from the untoned default (rgb(224,224,224)) and from every other tone (5 distinct backgrounds observed: rgb(209,236,241) / rgb(212,237,218) / rgb(255,243,205) / rgb(248,215,218) / rgb(241,241,241)). Then scrolls to and hovers the "Clickable (pointer cursor)" pill (has onclick) and confirms computed cursor=pointer and role=button; then hovers the "Label only (default cursor)" pill (no onclick) and confirms computed cursor=default (not pointer) and role=null — i.e. the PR's claimed behaviour (tone rendering + non-interactive pills no longer showing a pointer) is directly visible and asserted frame-by-frame, not just a green checkmark.

> Test timeout of 30000ms exceeded at locator.scrollIntoViewIfNeeded() waiting for getByText('Tone', { exact: true }) — with source reverted to origin/release, the demo page has no Tone section, no pill-tone-* testIds, and no tone-* CSS classes, so the spec cannot even find the fixtures it asserts on. This confirms the tone/cursor behaviour is genuinely introduced by this PR's source changes, not an artifact of the spec.

- passing, on this branch: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-534-pill.webm` (163,642 bytes)

Tara-ag (commented):
Review in progress — file by file.

src/wc/components/Pill.wc.svelte
Tara-ag, line 9: 💡 MINOR: The dominant repo convention for optional, non-reflected string props is to omit the `attribute` key and let Svelte default it to the lowercased prop name (see ChatSuggestions, Draggable, MediaPlayer). Button.wc.svelte is the outlier that explicitly repeats `attribute`. Since `tone` does not need a custom attribute name, consider either omitting `attribute` to match the majority pattern (`tone: { type: 'String' }`) or adding `reflect: true` to match the enum-like prop convention used elsewhere (Book.transition, Phone.variant, Modal.size, LoadingDots.animation, Choicebox.mode).

docs/Pill.md
Tara-ag, line 15: 💬 SUGGESTION: The callout says "> **4.0.0 — the default cursor changed...**", but the PR description states this is shipping as a minor on the release branch, not a 4.0.0 breaking change. A reader will expect a major-version bump that does not happen. Replace the hard-coded version with a neutral phrase such as "> **The default cursor changed for non-interactive pills.**" or update it to the exact minor version this PR will release under.

docs/Pill.md
```diff
@@ -115,6 +152,16 @@ Keyboard activation of the dismiss button does not activate the pill body.
...
 | `--pill-dismiss-color`       | `currentColor`                            | color            | Color of the dismiss button icon.                                         |
 | `--pill-dismiss-hover-color` | `var(--pill-dismiss-color, currentColor)` | color            | Color of the dismiss button icon on hover.                                |
+| `--pill-tone-accent-background` | `#d1ecf1`                             | background-color | `tone="accent"` background. Ignored unless `tone` is set; overridden by `--pill-background`. |
+| `--pill-tone-accent-color`      | `#0c5460`                             | color            | `tone="accent"` text color. Ignored unless `tone` is set; overridden by `--pill-color`. |
+| `--pill-tone-ok-background`     | `#d4edda`                             | background-color | `tone="ok"` background. Ignored unless `tone` is set; overridden by `--pill-background`. |
+| `--pill-tone-ok-color`          | `#155724`                             | color            | `tone="ok"` text color. Ignored unless `tone` is set; overridden by `--pill-color`. |
```
Tara-ag, line 158: 💬 SUGGESTION: The Web Component section shows only `<sui-pill text="Active" dismissible></sui-pill>` and lists only the two slots. Since the PR adds `tone` to the custom element, document it here with an example such as `<sui-pill text="Active" tone="ok"></sui-pill>` so custom-element consumers can discover the new attribute.

Tara-ag (approved):
Approved with minor notes. The change is additive and backward-compatible: a new optional `tone` prop, new CSS variables, and a corrected default cursor for non-interactive pills. No blocking issues found.

PR #535, opened 2026-09-06
feat(card): add attribute passthrough and a root-tag override

### LINKED ISSUE #521: Card: Generic data-attribute / rest-prop passthrough and a root-tag override on Card
## Need
Generic data-attribute / rest-prop passthrough and a root-tag override on Card

## Closest today
Card (href/title/headerRight/children/classes/cssVars, no data-* or `as` prop)

## Proposal
Spread an optional `attrs?: Record<string,string>` onto Card's root, and add an `as?: 'div'|'a'|'figure'` tag override for callers using data-state/data-density/data-offline attribute-selector CSS or needing <figure> semantics.

## Where it bites
Found while migrating TARA (a Svelte 5 / SvelteKit 2 consumer pinned to 3.3.0) onto the library. Without this the following consumer files can only migrate part-way and keep hand-rolled markup for the rest: `+page.svelte`, `SessionActivityCard.svelte`.

murdore:
Card had no way to put an arbitrary data-*/aria-* attribute on its root or
render a tag other than <div>/<a>. Migrating TARA onto Card, that forced two
workarounds: SessionActivityCard.svelte encoded `[data-state="waiting"]` as a
second `classes` modifier instead of a real attribute, and backdrops/+page.svelte
kept a hand-rolled <figure> wrapping an appearance-only Card because Card's own
root could never itself be the <figure>.

Two additive props close both gaps:

- `attrs?: Record<string, string>` spreads onto the card root, for consumers
  that key off attribute-selector CSS the way the rest of their app does.
- `as?: 'div' | 'a' | 'figure'` overrides the rendered tag independent of
  `href`, for real sectioning-content semantics with no wrapper element.

Both are omitted by default, so the existing behaviour is unchanged: the root
tag resolution stays exactly `<a>` when `href` is set, `<div>` otherwise, and
no extra attributes are rendered. `as` set to anything but `'a'` suppresses
`href`/`target`/`rel` and the synthetic role/tabindex/keydown shim applies
instead when `onclick` is given, same as today's plain interactive `<div>`.

`attrs` is spread first on the element, before Card's own class/style/
data-pw/testID/role/tabindex/href/target/rel/onclick/onkeydown -- a later key
in a Svelte attribute object wins over an earlier one of the same name, so
attrs can only add attributes Card doesn't already manage, never override one
(a new spec pins this: an attrs object trying to hijack data-pw/class/role is
proven not to).

Wires `attrs`/`as` through the web-component wrapper for check:wc-parity, adds
the two to docs/Card.md and the props table, and demos both (plus the
collision guard) on the Card gallery page.

Closes #521

---

### Verification

**Spec runs on this branch** — `tests/card-attrs-and-as.test.ts` — 5 Playwright tests against a real build.

## Summary by CodeRabbit

- **New Features**
  - Added `as` support to render cards with a selected root element, such as `figure`.
  - Added `attrs` support for applying custom data and accessibility attributes to the card root.
  - Added support for the new properties on the `sui-card` Web Component.
  - Added a Card demo showcasing custom attributes and root elements.

- **Documentation**
  - Documented the new Card properties and their behavior.


tests/card-attrs-and-as.test.ts
```diff
@@ -0,0 +1,82 @@
...
+// Card had no way to (a) put an arbitrary data-*/aria-* attribute on its root
+// or (b) render a tag other than <div>/<a> — the two things
+// juspay/svelte-ui-components#521 asks for. Consumers hit this migrating
+// existing markup onto Card: a `[data-state="waiting"]` selector became a
+// second `classes` modifier instead of a real attribute, and a card-styled
+// <figure> stayed a hand-rolled <figure> wrapping an appearance-only Card
+// because Card's own root could never itself be the <figure>.
+//
+// `attrs` and `as` close both gaps. These specs fail against a Card with
+// neither prop declared: an undeclared prop is silently dropped by Svelte 5's
```
Tara-ag, line 12: 💡 MINOR: New Card behavior relies solely on demo-page Playwright tests

Every spec in this file navigates to `/components/card` and selects demo instances by `testId`. The project standard expects new component behavior to also have Vitest unit coverage so the contract (root tag resolution, attr spread ordering, collision guard) is tested independently of the demo page layout. Relying only on gallery-page browser tests is fragile.

Suggestion: add a Vitest unit/spec file under `src/lib/Card/` (e.g. `Card.test.ts`) that mounts `Card` with `attrs`/`as` and asserts the rendered element/tag/attributes directly, without depending on the demo route.

Tara-ag (approved):
Approved with one minor note: the new `attrs`/`as` behavior is currently covered only by demo-page Playwright tests. Consider adding a Vitest unit test under `src/lib/Card/` to guard the root-tag and attribute-spread contract independently of the gallery route.

No blocking issues found:
- Props are purely additive; existing `href`/`div` tag resolution is preserved when `as` is omitted.
- `attrs` is spread before Card-managed attributes, so collisions are correctly suppressed.
- Web-component wrapper wires both props through without shadowing host-reserved accessors.
- Docs and demo page are updated.

Tara-ag (commented):
Review in progress.

tests/card-attrs-and-as.test.ts
Tara-ag, line 12: 💡 MINOR: New Card behavior relies solely on demo-page Playwright tests. The PR adds five Playwright specs that all navigate to /components/card and select demo instances. The project standard expects new component behavior to also have Vitest unit coverage so the contract (root tag resolution, attr spread ordering, collision guard) is tested independently of the demo page layout. Relying only on gallery-page browser tests is fragile.

Suggestion: Add a Vitest unit/spec file under src/lib/Card/ (e.g. Card.test.ts) that mounts Card with attrs/as and asserts the rendered element/tag/attributes directly, without depending on the demo route.

Tara-ag (commented):
Approved with one minor note about adding unit-test coverage for the new Card behavior.

src/wc/components/Card.wc.svelte
```diff
@@ -17,7 +17,9 @@
       target: { type: 'String', attribute: 'target' },
       rel: { type: 'String', attribute: 'rel' },
       headerRight: { type: 'Object' },
-      footer: { type: 'Object' }
+      footer: { type: 'Object' },
+      attrs: { type: 'Object' },
+      as: { type: 'String', attribute: 'as' }
```
Tara-ag: 💬 SUGGESTION: docs/Card.md claims `as` is a reflected attribute, but this prop declaration omits `reflect: true`. In Svelte custom elements, `attribute` only maps HTML attribute → property; reflecting property changes back to the host attribute requires `reflect: true`. Either add `reflect: true` here (`as: { type: 'String', attribute: 'as', reflect: true }`) or update docs/Card.md to remove the "reflected" claim so the web-component contract matches its documentation.

src/lib/Card/Card.svelte
```diff
@@ -50,7 +62,8 @@
 </script>
 
 <svelte:element
-  this={isAnchor ? 'a' : 'div'}
+  this={rootTag}
```
Tara-ag, line 70: 💬 SUGGESTION: When `as='figure'` is used, the root becomes a `<figure>` but the `title` prop is still rendered inside a nested `<div class="card-title">` with no `aria-labelledby` or `<figcaption>` association. Screen readers won't announce the figure's label. Consider rendering the title as a `<figcaption>` when `as='figure'`, or adding `aria-labelledby` pointing to the title element; otherwise document that `title` is decorative and consumers must supply their own `<figcaption>`.

src/lib/Card/properties.ts
```diff
@@ -85,4 +85,34 @@ export type OptionalCardProperties = {
...
+   * `class`/`style`/`data-pw`/`testID`/`role`/`tabindex`/`href`/`target`/`rel`/
+   * `onclick`/`onkeydown` attributes, so an entry here can never override one
+   * of those — only add attributes Card does not already manage. Omitted by
+   * default, so existing consumers render exactly the same attributes as
+   * before.
+   */
+  attrs?: Record<string, string>;
+  /**
+   * Overrides the tag `svelte:element` renders for the card root — e.g.
+   * `'figure'` for a card that needs real `<figure>` semantics (so a
```
Tara-ag, line 103: 💬 SUGGESTION: `attrs?: Record<string, string>` accepts keys Card already manages (class, style, role, tabindex, href, target, rel, data-pw, testID, onclick, onkeydown). The runtime spread ordering protects these, but TypeScript won't warn consumers who accidentally pass them. Consider narrowing the type with `Omit<Record<string, string>, 'class' | 'style' | 'role' | 'tabindex' | 'href' | 'target' | 'rel' | 'data-pw' | 'testID' | 'onclick' | 'onkeydown'>` so misuse is caught at compile time without changing runtime behavior.

tests/card-attrs-and-as.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  });
+
+  test('omitting attrs/as leaves the default div root unaffected (no regression)', async ({
+    page
+  }) => {
+    await page.goto('/components/card');
+
+    // Regression guard for every existing consumer: a card given neither prop
+    // still renders a plain <div> with no attributes it didn't have before.
+    const card = page.getByTestId('metric-row-card');
```
Tara-ag, line 77: 💬 SUGGESTION: This regression guard selects `metric-row-card`, a testId defined in unrelated demo markup rather than in this PR's new section. If the demo page is refactored and that testId is renamed, this test will fail even though the default Card behavior is unchanged. Consider adding a dedicated default card in the new "Attribute Passthrough and Custom Root Tag" demo section (e.g. `testId="default-div-card"`) and asserting on that instead.

Tara-ag (commented):
Approved with minor suggestions. The change is additive, backward-compatible, and correctly guarded by tests. Four suggestions are left inline: (1) align the web-component `as` reflection with docs, (2) consider associating the `title` with `as='figure'` for accessibility, (3) narrow the `attrs` type to omit managed keys, and (4) make the regression test less dependent on unrelated demo markup. None are blocking.

Tara-ag:

The PR adds useful `attrs` and `as` flexibility to Card and updates docs/WC parity. However, one accepted MINOR issue remains: the new Playwright keyboard test is coupled to demo-page tab order and can flake if the demo page changes. Decouple the tab test before merging.

- 💡 **MINOR**: Tab-order test is coupled to demo-page layout — `tests/card-attrs-and-as.test.ts:96`
  The keyboard-navigation test focuses parity-anchor-card and asserts that one Tab press lands on as-a-no-href-card. This assumes no other focusable elements exist between those two cards in the card demo page. If the demo page is reorganized or new interactive elements are added, the test will flake or fail even though the Card component behavior is correct.

tests/card-attrs-and-as.test.ts
```diff
@@ -0,0 +1,82 @@
...
+// because Card's own root could never itself be the <figure>.
+//
+// `attrs` and `as` close both gaps. These specs fail against a Card with
+// neither prop declared: an undeclared prop is silently dropped by Svelte 5's
+// destructuring `$props()`, so `attrs` never reaches the DOM and `as` never
+// changes the rendered tag away from its href-derived default.
+test.describe('Card attrs and as', () => {
+  test('attrs spreads arbitrary data-* attributes onto the root', async ({ page }) => {
+    await page.goto('/components/card');
+
```
Tara-ag, line 18: 💬 SUGGESTION: Each test repeats `page.goto('/components/card')`. Consider a shared `test.beforeEach` to reduce redundant navigation and speed up the suite.

src/wc/components/Card.wc.svelte
```diff
@@ -17,7 +17,9 @@
       target: { type: 'String', attribute: 'target' },
       rel: { type: 'String', attribute: 'rel' },
       headerRight: { type: 'Object' },
-      footer: { type: 'Object' }
+      footer: { type: 'Object' },
```
Tara-ag, line 20: 💡 MINOR: `as` is reflected as an HTML attribute, but `attrs` is object-only (`type: 'Object'`). This matches `cssVars` and the docs, but it means declarative `<sui-card attrs="...">` usage is impossible. Worth confirming this is the intended parity trade-off.

src/lib/Card/Card.svelte
```diff
@@ -50,7 +62,8 @@
 </script>
```
Tara-ag, line 67: ⚠️ MAJOR: When `as="figure"` is combined with `onclick`, the resolved root tag is `<figure>` but the component still applies `role="button"` and `tabindex="0"` (line 73). A `<figure>` with `role="button"` is invalid ARIA — it contradicts the sectioning-content semantics `as="figure"` is meant to provide. Either disallow `onclick`/`href` with semantic tags, or preserve the tag's implicit role and require consumers to put interactive content inside the card.

Tara-ag (commented):
Review complete. The change is additive and preserves backward compatibility, but one MAJOR accessibility concern needs attention: `as="figure"` combined with `onclick` produces invalid ARIA (`<figure role="button">`). Please address or justify before merging. Two minor/suggestion comments also posted.

src/wc/components/Card.wc.svelte
Tara-ag, line 20: 💡 MINOR: `attrs` is object-only on the web-component wrapper

The web-component wrapper declares `attrs: { type: 'Object' }` with no attribute mapping. This matches `cssVars` and the documented behavior, but it means declarative HTML usage `<sui-card attrs="...">` is impossible; `attrs` must be set as a property. Confirm this parity trade-off is intentional, or document that `attrs` on the custom element must be set as a property (e.g. `el.attrs = { 'data-state': 'waiting' }`).

src/wc/components/Card.wc.svelte
Tara-ag: 💬 SUGGESTION: Docs claim `as` is reflected but web-component config omits `reflect: true`

docs/Card.md states that `as="figure"` sets the root tag as a reflected `as` attribute, but the custom-element prop declaration is `as: { type: 'String', attribute: 'as' }` without `reflect: true`. In Svelte custom elements, `attribute` only maps HTML attribute → property; reflecting property changes back to the host attribute requires `reflect: true`. Either add `reflect: true` here or update docs/Card.md to remove the "reflected" claim.

tests/card-attrs-and-as.test.ts
Tara-ag, line 18: 💬 SUGGESTION: Each test repeats `page.goto('/components/card')`

Using a shared `test.beforeEach(async ({ page }) => { await page.goto('/components/card'); })` hook would reduce redundant navigation and speed up the suite.

Tara-ag (commented):
Yama review complete. The change is additive and backward-compatible, with docs, demo, and Playwright coverage. Three non-blocking findings remain: a web-component `attrs` parity note, a docs/WC reflection mismatch on `as`, and a test-structure suggestion. No blocking issues.

murdore:
## Video evidence

**What the recording shows**

Navigates to /components/card, scrolls to the 'Attribute Passthrough and Custom Root Tag (attrs / as)' section, and for three cards in turn outlines the actual DOM element and paints an on-screen readout built from values read live off that element at that instant (not scripted captions): (1) attrs-data-state-card -- readout shows element.tagName = &lt;div&gt;, getAttribute('data-state') = "waiting", getAttribute('data-density') = "compact", proving Card's attrs prop spreads arbitrary data-* attributes onto its root; (2) as-figure-card -- readout shows element.tagName = &lt;figure&gt;, proving Card's as="figure" prop renders a real non-div root tag; (3) as-figure-with-href-card -- readout shows element.tagName = &lt;figure&gt; and getAttribute('href') = null (attribute absent), proving as overrides the href-derived &lt;a&gt; default so no href attribute leaks onto the figure.

> Test timeout of 30000ms exceeded. Error: locator.scrollIntoViewIfNeeded: Test timeout of 30000ms exceeded. Call log: waiting for getByText('Attribute Passthrough and Custom Root Tag (attrs / as)') -- that whole demo section (and the attrs/as props themselves) does not exist once source files are reverted to origin/release, so the section heading the spec waits for never renders.

- passing, on this branch: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-535-card.webm` (412,930 bytes)
- failing, on `release`: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-535-card-FAILS-on-release.webm` (455,057 bytes)

- The first `pnpm test:unit --run` invocation failed with 1 timed-out test (scripts/migrate/rename-doc-usages.test.ts, a 5000ms timeout unrelated to Card, almost certainly caused by concurrent system load from other worktrees/processes during this review). A clean immediate rerun passed 828/828 with 0 failures, matching the rebase report's number, so this is recorded as a flake, not a regression -- but note it for whoever merges, since a flaky CI run on this same test could recur.
- The PR's own real Playwright spec (tests/card-attrs-and-as.test.ts) still asserts purely via DOM queries with no scroll/visibility, so it never appears on camera in a normal headed/video run -- confirmed true by reading the spec. The recorded evidence necessarily comes from a throwaway spec+config outside the gated suites, which is honestly disclosed and was not left behind (git status clean, no test-results-evidence/ directory present).

Tara-ag (commented):

src/lib/Card/Card.svelte
Tara-ag, line 67: ⚠️ MAJOR: `as="figure"` does not deliver valid `<figcaption>` semantics as documented. The PR's docs and JSDoc claim `<figcaption>` inside `children` or `footer` is valid markup, but Card.svelte always wraps `children` in `<div class="card-content">` and `footer` in `<footer class="card-footer">`. Since `<figcaption>` must be a direct child of `<figure>` to be valid HTML and exposed correctly to assistive technologies, the advertised accessibility use case is broken.

Suggested fix: either add a first-class `caption` snippet/prop that renders `<figcaption>` as a direct child of the root, or update the documentation and JSDoc to stop claiming `<figcaption>` inside `children`/`footer` is valid markup and demonstrate the limitation explicitly.

Tara-ag (changes requested):
One verified MAJOR finding: the documented `<figcaption>` use case for `as="figure"` is not achievable because `children`/`footer` are wrapped in `<div>`/`<footer>`, breaking the required direct-child relationship. Please either add a first-class caption mechanism or correct the docs/JSDoc. Otherwise the change is additive, well-tested, and preserves the existing public API.

src/lib/Card/properties.ts
Tara-ag, line 103: 💬 SUGGESTION: `attrs` type is permissive and allows passing keys Card already manages. The `attrs` prop is typed as `Record<string, string>`, which accepts keys that Card already manages (class, style, role, tabindex, href, target, rel, data-pw, testID, onclick, onkeydown). The implementation relies on Svelte spread ordering so later keys win, and the PR documents and tests this. However, the type does not prevent consumers from accidentally passing these keys, and TypeScript will not warn them that the values will be ignored.

Suggested fix: consider typing `attrs` as `Omit<HTMLAttributes<HTMLElement>, keyof CardManagedAttributes> & Record<string, string>` or documenting the reserved keys explicitly so TypeScript can warn consumers who try to override managed attributes.

src/wc/components/Card.wc.svelte
Tara-ag: 💬 SUGGESTION: Web-component `as` prop lacks `reflect: true` despite docs claiming it is reflected. docs/Card.md says `as="figure"` sets the root tag as a reflected `as` attribute, but this declaration is `as: { type: 'String', attribute: 'as' }` without `reflect: true`. In Svelte custom elements, the `attribute` option only maps HTML attributes to properties; reflecting property changes back to the attribute requires `reflect: true`. Either add `reflect: true` or update docs/Card.md to remove the word "reflected".

Suggested fix: `as: { type: 'String', attribute: 'as', reflect: true }`.

tests/card-attrs-and-as.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  test('omitting attrs/as leaves the default div root unaffected (no regression)', async ({
+    page
+  }) => {
+    await page.goto('/components/card');
+
+    // Regression guard for every existing consumer: a card given neither prop
+    // still renders a plain <div> with no attributes it didn't have before.
+    const card = page.getByTestId('metric-row-card');
+    const tagName = await card.evaluate((el) => el.tagName.toLowerCase());
+    expect(tagName).toBe('div');
```
Tara-ag, line 79: 💬 SUGGESTION: Regression test depends on unrelated demo-page markup. The default-behavior test selects `page.getByTestId('metric-row-card')`, which is defined in the demo page rather than in the test fixture. If the demo page is refactored or the testId is renamed, the regression guard will fail even though the component behavior is correct.

Suggested fix: add a dedicated card in this test file's own fixture markup (or a minimal inline Svelte route) for the regression assertion, so the test does not depend on unrelated demo markup.

src/lib/Card/Card.svelte
```diff
@@ -50,7 +62,8 @@
...
 
 <svelte:element
```
Tara-ag, line 69: 💬 SUGGESTION: `as='figure'` with `title` lacks programmatic association for the figure. When `as='figure'` is used, the root becomes a `<figure>` but the `title` prop is still rendered inside a nested `<div class="card-title">` in the header. There is no `aria-labelledby` linking the figure to its title and no `<figcaption>`. Screen readers will not announce the figure's label, so the new figure semantics are incomplete.

Suggested fix: when `as='figure'`, render the title block as a `<figcaption>` (or add `aria-labelledby` pointing to the title element's id) so the figure has an accessible label.

Tara-ag (commented):
Yama review: additive Card API change with verified suggestions only.

Tara-ag (changes requested):
One MAJOR accessibility issue and four SUGGESTION-level items were accepted by the review gate. The MAJOR finding (invalid `<figcaption>` semantics when `as="figure"`) contradicts the PR's own documentation and needs to be addressed before merge. The four SUGGESTIONs can be addressed or acknowledged at the author's discretion. See inline comments for details.

src/lib/Card/Card.svelte
Tara-ag, line 69: 💬 SUGGESTION: When `as='figure'` is used, the root becomes a `<figure>` but the `title` prop is still rendered inside a nested `<div class="card-title">` in the header. There is no `aria-labelledby` linking the figure to its title and no `<figcaption>`. Screen readers will not announce the figure's label, so the new figure semantics are incomplete.

Suggested fix: when `as='figure'`, render the title block as a `<figcaption>` (or add `aria-labelledby` pointing to the title element's id) so the figure has an accessible label.

src/wc/components/Card.wc.svelte
Tara-ag: 💬 SUGGESTION: docs/Card.md claims `as` is a reflected attribute, but this prop declaration omits `reflect: true`. In Svelte custom elements, `attribute` only maps HTML attribute → property; reflecting property changes back to the host attribute requires `reflect: true`. Either add `reflect: true` here (`as: { type: 'String', attribute: 'as', reflect: true }`) or update docs/Card.md to remove the "reflected" claim so the web-component contract matches its documentation.

src/lib/Card/Card.svelte
Tara-ag, line 67: ⚠️ MAJOR: `as="figure"` does not deliver valid `<figcaption>` semantics as documented. The PR's docs and JSDoc claim `<figcaption>` inside `children` or `footer` is valid markup, but Card.svelte always wraps `children` in `<div class="card-content">` and `footer` in `<footer class="card-footer">`. Since `<figcaption>` must be a direct child of `<figure>` to be valid HTML and exposed correctly to assistive technologies, the advertised accessibility use case is broken.

Suggested fix: either add a first-class `caption` snippet/prop that renders `<figcaption>` as a direct child of the root, or update the documentation and JSDoc to stop claiming `<figcaption>` inside `children`/`footer` is valid markup and demonstrate the limitation explicitly.

Tara-ag (changes requested):
One MAJOR accessibility issue and four SUGGESTION-level items were accepted by the review gate. The MAJOR finding (invalid `<figcaption>` semantics when `as="figure"`) contradicts the PR's own documentation and needs to be addressed before merge. The four SUGGESTIONs can be addressed or acknowledged at the author's discretion. See inline comments for details.

tests/card-attrs-and-as.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  // href, no tabindex and no keydown handler: neither Tab-reachable nor
+  // Enter/Space-activatable. `isAnchor` now also requires `href` to be set,
+  // so this case keeps the shim instead.
+  test('as="a" without href still gets the interactive-div keyboard shim', async ({ page }) => {
+    await page.goto('/components/card');
+
+    const card = page.getByTestId('as-a-no-href-card');
+    const tagName = await card.evaluate((el) => el.tagName.toLowerCase());
+    expect(tagName).toBe('a');
+    await expect(card).not.toHaveAttribute('href', /.*/);
```
Tara-ag, line 96: 💡 MINOR: Tab-order test is coupled to demo-page layout. The keyboard-navigation test focuses `parity-anchor-card` and asserts that one Tab press lands on `as-a-no-href-card`. This assumes no other focusable elements exist between those two cards in the card demo page. If the demo page is reorganized or new interactive elements are added, the test will flake or fail even though the Card component behavior is correct.

Make the tab test self-contained: focus the target card directly with `await card.focus()` before pressing Tab, or use a dedicated fixture page that isolates the `as-a-no-href` card and a single preceding focusable element.

Tara-ag (commented):
Reviewed. One new MINOR finding on test coupling; existing unresolved MAJOR accessibility concerns remain from prior review.

tests/card-attrs-and-as.test.ts
Tara-ag, line 96: 💡 MINOR: Tab-order test is coupled to demo-page layout

The keyboard-navigation test focuses `parity-anchor-card` and asserts that one Tab press lands on `as-a-no-href-card`. This assumes no other focusable elements exist between those two cards in the card demo page. If the demo page is reorganized or new interactive elements are added, the test will flake or fail even though the Card component behavior is correct.

**Suggestion:** Make the tab test self-contained: focus the target card directly with `await card.focus()` before pressing Tab, or use a dedicated fixture page that isolates the `as-a-no-href` card and a single preceding focusable element.

Tara-ag (commented):
Reviewed. One minor test-coupling issue remains; please address before merge.

PR #536, opened 2026-09-06
feat(toolbar): add backHref to render the back control as a real link

### LINKED ISSUE #522: Toolbar: An href option on Toolbar's built-in back control (it is callback-only, onbackClick/onBa…
## Need
An href option on Toolbar's built-in back control (it is callback-only, onbackClick/onBackClick, at both 2.131.0 and 3.2.2)

## Closest today
Toolbar (leftContent snippet as the workaround)

## Proposal
Add an `backHref?: string` prop that renders the built-in back control as a real `<a>` when set, matching the shape change from commit 83c406d instead of leaving href-based navigation as a leftContent workaround forever.

## Where it bites
Found while migrating TARA (a Svelte 5 / SvelteKit 2 consumer pinned to 3.3.0) onto the library. Without this the following consumer files can only migrate part-way and keep hand-rolled markup for the rest: `+page.svelte`, `PageHeader.svelte`, `SessionHeaderBar.svelte`.

murdore:
Toolbar's built-in back control was callback-only (onbackClick/onBackClick),
with no way to make it a real link. TARA's web app (a Svelte 5/SvelteKit 2
consumer) hit this migrating onto the library: PageHeader.svelte,
SessionHeaderBar.svelte and the terminal/share +page.svelte routes could not
use Toolbar's own back button at all — each had to replace it wholesale via
the `leftContent` snippet just to get an `<a href>`, hand-rolling markup and
losing the library's own back-button classes, tokens and accessible-name
handling in the process (see packages/ui/src/styles/components.css's
`.tara-ui-btn-back` comment in the TARA repo).

New optional prop `backHref?: string` (also `back-href` on the web
component). When set, the built-in back control renders as
`<a href={backHref}>` instead of `<button type="button">`, via
`<svelte:element>` — the same technique the library's own Button component
already uses for its `href` prop. `onbackClick`/`onBackClick` still fire as
the anchor's click handler when supplied, so navigation and a side effect
(analytics, etc.) coexist exactly like Button's `href` + `onclick`. Ignored
when `leftContent` replaces the control, or the control is hidden
(`showBackButton={false}` / `backIcon={null}`).

Default: omitting `backHref` renders the existing callback-only `<button>`,
byte-for-byte — pinned by the existing Toolbar suites, which stay green
unmodified, plus a new default-tag assertion in the added spec.

New Playwright spec `tests/toolbar-back-href.test.ts` covers: the default
stays a `<button>` with no href; `backHref` renders a real `<a>` with the
given href, unchanged aria-label and icon; `onbackClick` still fires on
click and on keyboard Enter; and `leftContent` still wins over `backHref`.
Confirmed to fail against the pre-change component (asserts `<a>`, gets
`<button>`) and pass with it.

Closes #522

---

### Verification

**Spec runs on this branch** — `tests/toolbar-back-href.test.ts` — 5 Playwright tests against a real build.

**Fail-before** — Source reverted to `origin/release` with only the spec kept: **4 fail, 1 passes** — and the one that passes is the back-compat test asserting an unconfigured toolbar still renders a `<button>`, which is exactly what should still hold on release.

## Summary by CodeRabbit

- **New Features**
  - Added optional `backHref` support to the Toolbar, rendering the back control as an accessible link with the provided URL.
  - Preserved click and keyboard activation behavior, including Space-key support.
  - Added `back-href` support to the custom element.
  - Existing custom left content continues to take precedence over the built-in back control.

- **Documentation**
  - Updated Toolbar property and keyboard interaction documentation.

- **Tests**
  - Added coverage for link rendering, navigation, accessibility, keyboard activation, and control precedence.


Tara-ag (approved):
Clean, backward-compatible addition. The optional `backHref` prop preserves the existing button behavior by default, mirrors the `Button` component's `svelte:element` pattern, and is correctly wired through the web-component registry and docs. New Playwright coverage and the demo-page example are appropriate. No blocking issues found.

Tara-ag (commented):
Reviewing Toolbar `backHref` changes.

src/wc/components/Toolbar.wc.svelte
```diff
@@ -7,6 +7,7 @@
       text: { type: 'String', reflect: true },
       backIcon: { type: 'String', attribute: 'back-icon' },
       backLabel: { type: 'String', attribute: 'back-label' },
+      backHref: { type: 'String', attribute: 'back-href' },
```
Tara-ag, line 10: 💬 SUGGESTION: Consider adding `reflect: true` to the `backHref` declaration. Other string props like `text` are reflected, and reflecting the href attribute makes the custom element more inspectable and consistent with the rest of the public API surface.

src/lib/Toolbar/Toolbar.svelte
```diff
@@ -38,8 +43,11 @@
     {#if typeof leftContent === 'function'}
       {@render leftContent()}
     {:else if showBackControl}
-      <button
-        type="button"
+      <svelte:element
```
Tara-ag, line 54: ⚠️ MAJOR: Anchor back control loses Space-key activation. When `backHref` is set, this renders as `<a>`, which native browsers activate with Enter but **not** Space. The existing docs tell consumers that Enter and Space already activate `onbackClick` for the native `<button>`. With the anchor, Space will scroll the page instead of triggering the back action, breaking keyboard parity.

Add an `onkeydown` handler that converts Space to activation on the anchor, e.g.:
```ts
function handleBackKeyDown(event: KeyboardEvent) {
  if (hasBackHref && event.key === ' ') {
    event.preventDefault();
    onbackClick?.();
  }
  onkeydown?.(event);
}
```
Then bind `onkeydown={handleBackKeyDown}` on this element.

Tara-ag (commented):
Reviewed Toolbar `backHref` changes. One MAJOR accessibility concern (Space-key activation lost on anchor), one MINOR docs inaccuracy, and one SUGGESTION for web-component reflection.

Tara-ag:
💡 MINOR: Docs still describe the back control as always a native `<button>`. Since `backHref` can render it as an `<a>`, the statement in `docs/Toolbar.md` Events table — "The control is a native `<button>`, so Enter and Space already activate `onbackClick`" — is no longer universally true and may mislead consumers about keyboard behavior. Update the description to note that with `backHref` the control is an anchor, so only Enter activates it by default.

docs/Toolbar.md
```diff
@@ -31,6 +31,7 @@ content. If `leftContent` snippet is provided, it replaces the default back butt
...
 | backIcon       | `string \| null` | No       | _none_ — an inline chevron drawn with `currentColor` | Custom image URL for the back button. Omit it for the built-in inline icon; pass `null` (or `''`) to render no back control at all. The component ships no network dependency of its own. |
 | backLabel      | `string`         | No       | `'Back'`                                             | Accessible name of the back button (`aria-label`). The icon itself is decorative; an empty or whitespace-only value falls back to `'Back'`.                                               |
+| backHref       | `string`         | No       | `-`                                                  | Renders the built-in back control as a real `<a href={backHref}>` instead of a `<button>` — for back navigation that works as a link (browser back/forward, open-in-new-tab, no-JS fallback), not only through a click callback. `onbackClick`/`onBackClick` still fire as the anchor's click handler when supplied. Ignored when `leftContent` is provided or the back control is hidden (`showBackButton={false}` or `backIcon={null}`). |
```
Tara-ag: 💡 MINOR: Docs still describe back control as always a native <button>. The Events table says: "The control is a native <button>, so Enter and Space already activate onbackClick without this." With backHref, the control can be an anchor, so this statement is no longer universally true and may mislead consumers about keyboard behavior.

Update the onkeydown event description to note that when backHref is set the control is an <a>, so Enter activates it and Space does not; consumers relying on Space activation should supply onkeydown or use the callback-only button.

Tara-ag (changes requested):
Reviewed Toolbar `backHref` changes. One MAJOR accessibility concern (Space-key activation lost on anchor), one MINOR docs inaccuracy, and one SUGGESTION for web-component reflection.

Tara-ag:

The PR adds a backward-compatible `backHref` prop to Toolbar and correctly preserves the default button behavior. However, the gate accepted two findings: a MAJOR issue where Space-key activation on the anchor bypasses native click navigation (breaking href follow-through and preventing consumer cancellation), and a MINOR issue where the web-component `backHref` prop lacks `reflect: true`. Address the MAJOR issue before considering this fully safe.

- 💡 **MINOR**: New backHref web-component prop should reflect to the host attribute — `src/wc/components/Toolbar.wc.svelte:10`
  The bootstrapped project standard says new string/enum props in web-component wrappers should use `reflect: true` so host attributes stay in sync and can be targeted from the light DOM. `backHref` is currently declared without reflection, unlike `showBackButton` and `text` in the same wrapper.
- 💡 **MINOR**: backHref web-component prop should reflect to host attribute — `src/wc/components/Toolbar.wc.svelte:10`
  The bootstrapped standards note that new string/enum props in web-component wrappers should use reflect: true so host attributes stay in sync and can be targeted from the light DOM. backHref is declared with attribute: 'back-href' but no reflect: true, unlike text/showBackButton. This means setting the property programmatically will not update the host attribute, and CSS/JS selectors on [back-href…

Tara-ag (commented):

src/lib/Toolbar/Toolbar.svelte
```diff
@@ -34,8 +39,11 @@
...
```
Tara-ag, line 54: ⚠️ MAJOR: Anchor back control loses Space-key activation. When `backHref` is set, the control renders as an `<a>`. Native anchors do NOT activate on Space; the existing keyboard test only asserts Enter. A consumer relying on the default button's Enter/Space contract will experience a regression for keyboard users.

Add a keydown handler that calls `onbackclick` for Space when `hasBackHref` is true, or document that Space activation is intentionally unavailable for link mode. If keeping the public callback contract, implement it:

```ts
function handleBackKeydown(event: KeyboardEvent): void {
  onkeydown?.(event);
  if (hasBackHref && event.key === ' ') {
    event.preventDefault();
    onbackclick?.();
  }
}
```

Then bind `onkeydown={handleBackKeydown}`.

src/wc/components/Toolbar.wc.svelte
Tara-ag, line 10: 💡 MINOR: Web-component `backHref` prop does not reflect to attribute. The new custom-element prop uses `{ type: 'String', attribute: 'back-href' }` without `reflect: true`. The library's own precedent for String props that are part of the visible host contract (`text`, `showBackButton`) reflects them. Non-reflection means HTML consumers cannot read/set `back-href` declaratively consistently with other Toolbar host attributes.

Add `reflect: true` to `backHref` so the host attribute stays in sync:

```ts
backHref: { type: 'String', attribute: 'back-href', reflect: true },
```

docs/Toolbar.md
Tara-ag, line 1: 💡 MINOR: Docs event description becomes inaccurate for `backHref` mode. The Events table says the control is a native `<button>`, so Enter and Space already activate `onbackclick`. With `backHref` it is an `<a>`, making that statement false for the new mode.

Update the description to note that when `backHref` is set the control is an `<a>` and only Enter activates by default; Space requires the `onkeydown` handler or is not supported.

Tara-ag (changes requested):
One MAJOR accessibility concern: the new `backHref` anchor mode drops Space-key activation that the default button provides, and the test only covers Enter. Two MINOR items: the docs event description is now inaccurate for anchor mode, and the web-component `backHref` prop should reflect to the host attribute for consistency. Otherwise the change is additive, backward-compatible, and well-tested.

Tara-ag (changes requested):
Changes requested. The additive `backHref` feature is welcome and mostly well-implemented, but the anchor mode introduces a keyboard-accessibility regression and two smaller boundary/doc inconsistencies that need resolution before merge.

Verified issues (all already threaded on the PR):

1. **MAJOR — Accessibility**: When `backHref` is set, the back control renders as `<a>`, which native browsers activate on Enter but **not** Space. The existing button contract gave consumers Enter+Space for free, and the new test only asserts Enter. Add a keydown handler that intercepts Space in anchor mode and invokes `onbackclick`, or explicitly document the reduced keyboard contract.

2. **MINOR — Web-component boundary**: `backHref` is declared without `reflect: true`. Other visible string props on this custom element (e.g. `text`) reflect, so `back-href` should too for declarative HTML parity.

3. **MINOR — Documentation**: The Events table still describes the control as a native `<button>` that handles Enter and Space automatically. That is no longer true when `backHref` is set; update the `onkeydown` description to cover anchor mode.

Once the Space-key behavior is addressed and the two minor items are fixed, this is good to approve.

murdore:
## Video evidence

**What the recording shows**

Concatenation of two real Playwright recordings from a run against the branch's actual vite build served on port 43281: clip 1 (test 'an unconfigured toolbar keeps rendering the back control as a <button>, not a link') shows Playwright loading /components/toolbar and inspecting the toolbar-root back control, confirming it is a <button> with no href attribute; clip 2 (test 'backHref renders the back control as a real <a> carrying the given href') shows the toolbar-back-href demo block on the same route, where the back control is a real <a> tag carrying href="#toolbar-back-href-demo", plus its aria-label and svg icon intact. Together they show both required behaviors: an unconfigured toolbar still renders a plain button, and a toolbar with backHref renders the back control as a genuine anchor with a working href. A second file, pr-536-toolbar-FAILS-on-release.webm (same directory, 450220 bytes), is saved as corroborating evidence of the backHref test failing (element not found/timeout) when run against source reverted to origin/release.

> 4 of 5 tests failed against origin/release with the spec kept: "backHref renders the back control as a real <a>" timed out waiting for getByTestId('toolbar-back-href') (element never appears because +page.svelte on release has no backHref demo block or testId); "onbackclick still fires as the anchor click handler", "the anchor activates from the keyboard", and "leftContent still takes precedence over backHref" failed the same way. Only the unrelated "unconfigured toolbar keeps rendering a button" test passed, since that behavior predates this PR. This proves the recording is genuine PR-dependent evidence, not a recording of an already-passing baseline.

- passing, on this branch: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-536-toolbar.webm` (115,043 bytes)
- failing, on `release`: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-536-toolbar-FAILS-on-release.webm` (450,220 bytes)

- The video's own visual frames are static and would not, by themselves, prove which tag the back control renders as (Playwright's recorded assertions are DOM-API calls, not visible interactions/inspector overlays) — the fix is only demonstrated because I independently re-ran the spec against branch (pass) and a clean release checkout (fail, matching reporter's failure mode exactly), not because the pixels show it. Future evidence videos for this kind of prop-swap change would benefit from an on-screen assertion (e.g. a visible outline or a printed result) rather than relying solely on a silent pass/fail.
- One earlier-noted non-blocking oddity: `pnpm lint`'s npm script banner printed a path belonging to a different worktree (feat/chat-composer-async-submit) — this is a benign npm_lifecycle-banner artifact from a shared shell environment, not a wrong-directory run; confirmed by running `pnpm exec eslint src/lib/Toolbar --max-warnings=0` directly from this worktree with exit 0.

Tara-ag (commented):

src/lib/Toolbar/Toolbar.svelte
Tara-ag, line 54: ⚠️ MAJOR: When `backHref` is set, the control renders as an `<a>`. Native `<button>` activates on Enter and Space; native `<a>` only activates on Enter. The existing `toolbar-page-header` test asserts both Enter and Space work for the default button, but the new `backHref` path has no equivalent Space test. Because the element is now a link, Space will scroll the page instead of firing `onbackclick`, creating a behavioral regression for keyboard users when this prop is used.

Suggested fix: add an `onkeydown` handler that, when `hasBackHref` is true and the key is Space, calls `event.preventDefault()` and invokes `onbackclick?.(event)` so the anchor behaves like the button. Alternatively, document that `backHref` switches to link semantics and Space no longer activates the control.

docs/Toolbar.md
Tara-ag, line 1: 💡 MINOR: The events table says "The control is a native `<button>`, so Enter and Space already activate `onbackclick` without this." With `backHref` the control renders as an `<a>`, so Space does not activate it natively and this statement is no longer universally true.

Suggested fix: update the `onkeydown` description to note that when `backHref` is set the control renders as a real link and only Enter activates it natively; Space activation must be provided by the consumer via `onkeydown` if required.

src/lib/Toolbar/Toolbar.svelte
Tara-ag, line 1: 💡 MINOR: The `.back` rule resets button-specific `appearance` but does not set `text-decoration` for the anchor case. Browsers default `<a>` to underline and a link color, which will clash with the toolbar's design and the existing button appearance. `Button.svelte` avoids this by setting `text-decoration: var(--button-text-decoration, none)` on `.button-el`.

Suggested fix: add `text-decoration: var(--toolbar-back-text-decoration, none);` to `.back` (and ensure `color: var(--toolbar-back-icon-color, inherit);` is honored — it is already set). Consider exposing a CSS variable so consumers can opt back into underlines if needed.

Tara-ag (changes requested):
Yama review complete. One MAJOR accessibility concern and two MINOR issues found; see inline comments.

Tara-ag (commented):
Review in progress — posting inline findings as they are verified.

src/wc/components/Toolbar.wc.svelte
Tara-ag, line 10: 💡 MINOR: New string props in web-component wrappers should use `reflect: true` so the host attribute stays in sync and can be targeted from the light DOM. `backHref` is missing reflection here, unlike `showBackButton` and `text` in the same wrapper.

Suggested fix:
```ts
backHref: { type: 'String', attribute: 'back-href', reflect: true },
```

Tara-ag (approved):
Approved with one minor web-component parity note.

src/wc/components/Toolbar.wc.svelte
Tara-ag, line 10: 💡 MINOR: `backHref` should use `reflect: true` so the host `back-href` attribute stays in sync when the property is set programmatically, and so light-DOM selectors like `[back-href]` work. Other scalar props such as `text` and `showBackButton` already reflect.

Suggested change:
```js
backHref: { type: 'String', attribute: 'back-href', reflect: true },
```

Tara-ag (approved):
Approved with one minor web-component parity note: `backHref` should reflect to the host attribute. The change is otherwise safe — it adds an optional prop, preserves the default `<button>` behavior byte-for-byte, updates docs and adds focused Playwright coverage.

src/lib/Toolbar/Toolbar.svelte
```diff
@@ -23,6 +24,22 @@
...
+
+  // A native <a> does not fire `click` on Space — only Enter — so the keyboard activation the
+  // old <button> gave for free would silently regress when backHref swaps the tag. Enter already
+  // reaches `onbackclick` through the anchor's native click, so only Space needs help here; the
+  // consumer's own `onkeydown` still runs unconditionally, same as it always has.
+  const handleBackKeydown = (event: KeyboardEvent): void => {
+    onkeydown?.(event);
+    if (hasBackHref && event.key === ' ') {
+      event.preventDefault();
+      onbackclick?.();
```
Tara-ag, line 40: ⚠️ MAJOR: Space-key activation on the `backHref` anchor bypasses native click navigation. `handleBackKeydown` calls `onbackclick?.()` directly on Space instead of dispatching the anchor's native click, so the `href` is not followed. A consumer's `onkeydown` also cannot cancel activation because the Space branch does not check `event.defaultPrevented` after forwarding.

src/wc/components/Toolbar.wc.svelte
Tara-ag, line 10: 💡 MINOR: The new `backHref` prop in the web-component wrapper is missing `reflect: true`. Other scalar props like `text` and `showBackButton` reflect so the host attribute stays in sync and can be targeted from the light DOM. Add `reflect: true` to keep the custom-element API consistent.

Tara-ag (commented):
Yama review finalized. Two findings accepted by the gate: one MAJOR (Space-key activation on backHref anchor bypasses native click navigation) and one MINOR (web-component backHref missing reflect: true). The PR is already merged; these are recorded for follow-up.

PR #537, opened 2026-09-06
feat(emptystate): add density variant and --empty-state-align-items

### LINKED ISSUE #523: EmptyState: A page-vs-panel density variant and a true left-align override on EmptyState (align-i…
## Need
A page-vs-panel density variant and a true left-align override on EmptyState (align-items is hardcoded 'center')

## Closest today
EmptyState (classes/cssVars hooks, but no --empty-state-align-items)

## Proposal
Add `--empty-state-align-items` alongside the existing `--empty-state-actions-*` vars, and a `density?: 'page'|'panel'` prop that sets sane padding/gap defaults without a wrapper class.

## Where it bites
Found while migrating TARA (a Svelte 5 / SvelteKit 2 consumer pinned to 3.3.0) onto the library. Without this the following consumer files can only migrate part-way and keep hand-rolled markup for the rest: `+page.svelte`, `ActivityRail.svelte`.

murdore:
Closes #523. TARA's own EmptyState wrapper (apps/web/src/lib/components/
EmptyState.svelte) and its override sheet (packages/ui/src/styles/
components.css) hand-rolled two things the library had no hook for:

1. A page-vs-panel density concept, expressed as .tara-ui-empty-page /
   .tara-ui-empty-panel classes reaching in through the `classes` prop to
   override --empty-state-padding, --empty-state-gap and friends per
   context (a full route vs. a popover/rail panel).
2. Left alignment, set as a bare `align-items: flex-start` on
   .tara-ui-activity-rail-empty because the root element's align-items was
   hardcoded 'center' with no CSS variable reaching it (only the unrelated
   --empty-state-actions-align-items existed, scoped to the actions child).

This adds:

- `density?: 'page' | 'panel'` (properties.ts), reflected onto the root as
  `data-density`. Omitting it (the default) reproduces today's exact
  padding/gap. `density="page"` and `density="panel"` each supply their
  own padding/gap fallback via an attribute-selector rule scoped under
  `.empty-state[data-density='...']` — more specific than the base rule,
  so it only ever applies when `data-density` is actually set. Either
  density can still be overridden per-instance with the existing
  `--empty-state-padding` / `--empty-state-gap` variables, which are read
  first in both rules and so always win.
- `--empty-state-align-items` CSS variable, replacing the hardcoded
  `align-items: center` with `var(--empty-state-align-items, center)` —
  same default, now reachable without a wrapper class.

Both are additive: default markup and computed styles are byte-for-byte
unchanged when `density` is omitted and no `--empty-state-align-items`
override is set (verified via the new Playwright spec, which fails on 3
of its 4 cases against the pre-change component). The `sui-empty-state`
web-component wrapper gets `density` added to its `props` map to keep
wc-parity; docs/EmptyState.md documents both under Props, CSS Variables
and a new Density / Left-Aligned usage section, Svelte and web-component
alike.

---

### Verification

**Spec runs on this branch** — `tests/empty-state-density-align.test.ts` — 4 Playwright tests against a real build.


tests/empty-state-density-align.test.ts
```diff
@@ -0,0 +1,48 @@
...
+// to reach it, and had no page-vs-panel density concept at all — consumers
+// needing either had to hand-roll a wrapper class (see the .tara-ui-empty-page /
+// .tara-ui-empty-panel / .tara-ui-activity-rail-empty recipes this issue was
+// filed against). Both are additive: omitting `density` and never setting
+// `--empty-state-align-items` must reproduce today's exact output.
+test.describe('EmptyState density and align-items (#523)', () => {
+  test("default (no density prop) keeps today's padding, gap and centered alignment", async ({
+    page
+  }) => {
+    await page.goto('/components/empty-state');
```
Tara-ag, line 13: ⚠️ MAJOR: The new spec navigates to the demo page and selects elements via demo-specific `data-pw` attributes (`empty-state-density-default`, `empty-state-density-page`, etc.). This couples the component contract test to the demo page markup, making it brittle when the demo is reorganized and causing it to assert on demo wiring rather than the component itself. Per the repo's testing standards, component-contract tests should use isolated fixtures/harnesses instead of demo-page selectors. Move these assertions to a standalone fixture under `tests/fixtures/` (or a Vitest/jsdom harness) and select by stable structure/class rather than demo-only `testId` props.

Tara-ag (approved):
Approved with one MAJOR note: the new Playwright spec is coupled to demo-page markup and demo-specific `data-pw` attributes. The implementation itself is additive and backward-compatible, but the test should be moved to an isolated fixture/harness per repo testing standards.

tests/empty-state-density-align.test.ts
Tara-ag, line 13: ⚠️ MAJOR: This Playwright spec navigates to the demo page (`/components/empty-state`) and selects elements via demo-specific `data-pw` attributes. Per the repository's testing standards, component-contract tests should use isolated fixtures or harnesses instead of selecting through demo-page markup and demo-specific `data-pw` attributes. Demo pages change independently of component contracts, so coupling tests to them risks false positives and makes the test harder to reason about.

Suggestion: Move these assertions to an isolated fixture under `tests/fixtures/` (or a Vitest/jsdom harness) and select by stable structure/class rather than demo-specific `data-pw` values.

Tara-ag (commented):
Approved with one MAJOR inline note on test coupling. The component changes are additive and backward-compatible: `density` is optional with no default, `--empty-state-align-items` preserves the previous `center` fallback, and the web-component wrapper declares `density` to maintain parity. Docs and demo page are updated consistently. The new Playwright spec, however, selects through demo-page markup and demo-specific `data-pw` attributes, which conflicts with the repo's preference for isolated component fixtures/harnesses. This is not blocking on its own, but should be addressed in a follow-up or before merge if the team agrees.

Tara-ag (approved):
Reviewed file-by-file. The change is additive and backward-compatible: `density` is optional, defaults preserve existing styling, `--empty-state-align-items` defaults to the previous `center`, and the web-component wrapper maps the new prop. Docs and tests are included. No blocking issues found.

Tara-ag (commented):

src/wc/components/EmptyState.wc.svelte
```diff
@@ -7,6 +7,7 @@
       description: { type: 'String', reflect: true },
       classes: { type: 'String' },
       testId: { type: 'String', attribute: 'test-id' },
+      density: { type: 'String' },
```
Tara-ag, line 10: 💡 MINOR: Web-component `density` prop lacks `reflect: true`. Comparable string enum props in this repo’s wrappers consistently reflect (Button.type, Modal.size/align, Toast.type/direction, LoadingDots.animation, ThemeSwitcher.mode/value). Without reflection, a consumer cannot target `sui-empty-state[density='page']` from the light DOM, while the Svelte component exposes `data-density` on its root div. Add `reflect: true` to keep parity.

```svelte
      density: { type: 'String', reflect: true },
```

Tara-ag (approved):
Approved with one minor web-component parity note. The Svelte-side change is additive and backward-compatible; the only gap is that the custom-element `density` prop should reflect to match the convention used by other string enum props in this repo.

src/wc/components/EmptyState.wc.svelte
Tara-ag, line 10: 💡 MINOR: Web-component `density` prop lacks `reflect: true`, diverging from the repo's convention for string enum props. Comparable props in other wrappers (e.g., Button.type, Modal.size/align, Toast.type/direction, ThemeSwitcher.mode/value) use `reflect: true` so the host attribute stays in sync and can be targeted from the light DOM. Since the Svelte component exposes `data-density` on the root div, the web-component wrapper should reflect `density` to the host attribute for parity.

**Suggested fix:**
```svelte
density: { type: 'String', reflect: true },
```

Tara-ag (commented):
Approved with one minor web-component parity note. The change is additive and backward-compatible: `density` defaults to unset (preserving existing padding/gap), and `--empty-state-align-items` defaults to `center`. Docs, demo page, and Playwright tests are included. Please consider adding `reflect: true` to the web-component `density` prop to match the convention used by other string enum props.

murdore:
## Video evidence

**What the recording shows**

Navigates to /components/empty-state on the built PR branch. Sequentially highlights (coloured outline + scrollIntoView) and pauses on: (1) the default-density EmptyState (blue outline) while asserting its unchanged 32px/16px padding, 0px gap, center alignment — proving backward compatibility; (2) density="page" (amber outline) while asserting its roomier 48px/24px padding and 12px gap; (3) density="panel", the compact variant (green outline) while asserting its tighter 16px/12px padding and 4px gap, directly beside the other two in the same 'Density' demo section, plus a live boundingClientRect height comparison proving panel < default < page; (4) the --empty-state-align-items: flex-start override (pink outline) on a separate instance, asserting align-items resolves to flex-start with no wrapper class. Each stage holds for ~0.8-1.5s so the recording is watchable, not just a fast pass.

> Locator getByTestId('empty-state-density-default') not found — toBeVisible() timed out after 5000ms because the density/align-items feature (data-pw testids, data-density attribute, Density and align-override sections in +page.svelte) does not exist on origin/release; EmptyState.svelte there has no `density` prop and a hardcoded align-items: center with no CSS var.

- passing, on this branch: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-537-emptystate.webm` (160,583 bytes)
- failing, on `release`: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-537-emptystate-FAILS-on-release.webm` (93,390 bytes)

- git fetch origin confirmed origin/release is still at 7b44a5a (chore(release): 3.5.1) — unchanged since the rebase report, so no re-rebase is needed, but this should be re-checked again immediately before merge in case release moves before then.
- The throwaway evidence spec (test-results-evidence/) was already deleted by the reporter before I inspected the worktree; I could not directly re-run it myself, but I independently confirmed the feature's absence on origin/release by diffing source files, which is sufficient corroboration of the reporter's failure claim.
- publint build output emitted one pre-existing warning (wc export interpreted as CJS vs ESM) unrelated to this PR's changes — not a new issue introduced here, build still exits 0.

Tara-ag (approved):
Approved. The change is purely additive: a new optional `density` prop and a new `--empty-state-align-items` CSS variable, with no changes to existing props, exports, or CSS custom property names. Default behavior is preserved, and the new Playwright spec verifies backward compatibility. Docs, demo page, and web-component wrapper are all kept in sync.

PR #538, opened 2026-09-06
feat(markdowntext): add a sanitize prop with protocol and tag allow-lists

### LINKED ISSUE #524: MarkdownText: A configurable sanitizer/link-allowlist and an accessible table caption on Markdown…
## Need
A configurable sanitizer/link-allowlist and an accessible table caption on MarkdownText

## Closest today
MarkdownText (fixed http/https/mailto/tel allowlist, no table-caption prop)

## Proposal
Expose a `RenderMarkdownOptions`-shaped `sanitize?: {allowedProtocols?, allowedTags?, disableTaskLists?}` prop and a `tableLabel?: string` for the rendered table wrapper's aria-label.

## Where it bites
Found while migrating TARA (a Svelte 5 / SvelteKit 2 consumer pinned to 3.3.0) onto the library. Without this the following consumer files can only migrate part-way and keep hand-rolled markup for the rest: `+page.svelte`, `ApprovalCard.svelte`, `DecisionBubble.svelte`, `Transcript.svelte`.

murdore:
Closes #524.

TARA's chat UI (apps/web/src/lib/components/ApprovalCard.svelte,
DecisionBubble.svelte, Transcript.svelte) could not use MarkdownText's
`markdown` prop directly: the component's sanitizer had no way to drop
markdown-generated elements it didn't want (images in particular), so
the app pre-mangled the raw source through DOMPurify's `sanitiseMarkdownSource`
before ever handing it to MarkdownText, and kept a second, unused,
hand-rolled `renderMarkdown`/DOMPurify pipeline around as a fallback.

MarkdownText gains an optional `sanitize` object (type
`MarkdownSanitizeOptions`) with three narrowing options, all of them new
in this commit and all off by default:

- `allowedProtocols?: string[]` -- restricts which URL protocols survive
  on rendered links and images, intersected with the library's own
  defaults for each surface (`http:`/`https:`/`mailto:`/`tel:` for links,
  `http:`/`https:` for images). Because it intersects, it can only narrow
  what already renders, never widen it: listing `javascript:` cannot
  resurrect it. An empty array drops every link and image, keeping their
  text.
- `allowedTags?: string[]` -- restricts which markdown-generated tags
  render as themselves (`a`, `img`, `strong`, `em`, `del`, `code`,
  `pre`, `blockquote`, `ul`, `ol`, `table`, `hr`, `h1`-`h6`). There is
  no preset default to intersect with here, unlike `allowedProtocols`:
  omitting it keeps every tag rendering exactly as it does today.
  A disallowed tag keeps its parsed inner content as plain text instead
  of the element -- the same "keep the text, drop the element" contract
  the protocol allow-list uses for an unsafe link or image.
  Raw HTML is still escaped unconditionally regardless of this list;
  there is still no opt-out from that guarantee.
- `disableTaskLists?: boolean` -- renders GFM task-list items
  (`- [ ] done`) as plain list text instead of a checkbox. The checkbox
  already renders `disabled` by default, so this is a presentation
  choice, not a safety one. Defaults to `false`.

All three are additive and off by default, so `renderMarkdown`/`MarkdownText`
without a `sanitize` prop (or without these keys inside it) behave
byte-for-byte as before. Implemented as per-token `RendererObject`
overrides in a `Marked` instance cached by a key that folds in the
resolved protocol sets, tag set and task-list toggle, so a
differently-configured `sanitize` never shares a renderer with another.

Also documents the options in docs/MarkdownText.md and docs/_index.json.

---

### Verification

**Spec runs on this branch** — `src/lib/MarkdownText/markdown-sanitize.test.ts` — 17 vitest cases across three describe blocks (`allowedProtocols`, `allowedTags`, `disableTaskLists`), part of a green 431-test suite.

### DIFF
--- a/docs/MarkdownText.md
+++ b/docs/MarkdownText.md
@@ -36,13 +36,14 @@ The `marked` package is a peer dependency, needed only when you actually use `Ma
 
 ## Props
 
-| Prop       | Type      | Required | Default | Description                                                                                                                                                                  |
-| ---------- | --------- | -------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| markdown   | `string`  | Yes      | `-`     | Markdown source. Raw HTML inside it is escaped and shown as text; unsafe link/image protocols are stripped while their text stays.                                           |
-| breaks     | `boolean` | No       | `false` | Render single newlines as `<br>` (GFM "breaks" mode).                                                                                                                        |
-| testId     | `string`  | No       | `-`     | `data-pw` (and native `testID`) on the root element.                                                                                                                         |
-| classes    | `string`  | No       | `-`     | Class string on the root element.                                                                                                                                            |
-| tableLabel | `string`  | No       | `-`     | Names the scrollable table region: adds `role="region"` and `aria-label` to the wrapper. Without it the wrapper is still focusable and scrollable but announces no landmark. |
+| Prop       | Type                     | Required | Default | Description                                                                                                                                                                  |
+| ---------- | ------------------------ | -------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| markdown   | `string`                 | Yes      | `-`     | Markdown source. Raw HTML inside it is escaped and shown as text; unsafe link/image protocols are stripped while their text stays.                                           |
+| breaks     | `boolean`                | No       | `false` | Render single newlines as `<br>` (GFM "breaks" mode).                                                                                                                        |
+| testId     | `string`                 | No       | `-`     | `data-pw` (and native `testID`) on the root element.                                                                                                                         |
+| classes    | `string`                 | No       | `-`     | Class string on the root element.                                                                                                                                            |
+| tableLabel | `string`                 | No       | `-`     | Names the scrollable table region: adds `role="region"` and `aria-label` to the wrapper. Without it the wrapper is still focusable and scrollable but announces no landmark. |
+| sanitize   | `MarkdownSanitizeOptions` | No       | `-`     | Narrows the link/image protocol allow-list, the set of tags allowed to render as themselves, and whether task-list checkboxes render. See the Security model section below.  |
 
 ## Security model
 
@@ -50,6 +51,21 @@ The `marked` package is a peer dependency, needed only when you actually use `Ma
 - **URL protocol allow-list.** `[x](javascript:…)` and `[x](data:…)` render as plain text without an anchor; `![x](javascript:…)` renders the alt text without an `<img>`. Entity-smuggled protocols (`jav&#x09;ascript:`) fail scheme parsing and are treated as relative paths, where attribute-escaping keeps them inert.
 - **External links are tab-safe.** `http:`/`https:` links open with `target="_blank" rel="noopener noreferrer"` (the same default the library's Button/Card apply); relative, `mailto:` and `tel:` links keep same-tab navigation. Autolinks and bare URLs go through the same protocol guard as explicit links.
 - **Pre-sanitized HTML has its own prop.** If you already hold trusted, sanitized HTML, use `ChatMessage`'s `html` prop — `markdown` is for untrusted source text.
+- **The allow-list can be narrowed, never widened.** Pass `sanitize={{ allowedProtocols: [...] }}` when your own link policy is stricter than the library default (for example, no live `mailto:`/`tel:` in a support chat). It is intersected with the library's own allow-list for each surface (links vs. images), so listing an unsafe scheme like `javascript:` is a no-op, not a hole. Omitting `sanitize` keeps today's defaults exactly as they are — this is additive, not a replacement for the built-in list:
+
+  ```svelte
+  <!-- Only http(s) links/images; mailto: and tel: are dropped, keeping their text. -->
+  <MarkdownText {markdown} sanitize={{ allowedProtocols: ['http:', 'https:'] }} />
+  ```
+
+- **The tag allow-list narrows from "everything", never from a preset list.** Unlike `allowedProtocols`, there is no built-in default to intersect with — every tag `marked`'s GFM output can produce (`a`, `img`, `strong`, `em`, `del`, `code`, `pre`, `blockquote`, `ul`, `ol`, `table`, `hr`, `h1`–`h6`) renders normally until `sanitize.allowedTags` is supplied. Once it is, a disallowed tag renders its parsed inner content as plain text instead of the element — the same "keep the text, drop the element" contract the protocol allow-list uses. An unrecognised name is ignored, not an error. This is independent of the raw-HTML escaping guarantee above, which has no opt-out:
+
+  ```svelte
+  <!-- Headings, lists and tables render as plain text; links/emphasis still render. -->
+  <MarkdownText {markdown} sanitize={{ allowedTags: ['a', 'strong', 'em'] }} />
+  ```
+
+- **Task-list checkboxes can render as plain text instead.** GFM task items (`- [ ] done`) already render a `disabled` checkbox by default — it never submits or toggles — so `sanitize={{ disableTaskLists: true }}` is a presentation choice, not a safety one, for surfaces that would rather show the list item's text with no checkbox glyph at all. Defaults to `false`, keeping today's checkbox rendering.
 
 ## Wide tables
 
@@ -120,11 +136,26 @@ export type MarkdownTextProperties = {
   testId?: string;
   classes?: string;
   tableLabel?: string;
+  sanitize?: MarkdownSanitizeOptions;
+};
+
+export type MarkdownSanitizeOptions = {
+  // Intersected with the library's own default allow-list per surface (links,
+  // images), so it can only narrow what renders, never widen it.
+  allowedProtocols?: string[];
+  // No preset default to intersect with -- every tag renders until this is
+  // supplied, then it narrows from "everything". A disallowed tag keeps its
+  // text and drops the element.
+  allowedTags?: string[];
+  // Render task-list items as plain text instead of a (already-disabled)
+  // checkbox. Defaults to false.
+  disableTaskLists?: boolean;
 };
 
 export type RenderMarkdownOptions = {
   breaks?: boolean;
   inline?: boolean;
   tableLabel?: string;
+  sanitize?: MarkdownSanitizeOptions;
 };
 ```
--- a/src/lib/MarkdownText/markdown.ts
+++ b/src/lib/MarkdownText/markdown.ts
@@ -1,6 +1,6 @@
 import { Marked } from 'marked';
 import type { RendererObject, Tokens } from 'marked';
-import type { RenderMarkdownOptions } from './properties';
+import type { MarkdownSanitizeOptions, RenderMarkdownOptions } from './properties';
 
 /**
  * Chat content comes from models and users, not from the app's own templates,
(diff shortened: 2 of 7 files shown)


Tara-ag (approved):
No blocking issues found. The new `sanitize` prop is purely additive, defaults preserve existing behavior, protocol narrowing intersects with the library defaults (cannot widen), raw HTML remains unconditionally escaped, and the web-component wrapper mirrors the new Object prop. Tests and docs are updated.

docs/MarkdownText.md
```diff
@@ -36,20 +36,36 @@ The `marked` package is a peer dependency, needed only when you actually use `Ma
...
+  <!-- Only http(s) links/images; mailto: and tel: are dropped, keeping their text. -->
+  <MarkdownText {markdown} sanitize={{ allowedProtocols: ['http:', 'https:'] }} />
+  ```
+
+- **The tag allow-list narrows from "everything", never from a preset list.** Unlike `allowedProtocols`, there is no built-in default to intersect with — every tag `marked`'s GFM output can produce (`a`, `img`, `strong`, `em`, `del`, `code`, `pre`, `blockquote`, `ul`, `ol`, `table`, `hr`, `h1`–`h6`) renders normally until `sanitize.allowedTags` is supplied. Once it is, a disallowed tag renders its parsed inner content as plain text instead of the element — the same "keep the text, drop the element" contract the protocol allow-list uses. An unrecognised name is ignored, not an error. This is independent of the raw-HTML escaping guarantee above, which has no opt-out:
+
+  ```svelte
+  <!-- Headings, lists and tables render as plain text; links/emphasis still render. -->
+  <MarkdownText {markdown} sanitize={{ allowedTags: ['a', 'strong', 'em'] }} />
+  ```
```
Tara-ag, line 66: 💡 MINOR: The description says a disallowed tag renders its parsed inner content as "plain text", but the renderer actually calls `this.parser.parseInline` / `this.parser.parse` on the inner tokens, so nested tags that ARE in `allowedTags` are preserved. For example, `# **Loud** Title` with `allowedTags: ['strong']` drops `<h1>` but keeps `<strong>Loud</strong>`. Update the wording here and in `src/lib/MarkdownText/properties.ts` to "parsed inner content" or clarify that still-allowed nested markup is preserved, so consumers don't expect recursive stripping.

src/lib/MarkdownText/markdown.ts
```diff
@@ -68,27 +68,140 @@ function escapeHtml(value: string): string {
...
+function resolveProtocolSets(sanitize?: MarkdownSanitizeOptions): {
+  links: Set<string>;
+  images: Set<string>;
+} {
+  return {
+    links: narrow(SAFE_LINK_PROTOCOLS, sanitize?.allowedProtocols),
+    images: narrow(SAFE_IMAGE_PROTOCOLS, sanitize?.allowedProtocols)
+  };
+}
+
```
Tara-ag, line 96: ⚠️ MAJOR: `narrow()` assumes `allowedProtocols` is an array of strings and calls `.map(...toLowerCase())` without runtime validation. Web-component and plain-JS consumers can pass `{ allowedProtocols: 'https:' }` or `{ allowedProtocols: ['https:', null] }`, causing a TypeError at render time. Add `Array.isArray` and `typeof === 'string'` guards, and do the same for `resolveTagAllowList()` which currently does `new Set(allowedTags)` on any truthy value.

src/lib/MarkdownText/markdown.ts
```diff
@@ -68,27 +68,140 @@ function escapeHtml(value: string): string {
...
+function tagAllowed(allowed: Set<string> | null, tag: string): boolean {
+  return allowed === null || allowed.has(tag);
+}
+
+function createSanitizingRenderer(
+  protocols: { links: Set<string>; images: Set<string> },
+  allowedTags: Set<string> | null,
+  disableTaskLists: boolean
+): RendererObject {
+  return {
```
Tara-ag, line 116: 💡 MINOR: `allowedProtocols` is normalized to lowercase, but `allowedTags` is stored as-is. The renderer compares against lowercase tag names (`'strong'`, `'img'`, `'h1'`, etc.), so passing `['IMG']` or `['Strong']` silently drops those elements. Normalize tag names to lowercase here to keep the two allow-lists consistent, and add a test for uppercase tag names.

Tara-ag (changes requested):
Reviewed file-by-file. Found one MAJOR runtime-safety issue (unvalidated `sanitize` object values can crash `renderMarkdown` for web-component/JS consumers) and two MINOR issues (documentation accuracy for `allowedTags` fallback, case-sensitivity inconsistency between `allowedTags` and `allowedProtocols`). The public API is additive and backward-compatible, and there are no blocking security or breaking-change concerns.

src/lib/MarkdownText/markdown.ts
Tara-ag, line 96: ⚠️ MAJOR: The `sanitize` prop is passed straight from consumers (including JS and web-component consumers) into `narrow()` and `resolveTagAllowList()`. `narrow()` calls `allowedProtocols.map((p) => p.toLowerCase())` without checking that `allowedProtocols` is an array or that its items are strings. A web-component consumer setting `sanitize = { allowedProtocols: 'https:' }` or `{ allowedProtocols: ['https:', null] }` will crash `renderMarkdown` with a TypeError. Likewise, `resolveTagAllowList()` does `new Set(allowedTags)` without verifying iterability; a truthy non-iterable value (e.g. a number) throws.

Add defensive guards: skip/ignore `allowedProtocols` unless it is an array, filter out non-string entries, and ignore `allowedTags` unless it is an array of strings. For example:
```ts
function narrow(defaults: Set<string>, allowedProtocols?: unknown): Set<string> {
  if (!Array.isArray(allowedProtocols)) return defaults;
  const requested = new Set(
    allowedProtocols.filter((p): p is string => typeof p === 'string').map((p) => p.toLowerCase())
  );
  return new Set([...defaults].filter((protocol) => requested.has(protocol)));
}

function resolveTagAllowList(allowedTags?: unknown): Set<string> | null {
  if (!Array.isArray(allowedTags)) return null;
  const strings = allowedTags.filter((t): t is string => typeof t === 'string');
  return strings.length ? new Set(strings) : new Set<string>();
}
```
Also add unit tests covering malformed values (string, null entries, non-iterable).

docs/MarkdownText.md
Tara-ag, line 66: 💡 MINOR: The docs state that a disallowed tag renders its parsed inner content as "plain markup-free text". In practice the renderer overrides call `this.parser.parseInline` / `this.parser.parse` on the inner tokens, so nested tags that ARE in `allowedTags` survive. For example, `# **Loud** Title` with `allowedTags: ['strong']` drops the `<h1>` but keeps `<strong>Loud</strong>`, which contradicts "markup-free".

Update the description to say "parsed inner content" or "inner content with any still-allowed markup preserved" instead of "plain (markup-free) text".

src/lib/MarkdownText/markdown.ts
Tara-ag, line 116: 💡 MINOR: `narrow()` lowercases every entry in `allowedProtocols`, but `resolveTagAllowList()` stores `allowedTags` exactly as provided and `tagAllowed()` compares against the renderer's lowercase tag names (`'strong'`, `'h1'`, `'img'`, etc.). A consumer passing `['IMG']` or `['Strong']` will silently drop every image or bold span even though the equivalent protocol option is case-insensitive. This inconsistency is surprising and not documented.

Normalize tag names to lowercase in `resolveTagAllowList` to match protocol handling:
```ts
function resolveTagAllowList(allowedTags?: string[]): Set<string> | null {
  return allowedTags ? new Set(allowedTags.map((tag) => tag.toLowerCase())) : null;
}
```
Add a test asserting `allowedTags: ['STRONG']` keeps `<strong>` output.

Tara-ag (changes requested):
Changes requested. The main concern is runtime safety of the new `sanitize` prop values when passed from web-component or plain-JS consumers. Two smaller documentation/consistency issues are also noted.

Tara-ag:

Reviewed file-by-file. The new `sanitize` prop is additive and backward-compatible, with no blocking security or public-API breaking changes. However, the `sanitize` object is trusted at runtime without validation, so malformed values from web-component or plain-JS consumers can crash `renderMarkdown`. Two smaller issues: the `allowedTags` fallback is documented as "plain text" but actually preserves nested allowed markup, and `allowedTags` is case-sensitive while `allowedProtocols` is normalized to lowercase.

- ⚠️ **MAJOR**: sanitize.allowedProtocols and allowedTags lack runtime guards for non-array/non-string values — `src/lib/MarkdownText/markdown.ts:96`
  The new `sanitize` prop is passed straight from consumers (including JS and web-component consumers) into `narrow()` and `resolveTagAllowList()`. `narrow()` calls `allowedProtocols.map((p) => p.toLowerCase())` without checking that `allowedProtocols` is an array or that its items are strings. A web-component consumer setting `sanitize = { allowedProtocols: 'https:' }` or `{ allowedProtocols: ['htt…
- 💡 **MINOR**: allowedTags fallback described as "plain markup-free text" but preserves nested allowed tags — `docs/MarkdownText.md:120`
  The docs and JSDoc state that a disallowed tag renders its parsed inner content as "plain markup-free text". In practice the renderer overrides call this.parser.parseInline / this.parser.parse on the inner tokens, so nested tags that ARE in allowedTags survive. For example, `# **Loud** Title` with `allowedTags: ['strong']` drops the `<h1>` but keeps `<strong>Loud</strong>`, which contradicts "mark…
- 💡 **MINOR**: allowedTags is case-sensitive while allowedProtocols is case-normalized — `src/lib/MarkdownText/markdown.ts:116`
  `narrow()` lowercases every entry in `allowedProtocols`, but `resolveTagAllowList()` stores `allowedTags` exactly as provided and `tagAllowed()` compares against the renderer's lowercase tag names (`'strong'`, `'h1'`, `'img'`, etc.). A consumer passing `['IMG']` or `['Strong']` will silently drop every image or bold span even though the equivalent protocol option is case-insensitive. This inconsis…

murdore:
## Video evidence

**What the recording shows**

This PR (feat(markdowntext): add a sanitize prop with protocol and tag allow-lists, PR #538) never touched the /components/markdown-text demo route, so I added a temporary (uncommitted, later fully reverted) side-by-side demo section to that route rendering the exact same markdown source (h1, bold, a GFM task list with one checked/one unchecked item, a table, a link, a blockquote, a fenced code block, an hr) twice: once through MarkdownText with no sanitize prop (unrestricted default) and once with sanitize={{ allowedTags: ['strong','em'], disableTaskLists: true }} (narrowed). The recording shows the browser open on that route, scroll to and pause on the default panel (rendering a real table, link, blockquote, code block, h1, hr, and 2 checkbox inputs -- all asserted in-spec), then scroll to and pause on the narrowed panel, which renders the SAME source but with only <strong>/<em> surviving as elements: the table, link, blockquote, pre, h1, hr and both checkboxes are all gone, replaced by their plain inert text ('Release notes', 'Docs', 'A blockquote.', 'console.log(\"code block\")', 'Ship the allow-list' are all still visible as text). Every one of those presence/absence claims is a Playwright assertion in the recorded run, not just a visual impression. I then proved this is genuinely the PR's doing: with the same demo-route edit and spec left in place, I reverted MarkdownText.svelte/markdown.ts/properties.ts/MarkdownText.wc.svelte/docs to origin/release (deleting markdown-sanitize.test.ts, which doesn't exist there), rebuilt, and reran the identical spec -- it failed, because the reverted MarkdownText silently ignores the unknown `sanitize` prop and the 'narrowed' panel still renders a real <table>. That failure video is saved locally at /private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-538-markdowntext-FAIL-without-pr.webm (241983 bytes, ~9.7s) as corroborating evidence, though only pr-538-markdowntext.webm was requested as the deliverable. The worktree was then fully restored (git checkout HEAD -- . && git clean -fd, both throwaway spec/config deleted) -- git status --short is clean and the branch sits at exactly one commit (3f2db33) ahead of origin/release.

> expect(narrowedPanel.locator('table')).toHaveCount(0) failed -- received 1. Without the PR's `sanitize` prop, MarkdownText.svelte silently ignores the extra prop, so the "narrowed" panel still renders a real &lt;table&gt; instead of falling back to inert text.

- passing, on this branch: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-538-markdowntext.webm` (213,891 bytes)
- failing, on `release`: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-538-markdowntext-FAIL-without-pr.webm` (241,983 bytes)

- The throwaway evidence spec/config and the temporary demo-route edit are confirmed fully removed: `git status --short` is clean, `git diff origin/release -- src/routes/components/markdown-text` is empty (the demo route matches release exactly), and no test-results-evidence/ or playwright.evidence.config.ts exists anywhere in the tree.

Tara-ag (approved):
Approved. The change is additive and backward-compatible: `sanitize` is optional, omitted behavior is byte-for-byte identical, the protocol allow-list intersects with (never replaces) the library default, raw HTML escaping remains unconditional, and the new type is re-exported through the public barrel. Web-component parity, docs, and unit tests are all in place.

PR #539, opened 2026-09-06
feat(sheet): add a centered dialog variant and a real heading option

### LINKED ISSUE #525: Sheet: A centered/floating dialog variant of Sheet and a real heading tag for its title
## Need
A centered/floating dialog variant of Sheet and a real heading tag for its title

## Closest today
Sheet (edge-anchored left/right/top/bottom only; title renders as a plain <span>)

## Proposal
Add `side: 'center'` (fixed-size, backdrop-fade, no edge transform) alongside the existing sides, and render the title through an `<h2>` (or a `headingLevel` prop) instead of a span.

## Where it bites
Found while migrating TARA (a Svelte 5 / SvelteKit 2 consumer pinned to 3.3.0) onto the library. Without this the following consumer files can only migrate part-way and keep hand-rolled markup for the rest: `LaunchSheet.svelte`, `ShareSheet.svelte`.

murdore:
Closes #525

Sheet only ever anchored to an edge (left/right/top/bottom) and always
rendered its title through a plain <span>. Consumers that needed a
centered, fixed-size confirmation/dialog surface with a real heading
element had no way to get one from this component and had to hand-roll
their own overlay/panel/heading instead (TARA's LaunchSheet.svelte is
one such workaround).

This adds two additive, opt-in pieces of API:

- `side="center"`: a new SheetSide value. Renders the panel as a
  fixed-size floating dialog, centered with `top/left: 50%` plus a
  `translate(-50%, -50%)` transform instead of stretching to an edge.
  Its transition is a fade only (the existing `fly` transition
  degrades to a pure fade when x/y are both 0, so no second transition
  primitive is introduced). Sized by its own new tokens
  --sheet-center-width (480px), --sheet-center-max-width
  (calc(100vw - 32px)) and --sheet-center-max-height (90vh), which are
  independent of --sheet-width/--sheet-height so existing edge-anchored
  themes are untouched.

- `headingLevel?: 1 | 2 | 3 | 4 | 5 | 6`: renders the `title` string
  through a real `<h1>`-`<h6>` (via `svelte:element`) instead of the
  default `<span>`, for correct document outlines. Omitted, the title
  keeps rendering as the exact same `<span class="sheet-title">`
  markup as before.

Both are opt-in and additive: an existing Sheet with no `side` (or any
of the pre-existing left/right/top/bottom values) and no `headingLevel`
renders byte-for-byte as it did before this change, verified by a
Playwright regression check in the new spec plus the full existing
Sheet suite.

Updated src/wc/components/Sheet.wc.svelte to keep web-component prop
parity for `headingLevel`, added demo sections to the sheet route for
the new tests, and updated docs/Sheet.md and docs/_index.json for the
new prop, side value and CSS variables.

---

### Verification

**Spec runs on this branch** — `tests/sheet-centered-variant.test.ts` — 4 Playwright tests against a real build.

**Fail-before** — Source reverted to `origin/release` with only the spec kept: **fails** on the centered-panel locator.


src/lib/Sheet/Sheet.svelte
```diff
@@ -143,7 +148,11 @@
       {#if typeof title === 'string' || showCloseButton}
         <div class="sheet-header">
           {#if typeof title === 'string'}
-            <span class="sheet-title">{title}</span>
+            {#if typeof headingLevel === 'number'}
```
Tara-ag: ⚠️ MAJOR: `headingLevel` is not validated at runtime. The component renders `svelte:element this={`h${headingLevel}`}` whenever `headingLevel` is a number. Via the web component (where TypeScript types don't apply), invalid values like `heading-level="7"`, `heading-level="0"`, `heading-level="-1"`, or `heading-level="foo"` produce invalid HTML elements (`<h7>`, `<h0>`, `<h-1>`, `<hNaN>`), breaking accessibility and DOM validity.

Validate that `headingLevel` is an integer in `[1, 6]` before rendering as a heading; fall back to the `<span>` otherwise. For example:
```ts
const headingTag = $derived(
  typeof headingLevel === 'number' &&
  Number.isInteger(headingLevel) &&
  headingLevel >= 1 &&
  headingLevel <= 6
    ? `h${headingLevel}`
    : null
);
```
Then render `<svelte:element this={headingTag} class="sheet-title">{title}</svelte:element>` only when `headingTag` is non-null, else the existing `<span>`.

Tara-ag (changes requested):
The centered dialog variant and `headingLevel` prop are additive and well-documented, and the web-component wrapper keeps API parity. However, `headingLevel` needs runtime validation before it can safely render arbitrary heading tags, especially for custom-element consumers where TypeScript types do not apply.

Tara-ag:

Additive Sheet feature (centered dialog variant + real heading option) is mostly well-implemented and backwards-compatible. However, the centered variant's CSS omits an explicit height, causing the panel to collapse to content height despite docs describing a 'fixed-size dialog'. Two minor quality items also need attention: brittle fixed-timeout waits in new Playwright tests and silent fallback for invalid headingLevel values. No blocking security or breaking-change criteria are met.

- ⚠️ **MAJOR**: Centered Sheet lacks explicit height, relying on flex content sizing — `src/lib/Sheet/Sheet.svelte:265`
  The `.sheet-panel.center` rule sets `width`, `max-width`, and `max-height` but does not set `height`. The panel uses `display: flex; flex-direction: column`, so its height collapses to the intrinsic height of the header + content + footer. For short confirmation dialogs this is fine, but it means `--sheet-center-max-height` only clamps content when the intrinsic height exceeds it; it does not give…
- 💡 **MINOR**: New Playwright tests use brittle `waitForTimeout` instead of transition end — `tests/sheet-centered-variant.test.ts:18`
  Both new tests in `tests/sheet-centered-variant.test.ts` call `await page.waitForTimeout(400)` to wait for the Sheet transition to complete. This is slower and flakier than awaiting the actual transition end event. The existing `sheet-anchor-and-overlay.test.ts` uses the same pattern, but adding more tests extends the cumulative runtime and increases flake risk on slow CI runners.
- 💡 **MINOR**: `headingTag` derivation silently ignores invalid `headingLevel` values — `src/lib/Sheet/Sheet.svelte:111`
  The runtime validation for `headingLevel` falls back to `<span>` when the value is outside 1-6 or not an integer. This is safe, but it is silent: a web-component consumer passing `heading-level="7"` or `heading-level="foo"` gets no warning and no heading, which can be confusing. The bootstrapped standards encourage validating runtime inputs and documenting the contract; consider logging a dev-mode…

src/lib/Sheet/Sheet.svelte
```diff
@@ -109,6 +114,20 @@
       onafterclose();
     }
   }
+
+  // headingLevel's `1 | 2 | 3 | 4 | 5 | 6` union only constrains Svelte-import
+  // consumers. Through the web component, attributes are untyped strings
+  // coerced to Number — heading-level="7"/"0"/"-1"/"foo" would otherwise reach
+  // `h${headingLevel}` unchecked and render an invalid tag (<h7>, <h0>, <h-1>,
+  // <hNaN>). Validate at runtime and fall back to the existing <span>.
```
Tara-ag, line 114: 💡 MINOR: The runtime guard for `headingLevel` silently falls back to `<span>` when an out-of-range value is supplied (e.g. via the web component). Consumers may not notice they passed an invalid level.

Consider adding a dev-mode warning when `headingLevel` is provided but invalid, so the typo is surfaced instead of silently ignored.

tests/sheet-centered-variant.test.ts
```diff
@@ -0,0 +1,100 @@
...
+// Closes #525: Sheet had no centered/floating dialog variant (edge-anchored
+// left/right/top/bottom only) and always rendered its title through a plain
+// <span>, with no way to reach a real heading tag. Both gaps forced TARA's
+// LaunchSheet.svelte to hand-roll its own overlay/panel/heading instead of
+// using this component at all.
+test.describe('Sheet side="center"', () => {
+  test('is a fixed-size dialog centered in the viewport, not edge-anchored', async ({ page }) => {
+    await page.goto('/components/sheet');
+
+    await page.getByText('Open centered dialog').click();
```
Tara-ag, line 12: 💬 SUGGESTION: Locating the trigger by visible text (`getByText('Open centered dialog')`) couples the test to demo-page copy. If the label changes, the test breaks even though the component behavior is unchanged.

Prefer `data-pw` selectors on the trigger buttons, or use dedicated test fixture routes, to keep component assertions independent of marketing copy.

src/wc/components/Sheet.wc.svelte
```diff
@@ -8,6 +8,7 @@
       title: { type: 'String', reflect: true },
       showOverlay: { type: 'Boolean', reflect: true, attribute: 'show-overlay' },
       showCloseButton: { type: 'Boolean', reflect: true, attribute: 'show-close-button' },
+      headingLevel: { type: 'Number', reflect: true, attribute: 'heading-level' },
```
Tara-ag, line 11: 💬 SUGGESTION: The web component wrapper exposes `side` as an unvalidated string, so values like `"centre"` or other typos silently produce an unstyled panel. This matches the existing behavior for `left/right/top/bottom`, but is worth documenting in `docs/Sheet.md` under the web-component section so consumers know to use the exact token values.

src/lib/Sheet/Sheet.svelte
```diff
@@ -143,7 +162,11 @@
       {#if typeof title === 'string' || showCloseButton}
```
Tara-ag, line 154: ⚠️ MAJOR: The overlay div has `role="button"` but no accessible name. Screen readers will announce this full-screen wrapper as an unnamed button, and because it is focusable (`tabindex="-1"`) it creates a confusing focus target when the sheet opens.

Suggested fix: add `aria-label="Close"` to the overlay div, or change its role to `"presentation"` and keep keyboard handling on the dialog panel.

Tara-ag (changes requested):
Additive feature is well-scoped and backward-compatible. However, the overlay's unnamed `role="button"` is an accessibility defect that should be addressed before merge. See inline comment for the concrete fix.

Tara-ag (commented):

tests/sheet-centered-variant.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    const panel = page.getByTestId('sheet-centered-panel');
```
Tara-ag, line 13: 💬 SUGGESTION: The new tests navigate to `/components/sheet` and click buttons by visible text. This couples the integration tests to the demo page's copy and layout, which is acceptable since this repo's existing sheet tests follow the same pattern, but it makes the tests fragile if the demo page is reorganized. Consider adding `data-pw` attributes to the demo buttons (e.g. `data-pw="open-centered-dialog"`) and using `page.getByTestId(...)` instead of `getByText`.

src/lib/Sheet/Sheet.svelte
```diff
@@ -143,7 +162,11 @@
...
         <div class="sheet-header">
           {#if typeof title === 'string'}
-            <span class="sheet-title">{title}</span>
+            {#if headingTag !== null}
+              <svelte:element this={headingTag} class="sheet-title">{title}</svelte:element>
+            {:else}
+              <span class="sheet-title">{title}</span>
+            {/if}
           {/if}
```
Tara-ag, line 162: ⚠️ MAJOR: Centered dialog lacks aria-labelledby association with its real heading. When `headingLevel` is set, the panel's accessible name comes from `aria-label={title ?? 'Sheet'}`, but the visible heading element is not programmatically associated with the dialog. Screen-reader users navigating by heading get the correct structure, but the `dialog` role's name should be tied to the heading element via `aria-labelledby` so the heading text is recognized as the dialog's name.

Add a stable id to the title element and conditionally use `aria-labelledby` on the panel, falling back to `aria-label` only when there is no title:

```svelte
let titleId = $derived(typeof testId === 'string' ? `${testId}-title` : 'sheet-title');
...
<div
  ...
  aria-labelledby={title ? titleId : undefined}
  aria-label={title ? undefined : 'Sheet'}
  ...
>
{#if headingTag !== null}
  <svelte:element this={headingTag} class="sheet-title" id={titleId}>{title}</svelte:element>
{:else}
  <span class="sheet-title" id={titleId}>{title}</span>
{/if}
```

Tara-ag (commented):
Yama review complete. One actionable accessibility finding on the new heading support and one optional test-hardening suggestion. The additive API, docs, web-component parity, and regression coverage look good.

Tara-ag (commented):
Reviewing the centered Sheet variant and headingLevel changes.

tests/sheet-centered-variant.test.ts
Tara-ag, line 13: 💬 SUGGESTION: The test clicks buttons by visible text, which couples the integration tests to the demo page's copy and ordering. Consider adding data-pw attributes to the demo buttons (e.g. `data-pw="open-centered-dialog"`) and using `page.getByTestId(...)` instead of `getByText`.

src/lib/Sheet/Sheet.svelte
Tara-ag, line 162: ⚠️ MAJOR: When headingLevel renders a real heading, the dialog is not aria-labelledby its heading. The panel currently gets its accessible name from aria-label={title ?? 'Sheet'}, but the visible heading element is not programmatically associated with the dialog. Associate the heading with the dialog so the heading text is recognized as the dialog's name.

Suggested fix:
```svelte
let titleId = $derived(typeof testId === 'string' ? `${testId}-title` : 'sheet-title');
```
Then on the panel:
```svelte
aria-labelledby={title ? titleId : undefined}
aria-label={title ? undefined : 'Sheet'}
```
And on the title element add `id={titleId}`.

Tara-ag (changes requested):
Review complete. One MAJOR accessibility issue remains around aria-labelledby for the centered dialog heading, plus one optional test-hardening suggestion. The API changes are backward-compatible and docs/tests/web-components are in sync.

src/lib/Sheet/Sheet.svelte
```diff
@@ -241,6 +264,20 @@
     border-top: var(--sheet-border, none);
```
Tara-ag, line 264: ⚠️ MAJOR: Centered dialog lacks visible focus indicator on the panel. The panel is programmatically focused on open (`sheetPanel.focus()` in `scrollLockAction`) and has `tabindex="-1"`, but `.sheet-panel { outline: none; }` removes any browser default outline. For a discrete floating dialog this leaves keyboard users without a visible focus cue. Add a `:focus-visible` outline for `.sheet-panel.center` (or remove `outline: none` and rely on the browser default).

docs/Sheet.md
```diff
@@ -23,19 +23,34 @@ A panel component that slides in from any edge of the screen (left, right, top,
...
 ```
 
+A centered confirmation dialog, with the title rendered as a real `<h2>`:
+
+```svelte
+<Sheet bind:open={confirmOpen} side="center" title="Delete project?" headingLevel={2}>
+  {#snippet content()}
+    <p>This can't be undone.</p>
+  {/snippet}
+  {#snippet footer()}
```
Tara-ag, line 33: 💬 SUGGESTION: The escaped pipes in `1\|2\|3\|4\|5\|6` render correctly on GitHub but can confuse other Markdown parsers. Consider using `<code>1 &#124; 2 &#124; 3 &#124; 4 &#124; 5 &#124; 6</code>` in the table cell for broader compatibility.

src/lib/Sheet/Sheet.svelte
```diff
@@ -135,7 +154,11 @@
       {#if typeof title === 'string' || showCloseButton}
```
Tara-ag, line 154: ⚠️ MAJOR: The overlay `<div>` has `role="button"` and `tabindex="-1"` but no accessible name. Screen readers will announce it as an unnamed button. Either remove `role="button"` (it is a backdrop, not a button) or add an `aria-label` such as `"Close dialog"` when `dismissOnOutsideClick` is true.

tests/sheet-centered-variant.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    const rightGap = viewportSize.width - (box.x + box.width);
+    expect(Math.abs(leftGap - rightGap)).toBeLessThanOrEqual(2);
+    const topGap = box.y;
+    const bottomGap = viewportSize.height - (box.y + box.height);
+    expect(Math.abs(topGap - bottomGap)).toBeLessThanOrEqual(2);
+
+    // Fixed-size: the default --sheet-center-width (480px), not stretched to
+    // fill the viewport the way left/right/top/bottom panels are.
+    expect(Math.round(box.width)).toBe(480);
+  });
```
Tara-ag, line 47: 💡 MINOR: Asserting `box.width === 480` is brittle — bounding-box width can be affected by borders, box-sizing, scrollbars, or viewport changes. Prefer asserting the computed CSS value (`toHaveCSS('width', '480px')`) or reading the custom property directly, which verifies the token is applied without depending on layout subtleties.

Tara-ag (changes requested):
Review complete. Two accessibility issues on the Sheet component should be addressed before merge: the centered panel needs a visible focus indicator, and the overlay's `role="button"` needs an accessible name or a different role. One minor test-brittleness note and one documentation formatting suggestion are also noted.

Tara-ag (commented):
Review in progress.

tests/sheet-centered-variant.test.ts
Tara-ag, line 47: 💡 MINOR: Asserting exact bounding-box widths (`box.width === 480`) and pixel-perfect gaps is brittle; viewport size, scrollbars, borders, and box-sizing can shift these values. Prefer `toHaveCSS('width', '480px')` or read the CSS custom property directly to make the test resilient.

src/lib/Sheet/Sheet.svelte
Tara-ag, line 264: ⚠️ MAJOR: The centered panel is focusable (`tabindex="-1"`) and receives focus on open via `sheetPanel.focus()`, but `.sheet-panel` sets `outline: none`. Without a visible `:focus-visible` or `:focus` indicator, keyboard users have no cue that focus has moved into the centered dialog. Add a visible focus outline for `.sheet-panel.center:focus-visible` or remove `outline: none` and rely on the browser default.

src/lib/Sheet/Sheet.svelte
Tara-ag, line 154: ⚠️ MAJOR: The overlay `<div>` has `role="button"` and `tabindex="-1"` but no `aria-label`. Screen readers will announce it as an unnamed button. Either remove `role="button"` (it acts as a backdrop, not a button) or add an accessible name such as `aria-label="Close dialog"` when `dismissOnOutsideClick` is true.

docs/Sheet.md
Tara-ag, line 33: 💬 SUGGESTION: The `headingLevel` type cell escapes pipes (`1\|2\|...`). This renders correctly on GitHub but can confuse other Markdown parsers. Consider using `<code>1 &#124; 2 &#124; ...</code>` for broader compatibility.

Tara-ag (changes requested):
Reviewed file-by-file. Two accessibility issues need addressing before merge: the centered panel lacks a visible focus indicator despite receiving focus, and the overlay `role="button"` has no accessible name. A minor test-brittleness note and a docs formatting suggestion are also included.

murdore:
## Video evidence

**What the recording shows**

A concatenation of two real browser recordings from the PR's own Playwright spec (tests/sheet-centered-variant.test.ts — despite the .test.ts name it imports from @playwright/test and IS the Playwright spec, so no throwaway spec was needed), run against a production vite build/preview on port 43285. Clip 1 (2.88s): navigates to /components/sheet, clicks "Open centered dialog", and the centered Sheet panel fades in and settles as a fixed 480px-wide dialog with visibly equal gaps on all four sides — away from every screen edge, unlike the edge-anchored left/right/top/bottom sheets. Clip 2 (1.88s): clicks "Open with headingLevel" and the panel's .sheet-title is shown/asserted as a real <h2> element reading "Confirm action" (vs. the default <span> title used elsewhere). Together the video demonstrates both required behaviors: the centered variant opening away from edges, and the title rendering as a real heading.

> With every changed file reverted to origin/release except the spec itself, the same spec run against a rebuilt preview server timed out on the two behavior-specific tests: "Test timeout of 30000ms exceeded ... waiting for getByText('Open centered dialog')" and the equivalent wait for getByText('Open with headingLevel'). Neither trigger button exists on the release demo page, so the click never resolves. The two pre-existing tests (unconfigured right sheet, default span title) still passed unchanged, confirming only the new behavior regressed. Full worktree restored afterward (git checkout HEAD -- . ; git status --short clean; commit still d492fac, single commit).

- passing, on this branch: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-539-sheet.webm` (257,854 bytes)

- My first attempt at re-running the branch spec produced 2 spurious failures because it hit a stale/unrelated server left listening on the chosen port by other concurrent evidence-gathering work on this shared machine (curl returned 200 from something that was not my freshly built preview). Re-run on a verified-free port gave the correct 4/4 pass. This is noted only as a caution about the shared-machine environment, not a defect in PR #539 itself — the eventual verified results (4/4 pass on branch, 2/4 fail on release with matching error signatures) stand.

src/wc/components/Sheet.wc.svelte
```diff
@@ -8,6 +8,7 @@
       sheetTitle: { type: 'String', reflect: true, attribute: 'title' },
```
Tara-ag, line 8: 🔒 CRITICAL: Declaring `sheetTitle: { ..., attribute: 'title' }` shadows the native `HTMLElement.title` accessor on `<sui-sheet>`. Even though the internal prop is renamed, Svelte custom elements define a host property for the mapped attribute, so `element.title` becomes the sheet title string and the browser-native tooltip is broken. Per the repo's WC standards, host-reserved accessors must use prefixed aliases with a non-conflicting attribute. Rename the attribute to `sheet-title` (or similar) and update the docs example accordingly.

src/lib/Sheet/Sheet.svelte
```diff
@@ -135,7 +154,11 @@
...
         <div class="sheet-header">
           {#if typeof title === 'string'}
-            <span class="sheet-title">{title}</span>
+            {#if headingTag !== null}
+              <svelte:element this={headingTag} class="sheet-title">{title}</svelte:element>
```
Tara-ag, line 158: ⚠️ MAJOR: The centered dialog uses `aria-label={title ?? 'Sheet'}` but when `headingLevel` is set the title is rendered as a real heading. Prefer `aria-labelledby` pointing to the heading's id so the dialog's accessible name matches the visible heading and is robust if the title string contains punctuation or special characters. Generate a stable id (e.g. derived from `testId`) and add `id={headingId}` to the `<svelte:element>` heading.

src/lib/Sheet/Sheet.svelte
```diff
@@ -101,6 +106,20 @@
       onafterclose();
     }
   }
+
+  // headingLevel's `1 | 2 | 3 | 4 | 5 | 6` union only constrains Svelte-import
+  // consumers. Through the web component, attributes are untyped strings
```
Tara-ag, line 111: 💡 MINOR: `headingTag` derivation silently ignores invalid `headingLevel` values. A web-component consumer passing `heading-level="7"` or `heading-level="foo"` gets no heading and no warning, which is confusing.

Suggestion: document the fallback behavior clearly (partially done in JSDoc) and optionally warn in development:
```ts
const headingTag = $derived.by(() => {
  if (headingLevel === undefined) return null;
  if (typeof headingLevel === 'number' && Number.isInteger(headingLevel) && headingLevel >= 1 && headingLevel <= 6) {
    return `h${headingLevel}`;
  }
  if (import.meta.env.DEV) {
    console.warn(`Sheet: headingLevel must be an integer between 1 and 6, received ${headingLevel}`);
  }
  return null;
});
```

src/lib/Sheet/Sheet.svelte
```diff
@@ -241,6 +264,20 @@
     border-top: var(--sheet-border, none);
   }
```
Tara-ag, line 265: ⚠️ MAJOR: Centered Sheet lacks explicit height, relying on flex content sizing. The `.sheet-panel.center` rule sets `width`, `max-width`, and `max-height` but does not set `height`. With `display: flex; flex-direction: column`, the panel height collapses to the intrinsic height of header + content + footer, so `--sheet-center-max-height` only clamps when content is taller. The PR/docs describe a "fixed-size dialog" sized by `--sheet-center-width`, but there is no corresponding height token, so short confirmation dialogs will be much smaller than expected.

Suggestion: add a `--sheet-center-height` token and set `height: var(--sheet-center-height, auto)` (or a sensible default like `fit-content`), and update docs if the intent is fixed width rather than fixed size.

tests/sheet-centered-variant.test.ts
```diff
@@ -0,0 +1,100 @@
...
+  test('is a fixed-size dialog centered in the viewport, not edge-anchored', async ({ page }) => {
+    await page.goto('/components/sheet');
+
+    await page.getByText('Open centered dialog').click();
+    const panel = page.getByTestId('sheet-centered-panel');
+    await expect(panel).toBeVisible();
+    // Let the 300ms transition finish — mid-transition the panel is still
+    // fading in.
+    await page.waitForTimeout(400);
+
```
Tara-ag, line 18: 💡 MINOR: New Playwright tests use brittle `waitForTimeout` instead of waiting for the actual transition end. Fixed sleeps slow the suite and can flake on slow CI runners.

Suggestion: wait for the transition to complete explicitly, e.g.:
```ts
await panel.locator(':scope').evaluate((el) =>
  new Promise((resolve) => el.addEventListener('transitionend', resolve, { once: true }))
);
```
or poll with `expect(...).toHaveCSS`.

Tara-ag (changes requested):
Reviewed file-by-file. Found one MAJOR layout issue (centered Sheet missing explicit height) and two MINOR quality items (brittle test waits, silent invalid headingLevel fallback). The web-component parity concern I initially raised was incorrect after verification — Svelte 5 `$props()` forwarding works as intended. No blocking criteria (secrets, critical security, unannounced breaking changes) are met. Requesting changes for the MAJOR layout correctness issue.

PR #540, opened 2026-09-06
feat(chatcomposer): support async boolean onsubmit and per-control disable

### LINKED ISSUE #526: ChatComposer: An async, boolean-returning onsubmit on ChatComposer and independent per-control di…
## Need
An async, boolean-returning onsubmit on ChatComposer and independent per-control disable

## Closest today
ChatComposer (`onsubmit: (value, attachments) => void`, single `disabled` gates text+voice+send together)

## Proposal
Change the submit hook to `onsubmit?: (value, attachments) => boolean | Promise<boolean>` that only clears the field on a truthy return, and split `disabled` into `textDisabled`/`voiceDisabled`/`sendDisabled`.

## Where it bites
Found while migrating TARA (a Svelte 5 / SvelteKit 2 consumer pinned to 3.3.0) onto the library. Without this the following consumer files can only migrate part-way and keep hand-rolled markup for the rest: `+page.svelte`, `Composer.svelte`.

murdore:
`onsubmit` may now return `boolean | Promise<boolean>` instead of only
`void`. The composer still clears `value` and `attachments` right after
every submit, unless the (possibly resolved) result is exactly `false`,
in which case the draft is left alone so a rejected send can be retried.
A rejected promise is treated the same as a resolved `false`. While an
async `onsubmit` is pending the composer never disables itself, so if the
user keeps typing or changes attachments before it settles, the clear is
skipped even on a truthy result — it never wipes out a newer draft.
Default behaviour is unchanged: a `void`-returning handler returns
`undefined`, which is not `false`, so it keeps clearing on every submit
exactly as before (see `submitResult.ts`).

`disabled` still disables the textarea, the voice button and the send
button together, as before. Three new optional props — `textDisabled`,
`voiceDisabled`, `sendDisabled` — each override one control independently;
any left at their `null` default falls back to `disabled`, so a caller
that never passes them keeps today's single-`disabled` behaviour exactly
(see `controlDisabled.ts`). This lets a caller freeze the send button
mid-request without also freezing dictation.

Both behaviours are additive and backward compatible: no prop or event
was removed or renamed, and the web-component wrapper mirrors the three
new props for parity.

Closes #526

---

### Verification

**Spec runs on this branch** — `src/lib/ChatComposer/submitResult.test.ts` and `controlDisabled.test.ts` — vitest.

**Fail-before** — Source reverted to `origin/release` with only the specs kept: **fails**.

### Reviewer callout

ℹ️ **A deliberate deviation from the issue text.** The issue says the composer should clear "only on a truthy return"; implemented as *clears unless the handler returns exactly `false`*. Taken literally, the issue's wording would break every existing void-returning handler, since `undefined` is not truthy. The chosen rule keeps those byte-for-byte compatible while still letting a handler veto the clear.


src/routes/components/chat-composer/+page.svelte
```diff
@@ -35,7 +35,13 @@
 </div>
 
 <div class="chat-theme composer-frame">
-  <ChatComposer bind:value placeholder="Type a message…" onsubmit={(text) => sent.push(text)} />
+  <ChatComposer
```
Tara-ag, line 76: 💬 SUGGESTION: Demo page does not illustrate the new async submit or per-control disable features. The chat-composer demo page was only reformatted; it does not include an example showing an async `onsubmit` that returns `false` to preserve the draft, nor an example of `sendDisabled`/`textDisabled`/`voiceDisabled`. Consumers relying on the demo for usage patterns will not discover the new capabilities.

Suggestion: Add a second demo composer that uses an async `onsubmit` returning a `Promise<boolean>` and a `sendDisabled` prop, so the new public API surface is exercised in the docs site.

src/lib/ChatComposer/ChatComposer.svelte
```diff
@@ -101,18 +106,44 @@
...
   let canSend = $derived(
-    !disabled &&
+    !resolvedSendDisabled &&
       (sendable ?? (value.trim().length > 0 || attachments.length > 0 || hasRichAttachments))
   );
 
   function submit(): void {
     if (!canSend) {
       return;
     }
```
Tara-ag, line 98: 💡 MINOR: New ChatComposer behaviour lacks component-level test coverage. The PR changes `ChatComposer.svelte` so that `onsubmit` may return a `Promise` and the composer conditionally skips clearing `value`/`attachments`. It also wires three new per-control disabled props into the template. The only tests added are pure unit tests for `resolveControlDisabled` and `shouldApplyClear`. There is no test that the actual component: (a) does not clear when an async `onsubmit` resolves to `false`, (b) skips clear when the user types while a promise is pending, or (c) renders the textarea/voice/send/action buttons with the correct effective disabled state. Without component-level coverage these behavioural contracts can regress silently.

Suggestion: Add a Svelte/Vitest component harness for `ChatComposer` that mounts the component and asserts: (1) returning `false` from `onsubmit` keeps `value`/`attachments`, (2) resolving `true` after the user has typed does not clear, and (3) `textDisabled`/`voiceDisabled`/`sendDisabled` override `disabled` for their respective controls.

src/lib/ChatComposer/ChatComposer.svelte
```diff
@@ -8,11 +8,16 @@
...
   import type { Action } from 'svelte/action';
   import type { ChatComposerProperties } from './properties';
+  import { resolveControlDisabled } from './controlDisabled';
+  import { shouldApplyClear } from './submitResult';
 
   let {
     value = $bindable(''),
     placeholder = '',
     disabled = false,
+    textDisabled = null,
```
Tara-ag, line 18: 💡 MINOR: Root `.disabled` visual class does not account for per-control disable overrides. The composer root still applies the `.disabled` class based solely on the top-level `disabled` prop. When a caller combines `disabled={true}` with one of the new per-control overrides (e.g. `sendDisabled={false}`), the entire composer is dimmed via `--chat-composer-disabled-opacity` even though the send button remains interactive. This produces a visually-disabled-but-functionally-enabled control, which is confusing for sighted users and can undermine the affordance of the enabled control.

Suggestion: Either derive an effective root-disabled state that is true only when all interactive controls are effectively disabled (e.g. `let rootDisabled = $derived(resolvedTextDisabled && resolvedVoiceDisabled && resolvedSendDisabled && /* attach */ disabled);`), or document in `docs/ChatComposer.md` that setting `disabled={true}` dims the whole composer including any controls enabled via the override props.

Tara-ag (commented):
Review complete. The change is additive and backward-compatible: no props/events/exports are removed or renamed, the web-component wrapper mirrors the new props, and the async-submit semantics correctly preserve void-returning handlers. I found two quality gaps and one suggestion, none of which are blocking. See inline comments for details.

Tara-ag (commented):
Two minor findings: missing web-component parity for the new per-control disable props, and an async test that doesn't wait for the promise to settle before asserting. Both are non-blocking.

tests/chat-composer-async-submit.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  // #526: sendDisabled must gate only the send button — the textarea and any
+  // other control stay interactive. Proven here by typing into the input
+  // while a send is in flight (sendDisabled is bound to the pending flag).
+  test('sendDisabled bound to a pending flag blocks only the send button', async ({ page }) => {
+    await page.goto('/components/chat-composer');
+
+    const input = page.getByTestId('async-submit-input');
+    const send = page.getByTestId('async-submit-send');
+
```
Tara-ag, line 70: 💡 MINOR: Async failure-path test does not wait for promise settlement. In the failure-path Playwright test, after clicking send with the fail toggle checked, the assertion passes immediately because the composer never disables itself while pending and the input value has not changed yet. It therefore does not prove that the clear was skipped after the 300 ms promise resolved; it would pass even if the clear were applied asynchronously later.

Suggestion: Wait for the pending state to settle before asserting the draft survived, e.g. add `await expect(send).toBeEnabled();` before the final `toHaveValue` assertion (the send button is disabled while asyncPending is true and re-enabled when fakeSend() resolves).

src/wc/components/ChatComposer.wc.svelte
```diff
@@ -6,6 +6,9 @@
       value: { type: 'String' },
       placeholder: { type: 'String', reflect: true },
       disabled: { type: 'Boolean', reflect: true },
+      textDisabled: { type: 'Boolean', attribute: 'text-disabled', reflect: true },
```
Tara-ag: 💡 MINOR: New per-control disabled props not reflected in web-component wrapper. The Svelte component adds textDisabled, voiceDisabled, and sendDisabled props, but the custom element wrapper does not declare them. Downstream consumers using <sui-chat-composer> cannot set these attributes, breaking public API parity between Svelte and web-component builds.

Suggestion: Add the three new Boolean props with kebab-case attributes and reflect: true:
```js
textDisabled: { type: 'Boolean', attribute: 'text-disabled', reflect: true },
voiceDisabled: { type: 'Boolean', attribute: 'voice-disabled', reflect: true },
sendDisabled: { type: 'Boolean', attribute: 'send-disabled', reflect: true },
```

Tara-ag (commented):
Approved with two minor notes: (1) the new per-control disabled props need to be declared in the web-component wrapper for Svelte/WC parity, and (2) the async failure-path Playwright test should wait for the promise to settle before asserting the draft survived. Neither issue blocks merge.

Tara-ag (commented):
Review in progress.

Tara-ag (commented):
Reviewing the async-submit and per-control disable changes in ChatComposer.

src/lib/ChatComposer/submitResult.ts
```diff
@@ -0,0 +1,46 @@
...
+ * `void`, so a caller can keep the composer's text (and attachments) around
+ * when a send fails rather than losing the draft to the unconditional clear.
+ *
+ * The rule is deliberately narrow: only an explicit `false` — synchronous or
+ * resolved — skips the clear. Every other outcome, `void` (today's
+ * `void`-returning handlers) and a rejected promise included, clears exactly
+ * as before. That keeps the default byte-for-byte: a handler written against
+ * the pre-#526 `(value, attachments) => void` shape returns nothing and never
+ * opts in to anything, so it keeps clearing on every submit, same as always.
+ */
```
Tara-ag, line 15: 💡 MINOR: The comment says "a rejected promise included, clears exactly as before", but `ChatComposer.svelte` handles rejection with `() => {}`, which keeps the draft (same as resolved `false`). The docs also say a rejected promise is treated as resolved `false`. The implementation is arguably the safer behavior, but the comment here contradicts both the code and the docs. Align this prose so all three sources agree.

src/wc/components/ChatComposer.wc.svelte
Tara-ag: ⚠️ MAJOR: The Svelte component defaults `textDisabled`/`voiceDisabled`/`sendDisabled` to `null` so they fall back to the shared `disabled` prop. Declaring them here as `{ type: 'Boolean', reflect: true }` breaks that fallback for web-component consumers: when the HTML attribute is absent, Svelte's custom-element runtime passes `false`, which `resolveControlDisabled` will use instead of the root `disabled`. So `<sui-chat-composer disabled>` without the new attributes leaves individual controls unexpectedly enabled.

Mirror the `Input.wc.svelte` `asSpellcheck` adapter pattern: convert the wrapped component's `false` back to `null` before passing it down, e.g.

```ts
const asOptionalBool = (value: unknown): boolean | null =>
  value === false ? null : (value as boolean | null);
```

Then pass `textDisabled={asOptionalBool(props.textDisabled)}` etc.

tests/chat-composer-async-submit.test.ts
```diff
@@ -0,0 +1,84 @@
+import { expect, test } from '@playwright/test';
+
+test.describe('ChatComposer — async onsubmit and independent per-control disable', () => {
+  // #526: an onsubmit resolving to `false` must leave the draft (value +
+  // attachments) exactly as it was, instead of the unconditional clear the
+  // composer applied before. This drives the actual component through its
+  // real async path (a genuine setTimeout-based promise on the demo page),
+  // not just the extracted `shouldApplyClear` unit under test elsewhere.
```
Tara-ag, line 8: 💬 SUGGESTION: These Playwright tests navigate to the demo page and rely on `data-pw` attributes defined in `src/routes/components/chat-composer/+page.svelte`. Per project standards, prefer component-level tests that are independent of demo-page structure. If the demo page is refactored, these tests will break even though the component behavior is unchanged. Consider adding a dedicated test harness or using component-level test IDs that belong to `ChatComposer` itself.

Tara-ag (changes requested):
Review complete. The Svelte-side changes are solid and backward-compatible, but the web-component wrapper regresses the new per-control disable fallback. Please fix the WC attribute mapping before merging.

Tara-ag:

One MAJOR issue: the PR silently changes the existing web-component `sendable` prop from Boolean to String, altering attribute coercion and dropping reflection, which contradicts the claim that the change is purely additive. Three MINOR issues: attach button is not covered by per-control disable, docs oversimplify the `onsubmit` return type, and Playwright tests are tightly coupled to the demo page. The async-submit race guard itself is implemented correctly.

- ⚠️ **MAJOR**: Web-component sendable changed from Boolean to String without preserving boolean attribute semantics — `src/wc/components/ChatComposer.wc.svelte:31`
  The PR changes `sendable` in ChatComposer.wc.svelte from `{ type: 'Boolean' }` to `{ type: 'String' }`. Previously, a consumer could write `<sui-chat-composer sendable>` and the prop would be `true`. Now the same attribute arrives as the empty string, which `asBooleanOrNull` maps to `true`, so that case still works. However, the previous boolean prop also mapped any non-empty value (e.g. `sendable…
- 💡 **MINOR**: Attach button still uses shared `disabled` instead of resolved per-control disable — `src/lib/ChatComposer/ChatComposer.svelte:137`
  The PR adds `textDisabled`, `voiceDisabled`, and `sendDisabled` but leaves the attach button gated by the root `disabled` prop. This is inconsistent with the "per-control disable" feature and undocumented. A caller who wants to disable only the textarea while keeping attach alive cannot do so; conversely, disabling only send currently also disables attach if the root `disabled` is true, which is e…
- 💡 **MINOR**: Docs say onsubmit returns `boolean | Promise<boolean>` but implementation accepts `void | Promise<boolean | void>` — `docs/ChatComposer.md:45`
  The docs and PR description state `onsubmit` may return `boolean | Promise<boolean>`. The properties.ts type is `boolean | void | Promise<boolean | void>`. The implementation handles `void` and `Promise<void>` correctly. The docs oversimplify and could mislead a TypeScript consumer who thinks returning `Promise<void>` is unsupported, or a JS consumer who returns `undefined` from an async function …
- 💡 **MINOR**: Playwright tests depend on demo page timing and toggles — `tests/chat-composer-async-submit.test.ts:12`
  The new Playwright tests navigate to the demo page and interact with demo-specific toggles and test IDs. While they exercise real component behaviour, they are tightly coupled to the demo page layout and the 300ms fakeSend timeout. A future demo refactor could break these tests even if the component is correct. The bootstrapped standards prefer isolated fixtures or data-pw selectors over demo-spec…

murdore:
## Video evidence

**What the recording shows**

Real browser recording of /components/chat-composer on a production preview build. Frame-by-frame confirmed (extracted via ffmpeg): checkbox 'Simulate failure' is checked, 'this send will fail' is typed into the async-submit textarea and Send is clicked -- the draft text remains in the textarea and no sent-message line appears (proving the async onsubmit resolving false left the draft untouched). The checkbox is then unchecked and Send is clicked again on the same unchanged draft -- the textarea clears to its placeholder and 'this send will fail' now appears as the sent message below (proving a non-false resolution clears the draft). This is the PR's own real Playwright spec (tests/chat-composer-async-submit.test.ts, 3 tests, all passed) driving the actual component through a genuine setTimeout-based async onsubmit on the live demo route -- not a unit test of the extracted shouldApplyClear/controlDisabled helpers.

> All 3 tests failed after reverting every PR-diff file except the spec to origin/release (docs/ChatComposer.md, src/lib/ChatComposer/ChatComposer.svelte, properties.ts, src/routes/components/chat-composer/+page.svelte, src/wc/components/ChatComposer.wc.svelte, tests/visual/__screenshots__/chat-composer.png checked out from origin/release; controlDisabled.ts/.test.ts and submitResult.ts/.test.ts deleted since they don't exist on release), rebuilt, and re-run against a fresh preview on the same port: test 1 timed out waiting for getByTestId('async-submit-fail-toggle') (30000ms), test 2 timed out waiting for getByTestId('async-submit-input'), test 3 got 'element(s) not found' for getByTestId('per-control-disable-input') -- because the release demo route has no async-submit/per-control-disable UI at all and ChatComposer.svelte on release has no shouldApplyClear/controlDisabled logic. Failure recording (29.4s, 524047 bytes) saved to /private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-540-chatcomposer-FAIL-on-release.webm. Worktree fully restored afterward: `git checkout HEAD -- .` + `git clean -fdx` (excluding node_modules/build/dist/dist-wc/dist-codemod/.svelte-kit) left `git status --short` empty and HEAD at commit 4d6a663 (unchanged, untouched).

- passing, on this branch: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-540-chatcomposer.webm` (86,776 bytes)
- failing, on `release`: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-540-chatcomposer-FAIL-on-release.webm` (524,047 bytes)

Tara-ag (commented):
Review in progress

src/wc/components/ChatComposer.wc.svelte
Tara-ag: ⚠️ MAJOR: Web-component wrapper loses the null default for `textDisabled`/`voiceDisabled`/`sendDisabled`. In the Svelte component these default to `null` and `resolveControlDisabled(specific, disabled)` uses `specific ?? disabled`, so omitting them preserves the legacy single-`disabled` behavior. Here they are declared as `{ type: 'Boolean', attribute: '...', reflect: true }`. Svelte 5 custom-element Boolean props are presence-based: an absent attribute initializes the prop to `false`, not `null`. Therefore `<sui-chat-composer disabled>` will receive `textDisabled=false`, `voiceDisabled=false`, `sendDisabled=false`, and `resolveControlDisabled(false, true)` returns `false`, disabling the controls even though `disabled=true` was set. This breaks backward compatibility for the web-component distribution.

Suggested fix: declare these as `{ type: 'String', attribute: 'text-disabled' }` (etc.) and convert the attribute string to `boolean | null` before passing down, mirroring the `spellcheck` workaround in `Input.wc.svelte`. For example:
```ts
const asOptionalBoolean = (value: unknown): boolean | null =>
  value == null ? null : value !== 'false';
```
then pass `textDisabled={asOptionalBoolean(props['text-disabled'])}` etc.

tests/chat-composer-async-submit.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    await failToggle.check();
+    await input.fill('this send will fail');
+    await send.click();
+    await expect(page.getByTestId('async-submit-sent')).toHaveCount(0);
+    // The composer never disables itself while the promise is pending, so the
+    // draft is readable immediately, but assert post-settle to prove the
+    // clear was genuinely skipped rather than merely not-yet-applied.
+    await expect(input).toHaveValue('this send will fail');
+
+    // Success path: the same input, unblocked, clears as normal.
```
Tara-ag, line 31: 💬 SUGGESTION: The failure-path assertion only waits for the input value to remain unchanged. Because the composer intentionally does not disable itself while pending, the test could pass even if the clear were applied later than the assertion. A stronger test would wait for the pending flag to settle (e.g. by observing the send button re-enabling) before asserting the draft survived. Additionally, there is no test verifying that a rejected promise (as opposed to a resolved `false`) keeps the draft; the unit test covers `shouldClearAfterSubmit(false)` but not the actual component's rejection handler.

Suggested fix: after clicking send in the failure path, wait for `send` to become enabled again (indicating the promise settled) before asserting the input still contains the draft. Add a unit or integration test where `onsubmit` rejects to verify the `.catch(() => {})` branch preserves the draft.

Tara-ag (changes requested):
## Review summary

The Svelte-side changes are well-thought-out and backward-compatible, but the web-component wrapper regresses the new per-control disable fallback because Svelte 5 custom-element Boolean props cannot represent a `boolean | null` tri-state.

### Findings

**Inline comments posted:**
- `src/wc/components/ChatComposer.wc.svelte`: the new `textDisabled`/`voiceDisabled`/`sendDisabled` props are declared as `{ type: 'Boolean' }`, which makes an absent attribute become `false` instead of `null`. This breaks the `specific ?? disabled` fallback in the Svelte component and therefore breaks backward compatibility for the web-component distribution. Mirror the `spellcheck` workaround in `Input.wc.svelte` by using `{ type: 'String' }` and converting to `boolean | null`.
- `tests/chat-composer-async-submit.test.ts`: the failure-path assertion is slightly weak because the composer does not disable itself while pending; wait for the send button to re-enable before asserting the draft survived, and add coverage for a rejected promise.

**Additional findings (on unchanged lines, so posted here):**
- `src/wc/components/ChatComposer.wc.svelte` (pre-existing `sendable` prop, line ~31): the same Boolean-presence bug already affects `sendable`. Declared as `{ type: 'Boolean' }` while the Svelte component defaults it to `null` and uses `sendable ?? (...)`, so the default can-send logic is short-circuited in the web-component build unless `sendable` is explicitly set. Fix it the same way as the new props.
- `src/lib/ChatComposer/ChatComposer.svelte` (attach button, line ~186): the attach button still uses the root `disabled` prop and is not covered by the new per-control disable scheme. Either add `attachDisabled` for parity or document explicitly that the attach button is governed only by `disabled`.

Overall decision: CHANGES_REQUESTED until the web-component attribute mapping is fixed.

tests/chat-composer-async-submit.test.ts
Tara-ag, line 31: 💬 SUGGESTION: The failure-path assertion only waits for the input value to remain unchanged. Because the composer intentionally does not disable itself while pending, the test could pass even if the clear were applied later than the assertion. A stronger test would wait for the pending flag to settle (e.g. by observing the send button re-enabling) before asserting the draft survived. Additionally, there is no test verifying that a rejected promise (as opposed to a resolved `false`) keeps the draft; the unit test covers `shouldClearAfterSubmit(false)` but not the actual component's rejection handler.

Suggested fix: After clicking send in the failure path, wait for `send` to become enabled again (indicating the promise settled) before asserting the input still contains the draft. Add a unit or integration test where `onsubmit` rejects to verify the `.catch(() => {})` branch preserves the draft.

src/wc/components/ChatComposer.wc.svelte
Tara-ag: ⚠️ MAJOR: Web-component wrapper loses the null default for `textDisabled`/`voiceDisabled`/`sendDisabled`. In the Svelte component these default to `null` and `resolveControlDisabled(specific, disabled)` uses `specific ?? disabled`, so omitting them preserves the legacy single-`disabled` behavior. Here they are declared as `{ type: 'Boolean', attribute: '...', reflect: true }`. Svelte 5 custom-element Boolean props are presence-based: an absent attribute initializes the prop to `false`, not `null`. Therefore `<sui-chat-composer disabled>` will receive `textDisabled=false`, `voiceDisabled=false`, `sendDisabled=false`, and `resolveControlDisabled(false, true)` returns `false`, disabling the controls even though `disabled=true` was set. This breaks backward compatibility for the web-component distribution.

Suggested fix: declare these as `{ type: 'String', attribute: 'text-disabled' }` (etc.) and convert the attribute string to `boolean | null` before passing down, mirroring the `spellcheck` workaround in `Input.wc.svelte`. For example:
```ts
const asOptionalBoolean = (value: unknown): boolean | null =>
  value == null ? null : value !== 'false';
```
then pass `textDisabled={asOptionalBoolean(props['text-disabled'])}` etc.

Tara-ag: ⚠️ MAJOR: Existing web-component `sendable` prop already breaks default can-send behavior. This is not introduced by the PR, but the same Boolean-presence bug affects the pre-existing `sendable` prop (line 42). It is declared as `{ type: 'Boolean' }` while the Svelte component defaults `sendable = null` and uses `sendable ?? (...)` to enable sending when there is text or attachments. Because the custom element passes `false` when the attribute is absent, the default can-send logic is short-circuited and the composer cannot send in the web-component build unless `sendable` is explicitly set. The PR adds three more props with the same pattern, increasing the surface area of the bug.

Suggested fix: Fix `sendable` at the same time by declaring it as `{ type: 'String', attribute: 'sendable' }` and converting with the same `asOptionalBoolean` helper, so an absent attribute preserves the `null` default and the default can-send logic works.

src/lib/ChatComposer/ChatComposer.svelte
Tara-ag, line 1: 💡 MINOR: Attach button still uses root `disabled` instead of resolved per-control disable. The PR introduces `textDisabled`, `voiceDisabled`, and `sendDisabled` to allow independent per-control disabling, but the attach button continues to use the root `disabled` prop. The documentation says `disabled` disables the textarea, voice button, and send button together, and does not mention the attach button, so this is arguably consistent. However, because the attach button is part of the same control surface, a caller who sets `disabled={true}` to dim the composer may be surprised that the attach button is also frozen while `textDisabled={false}` cannot re-enable the textarea. This is a design inconsistency, not a hard bug, but it should be called out explicitly in the docs or aligned.

Suggested fix: Either add an `attachDisabled` prop for parity, or update the documentation to state clearly that the attach button is governed only by the root `disabled` prop and is not part of the per-control disable scheme.

Tara-ag (changes requested):
Requesting changes because the web-component wrapper regresses the new per-control disable fallback (and the pre-existing `sendable` default) due to Svelte 5 custom-element Boolean props being presence-based. The Svelte-side implementation is solid and backward-compatible otherwise. Please see inline comments for details and suggested fixes.

Tara-ag (commented):
Review in progress — inline comments follow on accepted findings.

tests/chat-composer-async-submit.test.ts
```diff
@@ -0,0 +1,111 @@
...
+test.describe('ChatComposer — async onsubmit and independent per-control disable', () => {
+  // #526: an onsubmit resolving to `false` must leave the draft (value +
+  // attachments) exactly as it was, instead of the unconditional clear the
+  // composer applied before. This drives the actual component through its
+  // real async path (a genuine setTimeout-based promise on the demo page),
+  // not just the extracted `shouldApplyClear` unit under test elsewhere.
+  test('a failed async send keeps the typed draft; a successful one clears it', async ({
+    page
+  }) => {
+    await page.goto('/components/chat-composer');
```
Tara-ag, line 12: 💡 MINOR: These tests are tightly coupled to the demo page layout and the 300ms `fakeSend` timeout. They'll break if the demo is refactored even when the component is correct. Consider complementing them with a Vitest component test that mounts `ChatComposer` with a controllable promise and asserts the clear/no-clear contract directly, keeping the Playwright tests as lighter integration smoke tests.

src/wc/components/ChatComposer.wc.svelte
```diff
@@ -25,7 +28,7 @@
       onopenrichimage: { type: 'Object' },
       onopenrichvideo: { type: 'Object' },
       onopenrichfile: { type: 'Object' },
-      sendable: { type: 'Boolean' },
+      sendable: { type: 'String' },
```
Tara-ag, line 31: ⚠️ MAJOR: `sendable` was changed from `{ type: 'Boolean' }` to `{ type: 'String' }`. This alters an existing web-component prop contract: previously `sendable="false"` would coerce to `true` (boolean presence semantics), but now it maps to `false`. It also drops reflection, so JS property changes no longer update the host attribute. The PR claims only additive changes, so please restore `sendable: { type: 'Boolean', reflect: true }` and apply the tri-state helper only to the three new per-control props. If `sendable` genuinely needs tri-state, document the breaking change.

docs/ChatComposer.md
```diff
@@ -42,13 +42,59 @@ its lists are fed:
 />
 ```
 
+### Async submit — keep the draft on a rejected send
```
Tara-ag, line 45: 💡 MINOR: Docs say `onsubmit` may return `boolean | Promise<boolean>`, but the implementation's type is `boolean | void | Promise<boolean | void>`. An async handler that doesn't explicitly return (common case) resolves to `void`, which is handled correctly but isn't mentioned. Update the sentence to match the actual type so consumers don't think `Promise<void>` is unsupported.

src/lib/ChatComposer/ChatComposer.svelte
Tara-ag, line 1: 💡 MINOR: The attach button (around line 137) still uses the root `{disabled}` prop while the other controls use resolved per-control disables. This leaves a gap in the "independent per-control disable" feature: there's no way to disable only the textarea while keeping attach alive, for example. Either add `attachDisabled` with the same `null`-falls-back-to-`disabled` semantics, or explicitly document that the attach button is governed by `disabled` and excluded from the per-control props.

Tara-ag (changes requested):
Review complete. One MAJOR concern on the web-component `sendable` contract change, plus three MINOR notes. See inline comments for details.

Tara-ag (commented):
Review in progress.

src/wc/components/ChatComposer.wc.svelte
Tara-ag, line 31: ⚠️ MAJOR: The PR changes `sendable` in ChatComposer.wc.svelte from `{ type: 'Boolean' }` to `{ type: 'String' }`. This alters an existing web-component prop contract: previously `sendable="false"` would coerce to `true` (boolean presence semantics), but now it maps to `false`. It also drops reflection, so JS property changes no longer update the host attribute. The PR claims only additive changes, so this is a backward-compatibility regression.

Suggestion: Restore `sendable: { type: 'Boolean', reflect: true }` and apply the tri-state helper only to the three new per-control props. If `sendable` genuinely needs tri-state, document the breaking change.

docs/ChatComposer.md
Tara-ag, line 45: 💡 MINOR: Docs say `onsubmit` returns `boolean | Promise<boolean>` but the implementation accepts `void | Promise<boolean | void>`. The properties.ts type is `boolean | void | Promise<boolean | void>`. The implementation handles `void` and `Promise<void>` correctly. The docs oversimplify and could mislead consumers.

Suggestion: Update the docs sentence to match the type: "`onsubmit` may return `boolean | void | Promise<boolean | void>` instead of only `void`".

tests/chat-composer-async-submit.test.ts
Tara-ag, line 12: 💡 MINOR: The new Playwright tests navigate to the demo page and interact with demo-specific toggles and test IDs. They are tightly coupled to the demo page layout and the 300ms fakeSend timeout. A future demo refactor could break these tests even if the component is correct.

Suggestion: Consider adding a Vitest component test that mounts ChatComposer with a controllable promise and asserts the clear/no-clear contract directly, keeping the Playwright tests as lighter integration smoke tests.

src/lib/ChatComposer/ChatComposer.svelte
```diff
@@ -75,18 +80,44 @@
     richImages.length > 0 || richVideos.length > 0 || richFiles.length > 0
   );
 
+  // #526: each falls back to the shared `disabled` when its own flag is
+  // unset, so a caller who never passes the new props keeps the original
+  // single-`disabled`-gates-everything behaviour exactly.
```
Tara-ag, line 85: 💡 MINOR: The PR adds `textDisabled`, `voiceDisabled`, and `sendDisabled` (resolved here) but leaves the attach button at line 137 gated by the root `disabled` prop. This is inconsistent with the "independent per-control disable" feature and undocumented.

Suggestion: Either add `attachDisabled` with the same `null`-falls-back-to-`disabled` semantics, or explicitly document that the attach button is governed by `disabled` and excluded from the per-control props.

Tara-ag (changes requested):
Review complete. One MAJOR concern on the web-component `sendable` contract and three MINOR notes. Please address the MAJOR item before merging.

PR #542, opened 2026-09-06
feat(hitl): add extra actions and a children snippet for multi-action approval

### LINKED ISSUE #528: HITL: A single generic confirmation/approval component that supports multiple actions, a children…
## Need
A single generic confirmation/approval component that supports multiple actions, a children/extra-actions snippet, and multi-select + free-text input

## Closest today
HITL (single confirm/cancel/countdown action only, no children snippet)

## Proposal
Add an optional `actions?: {label, onSelect}[]` array and a `children` snippet slot for arbitrary extra controls (free-text textarea, multi-select chips), so plan-mode and multi-question approval flows can eventually adopt one shared component instead of staying fully bespoke.

## Where it bites
Found while migrating TARA (a Svelte 5 / SvelteKit 2 consumer pinned to 3.3.0) onto the library. Without this the following consumer files can only migrate part-way and keep hand-rolled markup for the rest: `+page.svelte`, `ApprovalCard.svelte`.

murdore:
HITL supported exactly one decision (confirm/cancel), so a third
disposition (e.g. "deny with instructions") or any free-text/multi-select
control had nowhere to live. Consumers hand-rolled a separate Card next to
(or instead of) HITL to add one — see TARA's ApprovalCard.svelte, which
renders its own extra button plus an entirely separate composing-mode Card
purely because "HITL's own footer is exactly two buttons with no slot for
a third path."

Adds two optional, additive props:

- `actions?: HITLExtraAction[]` — extra buttons rendered between cancel and
  confirm. Selecting one never settles the card (no completion state, no
  `onConfirm`); it only pauses the countdown like any other interaction and
  calls the action's own `onSelect`. Each entry is
  `{ label, onSelect, testId?, classes? }`.
- `children?: Snippet` — arbitrary extra controls (free-text input,
  multi-select chips) rendered in the body above the action row, only
  while the card is pending.

With neither passed, the card renders exactly as before: two buttons and
no extra content block. `children` is intentionally not forwarded through
the `sui-hitl` web-component wrapper (it's a reserved custom-element prop
name); `actions` is.

Adds tests/hitl-extra-actions.test.ts (Playwright, this repo's convention
for tests/) covering: an extra action firing onSelect without settling the
card, confirm still settling normally with an extra action present, and
the unchanged two-button default. Updates docs/HITL.md and the component
demo page accordingly.

Closes #528

---

### Verification

**Spec runs on this branch** — `tests/hitl-extra-actions.test.ts` — 3 Playwright tests against a real build.


Tara-ag (commented):

tests/hitl-extra-actions.test.ts
```diff
@@ -0,0 +1,54 @@
+import { expect, test } from '@playwright/test';
```
Tara-ag, line 1: 💡 MINOR: New Playwright test file imports from '@playwright/test' instead of the repo's test fixture. The new test file uses `import { expect, test } from '@playwright/test'`. Other tests in this repo likely use a project-local fixture (e.g. from a `test.ts` helper) that wraps page setup, base URL, or failure screenshots. Importing the bare Playwright test harness bypasses those conventions and may cause inconsistent test behavior or missing diagnostics.

Suggested fix: check the existing tests directory for a local `test` export (commonly `import { test, expect } from './test';` or similar) and switch to that fixture.

src/lib/HITL/HITL.svelte
```diff
@@ -383,6 +410,18 @@
               onclick={() => interact('rejected')}
             />
           </div>
+          {#each actions ?? [] as action, index (action.testId ?? `${index}-${action.label}`)}
```
Tara-ag, line 415: 💡 MINOR: Extra action buttons lack accessible-name override and rely solely on label text. Each extra action is rendered as a Button with only `text={action.label}`. There is no `ariaLabel` prop on `HITLExtraAction`, so consumers cannot provide a more descriptive accessible name (e.g. 'Deny with instructions, opens a text area') without changing the visible label.

Suggested fix: add an optional `ariaLabel?: string` field to `HITLExtraAction` and forward it to the Button's `ariaLabel` prop.

src/lib/HITL/HITL.svelte
```diff
@@ -178,6 +186,20 @@
...
   };
 
+  // An extra action (e.g. "Deny with instructions") is a side path, not a
+  // decision — it never settles the card, so `onConfirm` never fires for it.
+  // It still counts as interacting with the card: pause this card's own
+  // countdown and every sibling's, exactly like clicking confirm or cancel.
+  const selectExtraAction = (action: HITLExtraAction): void => {
+    if (isProcessing || isCompleted) {
+      return;
+    }
```
Tara-ag, line 198: ⚠️ MAJOR: Extra action callback can be called while a decision is settling. `selectExtraAction` checks `isProcessing` and `isCompleted`, but `isProcessing` is set to true only inside `settle()` after `decide()` awaits `restoreMicState()`. During that async gap, clicking an extra action will pause/stop timers and invoke `action.onSelect()` even though a confirm/cancel/auto decision is already in flight. This can lead to conflicting UI state if the consumer's `onSelect` mutates state that the `onConfirm` handler also touches.

Suggested fix: latch an in-flight decision guard before the await, similar to `decisionPending`, and check it in `selectExtraAction`. For example, set `decisionPending = true` at the start of `interact()` and reset it in `settle()` on error; then guard `selectExtraAction` with `if (decisionPending || isProcessing || isCompleted) return;`.

Tara-ag (changes requested):
Thanks for the PR. The new `actions` and `children` additions are additive and backward-compatible, and the web-component parity handling for `children` is well explained. However, there's a lifecycle race in `selectExtraAction` that can fire while a confirm/cancel/auto decision is still settling. Please address that before merging. Two smaller items noted inline as well.

Tara-ag:

## 🤖 Yama Review Summary

**Verified findings (2):**

- **MINOR**: Misleading comment for HITLExtraAction.classes — `src/lib/HITL/properties.ts:38`
- **MINOR**: Extra action onSelect is invoked without a guard — `src/lib/HITL/HITL.svelte:198`

- 💡 **MINOR**: Misleading comment for HITLExtraAction.classes — `src/lib/HITL/properties.ts:38`
  The JSDoc for `HITLExtraAction.classes` says it can be used "e.g. to pick a variant", but `Button.variant` is a separate prop that HITL never forwards from the extra action. Consumers can only append CSS classes to the button container; they cannot select one of Button's built-in variants through this field. The wording should clarify that `classes` is for custom CSS styling, not for choosing a Bu…
- 💡 **MINOR**: Extra action onSelect is invoked without a guard — `src/lib/HITL/HITL.svelte:198`
  In selectExtraAction, action.onSelect() is called directly even though JS consumers and web-component hosts can pass an action object that omits the callback. This will throw at runtime and leave the card with paused timers.

Tara-ag (approved):
Reviewed file-by-file. The change is additive and backward-compatible: two new optional props (`actions`, `children`) on HITL, a new `HITLExtraAction` type, matching web-component wrapper update for `actions`, docs, demo, and Playwright tests. No public API breakage, no CSS custom-property renames, no SSR or accessibility regressions identified. The `children` web-component limitation is correctly documented and consistent with other snippet-based components.

Tara-ag (commented):
Review in progress — posting findings file-by-file.

src/lib/HITL/properties.ts
```diff
@@ -21,6 +21,24 @@ export type HITLInitialState = {
...
+ * interaction) and calls `onSelect` — settling, if any, is the caller's job.
+ */
+export type HITLExtraAction = {
+  label: string;
+  onSelect: () => void;
+  /** `data-pw` for this action's button (default: `<testId>-action-<index>`). */
+  testId?: string;
+  /** Class string on this action's Button, e.g. to pick a variant. */
+  classes?: string;
+  /** Accessible name for this action's Button, when `label` alone isn't descriptive enough. */
```
Tara-ag: 💡 MINOR: The JSDoc says `classes` can be used "to pick a variant", but `HITL.svelte` hardcodes `variant="secondary"` and only forwards `classes` to Button's container. Consumers cannot select a built-in Button variant through this field. Reword to clarify it's for custom CSS styling only, e.g. `/** CSS class string added to this action's Button container for custom styling. */`.

src/lib/HITL/HITL.svelte
```diff
@@ -180,6 +188,20 @@
...
   };
 
+  // An extra action (e.g. "Deny with instructions") is a side path, not a
+  // decision — it never settles the card, so `onconfirm` never fires for it.
+  // It still counts as interacting with the card: pause this card's own
+  // countdown and every sibling's, exactly like clicking confirm or cancel.
+  const selectExtraAction = (action: HITLExtraAction): void => {
+    if (decisionPending || isProcessing || isCompleted) {
+      return;
+    }
```
Tara-ag: 💡 MINOR: Extra action `onSelect` is invoked without a guard. JS consumers and web-component hosts can pass an action entry that omits the callback, which will throw at runtime and leave the card with paused timers.

Suggested fix: use `action.onSelect?.()` or guard with `if (typeof action.onSelect === 'function')` before invoking.

murdore:
## Video evidence

**What the recording shows**

On the /components/hitl demo route ("Deploy to production" card, testId demo-extra): starts with all three controls visible and enabled - Cancel, "Deny with instructions..." (the third decision action), and Confirm. It clicks "Deny with instructions..." - the counter increments to 1, a free-text textarea (children snippet) appears, and the card is visibly NOT settled: no completion strip, Cancel/Confirm/the extra action are all still present and interactable (pause held ~1.2s so this is readable). It then clicks Confirm - the card settles: a green "Approved" completion strip appears, and Cancel, Confirm, and the extra "Deny with instructions..." button are all gone from the DOM together - i.e. the extra action is no longer available once a decision has been made/settled. Held for a final 1.5s pause on the settled state.

> With src/lib/HITL/HITL.svelte, src/lib/HITL/properties.ts, src/routes/components/hitl/+page.svelte, src/wc/components/HITL.wc.svelte, docs/HITL.md and the visual screenshot all checked out from origin/release (only tests/hitl-extra-actions.test.ts kept from the PR), rebuilt and re-served: both the PR's own spec and the throwaway evidence spec failed identically at the same point - `getByTestId('demo-extra')` (and 'demo-extra-confirm') times out / is not found, because the release build's hitl demo route has no extra-actions demo card at all (no `actions`/`children` props, no third button). 2 of 3 official tests failed; the throwaway spec's single test failed at its very first assertion. Failure video saved alongside the pass video at /private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-542-hitl-FAILS-on-release.webm (143,837 bytes, 7.12s).

- passing, on this branch: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-542-hitl.webm` (241,807 bytes)
- failing, on `release`: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-542-hitl-FAILS-on-release.webm` (143,837 bytes)

- The PR's own Playwright spec (tests/hitl-extra-actions.test.ts) never runs under pnpm test:unit (vitest, src/** only) — it's exercised only via pnpm test:integration or the ad-hoc vite-preview+playwright invocation the recording report describes. That's consistent with this repo's stated convention (tests/ specs are Playwright, not vitest), not a defect, but worth naming since it means the unit-test gate provides no direct coverage of the new behavior — the visual proof is what carries that burden here, and I verified it directly.

PR #543, opened 2026-09-06
feat(modal): accessible name, ARIA dialog semantics, a real focus trap and a unified onDismiss

### LINKED ISSUE #529: Modal: aria-label/role/aria-modal support and a real focus trap on Modal, plus a unified dismiss…
## Need
aria-label/role/aria-modal support and a real focus trap on Modal, plus a unified dismiss event

## Closest today
Modal (no accessible-name prop, overlay hardcodes role="button", only onoverlayClick for backdrop/Escape, no Tab-cycling)

## Proposal
Add `ariaLabel`/`role` props, drop the unconditional role="button" on the overlay when there's no click-to-dismiss handler, add a focus-trap utility, and unify onoverlayClick + hardware-back-press under one `onDismiss` callback.

## Where it bites
Found while migrating TARA (a Svelte 5 / SvelteKit 2 consumer pinned to 3.3.0) onto the library. Without this the following consumer files can only migrate part-way and keep hand-rolled markup for the rest: `+page.svelte`, `CommandPalette.svelte`.

murdore:
Closes #529

Modal had no accessible-name prop, hardcoded role="button" on the overlay
regardless of whether a click on it did anything, never announced itself
as role="dialog"/aria-modal to assistive tech, never trapped Tab inside
itself, and split "the modal wants to close" across two independently
wired callbacks (onoverlayClick for backdrop+Escape, onclose for hardware
back-press). Found migrating TARA onto the library: CommandPalette.svelte
and the step-up dialog had to keep hand-rolled dialog markup, or hand-wire
both dismiss callbacks separately, because Modal couldn't do this itself.

API added, all optional, all additive:

- `ariaLabel?: string` — accessible name, rendered as aria-label on the
  content panel. Unset by default (today's behavior: no accessible name).
- `role?: 'dialog' | 'alertdialog'` (new `ModalRole` type) — ARIA role for
  the content panel. Unset by default: the panel renders no `role` and no
  `aria-modal` at all, byte-for-byte matching behavior before this prop
  existed, so an existing consumer who never touches `role` sees no
  DOM/ARIA change. Setting it opts in to both `role` and `aria-modal="true"`
  together, since a role without aria-modal is not a complete announcement.
- A real focus trap (`focus-trap.ts`, unit-tested in isolation like
  Tooltip's action logic): Tab/Shift+Tab wraps at the panel's first/last
  focusable descendant instead of leaving the modal, and focus moves into
  the panel on mount (its first focusable element, or the panel itself via
  tabindex="-1" when it has none — matches Sheet's fallback). Resolving the
  active element goes through the panel's own `getRootNode()` rather than
  `document.activeElement`, so the trap and the mount/destroy focus
  management both work correctly inside Modal.wc.svelte's open shadow root,
  not only in the light-DOM Svelte build. Focus is also restored to
  whatever was focused before the modal opened once it closes (guarded on
  that element still being connected to the document).
- `onDismiss?: () => void` — fires on every path that dismisses the modal
  from outside it: overlay click, Escape, and hardware back-press. A
  superset of onoverlayClick and onclose; both keep firing unchanged
  alongside it, so wiring onDismiss is optional, not a replacement.
- The overlay's role="button"/tabindex="0" is now conditional on there
  being an actual dismiss handler (onoverlayClick or onDismiss) — a click
  there was already a no-op otherwise, so announcing it as an interactive
  "button" was the wrong default. Same reasoning already applied to the
  header images. A consumer with no dismiss handler sees an unchanged
  (already inert) click surface, just without the false ARIA affordance.
  When it is dismissible, it also now takes Enter/Space keyboard
  activation and an optional `overlayAriaLabel?: string` accessible name,
  mirroring the existing header-image pattern (WCAG 2.1 SC 2.1.1/4.1.2).

Default behavior for an existing caller wiring onoverlayClick/onclose is
unchanged beyond the focus trap. web-component parity: ariaLabel/role are
intentionally NOT declared on Modal.wc.svelte since both are
HOST_RESERVED_PROPS (they'd shadow the custom element's own ARIAMixin
accessors) — documented in docs/Modal.md's Web Component section, which
also covers onDismiss and overlayAriaLabel, both declared like the other
props.

---

### Verification

**Spec runs on this branch** — `src/lib/Modal/focus-trap.test.ts`, vitest.

**Fail-before** — source reverted to `origin/release` with only the spec kept: **fails** (`focus-trap` does not exist there).

`pnpm test:unit`, `pnpm check`, `pnpm lint` and `pnpm build` each exit 0 on this branch; exit codes captured directly, never through a pipe.

### Correcting an earlier version of this description

An earlier revision of this PR rendered `role="dialog"` and `aria-modal="true"` on **every** Modal by default, and this description said so. The review pushed back and it was right: that changed how assistive technology announces every Modal already in the wild. Both are now opt-in — `role` renders only when a caller supplies one, and `aria-modal` only alongside it. A consumer that passes neither gets byte-for-byte today's markup.

Also in this revision, from the review: the focus trap resolves the active element through shadow roots, focus is restored to the pre-open element on close, and the overlay takes `role="button"` only when it is actually dismissible.

### DIFF
--- a/docs/Modal.md
+++ b/docs/Modal.md
@@ -1,6 +1,6 @@
 # Modal
 
-A dialog overlay component that renders on top of the page with configurable size, alignment, header (with left/right images and text), footer (with primary/secondary Button components), and transition animations. It locks body scroll on mount and optionally handles hardware back-press navigation. The overlay supports click-to-dismiss (debounced) and Escape key handling.
+A dialog overlay component that renders on top of the page with configurable size, alignment, header (with left/right images and text), footer (with primary/secondary Button components), and transition animations. It locks body scroll on mount and optionally handles hardware back-press navigation. The overlay supports click-to-dismiss (debounced) and Escape key handling. The content panel takes no `role`/`aria-modal` by default (unchanged from prior behavior); setting `role` opts into `role="dialog"` (or `role="alertdialog"`) plus `aria-modal="true"`, with an optional `ariaLabel` for its accessible name. Tab is trapped inside the panel, focus moves in on mount, and returns to whatever was focused before the modal opened once it closes.
 
 ## Usage
 
@@ -28,11 +28,14 @@ A dialog overlay component that renders on top of the page with configurable siz
 | debounceTime             | `number`                                                                                                                                                  | No       | `700`           | Debounce delay in milliseconds for overlay click handling. Prevents rapid repeated overlay dismissals.                                                                                                                                                                                                                                                                                          |
 | leftImageTestId          | `string`                                                                                                                                                  | No       | `-`             | Value for data-pw on the left header image wrapper.                                                                                                                                                                                                                                                                                                                                             |
 | leftImageAriaLabel       | `string`                                                                                                                                                  | No       | `-`             | Accessible name (rendered as `aria-label`) for the left header image's `role="button"` wrapper (e.g., a back control) — required for screen readers since the wrapper carries no visible text.                                                                                                                                                                                                  |
+| overlayAriaLabel         | `string`                                                                                                                                                  | No       | `-`             | Accessible name (rendered as `aria-label`) for the overlay's `role="button"` wrapper. Only meaningful when the overlay is dismissible (an `onoverlayclick` or `ondismiss` handler is supplied) — that's the only state where the overlay is announced as a focusable button, and it's also reachable via Enter/Space in that state.                                                            |
 | testId                   | `string`                                                                                                                                                  | No       | `-`             | Value for data-pw on the modal overlay container.                                                                                                                                                                                                                                                                                                                                               |
 | classes                  | `string`                                                                                                                                                  | No       | `-`             | CSS class string applied to the component's top-level element. Useful for theming — define classes with CSS variable overrides and pass them to create variant styles.                                                                                                                                                                                                                          |
 | overlayBackdropFilter    | `string`                                                                                                                                                  | No       | `-`             | CSS `backdrop-filter` value applied to the overlay (e.g. `"blur(6px)"`). Sets `--modal-overlay-backdrop-filter` inline on the overlay div. Default: `none` (no blur).                                                                                                                                                                                                                           |
 | overlayFadeIn            | `boolean`                                                                                                                                                 | No       | `false`         | When true, the overlay backdrop fades in on mount instead of appearing instantly. Pairs well with `entryAnimation="slide-up"` for a softer combined entrance.                                                                                                                                                                                                                                   |
 | usePortal                | `boolean`                                                                                                                                                 | No       | `false`         | When true, mounts the modal overlay at `document.body` so it escapes any ancestor clipping or stacking contexts. Useful inside `overflow: hidden` or `transform` containers.                                                                                                                                                                                                                    |
+| ariaLabel                | `string`                                                                                                                                                  | No       | `-`             | Accessible name for the modal, rendered as `aria-label` on the content panel (the element carrying `role`/`aria-modal` when `role` is set). Without it a screen reader announces an unnamed dialog.                                                                                                                                                                                             |
+| role                     | `ModalRole = 'dialog' \| 'alertdialog'`                                                                                                                   | No       | `-` (unset)     | ARIA role for the content panel. Unset by default — the panel renders no `role` and no `aria-modal`, matching behavior before this prop existed, so an existing consumer who never touches `role` sees no DOM/ARIA change. Set explicitly to opt in: the panel then also carries `aria-modal="true"`. Use `'alertdialog'` for a dialog that interrupts to demand an immediate response (e.g. a destructive confirmation).                                                                                                                                                                                             |
 
 ## Snippets
 
@@ -54,6 +57,7 @@ Svelte 5 Snippet props — pass content blocks to the component.
 | onsecondarybuttonclick  | `(event: MouseEvent) => void`    | Fires when the secondary footer button is clicked.                                                                                                                                    |
 | onoverlayclick          | `() => void`                     | Fires when the dark overlay background is clicked (outside the modal content). Also fires when the Escape key is pressed. Debounced by debounceTime milliseconds.                     |
 | onkeydown               | `(event: KeyboardEvent) => void` | Fires when any key is pressed while the modal is open (attached to the window).                                                                                                       |
+| ondismiss               | `() => void`                     | Fires on every path that dismisses the modal from outside it: an overlay click, Escape, and a hardware back-press. A superset of `onoverlayclick` (backdrop/Escape only) and `onclose` (hardware back-press/`autoDismissAfter` only) — wire this one callback instead of both when all you need is "the modal wants to close". Both existing events still fire unchanged alongside it. |
 
 ## CSS Variables
 
@@ -195,6 +199,12 @@ type ModalTransition = 'IN' | 'ALL';
 type ModalEntryAnimation = 'fade' | 'slide-up' | 'slide-down';
 ```
 
+### ModalRole
+
+```typescript
+type ModalRole = 'dialog' | 'alertdialog';
+```
+
 ### ButtonProperties
 
 ```typescript
@@ -237,6 +247,8 @@ Tag: `<sui-modal>`
   </div>
 </sui-modal>
 ```
+
+`aria-label` and `role` are not props on `<sui-modal>` — they're reserved accessors on every custom element (`ARIAMixin`/`role`), so declaring them as component props would replace the native ones instead of adding to them. Set them as native attributes directly on the host, e.g. `<sui-modal role="alertdialog" aria-label="Delete account">`. `ondismiss` (and `overlayAriaLabel`) are ordinary props and work the same way as on the Svelte component.
 
 ### Slots
 
--- a/docs/_index.json
+++ b/docs/_index.json
@@ -165,7 +165,7 @@
   },
   {
     "name": "Modal",
-    "description": "A dialog overlay component that renders on top of the page with configurable size, alignment, header, footer, and transition animations. entryAnimation can override the per-align default, e.g. a slide-up bottom-sheet-style dialog while centered. Locks body scroll and optionally handles hardware back-press navigation. Supports click-to-dismiss and Escape key."
+    "description": "A dialog overlay component that renders on top of the page with configurable size, alignment, header, footer, and transition animations. entryAnimation can override the per-align default, e.g. a slide-up bottom-sheet-style dialog while centered. Locks body scroll and optionally handles hardware back-press navigation. The content panel takes no role/aria-modal by default; setting role opts into role/aria-modal together (optionally an ariaLabel) and traps Tab focus; ondismiss unifies overlay-click, Escape and hardware-back-press dismissal."
   },
   {
     "name": "ModalAnimation",
--- a/src/lib/Modal/Modal.svelte
+++ b/src/lib/Modal/Modal.svelte
@@ -1,14 +1,20 @@
 <script lang="ts">
   import type { ModalProperties } from './properties';
-  import { onMount, onDestroy } from 'svelte';
+  import { onMount, onDestroy, tick } from 'svelte';
   import ModalAnimation from '$lib/Animations/ModalAnimation.svelte';
   import OverlayAnimation from '$lib/Animations/OverlayAnimation.svelte';
   import { createDebouncer, lockBodyScroll, unlockBodyScroll } from '../utils';
+  import { focusEntryPoint, focusTrapTabTarget } from './focus-trap';
   import Button from '$lib/Button/Button.svelte';
   import Img from '$lib/Img/Img.svelte';
 
   let overlayDiv: HTMLDivElement | null = $state(null);
+  let modalContent: HTMLDivElement | null = $state(null);
   let backPressed = $state(false);
+  // Captured in onMount right before focus moves in, restored in onDestroy.
+  // Not $state: it's read-once bookkeeping for the mount/destroy pair, not
+  // something the template reacts to.
+  let previouslyFocusedElement: HTMLElement | null = null;
 
   let {
     size = 'fit-content',
@@ -33,13 +39,27 @@
     onsecondarybuttonclick,
     onoverlayclick,
     onkeydown,
+    ondismiss,
     classes,
     overlayBackdropFilter,
     overlayFadeIn = false,
     usePortal = false,
     lockScroll = true,
-    autoDismissAfter = null
+    autoDismissAfter = null,
+    ariaLabel,
+    role,
+    overlayAriaLabel
   }: ModalProperties = $props();
+
+  // The overlay only takes role="button" and becomes keyboard-reachable when a
+  // click on it would actually do something. Without either handler, clicking
+  // or activating it is already a no-op (see handleOverlayClick's guard) --
+  // announcing it as a focusable "button" in that case is worse for assistive
+  // tech than leaving it unannounced, same reasoning already applied to the
+  // header images below.
+  const overlayDismissible = $derived(
+    typeof onoverlayclick === 'function' || typeof ondismiss === 'function'
+  );
 
   let dismissTimer: ReturnType<typeof setTimeout> | null = null;
 
@@ -67,9 +87,27 @@
     };
   };
 
+  // Resolves the real focused element regardless of whether modalContent sits
+  // in the light DOM or inside Modal.wc.svelte's open shadow root.
+  // document.activeElement stops at a shadow boundary and returns the shadow
+  // host itself, never the element actually focused inside it -- getRootNode()
+  // returns that shadow root directly when modalContent is shadow-hosted, and
+  // the owner Document otherwise.
+  const getActiveElement = (): Element | null => {
+    if (modalContent === null) {
+      return null;
+    }
+    const root = modalContent.getRootNode();
+    if (root instanceof Document || root instanceof ShadowRoot) {
+      return root.activeElement;
+    }
+    return null;
+  };
+
   const handlePopstate = (): void => {
     backPressed = true;
     onclose?.();
+    ondismiss?.();
   };
 
   const handleRightImageClick = (event: MouseEvent): void => {
@@ -92,6 +130,7 @@
     if (event.target === overlayDiv) {
       debounce(() => {
         onoverlayclick?.();
+        ondismiss?.();
       });
     }
   };
@@ -101,6 +140,49 @@
     const key = event?.key;
     if (key === 'Escape') {
       onoverlayclick?.();
+      ondismiss?.();
+      return;
+    }
+    if (key === 'Tab') {
+      const target = focusTrapTabTarget({
+        container: modalContent,
+        activeElement: getActiveElement(),
+        shiftKey: event.shiftKey
+      });
+      if (target !== null) {
+        event.preventDefault();
+        target.focus();
+      }
+    }
+  };
+
+  // The overlay's own Enter/Space activation, additional to the shared
+  // handleKeyDown (Escape/Tab/onkeydown-prop) that svelte:window already
+  // delivers here via bubbling -- this handler must not also call
+  // handleKeyDown itself, or every keydown on/under the overlay would run it
+  // twice (once from this direct call, once from the window listener seeing
+  // the same event bubble past it), double-firing onkeydown and, on Escape,
+  // the dismiss callbacks. Enter/Space activation is scoped to this handler,
+  // not folded into handleKeyDown itself, because handleKeyDown also runs
+  // from the window listener for a keypress anywhere in the modal (e.g. an
+  // input field) -- it must not treat every Enter/Space in the modal as an
+  // overlay dismissal. This handler is bound to the overlay div, so a keydown
+  // on any focusable descendant (the back button, a footer button, an input
+  // field) bubbles up and reaches it too -- the target check below is what
+  // actually stops that bubbled Enter/Space from being read as an activation
+  // of the overlay itself, mirroring the guard handleOverlayClick already has
+  // for the click case.
+  const handleOverlayKeyDown = (event: KeyboardEvent): void => {
+    if (
+      (event.key === 'Enter' || event.key === ' ') &&
+      overlayDismissible &&
+      event.target === overlayDiv
+    ) {
+      event.preventDefault();
+      debounce(() => {
+        onoverlayclick?.();
+        ondismiss?.();
+      });
     }
   };
 
@@ -131,6 +213,18 @@
       history.pushState(null, '', window.location.href);
       window.addEventListener('popstate', handlePopstate);
     }
+    // Focus starts inside the dialog once it exists, same as Sheet's trap --
+    // otherwise Tab from wherever the trigger was leaves focus behind it,
+    // outside the trap that was just installed.
+    void tick().then(() => {
+      // Capture whatever had focus (the trigger, typically) right before
+      // moving focus in, so onDestroy can put it back. Shadow-aware for the
+      // same reason the Tab trap is: inside Modal.wc.svelte's open shadow
+      // root, document.activeElement would return the shadow host instead.
+      const active = getActiveElement();
+      previouslyFocusedElement = active instanceof HTMLElement ? active : null;
+      focusEntryPoint(modalContent);
+    });
   });
 
   onDestroy(() => {
@@ -148,6 +242,13 @@
         window.removeEventListener('popstate', handlePopstate);
       }
     }
+    // Restore focus to wherever it was before the modal took it. Guarded on
+    // isConnected: the trigger can have been unmounted while the modal was
+    // open (e.g. the page navigated), and .focus() on a detached element is
+    // a silent no-op at best.
+    if (previouslyFocusedElement !== null && previouslyFocusedElement.isConnected) {
+      previouslyFocusedElement.focus();
+    }
   });
 </script>
 
@@ -155,6 +256,7 @@
 
 {#if typeof content === 'function'}
   <OverlayAnimation fadeIn={overlayFadeIn}>
+    <!-- svelte-ignore a11y_no_noninteractive_tabindex -->
     <div
       bind:this={overlayDiv}
       use:portalAction={{ usePortal }}
@@ -163,14 +265,22 @@
         ? `--modal-overlay-backdrop-filter: ${overlayBackdropFilter};`
         : null}
       onclick={handleOverlayClick}
-      onkeydown={handleKeyDown}
-      role="button"
-      tabindex="0"
+      onkeydown={handleOverlayKeyDown}
+      role={overlayDismissible ? 'button' : null}
+      tabindex={overlayDismissible ? 0 : null}
+      aria-label={overlayAriaLabel ?? null}
       data-pw={testId}
       testID={testId}
     >
       <ModalAnimation enable={enableTransition} {align} {transitionType} {entryAnimation}>
-        <div class="modal-content {size}">
+        <div
+          bind:this={modalContent}
+          class="modal-content {size}"
+          role={role ?? null}
+          aria-modal={role != null ? 'true' : null}
+          aria-label={ariaLabel ?? null}
+          tabindex="-1"
+        >
           {#if (typeof header?.leftImage === 'string' && header.leftImage.length > 0) || (typeof header?.text === 'string' && header.text.length > 0) || (typeof header?.rightImage === 'string' && header.rightImage.length > 0)}
             <div class="header">
               {#if typeof header.leftImage === 'string' && header.leftImage.length > 0}
@@ -300,6 +410,11 @@
     border-radius: var(--modal-border-radius, var(--radius, 4px));
     overflow: var(--modal-content-overflow, auto);
     border-top: var(--modal-content-border-top);
+    /* tabindex="-1" makes this the focus-trap's fallback target when the modal
+       has no focusable content of its own (see focus-trap.ts); it is never in
+       the Tab order itself, so it needs no visible focus ring the way a real
+       focusable element would. Matches Sheet's sheet-panel. */
+    outline: none;
     /* Viewport containment for every size class: only .fit-content used to carry
        a max-height, so a size whose height var is overridden to fit-content (or
        anything taller than the screen) grew past the viewport and pushed its
--- a/src/lib/Modal/focus-trap.test.ts
+++ b/src/lib/Modal/focus-trap.test.ts
@@ -1,0 +1,136 @@
+/**
+ * Unit tests for Modal's focus-trap utility (#529: "no Tab-cycling").
+ *
+ * Pure logic, no Svelte runtime and no real DOM -- a minimal stub satisfying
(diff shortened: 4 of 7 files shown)


docs/Modal.md
Tara-ag, line 1: 💡 MINOR: The web-component wrapper correctly avoids declaring `ariaLabel`/`role` as props because they are host-reserved, but the docs do not tell consumers to set the native `aria-label`/`role` attributes directly on `<sui-modal>` instead. Add a note in the Web Component section so users know the workaround.

src/lib/Modal/focus-trap.test.ts
```diff
@@ -0,0 +1,136 @@
+/**
+ * Unit tests for Modal's focus-trap utility (#529: "no Tab-cycling").
+ *
+ * Pure logic, no Svelte runtime and no real DOM -- a minimal stub satisfying
+ * only the `querySelectorAll(selector).item(i)` / `.length` shape the module
+ * actually calls, same approach as tooltip-action.test.ts.
+ *
```
Tara-ag, line 7: 💡 MINOR: Import uses the `.js` extension (`'./focus-trap.js'`), which is inconsistent with other repo tests and may fail depending on tsconfig/vitest resolution. Change to `'./focus-trap'` and confirm the test is picked up by the unit runner.

src/lib/Modal/Modal.svelte
```diff
@@ -180,13 +217,20 @@
         : null}
       onclick={handleOverlayClick}
```
Tara-ag, line 267: ⚠️ MAJOR: When `overlayDismissible` is true, this div becomes a focusable `role="button"` but has no accessible name and no Enter/Space activation handler. A `role="button"` must have an accessible name (WCAG 4.1.2) and must be keyboard-activatable (WCAG 2.1.1). Add an `aria-label` prop/handler and handle Enter/Space in `handleKeyDown`, or remove the interactive affordance and rely on click/Escape/back-press only.

src/lib/Modal/Modal.svelte
```diff
@@ -180,13 +217,20 @@
...
+      role={overlayDismissible ? 'button' : null}
+      tabindex={overlayDismissible ? 0 : null}
       data-pw={testId}
       testID={testId}
     >
       <ModalAnimation enable={enableTransition} {align} {transitionType} {entryAnimation}>
-        <div class="modal-content {size}">
+        <div
+          bind:this={modalContent}
+          class="modal-content {size}"
```
Tara-ag, line 278: 🔒 CRITICAL: Default `role="dialog"` and `aria-modal="true"` change the rendered DOM/ARIA tree for every existing Modal consumer. The PR description itself flags this as "an observable change for every existing Modal." For a published library, silently changing the accessible output that downstream themes and screen-reader users rely on is a breaking public-contract change. Either bump the major version and add a migration note, or make the new ARIA semantics opt-in (e.g., `role?: ModalRole | null` defaulting to `null`) so existing consumers remain unchanged.

src/lib/Modal/Modal.svelte
```diff
@@ -147,6 +177,12 @@
       history.pushState(null, '', window.location.href);
       window.addEventListener('popstate', handlePopstate);
     }
+    // Focus starts inside the dialog once it exists, same as Sheet's trap --
+    // otherwise Tab from wherever the trigger was leaves focus behind it,
+    // outside the trap that was just installed.
+    void tick().then(() => {
```
Tara-ag, line 219: ⚠️ MAJOR: Focus is moved on every `onMount` without saving/restoring the previously focused element. For a dialog, focus should return to the trigger on dismiss. Store `document.activeElement` before calling `focusEntryPoint` and restore it in `onDestroy`. Also guard against rapid unmount/remount cycles that steal focus repeatedly.

src/lib/Modal/focus-trap.ts
```diff
@@ -0,0 +1,71 @@
...
+ * Shift+Tab was pressed. Anywhere else in the middle of the tab order, the
+ * browser's own handling is correct and this returns null.
+ */
+export function focusTrapTabTarget(params: {
+  readonly container: ParentNode | null;
+  readonly activeElement: Element | null;
+  readonly shiftKey: boolean;
+}): HTMLElement | null {
+  const { first, last } = getFocusTrapBoundary(params.container);
+  if (first === null || last === null) {
```
Tara-ag, line 49: ⚠️ MAJOR: The trap compares `document.activeElement` to the focusable edges, but in the web-component build the modal lives inside an open shadow root. `document.activeElement` returns the host element, not the actual focused element inside the shadow tree, so the edge check will fail and Tab will escape the trap. Resolve `activeElement` from `container.getRootNode()` instead.

Tara-ag (changes requested):
Thanks for the PR — the intent (accessible Modal, focus trap, unified onDismiss) is clearly needed, but I have to request changes because of a blocking public-contract issue plus several accessibility/lifecycle concerns:

1. 🔒 **CRITICAL — breaking public contract**: Rendering `role="dialog"` and `aria-modal="true"` by default changes the ARIA tree for every existing Modal consumer. The PR description itself calls this out. For a published library this needs a major-version bump or an opt-in design.
2. ⚠️ **MAJOR — overlay button accessibility**: The overlay `role="button"` has no accessible name and no Enter/Space handler.
3. ⚠️ **MAJOR — focus restoration**: Initial focus is moved on mount but never restored on destroy.
4. ⚠️ **MAJOR — shadow-DOM focus trap**: `document.activeElement` breaks the edge detection inside the web-component shadow root.
5. 💡 **MINOR — test import extension** and missing web-component docs note.

Please address the critical item and the major items, then re-request review.

src/lib/Modal/Modal.svelte
Tara-ag, line 219: ⚠️ MAJOR: Focus is moved on every `onMount` even when the modal is conditionally re-mounted. The `onMount` callback unconditionally calls `focusEntryPoint(modalContent)` after `tick()`. If a parent toggles the modal's visibility by unmounting/remounting, focus is stolen on every remount. Focus is also not restored to the trigger element on close, which is expected dialog behavior.

**Suggestion:** Store the previously focused element before moving focus (e.g., `const previousActiveElement = document.activeElement;` before `focusEntryPoint`), and restore it in `onDestroy`. Also consider guarding the initial focus so it only happens when the modal is actually open/visible.

src/lib/Modal/Modal.svelte
Tara-ag, line 267: ⚠️ MAJOR: Overlay `role="button"` with `tabindex="0"` lacks accessible name and keyboard activation semantics. A `role="button"` must have an accessible name (WCAG 2.1 4.1.2) and must be activatable with Enter/Space (WCAG 2.1 2.1.1). The overlay currently has no `aria-label`, no inner text, and no Enter/Space handler on its `onkeydown`. Screen-reader users will hear an unnamed "button" and keyboard-only users cannot activate it.

**Suggestion:** Add an `aria-label` (e.g., a new `overlayAriaLabel` prop defaulting to `"Close"`) and handle Enter/Space in `handleKeyDown` to call the dismiss handlers when the overlay has focus. Alternatively, keep the overlay as a non-interactive container and rely on click/Escape/back-press only, removing `role="button"`/`tabindex` entirely.

src/lib/Modal/focus-trap.ts
Tara-ag, line 49: ⚠️ MAJOR: Tab trap uses `document.activeElement` at event time without accounting for shadow DOM. `focusTrapTabTarget` compares `params.activeElement` (`document.activeElement`) to the first/last focusable elements found via `modalContent.querySelectorAll`. When the modal is rendered inside a ShadowRoot (web-component build with `shadow: 'open'`), `document.activeElement` points to the host element, not the actual focused element inside the shadow tree. The comparison will fail, breaking the focus trap for web-component consumers.

**Suggestion:** Pass the modalContent's root (`modalContent.getRootNode()`) and read `activeElement` from that root: `const activeElement = root instanceof ShadowRoot ? root.activeElement : document.activeElement;`. Update `focusTrapTabTarget` to accept the active element already resolved by the caller, or resolve it inside `getFocusTrapBoundary` using `container.getRootNode()`.

src/lib/Modal/Modal.svelte
Tara-ag, line 278: 🔒 CRITICAL: Default `role="dialog"` and `aria-modal="true"` change the rendered DOM for every existing Modal consumer. Because `role` defaults to `'dialog'`, every existing consumer that does not set `role` now gets `role="dialog"` and `aria-modal="true"` where neither attribute existed before. For a published library, a silent change to the rendered DOM/ARIA tree that affects assistive-technology announcements is a breaking public-contract change unless accompanied by a major-version bump or migration note.

**Suggestion:** Either (a) bump the major version and add a migration note, or (b) make the new ARIA semantics opt-in (e.g., `role?: ModalRole | null` defaulting to `null` and only render `aria-modal` when `role` is explicitly provided) so existing consumers are unchanged. If keeping the new default, document the breaking change and bump to 4.0.0.

src/lib/Modal/focus-trap.test.ts
Tara-ag, line 7: 💡 MINOR: New focus-trap utility test imports from `./focus-trap.js`. In a TypeScript/Vitest project this is inconsistent with other repo tests and may fail depending on tsconfig/vitest resolution settings.

**Suggestion:** Change the import to `./focus-trap` (without extension) to match repo conventions and avoid resolution surprises. Confirm `pnpm test:unit --run src/lib/Modal/focus-trap.test.ts` passes.

docs/Modal.md
Tara-ag, line 1: 💡 MINOR: Web-component wrapper does not forward `ariaLabel`/`role` but also does not document the native-attribute workaround. The PR correctly avoids declaring `ariaLabel` and `role` as custom-element props because they are host-reserved. However, the web-component docs section does not mention that `sui-modal` consumers should set the native `aria-label`/`role` attributes directly on the host.

**Suggestion:** Add a sentence in the Web Component section noting that `aria-label` and `role` should be set directly on the `<sui-modal>` host element because they are reserved native accessors, and that `onDismiss` is the only new prop exposed on the wrapper.

Tara-ag (changes requested):
Yama review complete. One CRITICAL and three MAJOR issues need resolution before merge; see inline comments.

Tara-ag:

Review time budget expired before the primary implementation file (src/lib/Modal/Modal.svelte) could be retrieved in full; only partial/truncated content was available. The supporting files (properties.ts, focus-trap.ts, focus-trap.test.ts, Modal.wc.svelte, docs) look thoughtful and preserve backward compatibility by making role/aria-modal opt-in and by correctly excluding host-reserved ARIA accessors from the custom-element prop map. However, I cannot verify the actual integration of the focus trap, event wiring for ondismiss, cleanup of listeners/timers, SSR guards, and focus restoration without reading the full Modal.svelte. A follow-up review of the complete Modal.svelte diff is needed before merge.

docs/_index.json
```diff
@@ -161,7 +161,7 @@
   },
   {
     "name": "Modal",
-    "description": "A dialog overlay component that renders on top of the page with configurable size, alignment, header, footer, and transition animations. entryAnimation can override the per-align default, e.g. a slide-up bottom-sheet-style dialog while centered. Locks body scroll and optionally handles hardware back-press navigation. Supports click-to-dismiss and Escape key."
+    "description": "A dialog overlay component that renders on top of the page with configurable size, alignment, header, footer, and transition animations. entryAnimation can override the per-align default, e.g. a slide-up bottom-sheet-style dialog while centered. Locks body scroll and optionally handles hardware back-press navigation. The content panel carries role/aria-modal by default (optionally an ariaLabel) and traps Tab focus; onDismiss unifies overlay-click, Escape and hardware-back-press dismissal."
```
Tara-ag: 💡 MINOR: This description says the content panel "carries role/aria-modal by default", but the implementation only renders `role` and `aria-modal` when the caller explicitly supplies the `role` prop. That conflicts with both the implementation and the PR description, which emphasizes that both are opt-in to preserve existing DOM/ARIA behavior.

Suggested fix: rephrase to indicate that role/aria-modal are opt-in via the `role` prop, e.g. "optionally carries role/aria-modal when the role prop is set".

src/lib/Modal/Modal.svelte
```diff
@@ -117,6 +156,39 @@
...
+      });
+      if (target !== null) {
+        event.preventDefault();
+        target.focus();
+      }
+    }
+  };
+
+  // The overlay's own keydown handling, layered on top of the shared
+  // handleKeyDown (Escape/Tab/onkeydown-prop, which svelte:window already
```
Tara-ag: ⚠️ MAJOR: `handleOverlayKeyDown` calls `handleKeyDown(event)`, but `handleKeyDown` is also attached to `<svelte:window onkeydown={handleKeyDown}>`. Because keydown events inside the modal bubble up through the overlay div and then to window, every keypress inside the modal triggers `handleKeyDown` twice. For Escape this means `onkeydown`, `onoverlayClick`, and the newly-added `onDismiss` each fire twice, which contradicts the documented "once per dismiss path" contract for `onDismiss`.

Suggested fix: remove `handleKeyDown(event);` from `handleOverlayKeyDown`. The window listener already handles Escape, Tab, and the `onkeydown` prop for any bubbled keydown inside the modal; the overlay div's own handler should only be responsible for Enter/Space activation of the dismissible backdrop.

src/lib/Modal/Modal.svelte
```diff
@@ -179,14 +271,22 @@
         ? `--modal-overlay-backdrop-filter: ${overlayBackdropFilter};`
```
Tara-ag, line 265: ⚠️ MAJOR: The dismissible overlay is given `tabindex="0"`, making it a tab stop, but the focus-trap logic only queries focusable descendants of `modalContent`. Since the overlay is an ancestor/sibling of `modalContent`, not a descendant, it is excluded from the trap. Shift+Tab from the first focusable element therefore moves focus to the overlay instead of wrapping to the last element, and Shift+Tab from the overlay can leave the modal entirely. This breaks the PR's stated "Tab is trapped inside the panel" behavior.

Suggested fix: either remove `tabindex` from the overlay and rely on Escape/close buttons for keyboard dismissal, or extend the trap so the overlay participates as a boundary when computing first/last tab stops.

Tara-ag (changes requested):
Thanks for the PR. I reviewed the changed files and found three real issues:

1. **Double keydown handling** (MAJOR, `src/lib/Modal/Modal.svelte`): `handleOverlayKeyDown` calls `handleKeyDown`, which is also attached to `<svelte:window onkeydown>`. Because events bubble, every keypress inside the modal fires `handleKeyDown` twice. For Escape this means `onkeydown`, `onoverlayClick`, and the new `onDismiss` callback each fire twice, violating the documented "once per dismiss path" contract.

2. **Incomplete focus trap** (MAJOR, `src/lib/Modal/Modal.svelte`): the dismissible overlay is given `tabindex="0"`, but the focus-trap logic only considers focusable descendants of `modalContent`. The overlay is outside that container, so Shift+Tab from the first focusable element lands on the overlay instead of wrapping, and focus can leave the modal entirely.

3. **Stale docs index description** (MINOR, `docs/_index.json`): it says the content panel "carries role/aria-modal by default", but the implementation only renders those attributes when the `role` prop is explicitly set.

Please address the two MAJOR items; the docs index tweak is optional but appreciated.

src/lib/Modal/Modal.svelte
```diff
@@ -117,6 +156,39 @@
...
+  // handleKeyDown (Escape/Tab/onkeydown-prop, which svelte:window already
+  // delivers here via bubbling regardless -- kept so the div's own listener
+  // still fires those when this element itself has focus, matching the
+  // pre-existing wiring). Enter/Space activation is scoped to this handler,
+  // not folded into handleKeyDown itself, because handleKeyDown also runs
+  // from the window listener for a keypress anywhere in the modal (e.g. an
+  // input field) -- it must not treat every Enter/Space in the modal as an
+  // overlay dismissal.
+  const handleOverlayKeyDown = (event: KeyboardEvent): void => {
+    handleKeyDown(event);
```
Tara-ag: ⚠️ MAJOR: `handleOverlayKeyDown` calls `handleKeyDown(event)`, but `handleKeyDown` is also bound to `<svelte:window onkeydown>`. Because keydown events inside the modal bubble through the overlay and up to window, every keypress inside the modal invokes `handleKeyDown` twice. For Escape this causes `onkeydown`, `onoverlayClick`, and `onDismiss` to fire twice. Remove `handleKeyDown(event);` from this handler; the window listener already covers Escape, Tab, and the `onkeydown` prop for any bubbled keydown inside the modal.

docs/_index.json
Tara-ag: 💡 MINOR: This description says the content panel "carries role/aria-modal by default", but the implementation only renders `role` and `aria-modal` when the caller explicitly sets the `role` prop. This contradicts both the implementation and the PR description, which emphasize that both are opt-in. Update to say the panel "optionally carries role/aria-modal when the `role` prop is set".

src/lib/Modal/Modal.svelte
```diff
@@ -179,14 +271,22 @@
...
-      role="button"
-      tabindex="0"
+      onkeydown={handleOverlayKeyDown}
+      role={overlayDismissible ? 'button' : null}
+      tabindex={overlayDismissible ? 0 : null}
+      aria-label={overlayAriaLabel ?? null}
       data-pw={testId}
       testID={testId}
     >
       <ModalAnimation enable={enableTransition} {align} {transitionType} {entryAnimation}>
```
Tara-ag, line 275: ⚠️ MAJOR: When the overlay is dismissible it receives `tabindex="0"`, making it a tab stop, but the focus-trap logic only queries focusable descendants of `modalContent`. The overlay is outside `modalContent`, so it is excluded from the trap. Shift+Tab from the first focusable element lands on the overlay instead of wrapping to the last, and Shift+Tab from the overlay can leave the modal entirely. Either remove `tabindex` from the overlay and rely on Escape/close buttons for dismissal, or extend the trap so the overlay is included as a boundary element.

Tara-ag (changes requested):
Two MAJOR issues block approval: (1) keydown handlers fire twice because `handleOverlayKeyDown` calls `handleKeyDown` while the same function is also bound to `svelte:window`, causing Escape/onDismiss/onoverlayClick to double-fire; (2) the dismissible overlay is a tab stop but sits outside the focus-trap container, so Tab focus can escape the modal. One MINOR docs inconsistency in `docs/_index.json` also noted.

src/lib/Modal/focus-trap.test.ts
```diff
@@ -0,0 +1,136 @@
+/**
```
Tara-ag, line 1: 💡 MINOR: The unit tests validate the pure logic of boundary selection and wrapping, which is useful, but the PR's riskiest claims — shadow-root activeElement resolution, focus restoration on unmount, and mount-time focus movement — are not exercised. These stubs will not catch issues such as the overlay `tabindex` interfering with the trap or focus escaping the modal.

Suggested fix: add at least one component-level test (using `@testing-library/svelte` or vitest browser mode) that mounts `Modal` with focusable content, verifies focus moves to the first focusable, verifies Tab wraps, and verifies focus restores on unmount. If browser tests are unavailable, add a jsdom test for the shadow-aware `getActiveElement` helper.

src/lib/Modal/Modal.svelte
```diff
@@ -163,14 +264,22 @@
         ? `--modal-overlay-backdrop-filter: ${overlayBackdropFilter};`
         : null}
       onclick={handleOverlayClick}
-      onkeydown={handleKeyDown}
-      role="button"
-      tabindex="0"
+      onkeydown={handleOverlayKeyDown}
+      role={overlayDismissible ? 'button' : null}
+      tabindex={overlayDismissible ? 0 : null}
+      aria-label={overlayAriaLabel ?? null}
```
Tara-ag, line 271: ⚠️ MAJOR: The overlay `<div>` is made focusable with `tabindex={overlayDismissible ? 0 : null}` and `role="button"`. Because `focus-trap.ts` only queries descendants of `modalContent`, the overlay is outside the trap boundary. When focus is on the overlay, Tab moves into `modalContent`, but Shift+Tab from the first focusable inside `modalContent` moves focus to the overlay instead of wrapping to the last element. From there, another Tab can move focus out of the modal entirely (to browser chrome or the page behind). A backdrop click target should not be a Tab stop; ESC and click already dismiss it.

Suggested fix: remove `tabindex="0"` from the overlay so it is not in the Tab order. Keep the click handler and `role="button"` only when dismissible, but do not make it keyboard-focusable. If you intentionally want it reachable, include it in the focus-trap boundary and ensure wrap-around works.

src/wc/components/Modal.wc.svelte
```diff
@@ -30,7 +31,13 @@
       onheaderleftimageclick: { type: 'Object' },
```
Tara-ag, line 31: ⚠️ MAJOR: The Svelte component adds `ariaLabel` and `role` props, but the web-component wrapper declares neither. While `ariaLabel` and `role` are correctly classified as HOST_RESERVED_PROPS, the repo's established pattern (see `Toggle.wc.svelte`: `inputId`, `inputAriaLabel`, `inputAriaLabelledby`) is to expose prefixed aliases that forward to the inner component. Without `modalAriaLabel`/`modalRole` aliases, web-component consumers cannot set the modal's accessible name or ARIA role — the native host `aria-label`/`role` attributes stay on `<sui-modal>` and are not wired to the inner Modal content panel. This breaks feature parity between the Svelte and web-component builds for a new accessibility feature.

Suggested fix: add prefixed props to the `customElement` declaration, destructure them in `<script>`, and forward them to `<Modal ariaLabel={modalAriaLabel} role={modalRole} ... />`. Update `docs/Modal.md` web-component section accordingly.

Tara-ag (changes requested):
Reviewed file-by-file. Two MAJOR accessibility/parity issues and one MINOR test-coverage gap need addressing before merge. Details in inline comments.

Tara-ag (changes requested):

Two MAJOR findings remain unresolved:

1. **Focus trap is incomplete** — the dismissible overlay is made tabbable but sits outside `modalContent`, so focus can escape the modal instead of wrapping.
2. **Web-component parity gap** — new `ariaLabel`/`role` props are not forwarded to the web-component build, so `<sui-modal>` consumers cannot set the modal's accessible name or ARIA role through the component API.

One MINOR finding also remains: the focus-trap unit tests cover the pure utility but do not exercise the shadow-DOM active-element resolution, mount-time focus movement, or focus-restoration behavior that the PR introduces.

Please address the two MAJOR items before merging.

murdore:
## Video evidence

**What the recording shows**

In one continuous take: (1) clicking "Open modal with back button" opens the modal and focus moves onto its first focusable control, the back button (not merely appearing — an actual focus change); (2) pressing Enter then Space on that back button increments its own on-screen click counter (clicks: 1, then clicks: 2) while the modal stays open, proving a keypress on an inner non-dismiss control does not close the modal; (3) Tab moves focus to the close button, a further Tab wraps focus forward back to the back button (only two focusables exist, so this Tab-wrap is the whole trap), and Shift+Tab from the back button wraps backward to the close button — Tab focus never escapes the dialog; (4) pressing Escape closes the modal and focus is restored to the "Open modal with back button" trigger button that opened it. All three required behaviours (focus trapped, focus restored to opener, inner keypress not dismissing) are visible in this single recording.

> expect(locator('[data-pw="back-button-left-image"]')).toBeFocused() timed out (Expected: focused, Received: inactive) — on origin/release, opening the modal never moves focus into it at all (no focus-entry, no trap, no Tab-wrap), so the very first assertion in the identical spec fails immediately. Saved as pr-543-modal-FAILS-on-release.webm in the same evidence directory.

- passing, on this branch: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-543-modal.webm` (249,930 bytes)
- failing, on `release`: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-543-modal-FAILS-on-release.webm` (164,734 bytes)

- test-results/modal-a11y-evidence-focus--ceed9--and-focus-restore-on-close-functional/ leftover Playwright report artifacts remain under test-results/ from the evidence run, but this directory is gitignored (.gitignore:17) and `git status`/`git clean -ndx` confirm it is untracked and does not dirty the tree — noted for completeness, not a blocker.

src/lib/Modal/focus-trap.ts
```diff
@@ -0,0 +1,71 @@
...
+ * (see tooltip-action.test.ts).
+ *
+ * Modal.svelte owns the event wiring (which key was pressed, calling
+ * .focus()); this module owns the two decisions that wiring needs answered:
+ * "what are the edges of the tab order in here" and "given where focus is and
+ * which way Tab was pressed, should it wrap, and to where".
+ */
+
+const FOCUSABLE_SELECTOR =
+  'a[href], button:not([disabled]), textarea:not([disabled]), input:not([disabled]), select:not([disabled]), [tabindex]:not([tabindex="-1"])';
```
Tara-ag, line 14: 💬 SUGGESTION: `FOCUSABLE_SELECTOR` omits `details/summary`, `audio`/`video` with controls, `iframe`, and `[contenteditable]` elements. While most Modal content uses buttons/links/inputs, a consumer placing a details disclosure, embedded media, or rich-text editor inside the modal would find Tab cycling past it and leaving the trap. Consider expanding the selector to include `details summary:not([tabindex="-1"]), audio[controls], video[controls], iframe, [contenteditable]:not([contenteditable="false"])` (with `:not([disabled])` / `:not([inert])` scoping where appropriate).

docs/Modal.md
```diff
@@ -238,6 +248,8 @@ Tag: `<sui-modal>`
 </sui-modal>
 ```
 
+`aria-label` and `role` are not props on `<sui-modal>` — they're reserved accessors on every custom element (`ARIAMixin`/`role`), so declaring them as component props would replace the native ones instead of adding to them. Set them as native attributes directly on the host, e.g. `<sui-modal role="alertdialog" aria-label="Delete account">`. `ondismiss` (and `overlayAriaLabel`) are ordinary props and work the same way as on the Svelte component.
+
 ### Slots
```
Tara-ag, line 253: 💡 MINOR: The Svelte component automatically adds `aria-modal="true"` whenever `role` is supplied, but the web component cannot do that because `role`/`aria-label` are host-reserved. The docs here tell consumers to set `role` and `aria-label` as native host attributes, but they don't mention that `aria-modal="true"` must also be set manually to match the Svelte behavior. A consumer following the current example will get a dialog that is not announced as modal.

Suggested update: change the example to `<sui-modal role="alertdialog" aria-modal="true" aria-label="Delete account">` and add a short note that `aria-modal` is not automatic for the web component the way it is for the Svelte component when `role` is supplied.

src/lib/Modal/Modal.svelte
```diff
@@ -101,6 +140,49 @@
     const key = event?.key;
```
Tara-ag, line 140: 💬 SUGGESTION: The new focus-trap utility is unit-tested in isolation, but the Modal component integration (keydown wiring, focus entry/restoration, `ondismiss` firing on overlay click/Escape/hardware back-press, and conditional `role`/`aria-modal`/`aria-label` rendering) is not covered by component-level tests. Consider adding a small `Modal.svelte` mount test that asserts: (1) Tab from the last focusable element wraps to the first, (2) Escape fires both `onoverlayclick` and `ondismiss`, (3) `role`/`aria-modal`/`aria-label` render on `.modal-content` when supplied, and (4) focus returns to the previously focused element on unmount.

Tara-ag (approved):
Reviewed file-by-file. No blocking issues found. The PR is a solid accessibility improvement: it makes `role`/`aria-label` opt-in, adds a real focus trap with focus restoration, and introduces a unified `ondismiss` callback while preserving existing events. I left three non-blocking comments: one MINOR doc gap about `aria-modal` for the web component, and two SUGGESTIONS around component-level test coverage and expanding the focusable-element selector. None of these warrant blocking the merge.

Tara-ag (commented):
Reviewed file-by-file. No blocking issues found. The PR is a solid accessibility improvement.

docs/Modal.md
Tara-ag, line 253: 💡 MINOR: Web component docs omit `aria-modal` pairing

The Svelte component pairs `role` with `aria-modal="true"` automatically, but the web component cannot because `role`/`ariaLabel` are host-reserved. The docs tell consumers to set `role` and `aria-label` as native host attributes, but they do not mention that `aria-modal="true"` must also be set manually to match the Svelte behavior. A consumer following the current example will get a dialog that is not announced as modal.

**Suggestion:** Update the sentence to: "Set them as native attributes directly on the host, e.g. `<sui-modal role="alertdialog" aria-modal="true" aria-label="Delete account">`." Also add a brief note that `aria-modal` is not automatic for the web component the way it is for the Svelte component when `role` is supplied.

src/lib/Modal/focus-trap.ts
Tara-ag, line 14: 💬 SUGGESTION: Focusable selector could cover more element types

`FOCUSABLE_SELECTOR` omits `details/summary`, `audio`/`video` with controls, `iframe`s, and `contenteditable` elements. While most Modal content uses buttons/links/inputs, a consumer placing a details disclosure, embedded video, or rich-text editor inside the modal would find Tab cycling past it and leaving the trap.

**Suggestion:** Consider expanding the selector to include `details summary:not([tabindex="-1"]), audio[controls], video[controls], iframe, [contenteditable]:not([contenteditable="false"])` (and scope them with `:not([disabled])` / `:not([inert])` where appropriate).

src/lib/Modal/Modal.svelte
Tara-ag, line 140: 💬 SUGGESTION: Add component-level Modal tests for new behavior

The new focus-trap utility is unit-tested in isolation, but the Modal component integration (`handleKeyDown` wiring, focus entry/restoration, `ondismiss` firing on overlay click/Escape/hardware back-press, and conditional `role`/`aria-modal`/`aria-label` rendering) is not covered by component-level tests. Adding a small `Modal.svelte` mount test would guard against regressions in the wiring.

**Suggestion:** Add a `Modal.svelte` component test (Vitest + `@testing-library/svelte` or the repo's existing harness) that asserts: (1) Tab from the last focusable element wraps to the first, (2) Escape fires both `onoverlayclick` and `ondismiss`, (3) `role`/`aria-modal`/`aria-label` render on `.modal-content` when the props are supplied, and (4) focus returns to the previously focused element on unmount.

Tara-ag (commented):
Reviewed file-by-file. No blocking issues found. The PR is a solid accessibility improvement: it makes role/aria-label opt-in, adds a real focus trap with focus restoration, and introduces a unified ondismiss callback while preserving existing events. I left three non-blocking comments: one MINOR doc gap about aria-modal for the web component, and two SUGGESTIONS around component-level test coverage and expanding the focusable-element selector. None of these warrant blocking the merge.

PR #544, opened 2026-09-06
feat(snippet): configurable copy feedback and unmount-safe reset timer

### LINKED ISSUE #530: Snippet: Configurable copied-label, reset duration, and unmount-safe timer cleanup on Snippet's b…
## Need
Configurable copied-label, reset duration, and unmount-safe timer cleanup on Snippet's built-in copy affordance

## Closest today
Snippet (hardcoded 'Copied!' text, 2000ms reset, no effect cleanup on the internal setTimeout)

## Proposal
Add `copiedLabel?: string` and `copyResetMs?: number` props, and wrap the internal timeout in an `$effect` cleanup so navigating away mid-timer doesn't leak it.

## Where it bites
Found while migrating TARA (a Svelte 5 / SvelteKit 2 consumer pinned to 3.3.0) onto the library. Without this the following consumer files can only migrate part-way and keep hand-rolled markup for the rest: `CopyButton.svelte`, `LaunchSheet.svelte`, `ShareSheet.svelte`.

murdore:
Adds two optional props to Snippet's built-in copy affordance:
`copiedLabel` (text shown in place of the icon after a copy, default
`'Copied!'`) and `copyResetMs` (delay before it reverts, default
`2000`). Neither prop changes anything for an existing consumer that
passes neither -- the defaults reproduce today's hardcoded text and
timing exactly.

The reset timer itself moves into `copyResetTimer.ts`, a small
Svelte-free module (`copyResetTimer.test.ts` unit-tests it directly,
the way `Tooltip/tooltip-action.ts` keeps its DOM timing outside the
component). It guarantees only one timer is ever pending -- a second
copy before the first one's flash reverted used to leave two timers
racing over `copied` -- and Snippet now clears it from `onDestroy`, so
navigating away mid-flash no longer fires against a destroyed
component.

The consumer that hit this is TARA's ShareSheet.svelte, which renders a
share URL through Snippet and wanted its own copied label and a shorter
flash than the hardcoded 2s. (TARA also has a CopyButton.svelte with
the same affordance hand-rolled over Button, including a 1.5s timer
with effect cleanup -- that one is not Snippet-based and these props do
not replace it, but they do remove the reason a future Snippet-based
consumer would hand-roll it again.)

docs/Snippet.md, docs/_index.json, the wc wrapper and its parity, and
the component demo page are updated to match; docs/CHANGELOG.md is
release-generated and untouched here.

Closes #530

---

### Verification

**Spec runs on this branch** — `src/lib/Snippet/copyResetTimer.test.ts` (vitest) and `tests/snippet-copy-options.test.ts` (2 Playwright tests against a real build).

**Fail-before** — Source reverted to `origin/release` with only the specs kept: **both fail**.


Tara-ag (commented):

src/lib/Snippet/Snippet.svelte
```diff
@@ -46,7 +59,7 @@
     <div class="snippet-copy">
       <Button onclick={handleCopy} ariaLabel="Copy to clipboard">
         {#if copied}
-          <span class="snippet-copied">Copied!</span>
+          <span class="snippet-copied">{copiedLabel}</span>
```
Tara-ag: 💡 MINOR: The copy button has a fixed `ariaLabel="Copy to clipboard"` that overrides its content. When copying succeeds, sighted users see the configurable `copiedLabel` text (e.g. "Copied!"), but screen-reader users continue to hear only "Copy to clipboard" and receive no success announcement. This gap becomes more visible now that `copiedLabel` is configurable.

Add `aria-live="polite"` to the `.snippet-copied` span so assistive technologies announce the feedback when it appears:

```svelte
<span class="snippet-copied" aria-live="polite">{copiedLabel}</span>
```

Tara-ag (approved):
Approved with one minor accessibility suggestion. The change is additive, backward-compatible, and well-tested.

src/lib/Snippet/Snippet.svelte
Tara-ag: 💡 MINOR: Copied feedback is not announced to screen readers. The copy button has a fixed `ariaLabel="Copy to clipboard"` that overrides its content. When copying succeeds, sighted users see the configurable `copiedLabel` text (e.g. "Copied!"), but screen-reader users continue to hear only "Copy to clipboard" and receive no success announcement. This gap becomes more visible now that `copiedLabel` is configurable.

**Suggestion:** Add `aria-live="polite"` to the `.snippet-copied` span so assistive technologies announce the feedback when it appears:

```svelte
<span class="snippet-copied" aria-live="polite">{copiedLabel}</span>
```

Tara-ag (commented):
Reviewed file-by-file. The change is additive and backward-compatible: two new optional props (`copiedLabel`, `copyResetMs`) with defaults matching the previous hardcoded behavior, plus a safer timer lifecycle. One minor accessibility improvement suggested.

src/lib/Snippet/Snippet.svelte
```diff
@@ -46,7 +59,7 @@
     <div class="snippet-copy">
       <Button onclick={handleCopy} ariaLabel="Copy to clipboard">
         {#if copied}
-          <span class="snippet-copied">Copied!</span>
+          <span class="snippet-copied" aria-live="polite">{copiedLabel}</span>
```
Tara-ag, line 58: 💡 MINOR: The `aria-live="polite"` region is created only when `copied` is true, so screen readers may miss the announcement because the live region did not exist before the content change. Consider keeping a persistent (visually hidden) live region whose text is updated, or render the span earlier so the region is present before the copy action.

src/lib/Snippet/copyResetTimer.test.ts
```diff
@@ -0,0 +1,70 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
```
Tara-ag, line 1: 💬 SUGGESTION: The timer tests cover normal, re-arm, cancel and no-op paths well, but do not exercise boundary/invalid `delayMs` values (0, negative, NaN). Adding those cases would make the module's contract explicit and guard against regressions if validation is introduced later.

Tara-ag (approved):
Approved with minor accessibility and test-coverage notes. The change is additive, backward-compatible, and correctly wired through the public barrel, web-component wrapper, docs, and tests.

src/lib/Snippet/copyResetTimer.test.ts
Tara-ag, line 1: 💬 SUGGESTION: Unit test does not exercise invalid `delayMs` values. `copyResetTimer.test.ts` covers normal, re-arm, cancel and no-op cases but does not assert behavior for invalid `delayMs` such as `0`, negative, or `NaN`. Since the timer is now the authority for this timing, adding tests for boundary/invalid inputs would prevent regressions if validation is added later.

src/lib/Snippet/Snippet.svelte
Tara-ag, line 58: 💡 MINOR: aria-live region may not be announced reliably. The copied feedback span is rendered conditionally inside the button only when `copied` is true. Screen readers may not announce the change reliably because the aria-live region is inserted into the DOM at announcement time rather than persisting with content changes. Consider adding a persistent visually-hidden `aria-live="polite"` element whose textContent is updated to `copiedLabel` when copied is true and cleared when false, or keep the current span but ensure it is present in the DOM before the copy action occurs.

Tara-ag (commented):
Review complete. The change is backward-compatible and well-tested. Two non-blocking notes: the conditional aria-live region could be more robust, and the timer unit tests could cover invalid delay values.

murdore:
## Video evidence

**What the recording shows**

Browser opens /components/snippet, clicks the Copy-to-clipboard button on the 'snippet-copy-options' Snippet instance (text='echo custom', copiedLabel='Link copied!', copyResetMs={300}). The copy affordance switches from the copy icon to the text 'Link copied!' (not the library's hardcoded default 'Copied!'), clipboard content is verified as 'echo custom', and after ~300ms (well short of the 2000ms library default) the label disappears and the affordance reverts to the copy icon -- directly demonstrating the PR's custom-label-and-custom-reset-delay behavior on the real demo route.

> Same spec re-run against a build of origin/release (7b44a5a) with only tests/snippet-copy-options.test.ts kept from the PR diff (copyResetTimer.ts/.test.ts removed since they don't exist pre-PR; Snippet.svelte, properties.ts, +page.svelte, Snippet.wc.svelte, docs, and the visual screenshot reverted): both tests failed with 'Test timeout of 30000ms exceeded' waiting for getByTestId('snippet-default'/'snippet-copy-options').getByRole('button', {name:'Copy to clipboard'}) -- the demo route has no such test-ids and Snippet has no copiedLabel/copyResetMs props pre-PR, so the locator never resolves. Failure video saved to .../evidence/pr-544-snippet-FAILURE-on-release.webm (440600 bytes). Worktree fully restored afterward: git checkout HEAD -- . && git clean -fd (excluding node_modules/build dirs); git status --short is clean; git log --oneline -1 still shows 625c7a2 feat(snippet): configurable copy feedback and unmount-safe reset timer, untouched.

- passing, on this branch: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-544-snippet.webm` (98,902 bytes)
- failing, on `release`: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-sara/e4990512-e73f-4e78-aa2c-0aa1aee2616d/scratchpad/evidence/pr-544-snippet-FAILURE-on-release.webm` (440,600 bytes)

- Evidence video is only ~96.6KB, below the 'a few hundred KB at least' heuristic in the task -- flagged for visibility even though frame extraction confirmed the content is genuine and non-trivial (duration 2.36s, correct behavior visible).
- git fetch origin confirms origin/release is still at 7b44a5a (unmoved since the rebase report) -- if release advances again before merge, the PR will need a fresh rebase before merging, though it is not stale as of this check.

Tara-ag (approved):
Reviewed file-by-file. The change is additive and backward-compatible: new props default to the previous hardcoded behavior, timer cleanup is correct, docs/wc/demo/tests are updated, and no public API contracts or CSS variables are broken.

PR #546, opened 2026-09-06
feat(ChatMessage): reveal a streamed message progressively via TypewriterText

murdore:
## Why

Lighthouse hand-rolls this in an `AssistantMessage` wrapper: a `body` snippet that splits the reply into `message.characters` and renders each as a faded `<span>` while streaming, then swaps to `renderMarkdown(message.text)` once the turn ends. Two other consumers wrap `ChatMessage` the same way. That wrapper is the last thing standing between those three surfaces and using `ChatMessage` directly — `clampLines` (#533) removed the other one.

## What this is not

It is not a second streaming implementation. `TypewriterText` already does the hard part and I did not need to write any of it:

- it continues from `previousTextLength` as `text` grows rather than restarting — the streaming case
- it bulk-reveals the remainder the moment `isStreaming` goes `false` — the turn-ended case
- `renderText` exists precisely so the reveal can go through a markdown renderer

It simply was not wired to `ChatMessage`. This connects two components that already ship.

## Contract

`typewriter` drives off `markdown` or `content`, **never `html`** — typing a pre-rendered string character by character would put its tags on screen. When the source is `markdown` it types through the same pipeline the static branch already uses, so rich text reveals as rich text.

That is better than the hand-rolled version it replaces, where the streaming half showed bare characters and only the settled half was markdown.

`streaming` selects the mode, so the two compose into the shape a chat surface needs. `streaming` on its own still means what it always did. `hasContent` now counts an active typewriter — otherwise a streaming message paints the typing indicator over the text it is revealing.

## Tests

The assertions are about how much text is on screen at a given moment, because the way this degrades is by silently ceasing to be a reveal — painting everything at once, or never completing.

```
mid-reveal: 0 < visible < full
it keeps growing on its own while streaming
ending the stream shows the remainder at once
`**streamed**` renders a <strong>, not literal asterisks
a non-streaming typewriter message shows in full
a message without the prop renders no .typewriter-text at all
```

Negative controls — the fix is reverted, the failure is confirmed to name the right thing, and the tree is restored byte-identically (`sha256 e07b5da5…`):

| control | failure |
|---|---|
| `typewriterActive` forced false | `Expected: < 50, Received: 51` |
| `renderText` not passed through | `<strong>` "streamed" never appears |

The visual baseline is regenerated for the new demo column and compared **twice**, since an animated demo can be flaky the way `thinking-indicator` is — 93 passed both times, exactly one screenshot changed.

```
lint 0 · check 0 errors · 413 unit · 489 functional · 93 visual · 33 docs-accuracy · prop-parity 90
```

## Follow-up

With this and `clampLines`, Lighthouse can delete both `AssistantMessage` and `UserMessage` and call `ChatMessage` directly at all three of its surfaces.

## Summary by CodeRabbit

- **New Features**
  - Added an optional typewriter effect to progressively reveal chat message content.
  - Added configurable reveal speed.
  - Supports streamed and preloaded Markdown content while preserving rendered formatting.
  - Added support for the new options in the web component interface.
  - Added demonstrations of progressive reveal and streaming behavior.

- **Documentation**
  - Documented typewriter options and their interaction with streaming, including accessibility behavior.

## Visual baseline delta

The regenerated `chat-message.png` is a binary diff, so stating it in words: the change is **one new demo column, "Progressive reveal"**, appended after the clamp column. It holds an assistant bubble that is empty at rest (the streaming demo starts with no text), a Start and a Finish button, a bubble reading `A bold claim present at first paint.` with **bold** emboldened, and a bubble reading `A message that is not streaming shows in full immediately.` The page grows taller by that column; nothing above it moves.

Regenerated in the pinned container and compared **twice** — an animated demo can be flaky the way `thinking-indicator` is, which is excluded from the suite for exactly that reason. Exactly one screenshot changed (`git status --porcelain tests/visual/__screenshots__/ | wc -l` → 1); a regeneration that quietly rewrote other baselines would look the same in a diff view.


src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -183,6 +202,15 @@
       >
         {#if hasBody}
           <div class="body" id={bodyId}>{@render body?.()}</div>
+        {:else if typewriterActive}
```
Tara-ag, line 233: ⚠️ MAJOR: Typewriter reveal inside ChatMessageList aria-live region announces every character. ChatMessageList renders its list with `role="log" aria-live="polite"`; inserting a progressively-updating `TypewriterText` here causes screen readers to announce each incremental text change as character-by-character chatter. Add `aria-live="off"` to this `.body` container so the reveal is not announced while the final text remains readable normally.

```svelte
<div class="body" id={bodyId} aria-live="off">
```

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -109,8 +112,24 @@
...
-  let hasContent = $derived(hasBody || hasHtml || content.length > 0);
+
+  /* The typewriter reveals SOURCE text, not rendered HTML: typing a pre-rendered
+     `html` string character by character would put its tags on screen. So it drives
+     off `markdown` or `content` and leaves an `html`-only message to the normal
+     path. `markdown` is typed through the same pipeline the static branch uses, so
+     rich text reveals as rich text rather than as raw syntax that reformats at the
+     end. */
+  let typewriterSource = $derived(hasMarkdown ? (markdown ?? '') : content);
+  let typewriterActive = $derived(typewriter && !hasBody && typewriterSource.length > 0);
```
Tara-ag: 💡 MINOR: `typewriterSpeed` is forwarded without validation. Negative values are accepted and become invalid `setTimeout` delays in `TypewriterText`. Clamp at the component boundary to match the repo's convention for numeric props:

```svelte
let typewriterPacing = $derived(
  typeof typewriterSpeed === 'number' ? { speed: Math.max(1, typewriterSpeed) } : {}
);
```

Tara-ag (approved):
Reviewed the wiring of `TypewriterText` into `ChatMessage`. The change is additive, backward-compatible, and correctly gates the new behavior behind the `typewriter` prop. Two issues noted:

1. **Accessibility (MAJOR):** The progressive reveal updates text inside a `ChatMessageList` that already has `aria-live="polite"`, which will cause screen readers to announce every character. The typewriter branch should set `aria-live="off"` on its container.

2. **Prop validation (MINOR):** `typewriterSpeed` is forwarded without clamping negative values, which would produce invalid `setTimeout` delays.

Neither is a blocking security/API-breaking issue, so approving with comments.

src/lib/ChatMessage/ChatMessage.svelte
Tara-ag, line 233: ⚠️ MAJOR: The typewriter branch renders inside a ChatMessageList that uses `role="log" aria-live="polite"`. Progressive DOM text updates here will be announced character-by-character by screen readers, creating noisy chatter instead of a coherent message.

Add `aria-live="off"` to this `.body` container so the incremental updates are not announced, while the final text remains normally readable:

```svelte
{:else if typewriterActive}
  <div class="body" id={bodyId} aria-live="off">
    <TypewriterText ... />
  </div>
```

src/lib/ChatMessage/ChatMessage.svelte
Tara-ag: 💡 MINOR: `typewriterSpeed` is typed as `number` but forwarded unvalidated. Negative values are accepted and passed to `TypewriterText`'s `setTimeout` delay, where they become invalid. The repo convention is to clamp numeric props at the component boundary.

Clamp to a sensible minimum before forwarding:

```svelte
let typewriterPacing = $derived(
  typeof typewriterSpeed === 'number' ? { speed: Math.max(1, typewriterSpeed) } : {}
);
```

If a minimum is chosen, update `docs/ChatMessage.md` to document it.

Tara-ag (commented):
Approved with two inline comments. The change is backward-compatible, correctly routes markdown through the existing sanitized pipeline, and covers the new behavior with Playwright tests. The remaining issues are an accessibility improvement (suppress live-region announcements during the progressive reveal) and a prop-validation nit (clamp `typewriterSpeed`); neither blocks the public contract.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -109,8 +112,26 @@
...
   let hasHtml = $derived(typeof effectiveHtml === 'string' && effectiveHtml.length > 0);
+
   let hasBody = $derived(typeof body === 'function');
-  let hasContent = $derived(hasBody || hasHtml || content.length > 0);
+
+  /* The typewriter reveals SOURCE text, not rendered HTML: typing a pre-rendered
+     `html` string character by character would put its tags on screen. So it drives
+     off `markdown` or `content` and leaves an `html`-only message to the normal
+     path. `markdown` is typed through the same pipeline the static branch uses, so
+     rich text reveals as rich text rather than as raw syntax that reformats at the
```
Tara-ag, line 134: ⚠️ MAJOR: `typewriterActive` becomes true as soon as `markdown` is non-empty, but `renderMarkdownFn` is loaded asynchronously in `onMount`. Until it resolves, `TypewriterText` receives no `renderText` and types the raw markdown source (`**streamed**`) as plain text. Once the pipeline loads, the partially revealed text is suddenly passed through the renderer, causing an abrupt visual jump from raw syntax to formatted HTML. This also breaks SSR for a streaming markdown message. Gate the active state on the pipeline being ready when `markdown` is the source:

```ts
let typewriterActive = $derived(
  typewriter &&
    !hasBody &&
    typewriterSource.length > 0 &&
    (!hasMarkdown || renderMarkdownFn !== null)
);
```

tests/chat-message-typewriter.test.ts
```diff
@@ -0,0 +1,76 @@
...
+// `typewriter` renders the message through TypewriterText instead of painting it at once.
+// The risk in wiring two components together is that the reveal silently stops being a
+// reveal — either showing everything immediately, or never completing — so the assertions
+// are about how much text is on screen at a given moment, not about which component ran.
+test.describe('ChatMessage — typewriter', () => {
+  test('reveals a streamed message progressively and completes when streaming stops', async ({
+    page
+  }) => {
+    await page.goto('/components/chat-message');
+
```
Tara-ag, line 12: 💡 MINOR: The test locates elements through demo-page `data-pw` attributes (`typewriter-start`, `typewriter-finish`, `typewriter-demo`). This couples the functional test to the demo layout; reorganizing the demo page will break the suite even if `ChatMessage` behavior is unchanged. Prefer a dedicated test harness or unit tests that render `ChatMessage` directly with controlled props, using stable component-level test IDs only for the component under test.

Tara-ag (changes requested):
The wiring between `ChatMessage` and `TypewriterText` is clean and the documentation is well synchronized, but there is one MAJOR correctness issue that needs to be addressed before merge: a `markdown`-driven typewriter starts revealing raw source before the async markdown pipeline is ready, violating the documented "rich text reveals as rich text" contract and producing an abrupt visual switch once the renderer loads. I also left a MINOR note about test fragility from relying on demo-page `data-pw` attributes.

Tara-ag:

Severity counts — 🔒 CRITICAL: 0 · ⚠️ MAJOR: 2 · 💡 MINOR: 4

Four gate-verified findings remain after review finalization. The change is additive and preserves the public API, but the accessibility silencing strategy for the typewriter inside ChatMessageList's polite live region is not robust across assistive technologies, the web-component prop reflection contract is inconsistent with existing boolean props, the visual baseline cannot be inspected, and the streaming docs should cross-link to ChatMessageList.

- ⚠️ **MAJOR**: aria-live="off" is not a reliable silencer inside a polite live region
  The PR adds `aria-live={streaming ? 'off' : null}` on the typewriter body to prevent a `role="log" aria-live="polite"` ancestor (ChatMessageList) from announcing each character. However, `aria-live="off"` is explicitly defined by WAI-ARIA as the default for elements with no live region role, and screen readers commonly treat it as equivalent to the absence of a live region — i.e. the ancestor's po…
- ⚠️ **MAJOR**: aria-live="off" may not reliably silence announcements inside ChatMessageList's polite live region — `src/lib/ChatMessage/ChatMessage.svelte:225`
  The typewriter body uses `aria-live={streaming ? 'off' : null}` to prevent screen readers from announcing each character as it appears. However, `aria-live="off"` is defined as the default for non-live-region elements, and some screen readers may still honor the ancestor `role="log" aria-live="polite"` declared by `ChatMessageList`. The documented "silence while revealing" contract is therefore no…
- 💡 **MINOR**: New web-component props lack `reflect` and use redundant attribute mapping
  The web-component config adds `typewriter` and `typewriterSpeed` but does not mark them as `reflect: true`, unlike `streaming`, `status`, and `role`. For Boolean props in Svelte custom elements, omitting `reflect` means the corresponding attribute is not updated when the property changes, which breaks consumers that read the attribute (e.g. CSS `[streaming]` selectors, tests, or external framework…
- 💡 **MINOR**: Visual baseline change is asserted only by a binary file diff
  The PR modifies `tests/visual/__screenshots__/chat-message.png` (a binary file). There is no textual description of what changed in the screenshot, and reviewers cannot verify the baseline from the diff. This is acceptable if the project's visual regression workflow regenerates and commits baselines automatically, but the PR description says the baseline was "regenerated ... and compared twice". W…
- 💡 **MINOR**: Web component typewriter prop omits reflect and uses redundant attribute mapping — `src/wc/components/ChatMessage.wc.svelte:12`
  The new `typewriter` boolean prop is declared as `{ type: 'Boolean', attribute: 'typewriter' }` without `reflect: true`. Other boolean props in the same wrapper, such as `streaming`, use `reflect: true`. The omission breaks CSS attribute selectors and external consumers that expect the host attribute to sync with the property. The `attribute: 'typewriter'` mapping is also redundant because it matc…
- 💡 **MINOR**: Visual baseline screenshot change cannot be inspected in the diff — `tests/visual/__screenshots__/chat-message.png`
  The PR modifies `tests/visual/__screenshots__/chat-message.png`, a binary file. Reviewers cannot see the expected visual delta in the GitHub diff, so there is no static way to confirm the new baseline matches the added "Progressive reveal" demo column.
- 💬 **SUGGESTION**: Docs mention ChatMessageList live-region behavior without cross-linking
  The new Streaming section explains that `ChatMessageList` is a live region and that `aria-live="off"` silences the typewriter body. It would help consumers to link to the ChatMessageList docs and note that this silencing only applies when the message is rendered inside a ChatMessageList (or any other live region).
- 💬 **SUGGESTION**: Streaming documentation should cross-link to ChatMessageList — `docs/ChatMessage.md:85`
  The new "Streaming" section in docs/ChatMessage.md explains that the body is `aria-live="off"` while revealing because `ChatMessageList` is a `role="log" aria-live="polite"` region. A reader who lands on the ChatMessage docs has no direct way to discover that this behavior is only meaningful when the component is used inside a live region such as ChatMessageList.

src/lib/ChatMessage/ChatMessage.svelte
murdore: Gated in `af8d13f`, taking your `(!hasMarkdown || renderMarkdownFn !== null)` exactly. But I could not reproduce the defect as described, and the difference matters, so here is what measuring it actually showed.

**First probe was wrong, in the direction that hides the bug.** A `page.evaluate` after `page.goto` runs *after* the dynamic import has resolved, so it reported clean and would have let a real race through. Redone with `addInitScript` so the observer is installed before any page script runs, polling every 10ms.

That probe reports `**` on screen — but the captured DOM says it is not the cause you identified:

```
<div class="body" aria-live="off"><div class="typewriter-text"><p>A **</p></div>
```

The `<p>` is the tell: the markdown renderer was already active in the first frame the probe could see. Comparing the same probe with and without the gate:

```
pre-fix   ...<div class="typewriter-text"><p>A **</p>...
post-fix  ...<div class="typewriter-text"><p>A **</p>...
```

Identical. So with a warm bundle the import resolves before `TypewriterText`'s first 30ms tick, and the window you describe does not open here. The gate is still worth having — it makes the behaviour deterministic instead of dependent on how fast a chunk loads, and it matches the degrade the static branch already documents — but I am recording it as insurance rather than claiming it repaired an observed fault.

**What the probe did find is different and inherent.** Mid-reveal the source is `A **`, which is not closed bold, so the renderer correctly emits a literal `**` until the closing token arrives. No gate fixes that; it is what progressively rendering markdown means. It is still the better of the two options — the alternative is typing plain source for the entire stream and reformatting at the end, which is exactly what the wrapper this replaces did, and that shows the syntax for longer.

**One test was deleted rather than kept green.** I had written `markdown present at first paint is never typed as raw syntax`; it passed, but only because it ran after the window closed, and the better probe shows the claim is simply false. It is replaced by one that asserts what is true: markdown set before mount settles as rendered markdown, `<strong>bold</strong>`, not the source.

lint 0 · check 0 errors · 413 unit · 491 functional · 93 visual (compared twice) · 33 docs-accuracy

tests/chat-message-typewriter.test.ts
murdore: Same answer as on #533, with the measurement rather than a preference — the premise does not hold for this repository.

```
$ grep -rl "goto('/components/" tests/*.ts | wc -l   ->  102
$ ls tests/*.ts | wc -l                            ->  108
$ ls src/routes/                                   ->  components  (only)
```

There is no `/tests/` route tree here; the demo pages under `/components/*` are the harness, and 102 of 108 test files drive one. A dedicated route for this component would be a pattern used by exactly one file.

Note also that this repo has vitest for unit tests but they cover scripts and parity checks, not rendered components — the component tier is Playwright against the demo routes. So "unit tests that render ChatMessage directly with controlled props" is not an existing tier here either.

What I have done is keep the assertions independent of demo *structure*: they are text content over time, computed style, and ARIA state, reached through `testId`s passed to the component itself. Every one is negative-controlled, so each fails when the behaviour is reverted.

Tara-ag (commented):
Review comments posted inline.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -183,6 +214,18 @@
       >
         {#if hasBody}
           <div class="body" id={bodyId}>{@render body?.()}</div>
+        {:else if typewriterActive}
+          <!-- ChatMessageList is a role="log" aria-live="polite" region, so a body whose text
+               grows one character at a time would be announced as it grows. Silence the reveal
+               and let the settled message be read normally. -->
+          <div class="body" id={bodyId} aria-live="off">
```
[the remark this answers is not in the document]
murdore: Correct, and this was a defect I introduced with the `aria-live` fix one round earlier — thank you for catching it.

I keyed the attribute off the branch rather than off the state. `typewriterActive` stays true once `streaming` goes false (the message still renders through `TypewriterText`), so the body would have been muted for the rest of its life and the settled message — the one announcement the live region exists for — would never have been read. Strictly worse than not having touched `aria-live` at all.

Fixed in `27c98be`:

```svelte
<div class="body" id={bodyId} aria-live={streaming ? 'off' : null}>
```

The test now covers both directions rather than only the silencing: `off` while revealing, absent after clicking Finish, and absent on a `typewriter` message that never streamed (which the previous version also muted permanently).

Negative control: pinning it back to a constant `"off"` fails with `Expected: not "off", Received: "off"` on `[data-pw="typewriter-demo"] .body`.

Worth noting the version before this one had a test asserting the attribute IS `off` — and it passed, because it only ever looked during the stream. A test that checks one direction of a two-directional contract is exactly how this shipped past me.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -109,8 +112,36 @@
...
+      typewriterSource.length > 0 &&
+      (!hasMarkdown || renderMarkdownFn !== null)
+  );
+  let typewriterRenderText = $derived(
+    hasMarkdown && renderMarkdownFn !== null ? { renderText: renderMarkdownFn } : {}
+  );
+  // Clamped at the boundary the way LoadingDots clamps `dots`, so the prop's floor is part
+  // of this component's contract rather than something a consumer discovers.
+  let typewriterPacing = $derived(
+    typeof typewriterSpeed === 'number' ? { speed: Math.max(1, typewriterSpeed) } : {}
```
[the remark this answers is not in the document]
murdore: Taken in `27c98be` — `Number.isFinite` guard added.

The reasoning holds exactly as stated: `typeof NaN === 'number'` is `true` and `Math.max(1, NaN)` is `NaN`, so the 1ms floor I had just documented was not actually enforced for a non-finite input. Documenting a floor and then not enforcing it is the worse half of that — the contract was wrong, not just the code.

`properties.ts` and `docs/ChatMessage.md` both now say a non-finite value is ignored, so `TypewriterText`'s own default applies.

No test, deliberately and for the same reason as the clamp on the sibling thread: the observable difference (a `NaN` delay becomes 0 and reveals instantly, versus the 15ms default) would need a demo case on the public component gallery that exists only to be asserted against. I would rather say that plainly than add a demo of a broken input, or claim coverage I do not have.

docs/ChatMessage.md
```diff
@@ -76,6 +78,21 @@ Only relevant when `clampLines` is set; without it the bubble is inert markup an
 - The bubble carries no handler either. An earlier revision let a click anywhere on it toggle, as a pointer convenience; that made it a control keyboard users could not reach, so it is gone. The Button is the only way to toggle, for every input method.
 - `aria-controls` on the button points at the clamped body, so assistive tech can reach the region the control governs. The id is generated per instance, so two clamped messages on one page never collide.
 
+## Streaming
+
+`typewriter` and `streaming` compose into the shape a chat surface actually needs: while a reply
+is arriving, `streaming` stays `true` and the text keeps typing as it grows; when the turn ends,
+`streaming` goes `false` and whatever is left is shown at once rather than typed out after the
```
Tara-ag, line 85: 💬 SUGGESTION: The Streaming section explains the live-region silencing but does not cross-link to `ChatMessageList`. Adding a note that this behavior is only meaningful when `ChatMessage` is inside a live region such as `ChatMessageList` (with a link to `ChatMessageList.md`) would help consumers understand when the silencing applies.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -183,6 +218,19 @@
       >
         {#if hasBody}
           <div class="body" id={bodyId}>{@render body?.()}</div>
+        {:else if typewriterActive}
+          <!-- ChatMessageList is a role="log" aria-live="polite" region, so a body whose text
+               grows one character at a time would be announced as it grows. Silence it only
+               WHILE it is growing: leaving it off once streaming ends would suppress the
+               settled message too, which is the announcement the live region exists for. -->
```
Tara-ag: ⚠️ MAJOR: `aria-live="off"` is not a reliable silencer inside a polite live region. WAI-ARIA defines `aria-live="off"` as the default for non-live-region elements, and screen readers often treat it as equivalent to no override — the ancestor `role="log" aria-live="polite"` in `ChatMessageList` may still announce each character as it appears. The documented contract (“silence while revealing”) is therefore not guaranteed across AT. Consider using `aria-hidden="true"` on the changing text while streaming, or buffering the final text outside the live region until streaming ends. If you keep `aria-live="off"`, verify with NVDA/JAWS/VoiceOver and document the tested matrix.

tests/visual/__screenshots__/chat-message.png
Tara-ag, line 1: 💡 MINOR: The visual baseline is a binary file change that reviewers cannot inspect in the diff. Please describe the expected visual delta in the PR body (e.g. “new Progressive reveal column added on the right”) or attach a perceptual diff if the workflow supports it.

src/wc/components/ChatMessage.wc.svelte
```diff
@@ -9,6 +9,8 @@
       markdown: { type: 'String' },
       body: { type: 'Object' },
       streaming: { type: 'Boolean', reflect: true },
+      typewriter: { type: 'Boolean', attribute: 'typewriter' },
```
Tara-ag: 💡 MINOR: New web-component props should mirror the reflection contract of existing ones. `streaming` uses `{ type: 'Boolean', reflect: true }`, but `typewriter` omits `reflect`. Without reflection, the `typewriter` attribute is not updated when the property changes, breaking CSS attribute selectors and external consumers. Also, `attribute: 'typewriter'` is redundant (it is the default mapping). Suggested:
```js
typewriter: { type: 'Boolean', reflect: true },
typewriterSpeed: { type: 'Number', attribute: 'typewriter-speed', reflect: true },
```

Tara-ag (changes requested):
One verified MAJOR accessibility concern and three smaller items. The typewriter/live-region silencing needs either a fix or AT verification before merge; the rest can be addressed in follow-up or here.

Tara-ag (approved):
Reviewed file-by-file. No blocking issues found:

- Public API: `typewriter` and `typewriterSpeed` are additive optional props; no existing props/exports/CSS variables changed.
- Accessibility: `aria-live="off"` on the revealing body correctly silences the parent `role="log"` live region while streaming; removed once settled.
- Reactivity: `typewriterActive` gates on the async markdown renderer, matching the existing static-branch degradation; `hasContent` now correctly suppresses the typing indicator during reveal.
- SSR safety: no new unguarded browser globals at module/init time; renderer still loads in `onMount`.
- Web-component symmetry: new props mapped with explicit kebab-case attributes.
- Tests cover progressive reveal, markdown pipeline, live-region behavior, at-load markdown, non-streaming full text, and the no-prop control.

Approving.

Tara-ag (commented):
Yama review verdict: CHANGES_REQUESTED. Four gate-verified findings remain: one MAJOR accessibility concern about `aria-live="off"` reliability inside `ChatMessageList`, two MINOR items (web-component prop reflection consistency and an uninspectable visual baseline), and one documentation cross-link suggestion. Inline comments are posted on each file.

src/lib/ChatMessage/ChatMessage.svelte
murdore: Hardened in `cfe1283`, but I am not going to claim the thing you asked for, because I cannot verify it.

`aria-live="off"` is kept — per ARIA, when live regions nest the innermost value applies, so it is the spec-correct mechanism and the standard one. Your point is about implementations diverging from that, which is a fair worry and one I have no way to settle here: I do not have NVDA, JAWS or VoiceOver in this environment, so **there is no tested matrix and I will not write one down**. Documenting an untested matrix would be worse than documenting nothing.

What I have done instead is stop the contract resting on a single behaviour. The body now also carries `aria-busy` while revealing:

```svelte
aria-live={streaming ? 'off' : null}
aria-busy={streaming ? 'true' : null}
```

`aria-busy` is the purpose-built mechanism for exactly this — "being modified, wait before exposing" — and it is already used in this library (`Progress.svelte:50`, `Button.svelte:98`), so it is not a novelty here. If an AT ignores the nested `off`, it still has `aria-busy` telling it to hold.

On the two alternatives suggested:

- **`aria-hidden="true"` while streaming** — I think this is worse, not better. It removes the text from the accessibility tree entirely, so a screen-reader user browsing the page mid-stream finds nothing where the reply is visibly appearing. Chatter is annoying; a hole is disorienting.
- **Buffering outside the live region until streaming ends** — this is the genuinely robust answer and I agree with it in principle, but it means the revealed text is not in the DOM where it is displayed, which is a restructure of `TypewriterText`'s output rather than a change to this integration. Worth its own issue if the AT behaviour turns out to matter in practice.

Test covers both attributes in both directions. Negative control: dropping `aria-busy` fails with `Expected: "true", Received: ""`.

src/wc/components/ChatMessage.wc.svelte
murdore: Taken in `cfe1283` — both now `reflect: true`, and the redundant `attribute: 'typewriter'` is gone.

Verified the premise rather than assuming it, because I declined a similar suggestion on #533:

```
$ grep -n "streaming\|typewriter" src/wc/components/ChatMessage.wc.svelte
11:      streaming: { type: 'Boolean', reflect: true },
```

The difference matters. On #533 the suggestion was to reflect `clampLines`, and I declined because there was no convention to follow — `Gauge` reflects `value` but not `max`. Here the **direct sibling** reflects, and `typewriter` is the same shape: a boolean mode flag on the same component, read alongside `streaming`. That is a local precedent rather than an absent one, so the same reasoning that rejected the first one accepts this.

`prop-parity` still passes at 90.

tests/visual/__screenshots__/chat-message.png
murdore: Fair — a binary diff is unreviewable, and "the baseline changed" is not a description.

The delta is one new demo column, **Progressive reveal**, appended after the clamp column. It contains: an assistant bubble that is empty at rest (the streaming demo starts with no text), a Start and a Finish button, an assistant bubble showing `A bold claim present at first paint.` with `bold` emboldened, and an assistant bubble showing `A message that is not streaming shows in full immediately.` The page grows taller by that column; nothing above it moves.

I will add this to the PR body.

Two things worth stating about how it was regenerated, since neither is visible in the diff either:

- It was compared **twice** after regenerating, not once. An animated demo can be flaky the way `thinking-indicator` is — that one is excluded from the suite precisely because its elapsed counter retimes the layout mid-capture. Two clean runs is the check that this column settles before capture.
- Exactly one screenshot changed (`git status --porcelain tests/visual/__screenshots__/ | wc -l` → 1). A regeneration that quietly rewrote other baselines would look identical in a diff view.

The suite runs inside the pinned Playwright container (`mcr.microsoft.com/playwright:v1.60.0-noble`) both locally and in CI, which is what makes a laptop-generated baseline and a CI comparison the same measurement.

docs/ChatMessage.md
murdore: Added in `cfe1283`, minus the link.

The Streaming section now says where the silencing is meaningful:

> The body is `aria-live="off"` and `aria-busy="true"` **while it is revealing**, and only then. This matters when the message sits inside a live region — `ChatMessageList` is `role="log" aria-live="polite"`, so without it a body growing one character at a time is announced as it grows. Outside a live region it is inert.

I tried it as a markdown link to `./ChatMessageList.md` first and it **broke the build**: the docs are served as routes, so the relative link resolved to `/components/ChatMessageList.md` and the prerenderer failed with `Error: 404 /components/ChatMessageList.md (linked from /components/chat-message)`. Checking afterwards, no doc in this repo links to another component's `.md` — mine would have been the first, and the mechanism does not work as-is. So the reference is by name.

If cross-linking between component docs is wanted, it needs a route-relative form (`/components/chat-message-list`) and should be done across the docs set rather than in one file.

docs/ChatMessage.md
```diff
@@ -76,6 +78,21 @@ Only relevant when `clampLines` is set; without it the bubble is inert markup an
...
+fact. Feed the accumulated text — `TypewriterText` continues from where it left off rather than
+restarting, so appending to `markdown`/`content` is all a consumer has to do.
+
```
Tara-ag, line 88: 💡 MINOR: The Streaming section repeats the same idea twice and the second sentence is broken across lines. Remove the duplication and keep one coherent paragraph.

Tara-ag:
💬 SUGGESTION (src/lib/ChatMessageList/ChatMessageList.svelte): `ChatMessageList` forwards `streaming` but not the new `typewriter`/`typewriterSpeed` props, and `ChatMessageData` lacks those fields. If progressive reveal is meant to work through the list abstraction, wire them through; otherwise note in `docs/ChatMessage.md` that `typewriter` is only available on direct `ChatMessage` usage.

Tara-ag (commented):
Reviewed. The feature is well-scoped and the implementation correctly wires `TypewriterText` into `ChatMessage` while preserving accessibility and SSR safety. I noted one documentation glitch and one suggestion about `ChatMessageList` forwarding; neither blocks the PR.

docs/ChatMessage.md
Tara-ag, line 88: 💡 MINOR: The Streaming section repeats the same sentence and has a broken line. The paragraph already says `ChatMessageList` is a live region, then restates it awkwardly across the line break. Remove the duplication and keep one coherent paragraph.

Tara-ag (commented):
Reviewed file-by-file. The typewriter feature is additive and safe; one non-blocking docs note posted inline. A second accepted finding about ChatMessageList forwarding could not be posted inline because that file is unchanged in this PR.

murdore:
## Proof of testing

Recorded against the built preview of this branch (`pnpm run build && pnpm run preview`), not a dev server and not a mock. The video walks the three behaviours this PR is responsible for, with a caption on each.

https://github.com/user-attachments/assets/64fd4c3c-71d7-4f72-b668-cf5d6383aacf

**1. `clampLines` — the control is the library `Button`.** Measured in the same run, not read off the diff:

```
clamped .body height  45px
expanded .body height 90px
collapsed again       45px   (round-trips, not one-way)
```

**2. `typewriter` on `ChatMessage`.** `streaming=true` reveals progressively; `streaming=false` shows the remainder at once rather than typing it out after the fact. The source is `**streamed**` and it renders as `<strong>` — the markdown pipeline runs, the reveal is not painting raw syntax.

**3. `ChatMessageList` forwards the per-message flag** (added in this round, from the review). Only the responder carries `typewriter`; the sender directly above it is identical data without it. Both go through the same render path, so the sender being complete while the responder is still typing *is* the forwarding.

### Why the third one has a negative control

A dropped prop does not look like a bug in a screenshot — the message still renders, just all at once. So I removed the two forwarding lines and re-ran only that test:

```
Expected: < 82
Received:   82        ← the whole string, painted at once
1 failed
```

Restored, `git diff --numstat` reads `2  0` — exactly the two lines back, nothing else touched.

### Full gate on this commit

```
lint          exit 0
check         exit 0
test:unit     413 passed (28 files)
test:integration  492 passed   (491 before; +1 is the new forwarding test)
```

`test:integration` also records video and trace for every test — those travel in the CI Playwright report artifact if you want the per-test evidence rather than this walkthrough.

docs/ChatMessage.md
murdore: Correct, and it was mine — I appended a sentence in the previous round without reading what was already there, so the paragraph stated the `ChatMessageList` live-region point twice and the second copy was broken across a line.

Rewritten as one paragraph in `56407f3`:

> The body is `aria-live="off"` and `aria-busy="true"` **while it is revealing**, and only then. That matters when the message sits inside a live region: `ChatMessageList` is `role="log" aria-live="polite"`, so without it a body growing one character at a time would be announced on every character. The settled text is still read normally, and outside a live region both attributes are inert.

murdore: Same finding as #discussion_r3943787886 (filed twice, five minutes apart) — fixed in `56407f3`. The duplicated sentence and the broken line are gone; the paragraph now states the live-region point once.

murdore:
> `ChatMessageList` forwards `streaming` but not the new `typewriter`/`typewriterSpeed` props, and `ChatMessageData` lacks those fields. If progressive reveal is meant to work through the list abstraction, wire them through; otherwise note in `docs/ChatMessage.md` that `typewriter` is only available on direct `ChatMessage` usage.

Wired through, in `56407f3`. `ChatMessageData` now carries `typewriter` and `typewriterSpeed`, and `ChatMessageList` forwards both.

I did check first whether wiring was the right half of that choice, because there is a collision the prop names hide: **`Chat` already reveals progressively, one layer down.** Its controller buffers the stream and appends to `content` on a timer (`REVEAL_CHARS_PER_TICK = 2`, `REVEAL_INTERVAL_MS = 22` in `controller.svelte.ts`), driven by its own `typewriter` option. Setting the new per-message flag on a message `Chat` owns would reveal an already-revealing string, and the two would compose into a much slower, uneven crawl.

Per-message opt-in on `ChatMessageData` is what makes those safe together: `Chat` never sets the field, so the two mechanisms cannot stack, and a consumer driving `ChatMessageList` directly — which is an exported component, with no controller behind it — still gets the reveal. Both halves of that are now written down in the docs under **Through `ChatMessageList` and `Chat`**.

There is a test, because this is the failure mode that looks fine in a screenshot — a dropped prop still renders the message, just all at once:

```
tests/chat-message-typewriter.test.ts
  ChatMessageList — forwards the per-message typewriter
    reveals only the message that carries the flag
```

The sender message directly above the responder is the in-page control: identical data, same render path, no flag. It must be complete at the same instant the responder is not.

Negative control — forwarding removed, that test alone re-run:

```
Expected: < 82
Received:   82        ← the whole string, at once
1 failed
```

Restored, `git diff --numstat` reads `2  0`. Suite is 492 passing (491 before).

Tara-ag (commented):
Review in progress — one accessibility/documentation consistency finding so far.

src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -183,6 +218,27 @@
...
+        {:else if typewriterActive}
+          <!-- ChatMessageList is a role="log" aria-live="polite" region, so a body whose text
+               grows one character at a time would be announced as it grows. Silence it only
+               WHILE it is growing: leaving it off once streaming ends would suppress the
+               settled message too, which is the announcement the live region exists for.
+               `aria-live="off"` is the spec answer (the innermost live setting wins), and
+               `aria-busy` is the purpose-built one — "being modified, wait before exposing" —
+               so the two together do not depend on a single behaviour being implemented. -->
+          <div
+            class="body"
```
Tara-ag, line 242: 💡 MINOR: The docs state the typewriter body is silenced "while it is revealing", but this only gates `aria-live="off"` / `aria-busy="true"` on `streaming`. A non-streaming `typewriter` message still types character-by-character after mount, so inside `ChatMessageList`'s `role="log" aria-live="polite"` region it will be announced incrementally. Either narrow the docs to "while `streaming` is true", or derive an `isRevealing` flag from `TypewriterText`'s `onProgress` and silence any active reveal.

tests/chat-message-typewriter.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    const body = page.locator('[data-pw="typewriter-demo"] .body');
+    const full = 'Here is a streamed reply that types as it arrives.';
+
+    await page.locator('[data-pw="typewriter-start"]').click();
+
+    // Mid-reveal: some text, but not all of it. This is the assertion that fails if the
```
Tara-ag, line 22: 💡 MINOR: The mid-reveal assertion snapshots text immediately after clicking Start, but the demo uses `typewriterSpeed={30}` and a 51-character string (full reveal ~1.5s). On a slow CI runner the first read could already be the complete string, failing `toBeLessThan(full.length)`. Add a dedicated test-speed variant (e.g. `typewriterSpeed={200}`) so the partial-reveal window is stable, and consider gating visual regression on `prefers-reduced-motion` or excluding the animated column to avoid the flakiness the PR description already calls out for `thinking-indicator`.

src/wc/components/ChatMessage.wc.svelte
```diff
@@ -9,6 +9,8 @@
       markdown: { type: 'String' },
       body: { type: 'Object' },
       streaming: { type: 'Boolean', reflect: true },
+      typewriter: { type: 'Boolean', reflect: true },
```
Tara-ag, line 12: 💡 MINOR: `typewriterSpeed` is declared with `reflect: true`, which serializes the numeric property to a string attribute and causes attribute churn on every speed change. It is not a boolean that needs DOM presence detection; the existing `clampLines` number prop on line 15 does not reflect. Drop `reflect: true` to match that pattern: `{ type: 'Number', attribute: 'typewriter-speed' }`.

Tara-ag (approved):
Approved with minor inline notes:

- The docs claim the typewriter body is silenced "while it is revealing", but the implementation only silences during `streaming`; a non-streaming typewriter still types character-by-character and will be announced inside `ChatMessageList`'s live region. Narrow the docs or extend the silencing.
- The Playwright mid-reveal assertions depend on wall-clock timing with `typewriterSpeed={30}`; give the test path a slower speed for stability.
- The web component reflects `typewriterSpeed` as a number attribute, which is unnecessary churn and inconsistent with `clampLines`; drop `reflect: true`.

No blocking issues: the change is additive, SSR-safe, and preserves the public API contract.

src/wc/components/ChatMessage.wc.svelte
murdore: Taken, in `a99b440` — `typewriterSpeed: { type: 'Number', attribute: 'typewriter-speed' }`.

You are right and my reasoning for adding it was wrong in a specific way worth naming: I justified `reflect` by pointing at the sibling `streaming`, which does reflect — but `streaming` is a **boolean**, where reflection buys presence detection. `typewriterSpeed` is a number, and the number prop to compare against is `clampLines`, which does not reflect. I applied a boolean's convention to a numeric prop.

`typewriter` keeps `reflect: true`, since that one really is the boolean sibling of `streaming`.

tests/chat-message-typewriter.test.ts
murdore: Taken, in `a99b440`. The arithmetic is right: 51 characters at 30ms is a ~1.5s partial window, and a stalled runner stepping over it fails the assertion for a reason unrelated to the behaviour under test.

Added `typewriter-slow` — same component, `typewriterSpeed={200}`, and the mid-reveal test now drives that instead. The human-paced demo stays at 30ms, because a demo page someone reads should not crawl. Comment on the test says which and why, so the next person does not "helpfully" repoint it at the fast one.

On the visual-regression half: the animated column is already excluded from the risk you describe, but not by intent I can point at, so I would rather not claim it. What I can say is measured — the `chat-message` baseline has been regenerated and re-verified twice in this round (93 passed both times, and `git diff --numstat` shows exactly one baseline file moving each time). If it does prove flaky in CI I will gate it rather than retry it.

src/lib/ChatMessage/ChatMessage.svelte
murdore: I took the docs half of this and measured the premise, which does not hold.

**"A non-streaming `typewriter` message still types character-by-character after mount"** — it does not. `TypewriterText` has an effect that fires whenever `isStreaming` is false: it clears the pending timeout, assigns `displayedText = text` and jumps `currentIndex` to the end (`TypewriterText.svelte:145-159`). The `onMount` call to `typeNextCharacter()` is unconditional, so I agree it *looks* like a race — which is why I measured it rather than reading it.

Probe: `addInitScript` installs a `requestAnimationFrame` sampler **before any app script runs**, so the window between mount and first paint is covered. A `page.evaluate` after `goto` would run after mount and report exactly the clean result that would let a real reveal through.

```
typewriter-static      (typewriter, streaming falsy)
  samples=418  distinct lengths=2  max=58
  INCREMENTAL REVEAL: NO  ->  0, 58        (one step, not 58)

typewriter-at-load     (typewriter, streaming falsy)
  samples=413  distinct lengths=1  max=37
  INCREMENTAL REVEAL: NO
```

418 rAF samples over 4s is ~16ms resolution. A character-by-character reveal would have produced tens of distinct lengths; both produced one transition. So there is no unannounced window for a non-streaming message, and `aria-live`/`aria-busy` gating on `streaming` is the correct condition rather than an approximation of one.

Deriving `isRevealing` from `onProgress` would therefore add state that is true exactly when `streaming` already is.

Your docs point stands on its own though, and I have taken it — "while it is revealing" invited precisely your reading. It now says **while `streaming` is true**, and states why that is not a narrowing:

> `streaming` is the right condition rather than an approximation of one, because a `typewriter` message with `streaming` false does not reveal incrementally at all: `TypewriterText` clears its pending timer and assigns the whole string, so the body goes from empty to complete in a single step. There is no unannounced window to cover.

Tara-ag (commented):
Reviewed file by file. One documentation suggestion found; no blocking issues.

Tara-ag (commented):
Reviewed file by file. The change is additive and backward-compatible: optional `typewriter`/`typewriterSpeed` props on `ChatMessage`, mirrored in `ChatMessageData` and forwarded by `ChatMessageList`, with matching web-component declarations and Playwright coverage. No blocking issues. One documentation suggestion noted.

Tara-ag:
💬 SUGGESTION: `docs/ChatMessageList.md` does not mention that each `ChatMessageData` item may include `typewriter` and `typewriterSpeed`, which are forwarded to the underlying `ChatMessage`. Consider adding a note in the Usage or Web Component section so consumers discover the per-message typing effect without reading `ChatMessage.md`.

Tara-ag (commented):
Reviewed file by file. The change is additive and backward-compatible: optional `typewriter`/`typewriterSpeed` props on `ChatMessage`, mirrored in `ChatMessageData` and forwarded by `ChatMessageList`, with matching web-component declarations and Playwright coverage. No blocking issues. One documentation suggestion noted.

PR #531, opened 2026-09-05
feat(events)!: remove the 3.x legacy event-prop spellings and host-reserved custom-element props

murdore:
> **Do not merge yet.** Phase 3 of `docs/EVENT_CASING_MIGRATION.md` may only land after a 3.x carrying #519's deprecation warnings and working codemod has been **released** (semantic-release → 3.5.0). This PR is stacked on #519 (base = `fix/close-audit-gaps`; GitHub retargets it to `release` when #519 merges). The `feat(events)!:` subject + `BREAKING CHANGE:` footer make semantic-release cut **4.0.0**.

One commit, 206 files, +1990 / −4624.

## Part A — the 142 legacy event-prop spellings are gone (57 components)

- Each `properties.ts` loses its `@deprecated` legacy prop; each component loses the `resolveDeprecatedProp` `$derived` alias and the `readDeprecatedProps` `$effect.pre`, and reads the corrected name directly. Done by two committed generators (`scripts/migrate/remove-event-aliases.ts`, `remove-wc-event-aliases.ts`, both with tests) so the change is reproducible, plus the hand fixes neither generator can see (Combobox → Input forwarding, FileDropzoneTrigger → Button, Chat's indexed type lookup, LottiePlayer.wc's internal `oncomplete`/`onerror` reads — the last one caught only by Playwright, invisible to svelte-check because that wrapper's `$props()` is untyped).
- `scripts/event-casing-baseline.json` deleted; `check-event-casing.js` now fails on **any** violation (0 today, no grandfathering).
- `src/lib/deprecation.ts`, `deprecation-version.ts` and the 3.x-phase generators (`alias-event-props`, `mark-deprecated`, `alias-wc-props`, `casing-report`, `codemod/generate-maps`) deleted with their tests. `casing.ts` / `signatures.ts` / `analyze.ts` stay (general-purpose).
- `scripts/codemod/legacy-pairs.ts` stays, frozen — it is the rename table consumers still run (`npx sui-codemod ./src` **before** upgrading).

## Part B — 24 host-reserved custom-element props renamed

`KNOWN_HOST_RESERVED_DECLARATIONS` is now empty and the ratchet fails on any new one. Renames follow the existing `sui-input` → `inputId` / `inputAriaLabel` precedent (`<component><ReservedName>`), each with an explicit kebab-case attribute:

| Custom element | Reserved | New prop | Attribute |
|---|---|---|---|
[...]

PR #547, opened 2026-09-06
feat(events)!: remove the 3.x legacy spellings and host-reserved element props

murdore:
Phase 2 of `docs/EVENT_CASING_MIGRATION.md`, and the half 3.x could not do while the old names still had to work: **one event, one prop name, lowercase throughout.** 190 `@deprecated` alias declarations are gone, along with every resolver that picked between two spellings and the dev-only warning that announced them.

Two other things a major is the only chance to fix ride along, both previously recorded as debt in the code that now pays it.

Consumers run `npx sui-codemod ./src`. **[docs/MIGRATION_4.0.md](docs/MIGRATION_4.0.md)** is the guide, with all 190 renames listed.

## What was removed

| | Count |
|---|---|
| `@deprecated` alias declarations | **190 → 0** |
| Wrapper alias declarations | 172 removed across 59 wrappers |
| Host-reserved element props | 27 → **0** (`KNOWN_HOST_RESERVED_DECLARATIONS` is now empty) |
| `src/lib/deprecation.ts` | deleted — nothing imported it |

Generated, not hand-edited. `remove-event-aliases.ts` rewrote 133 files; `remove-wc-alias-props.ts` (new) took the dead custom-element declarations, since an element that still declared a removed spelling would expose a setter reaching nothing — worse than not declaring it; `rename-host-reserved-props.ts` (new) took the ratchet list itself as its work list, so the two cannot drift.

## Three things needed a hand, and each was reported rather than guessed

**Three components would have lost a default.** `HITL`, `IframeViewer` and `Input` each had a resolver carrying `?? null` or `?? (() => {})`. Removing the resolver deletes the fallback with it, leaving an undefined handler called unguarded. `svelte-check` confirmed all three as real errors before the fix.

**`Chat` was corrected, not just stripped.** It shipped `onscrollstate` marked `@deprecated` in favour of `onScrollState` — backwards, and the only component pointing away from the rule. It got past the phase 0 gate because it borrows its type by indexed access (`ChatMessageListProperties['onscrollstate']`) and the gate treated only inline function types as events. The gate now recognises that shape, with a test for it and for the narrowing that keeps `onErrorMessage?: InputProperties['errorMessage']` out of scope. **The codemod was never affected** — the backwards pair never entered its table, so nobody was told to migrate towards camelCase.

**`Table.oncellchange` is deleted, which forced a codemod change.** It was accepted so a call site could type-check and never called; `docs/Table.md` said so and pointed at the `cell`-snippet pattern. Its entry also had to leave `legacy-pairs.ts`, because a rename table still pointing at it would move a consumer onto a prop that no longer exists. Hence 190, not 191.

## The host-reserved rename is property-only

24 declarations used names the platform already defines on every element — `title`, `id`, `role`, `hidden`, and the ARIAMixin `aria*` accessors — so the declaration replaced the platform's.

Each declaration now **pins the attribute it already observed**, so `<sui-card title="Sales">` and every other HTML usage is untouched. Only the JavaScript property moves:

```js
element.title = 'Sales';     // 3.x: the component prop. 4.0.0: the tooltip.
element.cardTitle = 'Sales'; // 4.0.0: the component prop.
```

A consumer who only ever sets attributes is unaffected.

**These were not 27 proven bugs, and the commit says so.** Only `children` was measured broken (`element.children` returned undefined instead of an HTMLCollection on three wrappers; that declaration is simply gone, since light DOM already carries slotted content). `sui-badge`'s `hidden` was measured *working* — Svelte's setter reflects to the attribute. The `aria*` entries have the clearest mechanism. They moved as a set anyway, because "the setter happens to reflect" is a property of Svelte's codegen rather than something this library promises, and a major is the only time the whole set can move at once.

## Coverage moved with the removal rather than thinning

- `src/legacy-spellings-removed.test.ts` replaces 3.x's alias matrix, asserting the inverse over the same pairs — each legacy spelling gone **and** its replacement still declared, so a removal that took the wrong one fails. Deleting the matrix without this would have quietly dropped that coverage.
- `legacy-pairs.test.ts` no longer asserts the table matches the tags, because there are none. It asserts the derivation now yields nothing, which is what proves the removal complete, and that the table stays frozen — regenerating it would silently empty the codemod.
- `wc-event-casing-parity.spec.ts` is rewritten, not deleted. Its dual-spelling cases are obsolete; its native-handler cases are not, because a lowercase prop colliding with a real `GlobalEventHandlers` accessor is a hazard of the rule itself.

## Gates

| Gate | Result |
|---|---|
| Lint | exit 0 — **0 violations, 0 aliases** across 94 `properties.ts` |
| Types | 864 files, **0 errors** |
| Unit | **828 passed** (36 files) |
| Build | exit 0 |
| Playwright | **486 passed** |
| Visual | **93 passed** (pinned container) |

Verified against the release workflow's own version-detection greps that this cuts a **major**: `feat(events)!:` matches the `!` rule and the `BREAKING CHANGE:` footer matches.

Supersedes #531, which was closed with no commits.


Tara-ag:

PR #547 is a deliberate, well-documented major breaking change (`feat(events)!:` with `BREAKING CHANGE:` footer) that removes the 190 deprecated camelCase event-prop aliases and standardizes on lowercase event props for v4.0.0. Verified: `src/lib/deprecation.ts` is removed and no resolver/deprecated references remain in `src/lib`; spot-checked components (Button, Input, IframeViewer, Table, ChatComposer) declare only lowercase event props; `IframeViewer` and `Input` now carry explicit no-op defaults (`onmessage = () => {}`, `onstatechange = () => {}`) replacing the removed resolver fallbacks; `src/wc/index.ts` registers the previously-missing `AttachmentChipRow`, `HITL`, and `TypewriterText` web components; `docs/MIGRATION_4.0.md` documents the 190 renames and the codemod path. All reported CI gates pass. No blocking criteria are triggered (no secrets, no critical security issues, and the breaking change is accompanied by major-version signals and a migration guide). Time constraints prevented fully reading the truncated `src/lib/HITL/HITL.svelte` to visually confirm that `onconfirm`/`onmictoggle` received the same explicit default treatment; the PR description and passing tests indicate they did, but a final human pass on that one file is recommended.

PR #549, opened 2026-09-06
fix(wc): restore the web-component type-check that gates every release

murdore:
## `release` is red, and nothing since 3.5.0 has published

`Release and Publish` has failed on **both** commits since 3.5.0:

| Commit | What it is | Release run |
|---|---|---|
| `9814b53` | #519 — every event prop lowercase, old spellings as aliases | ❌ failure |
| `190800b` | #547 — the 4.0.0 removal | ❌ failure |

npm still serves **3.5.0**, which predates both. So the lowercase rule is unpublished, and — noted independently by the session working on Lighthouse — the published 3.5.0 also declares `bin` as `./scripts/codemod/cli.ts`, a TypeScript file with no shebang. Verified: 3.5.0 has `"sui-codemod": "./scripts/codemod/cli.ts"`, release HEAD has `./dist-codemod/bin.js`. `npx sui-codemod` from the published version does not work; the fix is already on `release` and just needs to ship.

## The 15 errors, in two groups

**Six** — `src/lib/docs-index.test.ts` imports `node:fs` / `node:path`, and `tsconfig.wc.json` type-checks `src/lib/**/*.ts` with `types: ["vite/client"]`, so node globals are absent. The test is correct; this config was never meant to see it. Test files are now excluded from it.

**Nine** — from #547's host-reserved rename. Turning `let props = $props()` into `let { renamed, ...props } = $props()` narrows the rest element's type, so `<Chat {...props}>` stops satisfying `MandatoryChatProperties`. Seventeen wrappers go back to binding the whole object and reading the renamed prop off it (`title={props.chatTitle}`); Combobox keeps its destructure with an annotation that admits `comboboxAriaLabel`.

## Why both reached a release — the part worth keeping

`check:wc` is the **only** thing that type-checks `src/wc/**`, and it was not in `pnpm check`. Local `check` covered 864 files, none of them the wrappers being edited, so six clean local gate runs said nothing. The gap only appeared in CI, inside the release job, after merge.

`pnpm check` now runs it. The same mistake fails on a laptop, and in `ci.yml`, before it can reach `release`.

Also: `hITLTitle` → `hitlTitle`, from a rename script that lowercased only the first character of `HITL`.

## ⚠️ This publishes 4.0.0, and the footer is load-bearing

`release.yml` derives the version from **`git log -1` alone** — the latest commit, not the range since the last tag. With a plain `fix:` subject this commit would publish `190800b`'s breaking removal as **3.5.1**.

The `BREAKING CHANGE:` footer is therefore not decoration; it is what makes the bump correct, and it is truthful — this release does carry the removal. Replayed against the workflow's own logic on the real commit:

```
commit: fix(wc): restore the web-component type-check that gates every release
HAS_BREAKING_CHANGE=true
=> version_type=major
```

**Separately worth a decision:** deriving the bump from one commit is fragile in exactly this way — any `fix` or `chore` landing on top of an unreleased `feat!` silently downgrades the release. That is a change to release logic while the pipeline is red, so I have not touched it here.

## Gates

| Gate | Result |
|---|---|
| `pnpm check` | **0 errors** in both passes — 864 files and 432 files |
| `check:wc` alone | 15 errors → **0** |
| Lint | exit 0 |
| Unit | **828 passed** |
| Build | exit 0 |
| Playwright | **486 passed** |
| Visual | **93 passed** |

One Playwright run showed `typewriter-text-pacing-progress-render` at 7045ms against a 2500ms wall-clock ceiling. It passed **20/20** on repeat, and this branch does not touch `TypewriterText` — the diff is `package.json`, `tsconfig.wc.json` and 19 wrappers. Reported rather than omitted.

## Summary by CodeRabbit

- **Documentation**
  - Expanded the migration guide with clarification on lowercase prop naming versus camelCase callback keys, including configuration examples and Svelte usage.

- **Bug Fixes**
  - Corrected the web component property name from `hITLTitle` to `hitlTitle`.

- **Refactor**
  - Improved web component prop forwarding and type-checking consistency without changing component behavior.

- **Chores**
  - Extended validation to cover web-component type checking and excluded test files from that configuration.

### DIFF
--- a/src/wc/components/HITL.wc.svelte
+++ b/src/wc/components/HITL.wc.svelte
@@ -4,7 +4,7 @@
     shadow: 'open',
     props: {
       confirmationId: { type: 'String', attribute: 'confirmation-id' },
-      hITLTitle: { type: 'String', reflect: true, attribute: 'title' },
+      hitlTitle: { type: 'String', reflect: true, attribute: 'title' },
       description: { type: 'String' },
       sections: { type: 'Object' },
       functionArguments: { type: 'Object' },
@@ -37,7 +37,7 @@
 
 <script lang="ts">
   import HITL from '$lib/HITL/HITL.svelte';
-  let { hITLTitle, ...props } = $props();
+  let props = $props();
 </script>
 
-<HITL {...props} title={hITLTitle} />
+<HITL {...props} title={props.hitlTitle} />
--- a/docs/MIGRATION_4.0.md
+++ b/docs/MIGRATION_4.0.md
@@ -23,9 +23,29 @@ npx sui-codemod --dry-run ./src
 
 The codemod is component-aware. It rewrites `onClick` on `<Toggle>` because
 that is a prop this library renamed, and leaves `onClick` on your own
-components alone. It also leaves callback keys inside config objects untouched
-— `columns={[{ id: 'a', onToggle: fn }]}` on `<Table>` keeps its spelling,
-because that is a key you write inside a value, not a prop on a tag.
+components alone.
+
+### "Lowercase throughout" means props, not every callback
+
+The rule governs **props you write on a tag**. Callback keys inside a config
+object you build are a different thing — they are members of a value, not
+attributes — and they were never deprecated, so 4.0.0 does not touch them and
+neither does the codemod. They stay camelCase:
+
+| Where                          | Keys that stay camelCase                                                              |
+| ------------------------------ | ------------------------------------------------------------------------------------- |
+| `TableColumn`                  | `onToggle`, `onSelect`, `onInput`, `onButtonClick`, `onPrimaryAction`, `onMenuAction` |
+| `TableCheckboxSelectionConfig` | `onSelectionChange`                                                                   |
+| `TablePaginationConfig`        | `onPageChange`, `onPageSizeChange`, `onLoadMore`                                      |
+| `ComboboxAction`               | `onClick`                                                                             |
+
+```svelte
+<!-- onrowclick is a prop and was renamed; onToggle is a key in a value and was not. -->
+<Table onrowclick={fn} columns={[{ id: 'a', onToggle: fn }]} />
+```
+
+So `<Table>` legitimately shows both spellings at once. If you expected
+`onselectionchange` and did not find it, this is why.
 
 ### What it cannot do for you
 
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "package": "svelte-kit sync && svelte-package && publint",
     "prepublishOnly": "npm run build:wc && npm run build:codemod && npm run package",
     "test": "npm run test:integration && npm run test:unit",
-    "check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json && npm run check:codemod && npm run check:migrate && npm run check:wc-parity",
+    "check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json && npm run check:wc && npm run check:codemod && npm run check:migrate && npm run check:wc-parity",
     "check:codemod": "tsc -p scripts/codemod/tsconfig.json",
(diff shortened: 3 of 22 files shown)


Tara-ag (commented):

src/wc/components/HITL.wc.svelte
```diff
@@ -4,7 +4,7 @@
     shadow: 'open',
     props: {
       confirmationId: { type: 'String', attribute: 'confirmation-id' },
-      hITLTitle: { type: 'String', reflect: true, attribute: 'title' },
+      hitlTitle: { type: 'String', reflect: true, attribute: 'title' },
```
Tara-ag, line 7: 🔒 CRITICAL: The custom element property `hITLTitle` was renamed to `hitlTitle`. Existing web-component consumers assigning `.hITLTitle` will break silently. Because the web-component API is a public contract for a published library, this rename requires a major-version bump or a documented migration path. Keep `hITLTitle` as an alias or restore the original name, and update docs/MIGRATION_4.0.md if the break is intentional.

Tara-ag (changes requested):
One blocking issue found: a web-component property was renamed without preserving backward compatibility or documenting a migration path. Please address before merging.

Tara-ag:

Severity counts — 🔒 CRITICAL: 2 · ⚠️ MAJOR: 1 · 💡 MINOR: 0

## 🤖 Yama Review Summary

**Verified findings (3):**

- **CRITICAL**: HITL web-component prop renamed from hITLTitle to hitlTitle
- **MAJOR**: Combobox.wc.svelte loses two-way binding for value/inputValue/open/highlightedIndex/selected
- **CRITICAL**: HITL web-component property renamed without migration note — `src/wc/components/HITL.wc.svelte:7`

- 🔒 **CRITICAL**: HITL web-component prop renamed from hITLTitle to hitlTitle
  The custom element prop `hITLTitle` was renamed to `hitlTitle` in src/wc/components/HITL.wc.svelte. This is a breaking change for any web-component consumer already using `<sui-hitl h-i-t-l-title="...">` or setting `.hITLTitle`. The PR body frames this as a spelling fix, but it changes the public custom-element API surface without a major-version bump or migration note. Because the library is alre…
- 🔒 **CRITICAL**: HITL web-component property renamed without migration note — `src/wc/components/HITL.wc.svelte:7`
  The custom element property `hITLTitle` was renamed to `hitlTitle` in src/wc/components/HITL.wc.svelte. Existing web-component consumers assigning the property via JavaScript (`.hITLTitle = ...`) or relying on the previous property name will break silently. Because this is a published library and the web-component API is a public contract, renaming a property requires a major-version bump or a doc…
- ⚠️ **MAJOR**: Combobox.wc.svelte loses two-way binding for value/inputValue/open/highlightedIndex/selected
  The PR changes Combobox.wc.svelte from destructuring with $bindable defaults and explicit `bind:` directives to `let props = $props()` plus a plain spread `<Combobox {...props}>`. Because the inner Combobox reassigns these bindable props, a one-way spread means the host custom element's properties will not reflect updates made by the component (e.g. selecting an item, opening/closing the dropdown,…

murdore:
Superseded by `d6066e9`, which fixed the same 15 `check:wc` errors a better way.

This PR reverted the nine wrappers to whole-object `$props()` binding. That works, but only because untyped `$props()` is `any` and spreading `any` satisfies anything — it restores the check by removing what the check was looking at. `d6066e9` instead keeps the destructures and annotates them properly, which surfaced two real defects this PR would have re-hidden: Menu's wrapper called `props.trigger()` with no argument, making `MenuTriggerProps` unreachable through the custom element, and HITL's renamed `title` was passed where a required `string` is expected.

Both changes agreed on the other two points — excluding test files from `tsconfig.wc.json`, and moving `check:wc` into `pnpm check` so this class of failure can't reach `release` again.

Closing rather than rebasing; the branch is conflicting and has nothing left to contribute.

PR #551, opened 2026-09-06
fix(wc): restore the type coverage the host-reserved rename removed, and gate it

murdore:
## Why

`Release and Publish` has failed on **every commit since `9814b53`** — four runs, two of which carried merged features — at `Type-check web components`, with 15 errors.

```
03:41  5c8f209  SUCCESS   ← last publish (3.5.0)
07:13  9814b53  failure
11:43  190800b  failure
13:45  e96ec81  failure    #546 typewriter
16:59  713ec88  failure    #548 marker
```

```
npm latest = 3.5.0
```

`typewriter` and `marker` are both merged and **neither is installable**. The workflow status never said so — only the registry does.

### 1. The host-reserved rename removed an `any`

`190800b` renamed the props the platform already defines on every element (`title` → `pillTitle`, `role` → `bannerRole`). That changed nine wrappers from:

```svelte
let props = $props();              <Pill {...props}>
```
to
```svelte
let { pillTitle, ...props } = $props();     <Pill {...props} title={pillTitle}>
```

Untyped `$props()` is `any`, and spreading `any` satisfies anything. **Destructuring takes that away**, so svelte-check started checking the attribute object properly and found the mandatory prop (`text`, `src`, `value`, `items`, `messages`, `confirmationId`) not provably present.

The rename didn't break these wrappers. It removed an `any` that had been hiding what they never declared. `Combobox` is the same thing from the other side — it *does* annotate `$props()`, and the annotation just wasn't updated for its own renamed prop.

Each wrapper now states the contract that was implicit:

```ts
let { pillTitle, ...props }: Omit<PillProperties, 'title'> & {
  pillTitle?: PillProperties['title'];
} = $props();
```

The renamed property keeps **the component's own optionality** — which is why HITL's is required, not optional: `title` is mandatory on HITL, and that's the same claim `Omit<HITLProperties, 'title'>` already makes about `confirmationId`.

### Two defects it had been hiding

Restoring the check surfaced these. Both are fixed here rather than annotated around:

- **`Menu` dropped its trigger's props.** The wrapper rendered `props.trigger()` with no argument, and its own `{#snippet trigger()}` declared no parameter to forward. But `trigger` is `Snippet<[MenuTriggerProps]>`, and Menu hands it the interaction wiring precisely so a consumer whose trigger is itself interactive can spread it and set `interactiveTrigger`. Through the custom element that argument was **always `undefined`**, making the feature unreachable.
- **`HITL`** was passing its renamed `title` where a required `string` is expected.

### 2. A vitest test in the web-component scope

Unrelated, and four hours earlier. `docs-index.test.ts` arrived in `9814b53`. `tsconfig.wc.json` includes `src/lib/**/*.ts` while overriding `types` to `["vite/client"]` — so a vitest test that legitimately imports `node:fs` was being type-checked without node types. Six errors, in a pass that has no business reading it.

Test files are now out of that scope: **516 files checked → 432**.

## The gate

This is the whole reason it reached `release` four times:

| script | what it is | where it ran |
|---|---|---|
| `check:wc-parity` | `tsc` over the parity script | `pnpm check` → CI ✅ |
| `check:wc` | `svelte-check` over `tsconfig.wc.json` | release workflow only ❌ |

Similar names, different checks. A PR could be entirely green and still fail the release. `check:wc` now runs inside `check`, which CI and pre-commit already invoke.

## Controls

**Negative** — remove one wrapper's annotation and the gate fails, showing the gap directly:

```
COMPLETED 866 FILES 0 ERRORS   ← tsconfig.json — what PR CI already ran
COMPLETED 432 FILES 1 ERRORS   ← tsconfig.wc.json — the one that was missing
Property 'text' is missing ... but required in type 'MandatoryPillProperties'
```

**Runtime** — the five specs that mount these custom elements, run before any edit and after:

```
BEFORE: 40 passed
AFTER:  40 passed
```

**Visual** — 93 passed, **zero baselines moved**, which is what a types-only change should do.

## Gate

```
lint               exit 0
check              exit 0   866 files 0 errors  +  432 files 0 errors
test:unit          828 passed
test:integration   497 passed
test:visual         93 passed (0 baselines changed)
```

## Summary by CodeRabbit

* **Chores**
  * Expanded validation checks to include web component type checking.
  * Improved type safety and editor guidance across web component properties without changing runtime behavior.
  * Updated menu trigger handling to forward trigger properties correctly.
  * Excluded test files from web component TypeScript configuration.

### DIFF
--- a/src/wc/components/Banner.wc.svelte
+++ b/src/wc/components/Banner.wc.svelte
@@ -21,7 +21,18 @@
 
 <script lang="ts">
   import Banner from '$lib/Banner/Banner.svelte';
-  let { bannerRole, ...props } = $props();
+  import type { BannerProperties } from '$lib/Banner/properties';
+  // The element renames `role` to `bannerRole` because the platform already defines `role` on every
+  // HTMLElement. The rest still carries the component's own props -- saying so is what
+  // a destructured `$props()` no longer infers on its own.
+  let {
+    bannerRole,
+    ...props
+    // `title` is omitted as well as `role`: it is host-reserved, so the element never
+    // declares it and the wrapper's own title snippet is unconditional. Including it
+    // would advertise a prop nothing can set.
+  }: Omit<BannerProperties, 'role' | 'title'> & { bannerRole?: BannerProperties['role'] } =
+    $props();
 </script>
 
 <!--
--- a/src/wc/components/Menu.wc.svelte
+++ b/src/wc/components/Menu.wc.svelte
@@ -24,13 +24,20 @@
 
 <script lang="ts">
   import Menu from '$lib/Menu/Menu.svelte';
-  let { menuAriaLabel, ...props } = $props();
+  import type { MenuProperties } from '$lib/Menu/properties';
+  // The element renames `ariaLabel` to `menuAriaLabel` because the platform already defines `ariaLabel` on every
+  // HTMLElement. The rest still carries the component's own props -- saying so is what
+  // a destructured `$props()` no longer infers on its own.
+  let {
+    menuAriaLabel,
+    ...props
+  }: Omit<MenuProperties, 'ariaLabel'> & { menuAriaLabel?: MenuProperties['ariaLabel'] } = $props();
 </script>
 
 <!-- A property-assigned trigger wins; the slot is the fallback. The branch stays
      inside the body snippet so `<slot>` keeps its `$$props` scope. -->
 <Menu {...props} ariaLabel={menuAriaLabel}>
-  {#snippet trigger()}
-    {#if props.trigger}{@render props.trigger()}{:else}<slot name="trigger"></slot>{/if}
+  {#snippet trigger(triggerProps)}
+    {#if props.trigger}{@render props.trigger(triggerProps)}{:else}<slot name="trigger"></slot>{/if}
   {/snippet}
 </Menu>
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "package": "svelte-kit sync && svelte-package && publint",
     "prepublishOnly": "npm run build:wc && npm run build:codemod && npm run package",
     "test": "npm run test:integration && npm run test:unit",
-    "check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json && npm run check:codemod && npm run check:migrate && npm run check:wc-parity",
+    "check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json && npm run check:wc && npm run check:codemod && npm run check:migrate && npm run check:wc-parity",
     "check:codemod": "tsc -p scripts/codemod/tsconfig.json",
     "codemod": "node scripts/codemod/cli.ts",
     "check:wc": "svelte-check --tsconfig ./tsconfig.wc.json --compiler-warnings \"options_missing_custom_element:ignore\"",
--- a/src/wc/components/Chat.wc.svelte
+++ b/src/wc/components/Chat.wc.svelte
@@ -56,7 +56,14 @@
 
 <script lang="ts">
   import Chat from '$lib/Chat/Chat.svelte';
-  let { chatTitle, ...props } = $props();
+  import type { ChatProperties } from '$lib/Chat/properties';
+  // The element renames `title` to `chatTitle` because the platform already defines `title` on every
+  // HTMLElement. The rest still carries the component's own props -- saying so is what
+  // a destructured `$props()` no longer infers on its own.
+  let {
+    chatTitle,
+    ...props
+  }: Omit<ChatProperties, 'title'> & { chatTitle?: ChatProperties['title'] } = $props();
 </script>
 
 <Chat {...props} title={chatTitle} />
--- a/src/wc/components/Checkbox.wc.svelte
+++ b/src/wc/components/Checkbox.wc.svelte
@@ -20,7 +20,16 @@
 
 <script lang="ts">
   import Checkbox from '$lib/Checkbox/Checkbox.svelte';
-  let { checkboxAriaLabel, ...props } = $props();
+  import type { CheckboxProperties } from '$lib/Checkbox/properties';
+  // The element renames `ariaLabel` to `checkboxAriaLabel` because the platform already defines `ariaLabel` on every
+  // HTMLElement. The rest still carries the component's own props -- saying so is what
+  // a destructured `$props()` no longer infers on its own.
+  let {
+    checkboxAriaLabel,
+    ...props
+  }: Omit<CheckboxProperties, 'ariaLabel'> & {
+    checkboxAriaLabel?: CheckboxProperties['ariaLabel'];
+  } = $props();
 </script>
 
 <Checkbox {...props} ariaLabel={checkboxAriaLabel}>
--- a/src/wc/components/Combobox.wc.svelte
+++ b/src/wc/components/Combobox.wc.svelte
@@ -65,7 +65,9 @@
     highlightedIndex = $bindable(-1),
     selected = $bindable([]),
     ...rest
-  }: ComboboxProperties = $props();
+  }: Omit<ComboboxProperties, 'ariaLabel'> & {
+    comboboxAriaLabel?: ComboboxProperties['ariaLabel'];
+  } = $props();
 </script>
 
 <Combobox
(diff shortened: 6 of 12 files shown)


Tara-ag (commented):
Review in progress

src/wc/components/Menu.wc.svelte
```diff
@@ -24,13 +24,20 @@
...
+  // The element renames `ariaLabel` to `menuAriaLabel` because the platform already defines `ariaLabel` on every
+  // HTMLElement. The rest still carries the component's own props -- saying so is what
+  // a destructured `$props()` no longer infers on its own.
+  let {
+    menuAriaLabel,
+    ...props
+  }: Omit<MenuProperties, 'ariaLabel'> & { menuAriaLabel?: MenuProperties['ariaLabel'] } = $props();
 </script>
 
 <!-- A property-assigned trigger wins; the slot is the fallback. The branch stays
```
Tara-ag, line 37: 💡 MINOR: The fix in `src/wc/components/Menu.wc.svelte` now forwards `MenuTriggerProps` to a property-assigned trigger snippet, restoring the intended interaction wiring. This behavioral fix is not covered by an integration test; a regression could silently revert to a trigger that receives no `onclick`/`onkeydown`/`aria-expanded` wiring.

Suggestion: Add a Playwright test in `tests/wc-custom-elements.spec.ts` that mounts a `<sui-menu>` with `element.trigger = (triggerProps) => ...`, opens the menu, and asserts `triggerProps.onclick`/`onkeydown`/`ariaExpanded` are provided and functional.

Tara-ag (commented):
No blocking issues. One minor suggestion to add a regression test for the Menu trigger forwarding fix.

src/wc/components/Menu.wc.svelte
Tara-ag, line 37: 💡 MINOR: Menu trigger forwarding lacks regression test. The fix now forwards `MenuTriggerProps` to a property-assigned trigger snippet, restoring the intended interaction wiring. This behavioral fix is not covered by an integration test; a regression could silently revert to a trigger that receives no `onclick`/`onkeydown`/`aria-expanded` wiring.

Suggestion: Add a Playwright test in `tests/wc-custom-elements.spec.ts` that mounts a `<sui-menu>` with `element.trigger = (triggerProps) => ...`, opens the menu, and asserts `triggerProps.onclick`/`onkeydown`/`ariaExpanded` are provided and functional.

Tara-ag (approved):
Approved with one minor test-coverage note. No blocking issues found.

murdore:
## Regression test added for the Menu defect

`8d361c4` adds `tests/wc-menu-trigger-props.spec.ts`. The type check alone would catch a plain revert of the wrapper, but not a change that keeps the signature and stops forwarding — and this bug was invisible in every other way. The menu still opened; the trigger just never received its wiring.

The test property-assigns a `trigger` to `sui-menu` and captures what the snippet is handed:

```
keys            ["ariaExpanded", "ariaHaspopup", "onclick", "onkeydown"]
ariaHaspopup    "menu"
ariaExpanded    boolean
onclick         function
onkeydown       function
```

**Negative control** — the wrapper reverted to the exact pre-fix shape (`{#snippet trigger()}` + `props.trigger()`), that spec alone re-run:

```
Expected: true
Received: false     ← the trigger received nothing
1 failed
```

`git diff --numstat` on the revert reads `2  2`, and restoring returns the file byte-identical to the commit.

Full gate on this commit:

```
lint               exit 0
check              exit 0   867 files 0 errors  +  432 files 0 errors
test:integration   498 passed   (497 before; +1 is this spec)
```

One note on writing it: the first version tripped the repo's `no-restricted-syntax` ban on the `undefined` identifier, which is why the presence check is `typeof props === 'object' && props !== null` rather than the obvious spelling.

Tara-ag (approved):
No blocking issues found. The type annotations restore the implicit prop contracts that the host-reserved renames had hidden, the Menu wrapper now correctly forwards trigger wiring, and the gate fix ensures `check:wc` runs in CI.

src/wc/components/Banner.wc.svelte
```diff
@@ -21,7 +21,14 @@
...
   import Banner from '$lib/Banner/Banner.svelte';
-  let { bannerRole, ...props } = $props();
+  import type { BannerProperties } from '$lib/Banner/properties';
+  // The element renames `role` to `bannerRole` because the platform already defines `role` on every
+  // HTMLElement. The rest still carries the component's own props -- saying so is what
+  // a destructured `$props()` no longer infers on its own.
+  let {
+    bannerRole,
+    ...props
+  }: Omit<BannerProperties, 'role'> & { bannerRole?: BannerProperties['role'] } = $props();
```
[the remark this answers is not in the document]
murdore: Correct, taken in `6fffb5b` — `Omit<BannerProperties, 'role' | 'title'>`.

Verified rather than assumed, and the wrapper already documents it: of Banner's four snippets, three read a property branch (`{#if props.icon}`, `props.rightContent`, `props.dismissIcon`) and only `title` does not — its own comment says so, *"`title` keeps no property branch: it is host-reserved, never declared."* The element declares `text`, `dismissible`, `testId`, `bannerRole`, `ondismiss`, `rightContent`; `title` is not among them.

I checked the other eight wrappers for the same shape and Banner is the only one where the type needed narrowing.

That sweep did turn up something else, which I have **not** folded into this PR because it is a behaviour change rather than a type one: `Pill` declares `leadingIcon` and `dismissIcon` as element properties, and `Checkbox` declares `checkedIcon` and `indeterminateIcon`, but all four snippets render only their slot with no `props.` branch — so assigning those properties reaches nothing, exactly like the Menu trigger defect this PR fixes. Raising it separately.

src/wc/components/Menu.wc.svelte
murdore: Already added — `tests/wc-menu-trigger-props.spec.ts`, in `8d361c4`, which landed before this review. (Filed twice; same answer on both.)

It does what you describe: property-assigns a `trigger` to `sui-menu` and captures what the snippet is handed.

```
keys            ["ariaExpanded", "ariaHaspopup", "onclick", "onkeydown"]
ariaHaspopup    "menu"
ariaExpanded    boolean
onclick         function
onkeydown       function
```

I put it in its own file rather than `wc-custom-elements.spec.ts` as suggested, because that spec opens by declaring its scope as the four form elements (`sui-chip-input`, `sui-color-picker`, `sui-combobox`, `sui-split-input`) and loads a bundle gated on exactly those tags — adding a menu case would have meant rewriting its premise.

**Negative control**, since a test only ever observed green is indistinguishable from one asserting nothing. Wrapper reverted to the exact pre-fix shape (`{#snippet trigger()}` + `props.trigger()`), that spec alone re-run:

```
Expected: true
Received: false      ← the trigger received nothing
1 failed
```

`git diff --numstat` on the revert reads `2  2`; restoring returns the file byte-identical to the commit.

Worth adding that your reasoning for wanting it is right and the type check alone would not have been enough: `check:wc` catches a revert that changes the signature, but not one that keeps the parameter and stops forwarding it. The bug was invisible in every other way — the menu still opened, the trigger just never received its wiring.

murdore: Already added — `tests/wc-menu-trigger-props.spec.ts`, in `8d361c4`, which landed before this review. (Filed twice; same answer on both.)

It does what you describe: property-assigns a `trigger` to `sui-menu` and captures what the snippet is handed.

```
keys            ["ariaExpanded", "ariaHaspopup", "onclick", "onkeydown"]
ariaHaspopup    "menu"
ariaExpanded    boolean
onclick         function
onkeydown       function
```

I put it in its own file rather than `wc-custom-elements.spec.ts` as suggested, because that spec opens by declaring its scope as the four form elements (`sui-chip-input`, `sui-color-picker`, `sui-combobox`, `sui-split-input`) and loads a bundle gated on exactly those tags — adding a menu case would have meant rewriting its premise.

**Negative control**, since a test only ever observed green is indistinguishable from one asserting nothing. Wrapper reverted to the exact pre-fix shape (`{#snippet trigger()}` + `props.trigger()`), that spec alone re-run:

```
Expected: true
Received: false      ← the trigger received nothing
1 failed
```

`git diff --numstat` on the revert reads `2  2`; restoring returns the file byte-identical to the commit.

Worth adding that your reasoning for wanting it is right and the type check alone would not have been enough: `check:wc` catches a revert that changes the signature, but not one that keeps the parameter and stops forwarding it. The bug was invisible in every other way — the menu still opened, the trigger just never received its wiring.

PR #552, opened 2026-09-06
feat(Table): render row selection with the library Checkbox

murdore:
## What this changes

`Table` rendered its row-selection checkboxes as a hand-rolled `role="checkbox"` span with its own checkmark markup, duplicating the library's own `Checkbox` a few directories away. This replaces that markup with `Checkbox` itself, in a new controlled mode.

## Why controlled mode was needed

`Checkbox` owns its `checked` state and flips it on click. A table's selection is not the checkbox's to own: the parent holds the selected set, and it may decline a change (a row that cannot be selected, a page whose selection is reconciled against a consumer-owned set). Left uncontrolled, the box would flip locally and then disagree with the parent's next render, so the rendered box, the native input and `aria-checked` would drift apart.

`controlled` makes a click report the value it is asking for through `onclick` and change nothing locally, so the parent's `checked` and `indeterminate` stay the single source of truth. Uncontrolled behaviour is untouched and is still the default.

`attributes` spreads extra DOM attributes onto the checkbox element itself — the `role="checkbox"` box, not the wrapping label — which is what lets `Table` keep injecting the per-row `data-pw` values its consumers' end-to-end selectors already depend on. It is spread last, so a consumer value wins over the component's own.

## What this deliberately preserves

Consumers select table rows by `data-pw`, and those attributes are produced by `getRowAttributes`. Rewriting the markup without carrying them would have silently re-pointed every one of those selectors in the consuming repositories. That is why `attributes` exists rather than the migration simply dropping the hand-rolled span.

## Verification

Run on this branch after rebasing onto `release` (`d6066e9`):

| Gate | Result |
|---|---|
| `pnpm run lint` | 0, including the event-casing guard (0 violations, 0 deprecated aliases) |
| `pnpm run check` | 0 errors across 868 files |
| `pnpm run test:unit` | 828 passed, 36 files |
| `pnpm run test:integration` | 501 passed |
| `pnpm run test:visual` | 93 passed |

The `Table` visual baseline is unchanged, which is the point: the migration is not meant to alter how a table looks. The `Checkbox` demo baseline is intentionally refreshed, because the demo page gains controlled-mode examples.

## One thing the rebase caught

The branch predated the event-casing standardisation on `release`, so it passed `onClick` where the component now takes `onclick`. Nothing conflicted — the rename happened in files the branch did not touch — and the combination only failed at type-check. Fixed at all five call sites, and the custom element no longer advertises the old alias.

## Summary by CodeRabbit

- **New Features**
  - Added controlled checkbox behavior, allowing applications to accept or reject user-initiated changes.
  - Added support for custom attributes on checkbox elements, including custom identifiers and test attributes.
  - Updated the table selection controls to use the standard Checkbox component, including tri-state header selection.

- **Documentation**
  - Added Checkbox controlled-mode and custom-attribute examples.
  - Documented updated table checkbox styling and customization options.

- **Tests**
  - Added coverage for controlled interactions, keyboard behavior, and custom attribute precedence.


Tara-ag:

Severity counts — 🔒 CRITICAL: 0 · ⚠️ MAJOR: 4 · 💡 MINOR: 2

The PR consolidates Table row-selection onto the library Checkbox in a new controlled mode, which is a good architectural cleanup. However, three issues need resolution: (1) existing themes that customized --table-checkbox-border width/style lose those overrides in checked/indeterminate states because the bridge replaces the whole border; (2) getRowAttributes can override the internally-owned row checkbox id and sever the header aria-controls relationship; (3) the checkbox cell stops all keydown propagation instead of only the activation keys.

- ⚠️ **MAJOR**: attributes spread can override essential ARIA semantics — `src/lib/Checkbox/Checkbox.svelte:89`
  The `{...attributes}` spread is placed after `role`, `tabindex`, `aria-checked`, `aria-disabled`, `aria-controls`, and `aria-label` on the checkbox box. Because Svelte spreads apply in declaration order, a consumer (or Table's `getRowAttributes` escape hatch) can accidentally override the checkbox role or state. This breaks the accessibility tree and can render the control unrecognizable to assist…
- ⚠️ **MAJOR**: Checkbox cell stops propagation of every keydown, breaking table keyboard navigation — `src/lib/Table/Table.svelte:915`
  The `<td class="table-checkbox-col">` now calls `stopPropagation()` on all `keydown` events. The previous implementation only stopped propagation when Space or Enter activated the checkbox. Blocking every keydown prevents arrow-key navigation, Tab handling, and other keyboard events from bubbling to the row/table, regressing accessibility for keyboard and screen-reader users.
- ⚠️ **MAJOR**: Checked/indeterminate checkbox border no longer inherits width/style from --table-checkbox-border — `src/lib/Table/Table.svelte:1429`
  The migration to the library Checkbox bridges --table-checkbox-checked-border-color onto Checkbox's --checkbox-checked-border, which replaces the entire border in the checked state. Previously Table only changed border-color, so a consumer who customized --table-checkbox-border width/style saw that width/style preserved when checked. Now the checked state defaults to 2px solid unless the new --tab…
- ⚠️ **MAJOR**: getRowAttributes can override the row checkbox id and break header aria-controls — `src/lib/Table/Table.svelte:610`
  checkboxAttributes spreads consumer-provided getRowAttributes last, so a consumer returning an `id` attribute overrides the internally-generated `row-checkbox-${rowId}` id. The header Checkbox's aria-controls references exactly those ids (`row-checkbox-${rowId}` for every selectableRowIds), so overriding them severs the accessible relationship between the select-all control and the row checkboxes.
- 💡 **MINOR**: Default table checkbox border-radius changed from 3px to 4px — `docs/Table.md:713`
  The fallback for `--table-checkbox-border-radius` changed from `3px` to `var(--radius, 4px)`. While documented in docs/Table.md, this changes the default visual appearance for consumers who did not set the token, which is a subtle visual breaking change in a minor release.
- 💡 **MINOR**: Checkbox cell stops keydown propagation for all keys — `src/lib/Table/Table.svelte:915`
  The table-checkbox-col td uses an unqualified onkeydown={stopPropagation}. This prevents any table-level or document-level keyboard handlers from seeing keydown events while focus is inside a checkbox, and is broader than the previous implementation which only intercepted Space/Enter activation keys.

Tara-ag (commented):
Reviewing the Table/Checkbox controlled-mode changes.

src/lib/Table/Table.svelte
```diff
@@ -917,40 +910,22 @@
...
-                        }}
-                      >
-                        {#if rowSelected}
-                          <!-- eslint-disable svelte/no-at-html-tags -->
-                          <span class="table-checkbox-icon">{@html checkmarkSvg}</span>
-                        {/if}
-                      </span>
+                    <!-- The cell swallows the click and the key so a clickable row does not
+                         also fire its own handler when the checkbox is toggled. -->
+                    <td
```
Tara-ag, line 915: 💡 MINOR: Checkbox cell stops `keydown` propagation for all keys. The `table-checkbox-col` `<td>` uses an unqualified `onkeydown={(keyboardEvent) => keyboardEvent.stopPropagation()}`. This prevents any table-level or document-level keyboard handlers from seeing `keydown` events while focus is inside a checkbox, and is broader than the previous implementation which only intercepted Space/Enter activation keys.

Suggested fix: Only stop propagation for the activation keys that would also trigger the row's `onrowclick` handler, e.g. `onkeydown={(e) => { if (e.key === ' ' || e.key === 'Enter') e.stopPropagation(); }}`.

src/lib/Table/Table.svelte
```diff
@@ -596,11 +595,22 @@
...
+   * consumer's `getRowAttributes` is spread last so its own test attribute wins.
+   */
+  const checkboxAttributes = (rowId: string, rowIndex: number): Record<string, string> => {
+    const suffix = rowIndex === -1 ? 'select-all' : `row-checkbox-${rowId}`;
+    const own: Record<string, string> = rowIndex === -1 ? {} : { id: `row-checkbox-${rowId}` };
+    if (typeof testId === 'string') {
+      own['data-pw'] = `${testId}-${suffix}`;
+      own.testID = `${testId}-${suffix}`;
     }
+    const consumer = checkboxSelection?.getRowAttributes
```
Tara-ag, line 610: ⚠️ MAJOR: `getRowAttributes` can override the row checkbox `id` and break header `aria-controls`. `checkboxAttributes` spreads consumer-provided `getRowAttributes` last, so a consumer returning an `id` attribute overrides the internally-generated `row-checkbox-${rowId}` id. The header Checkbox's `aria-controls` references exactly those ids (`row-checkbox-${rowId}` for every `selectableRowIds`), so overriding them severs the accessible relationship between the select-all control and the row checkboxes.

Suggested fix: Restore the internally-owned `id` after spreading consumer attributes so `aria-controls` remains valid: `return { ...own, ...consumer, id: own.id };` (or strip `id` from consumer before merging).

src/lib/Table/Table.svelte
```diff
@@ -1444,64 +1419,35 @@
...
-    justify-content: center;
-    width: var(--table-checkbox-icon-size, 12px);
-    height: var(--table-checkbox-icon-size, 12px);
-    color: var(--table-checkbox-icon-color, #ffffff);
-  }
-
-  .table-checkbox-icon :global(svg) {
-    width: 100%;
-    height: 100%;
+    --checkbox-container-gap: 0;
```
Tara-ag, line 1429: ⚠️ MAJOR: Checked/indeterminate checkbox border no longer inherits width/style from `--table-checkbox-border`. The migration to the library Checkbox bridges `--table-checkbox-checked-border-color` onto Checkbox's `--checkbox-checked-border`, which replaces the entire border in the checked state. Previously Table only changed `border-color`, so a consumer who customized `--table-checkbox-border` width/style saw that preserved when checked. Now the checked state defaults to `2px solid` unless `--table-checkbox-checked-border` is explicitly set, silently changing the visual contract for existing themes.

Suggested fix: Either change `Checkbox.svelte` so checked/indeterminate states only override `border-color` by default (preserving width/style from `--checkbox-border`), or bridge decomposition variables such as `--table-checkbox-border-width`/`--table-checkbox-border-style` so existing `--table-checkbox-border` customizations continue to apply in checked/indeterminate states.

Tara-ag (changes requested):
Yama review complete. The PR introduces a clean consolidation of Table row-selection onto the library Checkbox, but three issues need addressing before merge:

1. **MAJOR — CSS backward compatibility**: checked/indeterminate states now replace the entire border via `--checkbox-checked-border`, ignoring existing `--table-checkbox-border` width/style customizations.
2. **MAJOR — Accessibility**: `getRowAttributes` can override the row checkbox `id`, breaking the header `aria-controls` relationship.
3. **MINOR — Accessibility**: the checkbox cell stops all `keydown` propagation instead of only Space/Enter.

See inline comments for file/line evidence and suggested fixes.

PR #555, opened 2026-09-06
fix(release)!: classify the version from the whole range, not the triggering commit

murdore:
## The defect

`Determine version bump type` read `git log -1`, so a release was versioned by whichever commit happened to trigger it. Any bump belonging to a commit whose own run failed — or that landed in the same push — was silently dropped.

## What it already cost

Three consecutive release runs failed at `Type-check web components`. The `fix(wc):` merge that finally succeeded was the only commit the classifier saw, so **3.5.1, a patch, shipped:**

- `feat(ChatMessage): typewriter`
- `feat(ChatMessage): marker`
- **`feat(events)!: remove the 3.x legacy spellings and host-reserved element props`**

That last one removes 190 event-prop spellings and renames 24 custom-element properties. Every consumer on `^3.x` takes it as a routine patch upgrade.

**Nothing catches it downstream.** Svelte does not error on an unknown prop — the handler just never fires. Measured in one consumer, against the published `.d.ts`:

```
Modal 3.5.1  →  onoverlayclick?: () => void
consumer     →  onoverlayClick={...}          82 files
```

| prop | files |
|---|---|
| `onoverlayClick` | 82 |
| `onToastHide` | 68 |
| `onheaderRightImageClick` | 56 |
| `onprimaryButtonClick` / `onsecondaryButtonClick` | 29 each |
| `onRowClick` | 21 |
| …6 more | 19 |
| **total** | **304** |

All of it type-checks and builds clean. Modal overlays stop closing; toasts stop dismissing; table rows stop responding.

## The fix

Both classifiers (the root job and the MCP job) now read `<last tag>..HEAD` and skip this workflow's own `chore(release):` bookkeeping commits.

## Control

Replaying the exact shell logic against the real `3.5.0..d6066e9` range:

```
OLD (git log -1)     ->  patch      the version that shipped
NEW (whole range)    ->  major      correct; the range contains feat(events)!
```

And for this PR's own range, which is what decides the version it cuts:

```
range 3.5.1..HEAD    ->  major      => 4.0.0
```

## What merging this does

Cuts **4.0.0**, republishing under a major the removals 3.5.1 shipped as a patch. **The code is unchanged from 3.5.1** — only the version now states what it contains.

- consumers on 3.4.x → migrate with `npx sui-codemod ./src`, see `docs/MIGRATION_4.0.md`
- consumers who took 3.5.1 → already on this code; re-pin
- 3.5.1 → to be deprecated on npm with a pointer here

YAML validated with `yaml.safe_load`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -230,12 +230,26 @@ jobs:
       - name: Determine version bump type
         id: version-type
         run: |
-          # Get the latest commit message
-          COMMIT_MSG=$(git log -1 --pretty=format:"%s")
-          echo "Latest commit: $COMMIT_MSG"
+          # Classify the WHOLE range since the last released tag, not just the commit that
+          # happened to trigger this run. Reading `git log -1` loses every bump a failed or
+          # skipped run should have carried: 3.5.1 shipped two features and a BREAKING
+          # change as a patch, because the three runs before it failed and the `fix(wc):`
+          # merge that finally succeeded was the only commit it looked at.
+          LAST_TAG=$(git describe --tags --abbrev=0 --match '[0-9]*' 2>/dev/null || true)
+          if [ -n "$LAST_TAG" ]; then
+            RANGE="$LAST_TAG..HEAD"
+            echo "Classifying range: $RANGE"
+          else
+            RANGE=""
+            echo "No release tag found; classifying full history"
+          fi
 
-          # Check for BREAKING CHANGE in commit body
-          COMMIT_BODY=$(git log -1 --pretty=format:"%b")
+          # Release commits are this workflow's own bookkeeping, never a reason to bump.
+          COMMIT_MSG=$(git log $RANGE --grep="^chore(release):" --invert-grep --pretty=format:"%s")
+          COMMIT_BODY=$(git log $RANGE --grep="^chore(release):" --invert-grep --pretty=format:"%b")
+          echo "Commits considered:"
+          echo "$COMMIT_MSG" | sed 's/^/  /'
+
           HAS_BREAKING_CHANGE=false
           if echo "$COMMIT_BODY" | grep -iE "^BREAKING CHANGE:" > /dev/null; then
             HAS_BREAKING_CHANGE=true
@@ -566,8 +580,13 @@ jobs:
         run: |
           # Use the latest non-release commit so the root job's auto "chore(release):"
           # commit doesn't drive the MCP version bump.
-          COMMIT_MSG=$(git log --grep="^chore(release):" --invert-grep -1 --pretty=format:"%s")
-          COMMIT_BODY=$(git log --grep="^chore(release):" --invert-grep -1 --pretty=format:"%b")
+          MCP_LAST_TAG=$(git describe --tags --abbrev=0 --match '[0-9]*' 2>/dev/null || true)
+          MCP_RANGE=""
+          if [ -n "$MCP_LAST_TAG" ]; then
+            MCP_RANGE="$MCP_LAST_TAG..HEAD"
+          fi
+          COMMIT_MSG=$(git log $MCP_RANGE --grep="^chore(release):" --invert-grep --pretty=format:"%s")
+          COMMIT_BODY=$(git log $MCP_RANGE --grep="^chore(release):" --invert-grep --pretty=format:"%b")
           echo "MCP release driven by commit: $COMMIT_MSG"
 
           HAS_BREAKING_CHANGE=false


.github/workflows/release.yml
```diff
@@ -566,8 +580,13 @@ jobs:
         run: |
           # Use the latest non-release commit so the root job's auto "chore(release):"
           # commit doesn't drive the MCP version bump.
-          COMMIT_MSG=$(git log --grep="^chore(release):" --invert-grep -1 --pretty=format:"%s")
-          COMMIT_BODY=$(git log --grep="^chore(release):" --invert-grep -1 --pretty=format:"%b")
+          MCP_LAST_TAG=$(git describe --tags --abbrev=0 --match '[0-9]*' 2>/dev/null || true)
```
Tara-ag, line 583: ⚠️ MAJOR: The MCP classifier uses `--match '[0-9]*'`, which matches root package tags (`3.5.1`) rather than MCP tags (`mcp-X.Y.Z`). After `git pull origin release --rebase`, HEAD is at the root release commit, so `MCP_RANGE` becomes `3.5.1..HEAD` (or the new root version), which is empty. The classifier then defaults to patch regardless of what actually changed in `mcp/`. Use `--match 'mcp-*'` so the range is anchored to the last MCP release.

.github/workflows/release.yml
```diff
@@ -230,12 +230,26 @@ jobs:
...
+          # skipped run should have carried: 3.5.1 shipped two features and a BREAKING
+          # change as a patch, because the three runs before it failed and the `fix(wc):`
+          # merge that finally succeeded was the only commit it looked at.
+          LAST_TAG=$(git describe --tags --abbrev=0 --match '[0-9]*' 2>/dev/null || true)
+          if [ -n "$LAST_TAG" ]; then
+            RANGE="$LAST_TAG..HEAD"
+            echo "Classifying range: $RANGE"
+          else
+            RANGE=""
+            echo "No release tag found; classifying full history"
```
Tara-ag, line 244: ⚠️ MAJOR: `git log --pretty=format:"%s"` concatenates all commit subjects with no separator. A range like `fix: a` followed by `feat: b` becomes the single string `fix: afeat: b`, so the fix branch matches first and the feature bump is silently lost. The same applies to `COMMIT_BODY` and the MCP job. Use `--pretty=tformat:"%s%n"` (and `%b%n`) so each commit is on its own line, or iterate commits with a NUL delimiter.

Tara-ag, line 244: 💡 MINOR: `git log $RANGE` expands the range variable unquoted. Quote it as `git log "$RANGE"` (and `git log "$MCP_RANGE"` in the MCP job) to avoid word splitting if the tag ever contains spaces or metacharacters. An empty quoted argument is harmless here.

Tara-ag (changes requested):
The fix to classify from the whole range is correct and important, but the implementation has two MAJOR issues that can still cause silent misclassification:

1. The MCP job matches root package tags (`[0-9]*`) instead of MCP-prefixed tags (`mcp-*`), so its range is wrong.
2. `git log --pretty=format:"%s"` concatenates commit subjects without separators, making the regex order-dependent and able to drop bumps.

Also one MINOR shell-quoting nit.

Please address the two MAJOR items before merge.

Tara-ag (changes requested):
The PR correctly fixes the root cause that caused 3.5.1 to ship as a patch despite containing a breaking change. However, the implementation introduces two major regressions in the MCP classifier and in multi-commit classification that can still silently choose the wrong bump type:

1. ⚠️ MAJOR (`.github/workflows/release.yml:583`): The MCP version classifier uses root package tags instead of MCP tags (`--match '[0-9]*'` vs `--match 'mcp-*'`), so the computed range is wrong.
2. ⚠️ MAJOR (`.github/workflows/release.yml:244`): `git log --pretty=format:"%s"` concatenates commit subjects without separators, causing the regex branch taken to depend on commit order.
3. 💡 MINOR (`.github/workflows/release.yml:244`): Unquoted `$RANGE` / `$MCP_RANGE` expansions in `git log`.

Please address the two MAJOR items before merging.

Tara-ag:

The PR correctly fixes the root cause that caused 3.5.1 to ship as a patch despite containing a breaking change. However, the implementation introduces two major regressions in the MCP classifier and in multi-commit classification that can still silently choose the wrong bump type. The MCP job matches root-package tags instead of mcp-* tags, and `git log --pretty=format:"%s"` concatenates commit subjects without separators so the regex branch taken depends on commit order. A minor quoting nit is also noted.

- ⚠️ **MAJOR**: MCP version classifier uses root package tags instead of MCP tags — `.github/workflows/release.yml:583`
  In the `publish-mcp` job, `git describe --tags --abbrev=0 --match '[0-9]*'` matches root package version tags (e.g. `3.5.1`), not MCP-prefixed tags (`mcp-X.Y.Z`). After `git pull origin release --rebase`, HEAD is at the root release commit, so the computed range `LAST_TAG..HEAD` is empty and the classifier silently defaults to patch regardless of actual MCP commits. When root release is skipped, t…
- ⚠️ **MAJOR**: Commit subjects concatenated without separators can misclassify the bump — `.github/workflows/release.yml:244`
  `git log --pretty=format:"%s"` concatenates all commit subjects with no delimiter. The regex `(^|: )feat(...)` then depends on commit order: a range like "fix: a" followed by "feat: b" becomes the single string "fix: afeat: b". The fix branch matches first and the classifier returns patch, silently dropping the feature bump. The same risk exists for `COMMIT_BODY` and the MCP job.
- 💡 **MINOR**: Unquoted variable expansion in git log range can split or glob — `.github/workflows/release.yml:244`
  `git log $RANGE` expands `$RANGE` without quotes. If `LAST_TAG` is empty, `RANGE` is empty and the command becomes `git log --grep=...`, which is fine. But if the tag contains spaces or shell metacharacters, or if the range is constructed with user-controlled input, the expansion can split into multiple arguments. The same applies to `MCP_RANGE` in the MCP job.

murdore:
Superseded by #558, which covers this plus two cases this approach misses. Closing rather than iterating here because #558 is already green.

This PR **does** fix the case it targets — replayed on the real 3.5.1 range it returns `major`. Two things it does not catch, verified by replaying its exact shell rather than reading it:

| Range | this PR | correct |
|---|---|---|
| the real 3.5.1 range | `major` | `major` ✅ |
| breaking `fix!` + ordinary `feat` | `minor` | `major` ❌ |
| breaking `docs!` + ordinary `feat` | `minor` | `major` ❌ |

Concatenating every subject and running the existing `if/elif` chain over the blob means the first matching **type** branch wins. Once any `feat` is in the range the chain enters the feat branch and only tests for `feat!`, so a `!` on any other type is invisible — the same failure mode as reading one commit, with a narrower trigger. #558 classifies each commit and takes the maximum, so the bang is honoured wherever it appears.

Second: the MCP boundary uses `--match '[0-9]*'`, which on this history resolves to `4.1.0` — the root package's tag — rather than `mcp-4.0.2`. MCP's bump would be driven by root-package commits. #558 uses distinct landmarks per package and path-filters MCP to `mcp/`.

Worth recording separately: this commit's subject is `fix(release)!:`, and under the rule that is still live the `!` matches `fix(\(.+\))?!:` → **major**. Merging it as-is would have published **5.0.0** from 4.1.0, for a workflow fix.

PR #556, opened 2026-09-06
docs(release)!: record that the 3.x removal shipped in 3.5.1, and cut it as 4.0.0

murdore:
## 3.5.1 published a breaking change as a patch

`release.yml` derives the version from **`git log -1` alone**. It read `d6066e9` (`fix(wc): …`, no breaking footer) and never saw `190800b` (`feat(events)!: remove the 3.x legacy spellings…`) two commits below. `pnpm version patch` from 3.5.0 gave 3.5.1, and that is what went to npm on **2026-09-06T18:39:40Z**.

Verified against the published tarballs rather than the branch:

| | 3.5.0 | 3.5.1 |
|---|---|---|
| `dist/Toast/properties.d.ts:27` | `onToastHide?: () => void;` | `ontoasthide?: () => void;` |
| camelCase names across all `dist/*/properties.d.ts` | **134** | **14** |

The removal is live, and a `^3.5.0` or `~3.5.0` range resolves to it.

**The failure mode is silence.** Svelte drops an unknown prop without erroring, so a caret upgrade that looks like a patch stops handlers firing with no build failure and no runtime error. Only a TypeScript consumer gets a signal, because the prop is gone from the component's `Properties` type.

## The docs said the opposite

That is what this PR fixes. No library code changes.

- **`README.md`** announced the `children` removal as "Coming in 4.0.0", under the words *"**Nothing below has happened yet** — the current 3.x line still accepts everything it always did."* That sentence tells a reader on 3.5.1 they are safe when they are not.
- **`DESIGN_PRINCIPLES.md`** §3 still had the aliases "remain accepted", present tense.
- **`docs/EVENT_CASING_MIGRATION.md`** was titled "completing in 4.0.0" and told consumers that through 3.x, "nothing immediately".
- **`docs/MIGRATION_4.0.md`** attributed the whole change to 4.0.0, with `element.title` commented `3.x:` / `4.0.0:`.

All four now name the release the change actually landed in. `MIGRATION_4.0.md` leads with how to tell whether you already have it:

```sh
npm ls @juspay/svelte-ui-components
```

…and states that **3.5.0 is the last release accepting the old spellings**. It recommends migrating over pinning to 3.5.0, and says what pinning costs: `TypewriterText`, `ChatMessage`'s `marker`, and the fix that made `MenuTriggerProps` reachable through `<sui-menu>`.

## ⚠️ Merging this publishes 4.0.0

4.0.0 ships the **same components 3.5.1 does**, under the version number the change should have carried. 3.5.1 stays published and undeprecated — a deliberate call, not an oversight.

The `!` in the subject and the `BREAKING CHANGE:` footer are load-bearing for exactly the reason this PR exists. Replayed against the workflow's own block, verbatim, on the real commit:

```
Latest commit: docs(release)!: record that the 3.x removal shipped in 3.5.1, and cut it as 4.0.0
HAS_BREAKING_CHANGE=true
version_type=major (breaking change)
current: 3.5.1  ->  major bump -> 4.0.0
```

They are independently sufficient: the `!` matches the `docs(…)!:` branch, and the footer forces major on every branch. Without either, `git log -1` reads a `docs:` subject and cuts 4.0.1.

**This trap is still armed.** Anything landing on `release` before the publish needs the same treatment, or it takes the release slot and buries the mis-versioned break one release deeper. Fixing the derivation to read the commit range since the last tag is deferred by decision until after 4.0.0 ships.

## Gates

| Gate | Result |
|---|---|
| Lint | 0 — prettier, eslint, **0 event-casing violations** across 94 `properties.ts` files |
| `pnpm check` | **exit 0 over both configs** — 867 files / 0 errors, 432 files / 0 errors |
| Unit | **828 passed** |

The 4 remaining `svelte-check` warnings (Modal, ThemeSwitcher, MediaPlayer caption, `moduleResolution` deprecation) predate this branch and are untouched by a docs change.

Worth recording: the README edit was caught by `rename-doc-usages`'s repository-wide guard, which flagged a legacy spelling I had named in prose. The prose was reworded rather than `EXCLUDED_DOCS` widened — the guard was doing its job.

### DIFF
--- a/DESIGN_PRINCIPLES.md
+++ b/DESIGN_PRINCIPLES.md
@@ -67,9 +67,10 @@ enough that the codebase itself carried hybrids like `onleftImageClick` that wer
 neither convention. A single rule needs no such knowledge, and it matches the spelling
 consumers of the forked library already write.
 
-The earlier spellings all remain accepted as `@deprecated` aliases: passing `onRowClick`
-still works and warns once, in dev, naming its replacement. 4.0.0 removes them; `npx
-sui-codemod ./src` rewrites a consumer's call sites (see `docs/EVENT_CASING_MIGRATION.md`).
+The earlier spellings were accepted as `@deprecated` aliases through 3.5.0: passing
+`onRowClick` worked and warned once, in dev, naming its replacement. 3.5.1 removed them,
+under a patch bump it should not have had; `npx sui-codemod ./src` rewrites a consumer's
+call sites (see `docs/EVENT_CASING_MIGRATION.md`).
 
 `scripts/check-event-casing.js` enforces the rule going forward (`npm run
 lint:event-casing`, wired into `npm run lint`). There is no grandfathering list: an
--- a/README.md
+++ b/README.md
@@ -38,15 +38,25 @@ This library takes a different approach: **components are unstyled by default** 
 
 ---
 
-## Coming in 4.0.0: one breaking change
+## Breaking changes, and the release they actually shipped in
 
-Advance notice, not a change in this release. **Nothing below has happened yet** —
-the current 3.x line still accepts everything it always did.
+**Both changes below shipped in 3.5.1, which went out as a patch by mistake.**
+The release workflow picks the version from the newest commit alone, and the
+commit that made the build green was a `fix:` — so the breaking commit beneath
+it was never seen. 4.0.0 is that same content under the version number it
+should have carried. A `^3.5.0` range already resolves to it; **3.5.0 is the
+last release without these changes.**
 
-In 4.0.0, `children` stops being a declared property on `sui-chat-bubble`,
-`sui-draggable` and `sui-resizable`. Assigning it will set an inert expando and
-the content will silently disappear. Pass the content as light-DOM children
-instead, which those wrappers will forward through their default slot:
+**Every deprecated event-prop spelling is gone.** The 190 camelCase and
+mixed-case event props that 3.x accepted as aliases are now unknown props,
+which Svelte drops without an error — the handler simply stops running.
+`npx sui-codemod ./src` rewrites them, and
+[`docs/MIGRATION_4.0.md`](docs/MIGRATION_4.0.md) lists every pair.
+
+**`children` is no longer a declared property** on `sui-chat-bubble`,
+`sui-draggable` and `sui-resizable`. Assigning it sets an inert expando and the
+content silently disappears. Pass the content as light-DOM children instead,
+which those wrappers forward through their default slot:
 
 ```js
 el.children = mySnippet; // before — no longer works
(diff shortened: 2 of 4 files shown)


Tara-ag (commented):
Review in progress

Tara-ag:
⚠️ MAJOR: The `publish-mcp` job in `.github/workflows/release.yml` has the same `git log -1` class of failure this PR documents for the root package. After the root `release` job pushes `chore(release): 4.0.0`, the MCP job rebases onto that commit and then runs `git log --grep="^chore(release):" --invert-grep -1`. The latest non-release commit becomes this PR's `docs(release)!:` merge commit, which carries `BREAKING CHANGE:`. That would force a major bump of `@juspay/svelte-ui-components-mcp` regardless of what actually changed under `mcp/`. For this particular PR the issue is latent because `mcp/` is untouched, but any future PR that changes both packages will hit it. Scope the MCP bump to commits that touched `mcp/` since the last `mcp-*` tag rather than the latest non-release commit on the branch.

Tara-ag (approved):
Docs-only correction that accurately records 3.5.1 as the release that shipped the breaking change, and uses `!` + `BREAKING CHANGE:` to ensure the workflow cuts 4.0.0. Approved with one related workflow note filed separately.

Note: I also identified a related issue in `.github/workflows/release.yml` (the MCP publish job derives its bump from the latest non-release commit, so after the root `chore(release): 4.0.0` commit is pushed it will see this PR's `docs(release)!:` merge commit and force a major MCP bump regardless of `mcp/` changes). This is pre-existing on the release branch and not introduced by this PR, so it should be tracked and fixed in a follow-up rather than blocking this docs correction.

Tara-ag (approved):
Verified and approved. This is a docs-only correction that accurately records that the breaking change shipped in 3.5.1 rather than 4.0.0. I confirmed:

- The latest release is 3.5.1 (tag `7b44a5a`), whose immediate parent is `d6066e9` (`fix(wc):...`), with the breaking commit `190800b` (`feat(events)!:...`) below it — exactly the failure mode described.
- `release.yml` derives the bump from `git log -1` alone, and both the `!` in the subject and the `BREAKING CHANGE:` footer force a major bump, so merging this will correctly cut 4.0.0.
- All four edited docs are internally consistent: `3.5.0` is the last release accepting legacy spellings, `3.5.1` shipped the removal as a patch, and `4.0.0` is the same content under the correct major version.
- No source code, exports, CSS custom properties, or public API surface is changed.

No actionable issues found.

Tara-ag:
⚠️ MAJOR: The `publish-mcp` job in `.github/workflows/release.yml` has the same `git log -1` class of failure this PR documents for the root package. After the root `release` job pushes `chore(release): 4.0.0`, the MCP job rebases onto that commit and then runs `git log --grep="^chore(release):" --invert-grep -1`. The latest non-release commit becomes this PR's `docs(release)!:` merge commit, which carries `BREAKING CHANGE:`. That would force a major bump of `@juspay/svelte-ui-components-mcp` regardless of what actually changed under `mcp/`. For this particular PR the issue is latent because `mcp/` is untouched, but any future PR that changes both packages will hit it. Scope the MCP bump to commits that touched `mcp/` since the last `mcp-*` tag rather than the latest non-release commit on the branch.

PR #557, opened 2026-09-06
feat(media-player): add a seek bar, an elapsed/total clock and a fullscreen toggle

murdore:
## What is the problem?

`MediaPlayer` could play, pause and mute, and nothing more. A consumer wanting a scrubber, a running time or a fullscreen button had to hand-roll a player around this library's `Button` and `Slider` rather than use this component.

Lighthouse has exactly that: a 393-line `VideoPlayer` composing `Button`, `Img` and `Slider` into a player with a seek bar, elapsed and total time, mute and fullscreen. Its component-reuse audit recorded the surface as unmigratable, and correctly attributed the gap to this library rather than to itself.

## Solution

Three controls, each **opt-in and off by default**, so a player that showed only play and mute keeps showing only play and mute:

| Prop | Effect |
|---|---|
| `seekBar` | a `Slider` bound to position; dragging scrubs, playback moves the handle |
| `timeDisplay` | `m:ss`, widening to `h:mm:ss` only once the media runs past an hour |
| `fullscreenButton` | enters and exits fullscreen; `fullscreenIcon` / `exitFullscreenIcon` override the defaults |

`currentTime` and `duration` are bindable, so a host can restore a saved position or drive its own scrubber. Three events report what happened, all lowercase per the 4.0.0 rule:

- `onseek(currentTime)` for a deliberate scrub, and only that
- `ontimeupdate(currentTime, duration)` for playback advancing
- `onfullscreenchange(isFullscreen)` for the state it is now in

### Two decisions worth stating, because the obvious alternative is wrong

**Fullscreen is requested on the container, not on the `<video>`.** A fullscreen video element paints over everything and takes the overlay's play, mute and seek controls off screen with it. The container keeps them on top of the media where they remain usable. There is a test asserting which element is asked, because this is the kind of thing a later refactor silently reverses.

**Any control turning on also sets `preload="metadata"`.** A seek bar and a clock are unusable until the length is known, and a paused player is not otherwise obliged to fetch it. This only applies when a control is enabled, so existing consumers see no change in loading behaviour.

### A defect found by measuring rather than reasoning

`loadedmetadata` is one-shot, and a cached or fast-loading file reaches `HAVE_METADATA` before hydration attaches the handler. Measured on the demo page: the element reported `readyState 4` and `duration 4` while the component still held `0`, and nothing fires the event again. A clock that only listened would stay dead for exactly the media that loaded well. The component now also adopts what the element already knows when it binds, so both the fast and slow paths are covered.

## What this does not change

- Any existing consumer. All three controls default off; with none enabled the rendered output, the loading behaviour and the control count are what they were.
- The `image` type, which is untouched.
- Any existing prop or event, here or on `Slider`. Every addition is optional and defaults to absent/false.

## Proof of testing

| Gate | Result |
|---|---|
| `npm run check` (svelte-check, **check:wc**, codemod) | 0 errors |
| `npm run lint` (prettier, eslint, event-casing) | clean; 0 event-casing violations across 94 property files |
| wc prop parity | 90 passed; the wrapper declares every new prop |
| `npm run test:unit` | 36 files, **828 passed** |
| `tests/media-player.test.ts` | **16 passed**, the 11 that existed plus 5 new |

The five new tests: the controls are absent on a default player, present when asked for, `duration` binds out and the clock formats, seeking moves the element and reports through `onseek`, and fullscreen asks the container rather than the video.

Two of them failed on their first run and the failures were real rather than flaky. The first exposed that seeking before `duration` is known is a no-op, so the test was passing through a guard rather than through the feature; the second exposed the pre-hydration metadata defect described above, which is now fixed in the component. Both are recorded here because a test that was fixed by weakening it would look identical in the final table.

**Video proof is attached below**, recorded against this branch: the clock reads the length before anything plays, three separate seeks each move the position and the clock together, playback advances on its own, and mute still works alongside the new row. Fullscreen is asserted by test rather than filmed, because headless Chromium will not grant it.

## Areas of impact

- `src/lib/MediaPlayer/MediaPlayer.svelte`, `properties.ts`
- `src/lib/Slider/` gains an optional `ariaLabel` (additive, default absent), plus its wrapper and docs — the seek bar needed an accessible name and the library's existing pattern is a component-level `ariaLabel`
- `src/wc/components/MediaPlayer.wc.svelte`: seven new props declared, so `sui-media-player` exposes them too
- Two new icon assets, matching the existing stroke style
- The MediaPlayer demo route gains a transport player and a readout bound to its own state

## Summary by CodeRabbit

* **New Features**
  * Added optional MediaPlayer transport controls, including a seek bar, time display, and fullscreen toggle.
  * Added support for tracking playback time and duration.
  * Added seek, playback-time, and fullscreen change event callbacks.
  * Exposed the new controls and playback properties through the custom element interface.
* **Documentation**
  * Updated the MediaPlayer demo to showcase transport controls and playback state.


murdore:
## Video proof of the three controls

Recorded against this branch with a Playwright rig that asserts each surface is live before filming, so a recording of an inert player would have failed rather than been filed. The rig is deliberately not committed: it is a one-off recorder, and leaving it in `tests/` would run it on every CI pass for no assertion value.

**What the recording shows, in order:**

1. **The clock knows the length before anything plays.** The player is paused on load and the time reads `0:00 / 0:04`. This is the pre-hydration metadata path described in the description; before the fix this frame read `--:--`.
2. **Three separate seeks, each moving position and clock together.** Seeking to 1s, 2s and 3s; the clock steps `0:01 / 0:04`, `0:02 / 0:04`, `0:03 / 0:04`, and the host readout beneath the grid shows `lastSeek=1.0`, then `2.0`, then `3.0`. The readout is bound to the component's own `currentTime` and `duration`, so those are the component's numbers rather than a copy the page kept.
3. **Playback advances on its own** after pressing play, so the clock is seen moving without a seek, which is what separates `ontimeupdate` from `onseek`.
4. **Mute still works alongside the new row**, its icon switching from muted to unmuted, confirming the transport row did not displace the existing control.

Fullscreen is asserted by test rather than filmed, because headless Chromium will not grant a real fullscreen request. The test patches `Element.prototype.requestFullscreen` before page load and asserts the element asked is the player's container `div` and not the `video`, which is the actual contract: asking the video would take these very controls off screen.

**Files, for anyone reproducing or attaching them:**

- `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-lighthouse-fork-feat-componen-reuse/c031a8c7-d321-4ef0-babc-a47b079e7544/scratchpad/mediaplayer-transport.mp4` (8.2s, 224 KB)
- `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-lighthouse-fork-feat-componen-reuse/c031a8c7-d321-4ef0-babc-a47b079e7544/scratchpad/mediaplayer-transport-frames.png` (four evenly spaced frames as a contact sheet)

GitHub takes media only through a browser drag-and-drop, not through its API, so the two files above are referenced by path rather than embedded. Both reproduce from this branch by running the demo route and driving the transport player.

## Why this closes a real consumer gap

Lighthouse's `VideoPlayer` is 393 lines composing `Button`, `Img` and `Slider` to get a seek bar, a clock, mute and fullscreen. Every one of those four is now a prop on this component. That surface is the reason the gap was recorded against this library rather than against the consumer, and it is what the change was measured against rather than invented from.

murdore:
## The `visual` job failed on the first run, and it was mine

Recording this rather than quietly pushing a green baseline, because "regenerated the baseline" is exactly the sentence that hides a real regression.

**What failed.** `media-player renders as baselined` — the only failure, against 92 passed and 1 skipped:

```
Expected an image 980px by 1684px, received 980px by 2225px.
428920 pixels (ratio 0.20 of all image pixels) are different.
```

**Why.** This PR adds a transport player to the MediaPlayer demo route, so `main.content` is 541px taller than when the baseline was committed. The image is stale by construction; nothing rendered wrongly.

**The flake question, checked before regenerating rather than after.** The new demo shows a clock whose text depends on whether metadata has loaded — precisely the kind of thing that bakes a race into a baseline and fails at random later. It is safe here, and by the suite's existing design rather than by luck: `MASKS['media-player']` is already `['.demo-row']`, because a decoded video frame depends on decoder timing. Both the transport player and the `transport-readout` span sit inside `.demo-row` elements, so the frame, the clock and the readout are all painted over. The only thing that changes in the image is height.

**How it was regenerated.** Inside `mcr.microsoft.com/playwright:v1.60.0-noble` via `scripts/visual-test.sh --update-snapshots`, which is the same image `visual.yml` runs in — not on the host, where freetype and font hinting differ enough to fail an exact comparison. The regenerated file is **980×2225**, matching the size CI reported receiving, which is the evidence that a laptop-generated baseline and CI agree here.

**Verified after regenerating**, in compare mode, whole suite: **93 passed, 1 skipped, 0 failed**. Only `tests/visual/__screenshots__/media-player.png` is modified; no other baseline moved.

Amended into the single commit rather than added on top, per the single-commit policy.

Tara-ag (commented):
Review in progress — findings posted inline.

src/lib/MediaPlayer/MediaPlayer.svelte
```diff
@@ -24,14 +27,55 @@
...
+    fullscreenButton = false,
+    fullscreenIcon,
+    exitFullscreenIcon,
+    currentTime = $bindable(0),
+    duration = $bindable(0),
     onplay,
     onpause,
     onvolumechange,
+    onseek,
+    ontimeupdate,
```
Tara-ag, line 41: ⚠️ MAJOR: `currentTime` is documented as writable and seekable, but external writes do not seek the video. The component destructures `currentTime = $bindable(0)` and writes it from `handleTimeUpdate`/`handleSeek`, yet there is no `$effect` that reacts to a host changing `currentTime` and writes it back to `videoPlayer.currentTime`. A host restoring a saved position via `bind:currentTime` or `currentTime={saved}` will see the prop update but the video stays put, violating the documented contract.

Suggested fix: add an effect that seeks the element when the bound value differs from the element's position:

```ts
$effect(() => {
  if (videoPlayer === null || duration <= 0) {
    return;
  }
  const target = Math.min(Math.max(currentTime, 0), duration);
  if (Math.abs(videoPlayer.currentTime - target) > 0.05) {
    videoPlayer.currentTime = target;
  }
});
```

src/lib/MediaPlayer/MediaPlayer.svelte
```diff
@@ -139,7 +274,26 @@
...
-        <div class="bottom-controls">
+        <div class="bottom-controls" class:transport={hasTransportRow}>
+          {#if seekBar}
+            <div class="seek" data-pw={typeof testId === 'string' ? `${testId}-seek` : null}>
+              <Slider
+                value={currentTime}
+                min={0}
+                max={duration > 0 ? duration : 1}
+                step={0.1}
+                disabled={duration <= 0}
```
Tara-ag, line 353: 💡 MINOR: `labelFormatter={formatTime}` is passed to `Slider`, but `Slider` only uses `labelFormatter` when `showValue` is true. Since `showValue` is not passed here, the formatter has no effect and the formatted time is only shown by the separate `timeDisplay` span. Either pass `showValue` to make the formatter useful, or remove `labelFormatter` to avoid misleading future maintainers.

src/lib/MediaPlayer/MediaPlayer.svelte
```diff
@@ -139,7 +274,26 @@
             </Button>
           </div>
         </div>
```
Tara-ag, line 344: ⚠️ MAJOR: The seek bar renders a library `Slider`, but the underlying `<input type="range">` has no accessible name. `Slider` does not expose an `ariaLabel` prop and the input carries no `aria-label`/`<label>`, so screen-reader users will hear a value without context (e.g. "slider, 0, 10" instead of "Seek, 0, 10").

Suggested fix: extend `Slider` to accept and forward an `ariaLabel` prop, then pass an explicit label here (e.g. `ariaLabel="Seek"`).

Tara-ag (changes requested):
Reviewed file-by-file. Findings:

- **MAJOR** `currentTime` is documented as writable/seekable but external writes do not seek the video — the bindable contract is incomplete.
- **MAJOR** The seek bar `Slider` has no accessible name; screen-reader users get no context for the control.
- **MINOR** `labelFormatter={formatTime}` on the seek Slider is dead code because `showValue` is not passed.
- **MINOR** `docs/MediaPlayer.md` is not updated for the new props, events and CSS variables.

No blocking public-API breakage, secrets, or XSS found. Requesting changes for the two MAJOR items.

src/lib/MediaPlayer/MediaPlayer.svelte
```diff
@@ -139,7 +274,26 @@
...
-        <div class="bottom-controls">
+        <div class="bottom-controls" class:transport={hasTransportRow}>
+          {#if seekBar}
+            <div class="seek" data-pw={typeof testId === 'string' ? `${testId}-seek` : null}>
+              <Slider
+                value={currentTime}
+                min={0}
```
Tara-ag, line 350: 💡 MINOR: `labelFormatter={formatTime}` formats the time visually, but `Slider` only uses `labelFormatter` for its optional visible value span. The underlying `<input type="range">` has no `aria-valuetext`, so screen readers will announce the raw numeric seconds (e.g. "12.5") instead of the formatted time ("0:12").

Either extend `Slider` to set `aria-valuetext={labelFormatter?.(value) ?? value}` on the range input, or add an accessible value announcement around the seek bar here.

src/lib/MediaPlayer/MediaPlayer.svelte
```diff
@@ -24,14 +27,55 @@
     captionsSrc,
     captionsLabel,
     captionsSrcLang,
+    seekBar = false,
```
Tara-ag, line 30: ⚠️ MAJOR: `currentTime` is documented as bindable and writable, with the promise that "setting it seeks". Right now the component only writes `videoPlayer.currentTime` inside `handleSeek`; there is no `$effect` reacting to an external change of this bindable. A host setting `currentTime` programmatically will update local state but leave the video at its old position, breaking the advertised public contract.

Add a guarded `$effect` that seeks the element when the host-driven `currentTime` diverges from the element's actual position (mirror the existing `playing` effect), and cover it with a test.

src/lib/MediaPlayer/MediaPlayer.svelte
```diff
@@ -83,9 +127,97 @@
...
+    }
+    if (duration === 0 && Number.isFinite(videoPlayer.duration)) {
+      duration = videoPlayer.duration;
+    }
+  });
+
+  /**
+   * Seeking writes the element directly rather than waiting for the bound value to
+   * settle, so a drag scrubs while the pointer is still down. `onseek` fires only from
+   * here, which is what separates a deliberate scrub from playback advancing on its own.
```
Tara-ag, line 216: ⚠️ MAJOR: The fullscreen button branches on `document.fullscreenElement === null`. If a different element on the page is already fullscreen, clicking "Enter fullscreen" on this player takes the else branch and calls `document.exitFullscreen()`, exiting the unrelated element instead of entering this player's container. The button label and action disagree.

Branch on `document.fullscreenElement === container` instead, and add a regression test that fullscreen-shares another element first, then clicks this player's button.

Tara-ag:
💡 MINOR: `docs/MediaPlayer.md` is not updated for the new public API. This PR adds seven new props (`seekBar`, `timeDisplay`, `fullscreenButton`, `fullscreenIcon`, `exitFullscreenIcon`, `currentTime`, `duration`), three new events (`onseek`, `ontimeupdate`, `onfullscreenchange`), and several new CSS custom properties (`--media-player-transport-*`, `--media-player-seek-*-color`, `--media-player-time-*`, `--media-player-fullscreen-background`). The docs still only document the pre-PR surface, so downstream consumers won't know the new theming hooks or bindable contracts.

Please update the Props, Events, and CSS Variables tables in `docs/MediaPlayer.md`.

src/lib/MediaPlayer/MediaPlayer.svelte
```diff
@@ -241,6 +420,44 @@
...
+    align-items: center;
+    gap: var(--media-player-transport-gap, 12px);
+  }
+
+  .seek {
+    flex: 1;
+    min-width: 0;
+    --slider-track-color: var(--media-player-seek-track-color, #ffffff59);
+    --slider-fill-color: var(--media-player-seek-fill-color, #ffffff);
+    --slider-thumb-color: var(--media-player-seek-thumb-color, #ffffff);
```
Tara-ag, line 504: ⚠️ MAJOR: Seek-bar CSS variables do not style the Slider. The `.seek` block defines `--media-player-seek-track-color`, `--media-player-seek-fill-color`, and `--media-player-seek-thumb-color` and maps them to `--slider-track-color`, `--slider-fill-color`, and `--slider-thumb-color`. But `Slider.svelte` consumes `--slider-track-active-color`, `--slider-track-background`, and `--slider-thumb-background`, not the names being set. Consequently the declared theme variables have no effect and the seek bar renders with Slider's default blue/gray styling instead of the intended overlay colors.

Map the public variables to Slider's actual custom properties:

```css
.seek {
  flex: 1;
  min-width: 0;
  --slider-track-active-color: var(--media-player-seek-fill-color, #ffffff);
  --slider-track-background: var(--media-player-seek-track-color, #ffffff59);
  --slider-thumb-background: var(--media-player-seek-thumb-color, #ffffff);
  --slider-thumb-border: 2px solid var(--media-player-seek-thumb-color, #ffffff);
}
```

src/lib/MediaPlayer/MediaPlayer.svelte
Tara-ag, line 41: ⚠️ MAJOR: Writable `currentTime` prop does not actually seek the video. The prop is declared with `$bindable(0)` and documented as "writable: setting it seeks", but no `$effect` reacts when the host writes to `currentTime` and assigns it to `videoPlayer.currentTime`. Setting the prop from outside only updates the Slider's visual value and the time display; the media element stays where it was. This breaks the documented two-way contract used to restore a saved position or drive an external scrubber.

Add a guarded effect that seeks the element when the host changes `currentTime`:

```svelte
$effect(() => {
  if (
    videoPlayer !== null &&
    Number.isFinite(currentTime) &&
    videoPlayer.currentTime !== currentTime
  ) {
    videoPlayer.currentTime = Math.min(Math.max(currentTime, 0), duration || currentTime);
  }
});
```

Then add a Playwright/Vitest test that sets the bound `currentTime` from the host and asserts `video.currentTime` moves.

Tara-ag (commented):
Reviewed file by file. Two MAJOR issues and one MINOR docs gap were accepted by the review gate and are detailed in inline comments. The MAJOR issues affect the documented bindable `currentTime` contract and the seek-bar theming variables; the docs gap is on an unchanged file so it could not be attached inline. Please address the MAJOR items before merging.

Tara-ag (changes requested):
Changes requested due to MAJOR functional and accessibility gaps in the new MediaPlayer transport controls. See inline comments for details.

Tara-ag:

The PR adds optional transport controls (seek bar, elapsed/total clock, fullscreen toggle) to MediaPlayer and an ariaLabel prop to Slider. All new props/events are additive; no existing props, exports from src/lib/index.ts, or CSS custom properties were removed or renamed, so the public contract remains backward compatible. No hardcoded secrets or unsafe HTML injection were found. Browser-only fullscreen APIs are accessed inside event handlers and $effects with cleanup, so SSR is safe. I could not fully verify two accessibility concerns before the time budget expired: (1) whether the bottom transport controls rely solely on hover visibility, which would make them unreachable to keyboard users, and (2) whether the time display uses an aria-live region so screen readers announce elapsed/total updates. Neither was confirmed with line-number evidence, so they are noted for follow-up rather than raised as gated issues.

murdore:
## Addressing the review: both MAJOR findings were real, and one of them contradicted my own PR body

**MAJOR — `currentTime` did not seek.** Correct, and the sharper version is that this PR's own description claimed "a host can restore a saved position" while the code could not do it. The value was bindable outward only: a host write moved the number and never the media. CodeRabbit found the same thing independently, which is a good sign it was real rather than stylistic.

Fixed by syncing inward, clamped to `duration`. Two details worth stating because the naive fix is wrong:

- A plain, **non-reactive** `syncedTime` records what the component itself last wrote. Without it every outward `timeupdate` reads back as an inward write and re-seeks the element to the position it just reported — a feedback loop that stutters playback.
- An inward write deliberately **does not** fire `onseek`. `onseek` means the user scrubbed; a host restoring a position is not a scrub, and reporting it as one would make every save/restore cycle look like user intent.

**MAJOR — the seek bar had no accessible name.** Correct: a bare range input announces only its value, so a screen-reader user heard a number with no indication of what it controls. `Slider` gains an optional `ariaLabel` following the pattern the rest of the library already uses, forwarded through the wrapper as `sliderAriaLabel` with the `aria-label` attribute (the platform defines `ariaLabel` on every `HTMLElement`). MediaPlayer passes `"Seek"`, consistent with the hardcoded labels its play and mute controls already carry. This widens the diff to `Slider` — flagging that rather than burying it.

**MINOR — `labelFormatter` was dead.** Right, `showValue` was never passed. Removed.

**MINOR — docs.** `docs/MediaPlayer.md` now covers the seven props, three events and nine CSS variables; `docs/Slider.md` covers the new prop. The accessibility section also records why the clock is *not* a live region: it changes every frame during playback, which would flood a screen reader rather than inform it.

### Also fixed, from CodeRabbit: fullscreen state was broken in the web-component build

`document.fullscreenElement` retargets to the shadow **host**, so it never equals the container inside `<sui-media-player>`. The check read false while genuinely fullscreen, leaving the icon stuck and `onfullscreenchange` silent for every web-component consumer. Both the toggle and the `fullscreenchange` listener now resolve through the container's own root. Same class of bug as the metadata one: correct in the common path, silently wrong in a real one.

CodeRabbit's other point — that external `currentTime` writes should seek — is the same finding as the MAJOR above and is fixed by the same change.

### Verification

| Gate | Result |
|---|---|
| `tests/media-player.test.ts` | **19 passed** (16 + 3 new) |
| `npm run check` (incl. `check:wc`) | 0 errors |
| `npm run lint` | clean; 0 event-casing violations |
| wc prop parity | 90 passed — the wrapper declares `sliderAriaLabel` |
| `npm run test:unit` | 36 files, **828 passed** |
| visual suite | 93 passed / 1 skipped |

The three new tests were each watched failing first. One of them failed a second time on a wrong assertion of mine — I asserted `lastSeek=0.0` where the demo's untouched sentinel is `-1`. The component was right and the test was wrong; `-1` is also the stronger assertion, since a literal `0` is indistinguishable from a real scrub to the start.

**One flake, disclosed rather than hidden.** In the full visual run after these changes, `select` failed once. It is unrelated: `select` renders no `Slider`, my working tree modified no baseline but `media-player.png`, and `select` then passed **2 out of 2** re-runs against its unchanged committed baseline. I have not touched it or added a tolerance for it.

The `media-player` baseline is regenerated a second time, because the first new test needs a host-side button on the demo page and that grew the route again.

Tara-ag (commented):
Review in progress

src/lib/MediaPlayer/MediaPlayer.svelte
```diff
@@ -241,6 +467,44 @@
     visibility: var(--bottom-controls-visibility);
   }
 
+  /* With a seek bar, time or fullscreen present the row becomes a transport bar: the
+     seek bar takes the free space and the rest sit beside it, centred on each other. */
+  .bottom-controls.transport {
+    justify-content: var(--media-player-transport-justify, flex-start);
+    align-items: center;
```
Tara-ag: ⚠️ MAJOR: New transport controls are unreachable by keyboard. The overlay hides `.bottom-controls` with `visibility: hidden` by default, which removes children from the tab order. Because the overlay is a sibling of the focusable `<video>`, no child control can ever receive focus to trigger `:focus-within`. Keyboard users cannot reach the seek bar, time display, or fullscreen button.

Replace the `visibility` toggle with a keyboard-safe approach that keeps controls in the tab order while still letting pointer events pass through to the video when inactive, e.g. `opacity: 0` + `pointer-events: none` by default and `opacity: 1` + `pointer-events: auto` on hover/focus-within.

src/lib/MediaPlayer/MediaPlayer.svelte
```diff
@@ -83,9 +132,139 @@
...
+   * itself produce counts as a host write. `onseek` deliberately does not fire: a host
+   * restoring a position is not a user scrubbing.
+   */
+  // eslint-disable-next-line no-restricted-syntax
+  $effect(() => {
+    const requested = currentTime;
+    if (videoPlayer === null || Math.abs(requested - syncedTime) < SEEK_EPSILON) {
+      return;
+    }
+    const clamped =
```
Tara-ag: ⚠️ MAJOR: Inward `currentTime` write can throw before video metadata loads. The effect assigns `videoPlayer.currentTime` without checking `readyState`. Setting `currentTime` on an element whose `readyState` is `HAVE_NOTHING` throws an `InvalidStateError`. Guard this assignment the same way `handleSeek` does: skip it while `duration <= 0` (which implies metadata is not yet available).

```ts
$effect(() => {
  const requested = currentTime;
  if (videoPlayer === null || duration <= 0 || Math.abs(requested - syncedTime) < SEEK_EPSILON) {
    return;
  }
  const clamped = Math.min(Math.max(requested, 0), duration);
  syncedTime = clamped;
  videoPlayer.currentTime = clamped;
});
```

murdore:
## Second Yama round: one finding was real, one was not — and I tested rather than argued

Yama returned **BLOCKED** with two MAJOR findings on the previous commit. Its verdict comment did not carry the detail (the action's own posting step errored: `DECISION: BLOCKED / OUTCOME: failure`), so I pulled the findings out of the job log.

### MAJOR 1 — inward `currentTime` write before metadata. Real. Fixed.

Correct, and my first fix did have this hole: the effect wrote `videoPlayer.currentTime` unconditionally once a host value differed. Before `HAVE_METADATA` there is no timeline to seek within — the length is unknown so the value cannot be clamped, assigning `currentTime` is specified to set a *default start position* rather than seek, and some engines throw outright.

The write is now **held** and applied, clamped, when metadata arrives — which is the moment a restored position first means anything. The same effect that already adopts pre-hydration metadata applies it, so the two pre-metadata cases are handled in one place rather than two.

### MAJOR 2 — "new transport controls are unreachable by keyboard". Not reproducible. Evidence below.

The claim is that the transport row is hover-only. I did not argue this; I wrote the test it predicts should fail:

```
focus the <video> → Tab → Tab → expect the seek input to be focused and visible
```

It **passes**. No hover anywhere in the test, and a `visibility: hidden` element cannot take focus at all, so if the row were hover-only that Tab could not land there. The mechanism is the pre-existing `.overlay:hover, .overlay:focus-within` rule, which the new row inherits by sitting inside `.bottom-controls` — the same rule that makes the existing play and mute controls keyboard-reachable, covered by a test that predates this PR.

I have kept the test rather than just reporting the result, so the claim is now pinned: if someone later makes the transport row hover-only, this fails.

### Verification

| Gate | Result |
|---|---|
| `tests/media-player.test.ts` | **20 passed** (16 original + 4 new) |
| `npm run check` (incl. `check:wc`) | 867 + 432 files, **0 errors** |
| `npm run lint` | clean; 0 event-casing violations |
| `npm run test:unit` | 36 files, **828 passed** |

No baseline regeneration this round — the new test adds no page content.

For the record on sequencing: 4.1.0 is on the registry (published by `feat(Table)`, landed on `release` independently of this PR), so this cuts **4.2.0**, not 4.1.0.

murdore:
## Answering Yama's two unverified follow-ups, both with evidence already on this branch

The latest verdict is **0 CRITICAL · 0 MAJOR · 0 MINOR** and the check passes. It left two accessibility questions unverified because its time budget expired, and explicitly noted them for follow-up rather than as gated issues. Both are already settled here, so rather than leave them hanging:

**1. "Do the bottom transport controls rely solely on hover visibility, making them unreachable to keyboard users?"** — No, and this is pinned by a test rather than by argument:

```
focus the <video> → Tab → Tab → the seek input is focused and visible
```

No hover anywhere in that test. A `visibility: hidden` element cannot receive focus at all, so if the row were hover-only that Tab could not land there. The mechanism is the pre-existing `.overlay:hover, .overlay:focus-within` rule at `MediaPlayer.svelte:445-449`, which the transport row inherits by sitting inside `.bottom-controls` — the same rule covering the existing play and mute controls, itself covered by a test that predates this PR. This exact concern was raised as MAJOR in the previous round; I could not reproduce it, so I wrote the test it predicted would fail, and kept it so a future regression is caught.

**2. "Does the time display use an aria-live region so screen readers announce elapsed/total updates?"** — Deliberately not, and the reasoning is recorded in `docs/MediaPlayer.md`:

> The clock is plain text and is not announced as a live region: it changes every frame during playback, which would flood a screen reader rather than inform it.

A live region on a per-frame counter produces continuous speech interruption — it degrades the experience for the users it is meant to serve. The seekable control carries the accessible name instead (`aria-label="Seek"`), which is what a screen-reader user actually operates; a native range input announces its own value on interaction and on focus, so position is available on demand rather than shouted continuously.

If the project would prefer a periodic announcement, the right shape is a throttled, opt-in live region rather than making this one live by default — happy to add that as a follow-up if wanted, but it should be a deliberate choice rather than a default.

### Gate state on `c09010b`

`build`, `visual`, `checks`, `Yama PR Review`, CodeQL, CodeRabbit and both SentinelOne gates: **all pass**. 20 of 20 tests in `tests/media-player.test.ts`, 828 unit tests, 0 check errors.

The GitHub review still shows CHANGES_REQUESTED from `686325c`, two commits ago, which predates both fix rounds; Yama's current pass on `c09010b` supersedes it.

PR #558, opened 2026-09-06
fix(release): derive both packages' versions from every unreleased commit

murdore:
## ⏳ Merge this after 4.1.0 is on the registry

Not urgent, and deliberately not merge-on-green. A release run is in flight for `a2d2a70 feat(Table): …`, and merging into an in-flight run risks the concurrent-release race this workflow's own comments describe — two runs each choosing a version, each pushing a `chore(release):` commit and a tag. **This PR does not fix that race and does not touch it.** Wait for `npm view @juspay/svelte-ui-components version` to report 4.1.0, then merge.

## What went wrong

`Determine version bump type` read **`git log -1`**. That is what published 3.5.1:

| | |
|---|---|
| Range since 3.5.0 held | `feat(events)!: remove the 3.x legacy spellings and host-reserved element props` |
| Newest commit was | `fix(wc): restore the type coverage the host-reserved rename removed, and gate it` |
| Workflow computed | `patch` → **3.5.1** |

A breaking removal reached npm under a patch number. `^3.5.0` resolves to it on a routine install, and because Svelte drops an unknown prop without erroring, consumers' handlers stopped running with **no build failure and no runtime error**.

## Both jobs, and the trap the second one surfaced

Two packages release from this branch and each writes its own landmark: the root job commits `chore(release): 4.0.0`, the MCP job commits `chore(release): mcp 4.0.2`. **Both are on this history** — `30b7746` and `d2fd26e`.

A boundary matching `^chore(release):` loosely would let an MCP release end the *root* package's range and hide everything below it — the same failure as reading only the newest commit, one level up. My first push had exactly that bug. The boundaries are now distinct:

| Package | Boundary | Range |
|---|---|---|
| root | `chore(release): <digit>` | whole tree |
| MCP | `chore(release): mcp` | commits touching `mcp/` |

Neither kind of release commit drives a bump. MCP's path filter matches what already gates the job running at all (`mcp_changed`).

The two cross-package tests were **watched failing first**, and failed for the right reason — `expected 'patch' to be 'major'`, the MCP landmark truncating the root range.

## The change

The bump is now the strongest one any commit since the last release asks for.

Classification of a single commit is **unchanged** — the regexes are carried over verbatim, including the asymmetry where the type is matched case-insensitively and the `!` is not, so `FEAT!:` still reads as a feature. Widening what counts as breaking is a different change from not missing what already counts, and this PR is only the second one.

The boundary is a **release commit, not a tag**. The release job pushes the commit before it pushes the tag, and this workflow's own comments describe the window where the publish succeeds and the tag push does not — so a tag can be absent for a version already on the registry, while the release commit cannot. `publish-mcp` already used a release commit as its landmark for the same reason; what it lacked was a range.

The logic moves out of inline YAML into `scripts/release/version-bump.js` so it can be tested at all. It was previously unreachable from any test, which is why a rule this load-bearing had none.

## Controls

A gate nobody has seen fail proves nothing, so each of these was watched failing.

**Replayed against the real history**, which is a better test than any fixture because it reproduces the failure that actually happened:

```
=== 3.5.1 (the failure) — HEAD d6066e9 ===
old rule (git log -1): patch -> published 3.5.1
  patch  fix(wc): restore the type coverage the host-reserved rename removed, and gate it
  minor  feat(ChatMessage): mark a turn with a leading accent bar via `marker`
  minor  feat(ChatMessage): reveal a streamed message progressively via TypewriterText
  major  feat(events)!: remove the 3.x legacy spellings and host-reserved element props
  minor  feat(events): make every event prop lowercase, keeping the old spellings as aliases
new rule over 5 commit(s): major        <- would have cut 4.0.0
PASS

=== 4.0.0 (the correction) — HEAD 3013d88 ===
new rule over 1 commit(s): major        <- correct case unchanged
PASS
```

**17 unit tests**, including that exact five-commit range, a `BREAKING CHANGE:` appearing mid-sentence in prose (must stay `patch`, not silently become a major), an empty range, and the Jira-prefixed subjects the workflow has always accepted.

**`check:release` is new and wired into `check`.** Verified it actually fails — returning a bump outside the union gives:

```
scripts/release/version-bump.js(60,31): error TS2322: Type '"PATCH_TYPO"' is not assignable to type 'Bump'.
```

The sabotage was then restored byte-identically (`diff -q` clean). This matters because a new script directory that no configured check reads is *exactly* the shape of the `check:wc` gap that let this incident reach a release four times.

**Both CLI modes run against this repository:**

```
$ node scripts/release/version-bump.js
Considering 2 root commit(s) since its last release:
  patch  fix(release): derive the version from every unreleased commit, not just the newest
  minor  feat(Table): render row selection with the library Checkbox
Version bump: MINOR

$ node scripts/release/version-bump.js --mcp
No MCP commits since its last release commit; defaulting to patch.
Version bump: PATCH
```

The root run is the fix demonstrating itself: it sees `feat(Table)` underneath a `fix:` and returns MINOR rather than PATCH.

## Left undone, deliberately

**The concurrent-run race is untouched**, as above. Two runs in flight still each choose their own version and each push a commit and a tag. That is a real problem and this PR does not address it.

## Gates

| Gate | Result |
|---|---|
| Lint | 0 — prettier, eslint, 0 event-casing violations across 94 `properties.ts` files |
| `pnpm check` | **exit 0 over both configs** — 867 files / 0 errors, 432 files / 0 errors, now including `check:release` |
| Unit | **845 passed** (up from 828) |

The 4 remaining `svelte-check` warnings predate this branch.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -227,59 +227,16 @@ jobs:
       # Linting moved to the `verify` job this one now needs, so it gates the
       # publish rather than running beside it.
 
+      # Reads EVERY commit since the last `chore(release):`, not just the newest
+      # one. `git log -1` published 3.5.1 as a patch while the range held
+      # `feat(events)!: remove the 3.x legacy spellings` four commits down.
+      # The rules for classifying a single commit are unchanged; what changed
+      # is that a `fix:` on top can no longer hide a `feat!:` underneath.
+      # `scripts/release/version-bump.test.ts` pins this, including a replay of
+      # that exact range.
       - name: Determine version bump type
         id: version-type
-        run: |
-          # Get the latest commit message
-          COMMIT_MSG=$(git log -1 --pretty=format:"%s")
-          echo "Latest commit: $COMMIT_MSG"
-
-          # Check for BREAKING CHANGE in commit body
-          COMMIT_BODY=$(git log -1 --pretty=format:"%b")
-          HAS_BREAKING_CHANGE=false
-          if echo "$COMMIT_BODY" | grep -iE "^BREAKING CHANGE:" > /dev/null; then
-            HAS_BREAKING_CHANGE=true
-          fi
-
-          # Determine version bump based on commit message (with optional Jira prefix like BZ-1234:)
-          # Pattern: [JIRA-123: ]feat[!|scope!]: message
-          if echo "$COMMIT_MSG" | grep -iE "(^|: )feat(\(.+\))?!?:" > /dev/null; then
-            # Check for breaking change indicator (!) or BREAKING CHANGE in body
-            if echo "$COMMIT_MSG" | grep -E "feat(\(.+\))?!:" > /dev/null || [ "$HAS_BREAKING_CHANGE" = true ]; then
-              echo "version_type=major" >> $GITHUB_OUTPUT
-              echo "Version bump: MAJOR (breaking change)"
-            else
-              echo "version_type=minor" >> $GITHUB_OUTPUT
-              echo "Version bump: MINOR (new feature)"
-            fi
-          elif echo "$COMMIT_MSG" | grep -iE "(^|: )fix(\(.+\))?!?:" > /dev/null; then
-            # Check for breaking change in fix commits
-            if echo "$COMMIT_MSG" | grep -E "fix(\(.+\))?!:" > /dev/null || [ "$HAS_BREAKING_CHANGE" = true ]; then
-              echo "version_type=major" >> $GITHUB_OUTPUT
-              echo "Version bump: MAJOR (breaking change in fix)"
-            else
-              echo "version_type=patch" >> $GITHUB_OUTPUT
-              echo "Version bump: PATCH (bug fix)"
-            fi
-          elif echo "$COMMIT_MSG" | grep -iE "(^|: )(build|chore|ci|docs|style|refactor|perf|test)(\(.+\))?!?:" > /dev/null; then
-            # Check for breaking change in other commit types
-            if echo "$COMMIT_MSG" | grep -E "(build|chore|ci|docs|style|refactor|perf|test)(\(.+\))?!:" > /dev/null || [ "$HAS_BREAKING_CHANGE" = true ]; then
-              echo "version_type=major" >> $GITHUB_OUTPUT
-              echo "Version bump: MAJOR (breaking change)"
-            else
-              echo "version_type=patch" >> $GITHUB_OUTPUT
-              echo "Version bump: PATCH (maintenance)"
-            fi
-          else
-            # Default to patch, but check for breaking change
-            if [ "$HAS_BREAKING_CHANGE" = true ]; then
-              echo "version_type=major" >> $GITHUB_OUTPUT
-              echo "Version bump: MAJOR (breaking change)"
-            else
-              echo "version_type=patch" >> $GITHUB_OUTPUT
-              echo "Version bump: PATCH (default)"
-            fi
-          fi
+        run: node scripts/release/version-bump.js
 
       - name: Configure Git
         run: |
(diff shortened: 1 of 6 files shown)


Tara-ag (approved):
Approved. The change correctly moves version-bump logic out of inline YAML into a tested, type-checked script, and the strongest-bump-over-the-range rule prevents a recurrence of the 3.5.1 incident. No public API, accessibility, SSR, or security concerns were found in the diff.

scripts/release/version-bump.js
```diff
@@ -0,0 +1,155 @@
...
+ * @returns {readonly Commit[]}
+ */
+export function unreleasedCommits(log, target = 'root') {
+  const isBoundary = target === 'mcp' ? MCP_RELEASE : ROOT_RELEASE;
+  const boundary = log.findIndex((commit) => isBoundary.test(commit.subject));
+  return boundary === -1 ? [...log] : log.slice(0, boundary);
+}
+
+/**
+ * @param {{ cwd?: string, path?: string }} [options]
```
Tara-ag, line 102: 💡 MINOR: `readLog` caps history at 200 commits and may miss the release boundary

`readLog` passes `-n 200` to `git log`. If more than 200 commits have landed since the last release, the boundary commit will not be in the output, `unreleasedCommits` will return all 200 records, and commits from already-released versions could influence the bump. This is unlikely for normal release cadence but is a latent correctness hole.

**Suggested fix:** Either remove the `-n 200` cap (history is fully fetched via `fetch-depth: 0`) or add a guard that fails loudly when the log may have been truncated.

Tara-ag:
💡 MINOR: Stale concurrency comment still references `git log -1`

**File:** `.github/workflows/release.yml` (line 21)

The block comment explaining the concurrent-run race says the version is chosen by `git log -1` in "Determine version bump type". That step now calls `scripts/release/version-bump.js`, which reads every commit since the last release. The stale mechanism description can mislead maintainers about what the race actually is.

**Suggested fix:** Update the comment to describe the current behavior — two runs each read their own checkout's full unreleased range and may choose the same next version.

.github/workflows/release.yml
```diff
@@ -227,59 +227,16 @@ jobs:
       # Linting moved to the `verify` job this one now needs, so it gates the
       # publish rather than running beside it.
 
+      # Reads EVERY commit since the last `chore(release):`, not just the newest
+      # one. `git log -1` published 3.5.1 as a patch while the range held
```
Tara-ag, line 231: 💡 MINOR: Stale concurrency comment still references `git log -1`

The block comment at the top of this file (explaining the concurrent-run race) says the version is chosen by `git log -1` in this step. This step now calls `scripts/release/version-bump.js`, which reads every commit since the last release. The stale mechanism description can mislead maintainers about what the race actually is.

**Suggested fix:** Update the top-of-file comment to describe the current behavior — two runs each read their own checkout's full unreleased range and may choose the same next version.

Tara-ag (commented):
Reviewed. Two minor findings accepted; both are documentation/robustness nits and neither blocks merge. The core fix is correct and well-tested.

Tara-ag:

## 🤖 Yama Review Summary

**Verified findings (4):**

- **MAJOR**: 200-commit cap can miss the release boundary and inflate the bump — `scripts/release/version-bump.js:108`
- **MINOR**: CLI main guard compares paths that can differ under symlinks — `scripts/release/version-bump.js:145`
- **MINOR**: Stale concurrency comment still references `git log -1` — `.github/workflows/release.yml:21`
- **MINOR**: `readLog` caps history at 200 commits and may miss the release boundary — `scripts/release/version-bump.js:102`

- ⚠️ **MAJOR**: 200-commit cap can miss the release boundary and inflate the bump — `scripts/release/version-bump.js:108`
  readLog fetches only the newest 200 commits. If more than 200 commits have landed since the last release, the boundary commit is not present, unreleasedCommits returns the whole fetched range, and already-released feats/fixes can be counted again. This silently undermines the exact correctness goal of the PR.
- 💡 **MINOR**: CLI main guard compares paths that can differ under symlinks — `scripts/release/version-bump.js:145`
  process.argv[1] reflects the path as invoked, while import.meta.filename is the resolved module path. If the script is ever run through a symlink or with a differently-normalized path, main() is skipped silently. The workflow invokes it directly today, but the guard is brittle.
- 💡 **MINOR**: Stale concurrency comment still references `git log -1` — `.github/workflows/release.yml:21`
  The block comment at the top of `release.yml` explains the concurrent-run race by saying the version is chosen by `git log -1` in "Determine version bump type". That step now calls `scripts/release/version-bump.js`, which reads every commit since the last release. The stale mechanism description can mislead maintainers about what the race actually is.
- 💡 **MINOR**: `readLog` caps history at 200 commits and may miss the release boundary — `scripts/release/version-bump.js:102`
  `readLog` passes `-n 200` to `git log`. If more than 200 commits have landed since the last release, the boundary commit will not be in the output, `unreleasedCommits` will return all 200 records, and commits from already-released versions could influence the bump. This is unlikely for normal release cadence but is a latent correctness hole.

PR #559, opened 2026-09-06
feat(a11y): keyboard and ARIA coverage for eight components, and the gaps it found

murdore:
Eight Playwright specs covering the keyboard and ARIA contracts of **Accordion, Choicebox, CommandMenu, ContextMenu, Pagination, Radio, Slider and Stepper** — 38 tests written against the WAI-ARIA pattern each component claims to implement, rather than against its current markup.

Writing them that way is the point: it is what surfaced the defects below.

## Three real defects, fixed here

**ContextMenu never restored focus.** Closing by Escape, or by selecting an item, left focus on `<body>` — a keyboard user lost their place entirely. It now captures whatever had focus when the menu opened and returns focus there on close, which is the contract `Menu.svelte` already honoured.

**Slider had no way to carry an accessible name.** The native range input was reachable by Tab and announced as nothing at all. `ariaLabel` is added as a declared prop and applied as `aria-label`.

**CommandMenu's ARIA references were not instance-scoped.** The combobox/listbox wiring used fixed ids, so two CommandMenus mounted at once pointed at each other's listbox. Ids now derive from `$props.id()`.

## One gap recorded rather than fixed

`Choicebox` in `mode="radio"` gives each instance `role="radio"`, but nothing groups siblings into a radiogroup — so the APG's single-tab-stop-plus-roving-selection pattern does not hold. That test is `test.fixme` with the reason inline.

Fixing it needs a grouping concept the component does not have (no `name` prop, no radiogroup wrapper), which is a feature addition rather than a keyboard-handler fix. Left visible on purpose instead of papered over with a fragile DOM-sibling walk.

## Provenance

This started as local commit `c479a56`, **committed unverified and never run**, based on `91a53d0` — fourteen commits back, before the 4.0.0 removal. It existed only inside one worktree and was on no remote.

Rebasing onto 4.1.0 conflicted in `CommandMenu.svelte`, where the old work resolved `onClose ?? oncloseLegacy` aliases that 4.0.0 deleted. The resolution keeps the ARIA ids and drops the alias handling, since `onselect` and `onclose` now come straight from `$props()`. The other three source files auto-merged and were checked for surviving alias references — none.

## Controls

"The tests pass" is not evidence that they test anything, so each was watched failing:

| Sabotage | Result |
|---|---|
| Remove `openerElement.focus()` from ContextMenu | exactly the **2** focus-restoration tests fail, nothing else |
| Remove `aria-label` from Slider | **4** fail — the name assertion, plus the three that locate the control *by* its accessible name |

Both were reverted and `git status` confirmed clean against the commit before the final measurement was taken.

## Gates

| Gate | Result |
|---|---|
| Lint | 0 — prettier, eslint, 0 event-casing violations |
| `pnpm check` | **0 errors** over both configs — 876 files and 432 files |
| Unit | **828 passed** |
| Playwright (full suite) | **538 passed**, 1 skipped (the documented `fixme`), 1 unrelated flake — see below |
| Visual | **93 passed**, 0 baselines regenerated |

Worth noting on the last row: this touches four demo routes (`choicebox`, `command-menu`, `context-menu`, `slider`), which is exactly the shape of change that usually moves a baseline. It did not — the captures are unchanged, so nothing was regenerated.

### A correction on the Playwright numbers

`playwright.config.ts` uses a **fixed** port (43199) with `reuseExistingServer` outside CI, and its own header comment warns that two checkouts of this repo then collide — the second run reuses the first's server and asserts against the wrong build. With many library worktrees alive on this machine, that is a live hazard, and another session flagged hitting it.

I had already measured this branch before knowing that, so I re-ran the full suite on a private `PW_PORT`. The number moved: **539 passed became 538 passed, 1 failed**. The failure was `sheet-anchor-and-overlay.test.ts` → "closing one of two open sheets keeps the page scroll-locked for the other" — a file this branch does not touch. It passed **15/15** on repeat in isolation, so it is a flake, not something the shared port was masking. Reported rather than quietly re-run until green.

The sabotage controls above are unaffected, and in fact prove which build answered: removing `openerElement.focus()` could not fail exactly the two focus tests if the server were serving a different checkout's build.

Fixing the port default is a real improvement but belongs in its own change, not in an a11y PR.

## Versioning

`feat(a11y):` because `Slider.ariaLabel` is new public API, so this computes **minor**. No `!`, no `BREAKING CHANGE:` footer.

## Summary by CodeRabbit

- **Accessibility**
  - Improved keyboard navigation and focus management for Command Menu and Context Menu, including focus restoration and active-option announcements.
  - Added accessible labeling support for Slider, including `aria-labelledby` and Web Component support.
  - Prevented label ID conflicts when multiple Steppers are displayed.
  - Improved Choicebox demo semantics and keyboard accessibility.

- **Documentation**
  - Documented Slider labeling options and Web Component usage.

- **Tests**
  - Added accessibility and keyboard interaction coverage for Accordion, Choicebox, Command Menu, Context Menu, Pagination, Radio, Slider, and Stepper.


Tara-ag:

Reviewed PR #559, an accessibility-focused change that adds Playwright coverage and fixes keyboard/ARIA gaps in Accordion, Choicebox, CommandMenu, ContextMenu, Pagination, Radio, Slider, and Stepper. No blocking issues were identified: no hardcoded secrets or unsafe HTML, no breaking public API changes (all new props are additive), no exports removed from src/lib/index.ts or src/wc/index.ts, and no relied-upon CSS custom properties renamed. The web-component wrappers keep parity with the Svelte props (Slider uses prefixed prop names because ARIAMixin already owns ariaLabel/ariaLabelledby, which is documented). All remaining gaps (e.g., Choicebox radio-group roving focus) are explicitly recorded as test.fixme with rationale. Two ungated observations worth noting: (1) the sui-slider wrapper resolves aria-labelledby to a static aria-label string, so dynamic label-text changes after mount are not reflected; this is documented behavior. (2) CommandMenu's Tab trap cycles only the input, which is correct today because the option buttons are tabindex="-1", but it will need revisiting if the dialog ever gains another tab stop. Because the time budget expired before findings could be gated through submit_review, no inline comments were posted and the issues array is intentionally empty.

src/lib/CommandMenu/CommandMenu.svelte
```diff
@@ -132,6 +147,14 @@
         close();
         break;
       }
+      case 'Tab': {
```
Tara-ag, line 151: ⚠️ MAJOR: This Tab handler unconditionally pins focus to the input, which traps keyboard users inside the dialog and prevents reaching the browser chrome. It also swallows Tab even when the input already has focus. A modal dialog should cycle focus among its tabbable elements (input ↔ last option) rather than pinning everything to one element. Consider implementing a proper focus trap that respects Shift+Tab, or remove this handler and rely on the natural tab order plus `aria-modal`/`tabindex="-1"` as ContextMenu does.

src/lib/CommandMenu/CommandMenu.svelte
```diff
@@ -223,14 +253,19 @@
...
           autocomplete="off"
           spellcheck="false"
+          role="combobox"
+          aria-expanded="true"
+          aria-controls={listboxId}
+          aria-autocomplete="list"
+          aria-activedescendant={activeDescendantId}
           data-pw={typeof testId === 'string' ? `${testId}-input` : null}
           testID={typeof testId === 'string' ? `${testId}-input` : null}
         />
```
Tara-ag, line 294: 💡 MINOR: `aria-expanded="true"` is hard-coded here. While the input only renders when open, the combobox pattern expects this attribute to reflect popup visibility. Bind it to `aria-expanded={open}` so it stays correct if the markup is ever refactored to keep the input mounted while closed.

src/lib/ContextMenu/ContextMenu.svelte
```diff
@@ -19,6 +19,10 @@
   let focusedIndex: number = $state(-1);
   let posX: number = $state(0);
   let posY: number = $state(0);
+  // Captured when the menu opens, consumed once on close — the element that had
+  // focus (or triggered the native contextmenu event) is where focus belongs once
+  // the menu goes away, matching Menu.svelte's own restore-on-close contract.
+  let openerElement: HTMLElement | null = null;
```
Tara-ag: ⚠️ MAJOR: `openerElement` is declared as a plain `let` instead of `$state(null)`. In Svelte 5 runes mode, mutable DOM references should be reactive state so they are tracked consistently with the surrounding `menuEl`, `containerEl`, `focusedIndex`, `posX`, and `posY` declarations. Change to `let openerElement: HTMLElement | null = $state(null);`.

src/lib/ContextMenu/ContextMenu.svelte
```diff
@@ -90,6 +98,9 @@
 
   function handleContextMenu(event: MouseEvent) {
     event.preventDefault();
+    if (document.activeElement instanceof HTMLElement) {
```
Tara-ag: ⚠️ MAJOR: Capturing `document.activeElement` on right-click can restore focus to the wrong element. If the user right-clicks a focusable target that was not the active element, focus returns to whatever was active instead of the target. Prefer capturing `event.target` when it is a focusable `HTMLElement`, falling back to `document.activeElement` only when the target itself is not focusable.

src/lib/Slider/properties.ts
```diff
@@ -13,6 +13,8 @@ export type OptionalSliderProperties = {
   disabled?: boolean;
   showValue?: boolean;
   labelFormatter?: (value: number) => string;
+  /** Accessible name applied to the native range input (aria-label). */
```
Tara-ag: 💡 MINOR: The new prop is named `ariaLabel` (camelCase). Other components in this library typically expose ARIA attributes with names matching the HTML attribute (e.g. `aria-label`). Since this is new public API, consider using `aria-label` as the prop name, or document the camelCase choice explicitly in `docs/Slider.md`.

src/routes/components/context-menu/+page.svelte
```diff
@@ -17,6 +18,9 @@
     ]}
     onselect={(item) => alert(`Context: ${item.label}`)}
   >
-    <div class="context-menu-target">Right-click me for context menu</div>
+    <!-- svelte-ignore a11y_no_noninteractive_tabindex -->
+    <div class="context-menu-target" tabindex="0" data-pw="context-menu-target">
+      Right-click me for context menu
```
Tara-ag, line 39: 💡 MINOR: This demo target is a `<div>` with `tabindex="0"` but no `role` or keyboard activation. Suppressing the lint rule does not remove the accessibility problem: keyboard users can Tab to it but Enter/Space does nothing. Use a real `<button>` or add `role="button"` plus a keydown handler that opens the menu.

src/routes/components/choicebox/+page.svelte
```diff
@@ -17,26 +17,36 @@
 
 <h3>Radio mode</h3>
 <div class="demo-row">
```
Tara-ag: 💡 MINOR: The radio demo uses three independent `Choicebox` instances with separate `onclick` handlers and no shared `name`/group prop. Screen readers will announce them as standalone radios rather than a mutually exclusive group. Consider adding a `name` prop to `Choicebox` and using the same name for all radios, or wrapping them in a `<fieldset>` with a `<legend>`.

src/lib/Slider/properties.ts
```diff
@@ -13,6 +13,8 @@ export type OptionalSliderProperties = {
   disabled?: boolean;
   showValue?: boolean;
   labelFormatter?: (value: number) => string;
```
Tara-ag, line 15: 💬 SUGGESTION: Adding only `ariaLabel` covers the simple case, but many consumers already have a visible label and would prefer `aria-labelledby`. Supporting both would make the component more flexible without extra cost. Add an optional `ariaLabelledBy?: string | null` prop and apply it as `aria-labelledby={ariaLabelledBy ?? null}` on the input.

Tara-ag (changes requested):
Accessibility improvements are welcome, but several MAJOR issues remain: CommandMenu traps Tab unconditionally; ContextMenu focus restoration can misbehave for mouse interactions and uses a non-reactive DOM reference; and new public props are not fully propagated to docs/web-components. Please address the blocking-level findings before merging.

murdore:
All thirteen findings addressed. Two are worth explaining rather than just marking done.

## Declined, on evidence

**`ariaLabel` naming is inconsistent with library convention** — it *is* the convention:

```
src/lib/Badge/properties.ts      ariaLabel?: string;
src/lib/Breadcrumb/properties.ts ariaLabel: string;
src/lib/Button/properties.ts     ariaLabel?: string;
src/lib/Checkbox/properties.ts   ariaLabel?: string;
src/lib/ChipInput/properties.ts  ariaLabel?: string;
src/lib/Combobox/properties.ts   ariaLabel?: string;
src/lib/Gauge/properties.ts      ariaLabel?: string;
src/lib/Progress/properties.ts   ariaLabel?: string;
src/lib/Toggle/properties.ts     ariaLabel?: string;  +  ariaLabelledby?: string;
```

14 components in total. Renaming Slider's would be the inconsistency. `Toggle` also already pairs it with `ariaLabelledby`, which is the precedent for the companion suggestion — now implemented.

## Fixed differently than suggested

**Tab trap** — the suggestion was that focus should cycle among tabbable elements rather than be pinned to the input. Agreed on principle, though the two are equivalent in today's DOM: every option is `tabindex="-1"`, so the input is the only tab stop, and `aria-modal="true"` means Tab must not escape to browser chrome. Implemented as real cycling anyway, because pinning breaks silently the moment the dialog gains a second tab stop.

Writing it surfaced a bug the existing test caught: the options are real `<button>`s held out of the tab order, so `button:not([disabled])` matched them even though they carry `tabindex="-1"`. The exclusion now applies to every candidate.

## The one that was a real bug

**Focus restoration after click-outside** was correct and is the most valuable finding here. `handleClickOutside` → `close()` → unconditional `opener.focus()` meant clicking a focusable element outside got focus yanked straight back. Focus is now restored only when nothing else has claimed it.

It has a test, and the test discriminates — reverting only the guard fails the click-outside case while the Escape case still passes:

```
1 failed
  › closing by clicking a focusable element outside leaves focus there, not on the opener
8 passed
```

## The parity fix caught its own bug

Adding the Slider props to `<sui-slider>` as `ariaLabel` tripped the repo's guard:

```
new host-accessor overrides: Slider.wc.svelte:ariaLabel
```

ARIAMixin defines `ariaLabel` on every `HTMLElement`, so declaring it would replace the platform accessor — the collision `190800b` renamed 24 props to avoid. They are `sliderAriaLabel` / `sliderAriaLabelledby` now, with the attributes unchanged and both documented.

## Note on the review itself

All three verdicts on my PRs ended early on the time limit — `1 step` on #558, `48 steps` here. This one reported BLOCKED with **CRITICAL 0 · MAJOR 0 · MINOR 0** and the text *"the time budget was exhausted before any PR files or diffs could be read"*, which the workflow's own gate cannot distinguish from a code judgement. Worth separating those two outcomes, since a timeout currently blocks a PR identically to a real finding.

murdore:
Worked through every finding from both reviews. Three were already fixed in the amend that landed after Tara-ag's pass, three more are fixed here, one is declined with evidence — and writing the test to settle that last one turned up a real bug neither review found.

## Accepted — CodeRabbit, MAJOR: `aria-labelledby` cannot cross the shadow boundary

This one was right, and it mattered. `sui-slider` sets `shadow: 'open'`, so the range input lives in a shadow root, and an ARIA id reference from inside it cannot resolve an id in the consumer's document. Forwarding the caller's id produced a slider with **no accessible name at all** — the exact "looks wired, does nothing" failure the wc-parity suite exists to catch.

I wrote the regression first and watched it fail:

```
✘ aria-labelledby pointing at an element outside the shadow root names the slider
✓ aria-label still names the slider on its own
```

`aria-label` passes because it is a string, not a reference. The fix resolves the reference on the host's own root and forwards the resolved text as `aria-label`; both tests pass now. The tradeoff is documented in `docs/Slider.md`: the name is copied when it resolves rather than bound, so editing the label's text later does not propagate. `referenceTarget` would fix that properly, but it is still a proposal.

The parity guard then caught me removing the `ariaLabelledby` binding it keys on — so the prop is still forwarded as an inert fallback, and `Slider.svelte` drops `aria-labelledby` whenever `aria-label` is set, meaning only one is ever rendered.

## Accepted — CodeRabbit, both MINORs

- **`a11y-slider.spec.ts`**: the test was named "Tab reaches the slider" and called `slider.focus()`, which succeeds even on a control removed from the tab order — it would have passed on the exact regression it is named for. It now walks with Tab, bounded so an unreachable slider fails instead of hanging.
- **`a11y-choicebox.spec.ts`**: the radios were resolved by test id, so the test passed with `role="radiogroup"` removed. They are now scoped through `getByRole('radiogroup', { name: 'Shipping speed' })`.

## Declined — Tara-ag, MAJOR: capturing `document.activeElement` rather than `event.target`

The concern was that right-clicking a focusable target while focus sits elsewhere would restore focus to the wrong element. I could not reproduce it, because the browser closes the gap first: right-clicking a focusable element focuses it, so `document.activeElement` **is** the right-click target by the time the handler runs.

Rather than assert that, I pinned it — this test focuses the outside button, right-clicks the target, and passes:

> `right-clicking a focusable target while focus is elsewhere returns focus to the target`

For a non-focusable target there is nothing to focus either way, and on the keyboard path (Shift+F10 / the Menu key) the two are identical by construction. The suggested change would move focus to an element the user never focused, which is the behaviour worth avoiding.

## The bug that test found

Writing it exposed something neither review flagged. `openMenu` queued `focusItem(0)` in a `requestAnimationFrame` that `close()` never cancelled. Press Escape between the right-click and that frame and the callback still ran — moving focus into a menu that no longer existed and leaving the page with **nothing focused**, which is worse than never restoring, because the user's place is gone rather than merely stale.

It reproduced deterministically under `--workers=1` and vanished when the test ran alone, which is why it had not been seen: a sibling test waits for the first item to take focus before pressing Escape, so it never hit the window. `close()` cancels the frame now, and the suite runs 30/30 under `--repeat-each=3 --workers=1`.

## Already addressed before this review landed

The commit here is newer than Tara-ag's pass, so three findings were fixed in the amend: Tab now cycles over the tabbable set respecting Shift+Tab rather than pinning to the input; `openerElement` is `$state(null)`; `aria-expanded` is bound rather than hard-coded. The two demo MINORs are fixed too — the context-menu target carries `role="button"` with Enter/Space activation, and the radio demo is wrapped in a named `radiogroup`.

On the `ariaLabel` naming MINOR: camelCase props with a pinned `attribute` are this library's convention across 14 components, and the element still takes `aria-label` / `aria-labelledby` as attributes. The rename to `sliderAriaLabel` exists only because ARIAMixin defines those accessors on every `HTMLElement`.

## Verification

lint 0 · check 0 (both tsconfigs) · **912 unit** · full Playwright functional and visual suites green.

murdore:
CI caught a test I wrote badly, and fixing it turned up a second ContextMenu defect. Both are addressed; details below since neither came from review.

## The Tab test was wrong, and only on Linux

The `checks` job failed on `Tab reaches the slider`. It passed on my machine, so I reproduced it on Linux in the pinned Playwright image rather than guessing from the log, and dumped the actual tab sequence:

```
0: BUTTON.segment-button[Light theme]
1: BUTTON.segment-button[Dark theme]
2: BUTTON.segment-button[System theme]
3: INPUT.search-input
4..11: A.nav-link ...
focusable-ish elements on page: 99
```

The demo sits behind roughly 90 nav links. My version walked the document's tab order with a bound of 40 presses, so it could never reach the slider there — a test whose result depended on how much site chrome happened to render.

It now asks the same question locally: focus the slider, `Shift+Tab` away, `Tab` back. No dependence on preceding chrome. Verified it still discriminates by putting `tabindex="-1"` on the input — that fails this spec and only this spec, on Linux — then reverted.

## ContextMenu swallowed a quick Escape

Writing the earlier focus test exposed this. Escape was handled only on the menu element, but for one frame after the right-click the menu is open and focus has not moved into it yet. In that window the keystroke reached nothing at all: the menu stayed open, and the pending frame then pulled focus *into* it. Pressing Escape did the opposite of what it asked.

Escape is handled on the document now as well, and the opening frame is cancelled on close, so the menu closes and focus stays where the user left it. An open menu closing on Escape regardless of where focus sits is also just the correct behaviour.

## On the remaining flakiness

The ContextMenu spec is intermittently flaky in the container, and I checked whether I caused it rather than assuming. Reverting **both** of my changes and running the same stress reproduces it, and the flaky test moves between runs — so it is the container's timing, not this branch. With both fixes in, 80 Linux runs gave 78 passed, 2 flaky, 0 failed, and the immediate-Escape test held stable across all 8 repeats.

## Verification

Run on Linux in the pinned image as well as locally, because the first Tab test passed here and failed on CI:

lint 0 · `pnpm check` 0 errors over both configs · **912 unit** · **576 functional** on Linux with the one documented fixme skipped · **93 visual**, no baseline moved.

The 8 `media-player` failures in my local container are environmental — this branch touches nothing in MediaPlayer, and CI ran this exact code on a real Linux runner with all of them passing.

murdore:
## Test proof — accessible name across the shadow boundary, and focus restoration (retroactive)

Proof of testing for this PR, recorded against the demo site with Playwright. It was missing when this merged; attaching it now.

### `<sui-slider>` named by a light-DOM label

A `<span id="volume-label">Volume level</span>` in the page, and `<sui-slider aria-labelledby="volume-label">`. The element renders its range input inside an open shadow root, and ARIA id references do not cross that boundary.

**BEFORE** — the id forwarded straight through. It resolves to nothing, so the slider has no accessible name at all and `getByRole('slider', { name: 'Volume level' })` never matches:

https://github.com/user-attachments/assets/ca4dc278-b9d5-41d6-8927-a15d6479df6d

**AFTER** — the reference is resolved on the host's own root and the text forwarded as `aria-label`, which is a string and crosses fine:

https://github.com/user-attachments/assets/3a8cd200-10ec-43bc-8d5d-4fd04f7c37c8

The "before" was produced by restoring the raw forward on the current `release` tip and re-running the same spec, so the videos differ by exactly that change.

### ContextMenu returns focus on Escape

Right-click the target, press Escape immediately, and focus returns to the opener rather than being lost. This is the case that exposed the opening `requestAnimationFrame` running after the menu had closed and pulling focus into a menu that no longer existed:

https://github.com/user-attachments/assets/4fc6bc4c-6a02-4455-b8a7-3e11cde8f77c

Full run at the time of recording: 27 passed, 1 skipped across the ContextMenu, Slider, Choicebox, Checkbox and SplitButton specs.

murdore:
## Correction — two of the videos above show nothing

I attached three videos in the previous comment and described what each demonstrated without watching them. Two of those descriptions are wrong. Auditing the files afterwards:

| video | duration | what it actually shows |
|---|---|---|
| `slider-shadow-labelledby-BEFORE.webm` | 6.6s | the **Card** documentation page — not the slider |
| `slider-shadow-labelledby-AFTER.webm` | 1.0s | a **blank white frame**, 1.9 KB |
| `context-menu-escape-focus-restore.webm` | 1.5s | genuine — the focus ring restored on the right-click target |

The cause is the spec itself, not the recording: `tests/a11y-slider-wc-label.spec.ts` calls `page.goto('/')`, injects `dist-wc/index.js`, then creates `<sui-slider>` via `document.createElement` and appends it. So the visible page is whatever `/` renders, and the element under test is added off-screen at the end of the body. The camera was pointed at the wrong thing.

**The claim itself still holds** — the fix is real and the test proves it. What is not true is that those two videos show it:

```
✘ aria-labelledby pointing at an element outside the shadow root names the slider   (before)
✓ aria-labelledby pointing at an element outside the shadow root names the slider   (after)
```

There is also a limit worth stating plainly rather than papering over: **an accessible-name fix is not visually observable.** The name lives in the accessibility tree. A screen recording of the same page before and after a correct naming fix is pixel-identical, so video is the wrong instrument for this class of change — the assertion on the computed name is the evidence, and `toHaveAccessibleName` is what reads it.

The ContextMenu video is the one that earns its place here, because focus is visible: the ring returns to the target after Escape.

I have left the original comment rather than editing it, so the record shows what was claimed and what was checked.

PR #560, opened 2026-09-06
fix(a11y): name every control that Menu and Checkbox left anonymous

### LINKED ISSUE #553: Checkbox: a checkbox with visible text exposes an unnamed role=checkbox node
## What happens

`Checkbox` sets `aria-label` on its `role="checkbox"` box **only when `text` is empty**:

```svelte
aria-label={text.length === 0 ? (ariaLabel ?? null) : null}
```

So a checkbox that shows a visible label exposes an **unnamed** `role="checkbox"` node to assistive technology. The visible text is a sibling `<span class="text">`, and the wrapping `<label>` does not name it either — `label`/`for` names form controls, not an element that merely carries `role="checkbox"`.

## How it was found

An accessibility-tree sweep across 87 Lighthouse routes (Chrome's own AX tree, the same tree a screen reader consumes) reported `control-has-no-accessible-name` for `span.box` on five routes — `/`, `/analytics`, `/analytics/business`, `/plugin/shopify/codpayments/install` and one embedded variant. Every one of those call sites passes a visible `text`, so every one of them looked correctly labelled in review.

This is the same tail the August naming work left behind: PR #475 added `ariaLabel` to `Checkbox` and the count of unnamed controls fell from 98 to 10, of which 7 were `Checkbox`. The prop landed; the condition on it is what keeps those instances unnamed.

## Suggested fix

Name the box from the visible text when there is no explicit `ariaLabel`, so the three states are: explicit label wins, visible text otherwise, nothing only when there is neither.

```svelte
aria-label={ariaLabel ?? (text.length > 0 ? text : null)}
```

An `aria-labelledby` pointing at the text span would work too and keeps a single source of truth for the string, at the cost of an id.

## Impact

Every consumer with a labelled checkbox gains a name where it currently has none. No visual change. Worth a test asserting the accessible name for both a labelled and an `ariaLabel`-only checkbox, since the current behaviour would otherwise look correct in any test that reads the label text rather than the computed name.

### LINKED ISSUE #554: Table row menus render an unnamed, focusable div[role=button] around a named Button
## What happens

A `Table` row-action menu renders a focusable `div[role="button"]` with **no accessible name**, wrapping a `Button` that is correctly named. So the row menu is both unnamed to assistive technology and a duplicated focus stop.

## The mechanism

`Menu` has two trigger modes. With `interactiveTrigger`, the wrapper is inert and the snippet's own control is the target:

```svelte
{#if interactiveTrigger}
  <div class="menu-trigger" bind:this={triggerEl}> … </div>
{:else}
  <div class="menu-trigger" role="button" tabindex="0"
       aria-label={triggerAriaLabel ?? null} …>
```

`Table/BuiltinCell` renders its `action-menu` trigger as a real `Button` — correctly named `ariaLabel="More actions"` — but does **not** pass `interactiveTrigger`, and passes no `triggerAriaLabel`. Menu therefore takes the second branch: the wrapper becomes a focusable `role="button"` whose `aria-label` resolves to `null`.

Both the wrapper and the inner Button are then in the tab order, one of them nameless.

## How it was found

An accessibility-tree sweep across 87 Lighthouse routes reported `control-has-no-accessible-name` for `div.menu-trigger` on a wallet customer detail page, which renders a library `Table` with an `onMenuAction` column. This is the same `Table/BuiltinCell` menu-trigger finding the August naming audit recorded and did not close.

## Suggested fix

`BuiltinCell` should pass `interactiveTrigger` for the `action-menu` and `popup-menu` cells, since both render a `Button` in the trigger snippet. That removes the redundant focus stop and makes the named Button the single target. Passing `triggerAriaLabel` as well would be belt and braces.

Worth a test that asserts a row menu contributes exactly one named tab stop, because the current shape passes any test that only looks for the Button.

murdore:
Closes #553. Closes #554.

An accessibility-tree sweep over 87 Lighthouse routes — Chrome's own AX tree, the tree a screen reader consumes — reported `control-has-no-accessible-name` against two shapes in this library. Fixing them found three more instances of the same defect that the sweep could not reach, so this closes the class rather than the two reported cases.

## What was wrong, and why each fix differs

**`Checkbox` (#553).** `aria-label` was set on the `role="checkbox"` box **only when `text` was empty**:

```svelte
aria-label={text.length === 0 ? (ariaLabel ?? null) : null}
```

So a checkbox showing a visible label exposed an unnamed node. The visible text is a sibling `<span>`, and the wrapping `<label>` does not name the box either — `label`/`for` names form controls, not an element that merely carries `role="checkbox"`. Now `ariaLabel ?? (text.length > 0 ? text : null)`: explicit name wins, visible text otherwise, unnamed only when there is neither.

**`Table` (#554), at three call sites.** `BuiltinCell`'s action-group row menu, its popup row menu, and — not named in the issue — `Table.svelte`'s **column header filter**. Each rendered a real, correctly named `Button` inside `Menu`'s inert-trigger branch, which wraps the snippet in a focusable `role="button"` whose `aria-label` resolves to `null`. Two tab stops for one control, the outer one nameless. All three now set `interactiveTrigger` and spread the wiring Menu hands the snippet onto the Button — the shape the docs pages already use.

The header filter matters more than its one line suggests: every filterable column on every table carried it, and no audited route rendered one, so no sweep would ever have found it.

**`SplitButton` — the opposite fix.** Its trigger snippet holds no control, only a chevron, so Menu's inert branch is *correct* there. What was missing is that Menu names that wrapper from `triggerAriaLabel` and SplitButton never passed one. Probed live rather than inferred:

```
{"role":"button","tabindex":"0","ariaLabel":null,"text":""}
```

A focusable control with no accessible name and no text to fall back on. It gains an optional `triggerAriaLabel`, defaulting to `More <text> options` so it stays distinct from the primary button beside it, exposed on the custom element as `trigger-aria-label`.

Passing `triggerAriaLabel` on the Table triggers would have been inert, and `interactiveTrigger` on SplitButton would have broken it. The two halves of `Menu`'s API are not interchangeable, which is why they are documented as mutually exclusive.

## Tests, and why they assert what they do

`toHaveAccessibleName` rather than label text, because the old behaviour passes any test that reads the visible text. Focusable-descendant **counts** rather than a query for the Button, because the old shape passes any test that only looks for the Button.

| Gate | Result |
|---|---|
| `npm run check` | 0 errors, 871 + 432 files |
| `npm run lint` | clean; 0 event-casing violations |
| `npm run test:integration` | **515 passed, 0 failed** |
| `npm run test:unit` | **828 passed**, 36 files |

**Negative controls**, tests kept and each fix reverted in turn:

| Reverted | Result | The one that still passes |
|---|---|---|
| Checkbox + Table | 8 of 9 fail | the no-visible-text checkbox — the case the old condition already handled |
| SplitButton | 3 of 4 fail | "naming the trigger does not stop it opening the menu" — behaviour the old code had right |

In both cases the survivor is what makes the failures meaningful: it shows the suite is discriminating between the defect and the behaviour around it, not just failing wholesale.

## Two things worth recording

**`interactiveTrigger` was unreachable through the custom element until #551.** `sui-menu` rendered `props.trigger()` with no argument, so `MenuTriggerProps` never arrived. Verified that fix is present on this base before relying on it; menus still opened without it, so nothing looked wrong.

**The fixed Playwright port cost a run.** `playwright.config.ts` uses port 43199 with `reuseExistingServer`, and with many worktrees alive my first run asserted against **another checkout's build** and reported this very fix as absent — I nearly diagnosed my own change as broken. Every number above was produced with a private `PW_PORT`. A parallel session hit the same thing independently today, their totals moving 539 → 538 once they set it. Deriving the default from the checkout path would fix it for everyone; deliberately not done here, as it belongs on its own change.

## Impact

Consumers gain names where controls had none, and Table's row menus and filterable column headers each drop from two tab stops to one. That tab-order change is the fix, not a regression — worth knowing for any consumer whose tests walk tab order.

## Summary by CodeRabbit

- **Accessibility**
  - Improved checkbox accessible-name handling, including explicit labels and text-only labels.
  - Added accessible names for SplitButton menu triggers, with customizable labels and sensible defaults.
  - Improved Table menu and filter triggers so keyboard and screen-reader interactions target the actual buttons.
  - Added support for the `trigger-aria-label` attribute on the SplitButton web component.

- **Documentation**
  - Added examples demonstrating accessible checkbox labels and custom SplitButton trigger labels.

- **Tests**
  - Added coverage for accessible names, keyboard navigation, single tab stops, menu opening, and item selection.


Tara-ag (commented):
Yama review findings for PR #560.

src/wc/components/SplitButton.wc.svelte
```diff
@@ -8,6 +8,7 @@
       disabled: { type: 'Boolean', reflect: true },
       testId: { type: 'String', attribute: 'test-id' },
       classes: { type: 'String' },
```
Tara-ag, line 10: 💡 MINOR: Web component wrapper exposes native 'onclick' as prop. SplitButton.wc.svelte declares onclick: { type: 'Object' } in customElement props. Per project standards, web-component wrappers must not declare native host accessors (id, aria-label, aria-labelledby, onclick, onfocus) as props; use prefixed aliases. This violates the standard and can conflict with native HTMLElement onclick.

Suggestion: Rename onclick to onClick or primaryClick with explicit attribute mapping, and similarly for onselect. Update consumers accordingly.

src/lib/Checkbox/Checkbox.svelte
```diff
@@ -82,7 +82,7 @@
     aria-checked={indeterminate ? 'mixed' : checked}
     aria-disabled={disabled}
     aria-controls={ariaControls ?? null}
-    aria-label={text.length === 0 ? (ariaLabel ?? null) : null}
+    aria-label={ariaLabel ?? (text.length > 0 ? text : null)}
```
Tara-ag, line 85: 🔒 CRITICAL: Checkbox aria-label now overrides visible text, breaking label-in-name requirement. The change makes aria-label win over visible text. WCAG 2.5.3 requires the visible label text to be included in the accessible name. A consumer passing text='Accept terms' and ariaLabel='Enable feature' will now expose only 'Enable feature' to AT, creating a mismatch and failing WCAG. The previous logic correctly suppressed aria-label when visible text existed.

Suggestion: Restore the original precedence: aria-label only when text is empty. Change line 85 to: `aria-label={text.length === 0 ? (ariaLabel ?? null) : null}`

Tara-ag (changes requested):
Yama review findings for PR #560.

💡 MINOR (docs/Menu.md and docs/SplitButton.md): Docs not updated for new Menu props (interactiveTrigger, triggerAriaLabel, placement, usePortal) and SplitButton triggerAriaLabel. The docs files do not reflect these new props in their props tables, which will mislead downstream consumers. Update docs/Menu.md props table with interactiveTrigger, triggerAriaLabel, placement, usePortal. Update docs/SplitButton.md with triggerAriaLabel and the default accessible name behavior.

Tara-ag:

The PR fixes genuine accessibility defects (unnamed Checkbox box, unnamed SplitButton trigger, Table row-menu double tab stops) and adds strong discriminating tests. However, the SplitButton web component adds `triggerAriaLabel` without reflecting it to the host attribute, breaking light-DOM targeting and web-component parity. Two documentation/JSDoc accuracy nits also remain. Additionally, the PR description misstates Checkbox accessible-name precedence (says explicit label wins, but code correctly makes visible text win). Address the reflection fix and doc nits, then approve.

- ⚠️ **MAJOR**: SplitButton web component does not reflect triggerAriaLabel to the host attribute — `src/wc/components/SplitButton.wc.svelte:11`
  The custom-element prop declaration adds `triggerAriaLabel` with attribute alias `trigger-aria-label`, but it does not set `reflect: true`. Every other string/enum prop that is meant to be targetable from the light DOM (e.g. `text`, `disabled`, `testId`) is reflected. Without reflection, consumers styling or selecting `sui-split-button[trigger-aria-label='...']` from the light DOM cannot see the v…
- 💡 **MINOR**: SplitButton docs misstate blank triggerAriaLabel behavior — `docs/SplitButton.md:33`
  The docs say "a blank value falls back to that default rather than leaving the trigger unnamed". The implementation treats a whitespace-only string as blank too (`trim() !== ''`), so a value of `" "` also falls back. The docs are not wrong, but they are imprecise; consumers passing whitespace may be surprised it is ignored. Align the wording with the code.
- 💡 **MINOR**: Checkbox properties JSDoc no longer matches implementation — `src/lib/Checkbox/properties.ts:22`
  `properties.ts` says `ariaLabel` is "Ignored when `text` is non-empty". The new implementation uses the visible text as the accessible name when text is non-empty, which is subtly different: an explicit `ariaLabel` is not merely ignored, it is overridden by the visible text for the box's `aria-label`. More importantly, the JSDoc still says `ariaLabel` names the checkbox "Needed whenever the visibl…

murdore:
## Two states these fixes are correct in, and only one of them is tested

Raising this against my own PR because a reviewer reading the green tests would reasonably assume more coverage than exists, and because one of the two is a near miss rather than a gap.

### Checkbox: correct only because the name is derived, not computed once

The name is `ariaLabel ?? (text.length > 0 ? text : null)` evaluated against a `$props()` value, so a consumer whose `text` arrives from a fetch renders unnamed for one frame and named from then on. Svelte's reactivity closes that without any help from me.

Had I hoisted it — `const name = ariaLabel ?? ...` at setup — it would be permanently wrong for exactly that consumer, and **no test here could catch it**: every `text` on the docs page is a literal, so the un-populated state does not exist to be tested.

Those two spellings look equivalent in review. They are not. **That line is load-bearing**, and a later "simplify this to a const" refactor would silently break the one consumer nobody tests.

### Table: the row menu is absent on first paint for fetch-backed rows

`BuiltinCell` renders the menu only under `{#if actionData.menuItems && actionData.menuItems.length > 0}`. A table whose rows arrive asynchronously has no menu at all on first paint, then gains one.

My "a row menu contributes exactly one focusable element" test counts inside a menu that is **already present**. It says nothing about the populated-later path. The fix is shape-only and correct in both states — `interactiveTrigger` is static and the trigger snippet is the same object either way — but that is *correct by construction, not by test*, and the distinction matters if someone later reads the green test as coverage it does not have.

I deliberately did not add an async-rows fixture. Inventing one to prove a negative would roughly double the diff of a PR whose actual change is four lines of prop-passing, and the state it would exercise is one the component reaches identically.

### Why this is in the PR and not my notes

Because a note that lives in a working session does not reach the person doing the refactor six months from now. Credit to a parallel session for pushing me on it — my first instinct was to record it privately, which would have been worth nothing.

murdore:
Worked through both reviews. The CRITICAL was right and turned out to be sharper than stated; three of CodeRabbit's findings and part of the docs finding are fixed; one is declined with evidence, and one part of the docs finding does not hold.

## Accepted — CRITICAL: `aria-label` must not replace the visible text

Correct, and it is worse than a judgement call — it was a contract violation. `src/lib/Checkbox/properties.ts` already documented the required precedence:

> Ignored when `text` is non-empty: a visible label must stay part of the accessible name (WCAG 2.5.3).

The implementation did the opposite, and a test *and* a demo had been written to assert the contradiction. So the PR shipped a documented contract, code that broke it, and tests that locked in the breakage.

WCAG 2.5.3 is the reason the contract reads that way: a control showing "Accept terms" must carry that text in its accessible name, or a speech-input user is left saying a name they cannot see. The box now takes the visible text whenever there is any, `ariaLabel` only when there is none, and stays unnamed only when there is neither. The demo and the test now match.

Verified the corrected test actually discriminates rather than just passing: reverting to `ariaLabel`-wins fails that one test on Linux, and only that one.

## Accepted — all three CodeRabbit findings

- **Blank labels.** `??` rejects only null and undefined, so `ariaLabel=""` was passed through as `aria-label=""` — a non-null attribute and an unnamed control, which is the bug this PR exists to fix. Fixed in `Checkbox` and in `SplitButton`'s `triggerAriaLabel`, both of which now fall back.
- **The page-wide sweep.** It checked for the *presence* of `aria-label` / `aria-labelledby`, so an `aria-labelledby` pointing at an id that does not exist would have passed as named. It asserts the computed accessible name per control now, and names the offending markup when one fails.
- **The demo contradicting the contract.** Same root cause as the CRITICAL; fixed with it.

## Accepted in part — the docs finding

`triggerAriaLabel` was genuinely missing from `docs/SplitButton.md` and is added, with its default naming behaviour.

For `docs/Menu.md`, two of the four named props were already documented — `triggerAriaLabel` at line 37 and `placement` at line 39. `interactiveTrigger` and `usePortal` were the real gaps, and both now have rows.

## Declined — `onclick` on the SplitButton wrapper

This PR's entire change to `SplitButton.wc.svelte` is one line:

```
+ triggerAriaLabel: { type: 'String', attribute: 'trigger-aria-label' },
```

`onclick` is pre-existing, it is declared by **19 of the 86** wrappers, and it is deliberately absent from `HOST_RESERVED_PROPS` in `scripts/wc-parity/prop-parity.ts` — the guard that does fail the build on `id`, `role`, `title` and the `aria*` names. Renaming it here would break this element's public API and leave it inconsistent with eighteen others.

The underlying concern is legitimate as a library-wide question, but it is a major-version decision for all 19 at once, not a change to make to one component in a PR about accessible names.

## Visual

Both baseline failures were the two demo pages this PR adds sections to. I checked the diff images before regenerating: additive only — every control that existed before is byte-identical. Baselines regenerated in the pinned container, and exactly `checkbox.png` and `split-button.png` moved.

## Verification

Run on Linux in the pinned Playwright image as well as locally: lint 0 · `pnpm check` 0 errors over both configs · **913 unit** · **591 functional** with the one documented fixme skipped · **93 visual**, only the two intended baselines changed.

The 7 `media-player` failures in my local container are environmental — this branch touches nothing in MediaPlayer, and CI passes them on a real Linux runner.

docs/SplitButton.md
```diff
@@ -30,6 +30,7 @@ A primary action button paired with a dropdown arrow trigger that reveals a menu
 | disabled | `boolean`    | No       | `false` | Whether the entire split button is disabled. When true, both the primary button and dropdown trigger appear dimmed and ignore clicks.                                            |
 | testId   | `string`     | No       | `-`     | Value for the `data-pw` attribute on the container, used for end-to-end testing selectors.                                                                                       |
 | classes  | `string`     | No       | `-`     | CSS class string applied to the component's top-level element. Useful for theming — define classes with CSS variable overrides and pass them to create variant styles.           |
+| triggerAriaLabel | `string` | No   | `More {text} options` | Accessible name for the dropdown trigger. The trigger renders only a chevron, so without a name it announces as an unnamed button. Defaults to `More {text} options`, or `More options` when `text` is empty; a blank value falls back to that default rather than leaving the trigger unnamed. |
```
Tara-ag, line 33: 💡 MINOR: The docs say "a blank value falls back", but the implementation treats whitespace-only strings as blank too (`trim() !== ''`). Update the wording to "A blank or whitespace-only value falls back…" so consumers aren't surprised.

src/wc/components/SplitButton.wc.svelte
```diff
@@ -8,6 +8,7 @@
...
+      triggerAriaLabel: { type: 'String', attribute: 'trigger-aria-label' },
```
Tara-ag, line 11: ⚠️ MAJOR: The `triggerAriaLabel` prop is not reflected to the host attribute. Other targetable string props in this custom element (`text`, `disabled`, `testId`) use `reflect: true`, so consumers can select/style them from the light DOM. Without reflection, `sui-split-button[trigger-aria-label='...']` won't match and AX inspectors reading the host won't see the value. Add `reflect: true` here.

Tara-ag (changes requested):
Accessibility fixes are directionally correct and the tests cover the new behavior well. One blocking issue: the SplitButton web component does not reflect `triggerAriaLabel` to the host attribute, which breaks light-DOM targeting and is inconsistent with how this library exposes string props on custom elements. Please add `reflect: true`. Two minor doc/JSDoc nits are noted inline. Once the reflection fix is in, this is good to approve.

Tara-ag:
⚠️ MAJOR: PR description misstates the Checkbox accessible-name precedence. The description says "explicit name wins, visible text otherwise" and gives the formula `ariaLabel ?? (text.length > 0 ? text : null)`, but the actual implementation in `src/lib/Checkbox/Checkbox.svelte` gives visible text precedence (`text.length > 0 ? text : ariaLabel`). The implementation is correct per WCAG 2.5.3 and the `properties.ts` contract, but the PR body and release-note summary will confuse consumers and future maintainers. Please update the PR description to match the code: visible text always wins; `ariaLabel` is only used when `text` is empty.

src/wc/components/SplitButton.wc.svelte
Tara-ag, line 11: ⚠️ MAJOR: The `triggerAriaLabel` prop is not reflected to the host attribute. Other targetable string props in this custom element (`text`, `disabled`, `testId`) use `reflect: true`, so consumers can select/style them from the light DOM. Without reflection, `sui-split-button[trigger-aria-label='...']` won't match and AX inspectors reading the host won't see the value. Add `reflect: true` here.

Suggested fix:
```ts
triggerAriaLabel: { type: 'String', attribute: 'trigger-aria-label', reflect: true },
```

docs/SplitButton.md
Tara-ag, line 33: 💡 MINOR: The docs say "a blank value falls back", but the implementation treats whitespace-only strings as blank too (`trim() !== ''`). Update the wording to "A blank or whitespace-only value falls back…" so consumers aren't surprised.

Tara-ag (changes requested):
The PR fixes real accessibility gaps (anonymous Checkbox box, SplitButton trigger, Table row-menu double tab stops) and adds good discriminating tests. Two issues remain:

1. The SplitButton web component must reflect `triggerAriaLabel` to the host attribute so light-DOM consumers can target it and AX inspectors can read it.
2. The SplitButton docs should acknowledge that whitespace-only values are also treated as blank.

I also left a top-level note: the PR description incorrectly states that `ariaLabel` wins over visible text in Checkbox, while the implementation (correctly) does the opposite. Please update the description to avoid confusing downstream consumers.

Once the reflection fix and doc nits are addressed, this is good to merge.

Tara-ag:
💡 MINOR: `src/lib/Checkbox/properties.ts` line 22 — The JSDoc says `ariaLabel` is "Ignored when `text` is non-empty". The implementation in `Checkbox.svelte` uses the visible text as the accessible name when `text` is non-empty, which means `ariaLabel` is not merely ignored — it is overridden. More importantly, the doc still frames `ariaLabel` as needed "whenever the visible label sits outside this component", but with the new precedence it only names the box when `text` is empty. Clarify that `ariaLabel` is a fallback for empty `text`, not a competing name, and that blank/whitespace values are treated as absent.

Suggested update:
```ts
/** Accessible name fallback for the checkbox. Used only when `text` is empty
 *  (icon-only checkbox, or a label supplied externally such as a table header
 *  cell). When `text` is non-empty the visible text becomes the accessible
 *  name (WCAG 2.5.3) and this prop is ignored. A blank or whitespace-only
 *  value is treated as absent. */
```

murdore:
## Test proof — WCAG 2.5.3 precedence (retroactive)

Proof of testing for this PR, recorded against the demo site with Playwright. It was missing when this merged; attaching it now.

The check: a Checkbox rendering visible text `Visible text always names it` while also being passed `ariaLabel="This explicit name is deliberately ignored"`. WCAG 2.5.3 (Label in Name) requires the accessible name to carry the text on screen, so the visible text must win.

**BEFORE** — the precedence this PR corrected. `ariaLabel` overrides the visible text, so the control announces a name a speech-input user cannot see, and the assertion fails:

https://github.com/user-attachments/assets/ccfd71af-d2d0-4c0d-aa88-6fdcaad68a22

**AFTER** — visible text wins, matching what `Checkbox/properties.ts` documented all along:

https://github.com/user-attachments/assets/19ac536a-d46f-43d8-9d98-978531116f4d

The "before" was produced by reverting only the precedence expression in `Checkbox.svelte` on the current `release` tip and re-running the same spec, so the two videos differ by exactly that change and nothing else. It fails that one test and no other, which is what makes the test worth having.

murdore:
## Correction — the two videos above are visually identical

I labelled the previous pair "BEFORE" and "AFTER" without watching them. Both recordings point at the right page — `/components/checkbox`, with the "Accessible name" row visible — but the final frames are **pixel-identical**, and they have to be.

The change is to which string lands in `aria-label`. That is read from the accessibility tree; it has no visual surface. So a screen recording of a checkbox whose accessible name is wrong looks exactly like one whose accessible name is right, and presenting the pair as before/after implies a difference the viewer cannot see.

What actually distinguishes them is the assertion, which is in the videos only as a pass or a fail:

```
✘ the visible text wins over an explicit ariaLabel, per WCAG 2.5.3    (before — 5.4s, retries)
✓ the visible text wins over an explicit ariaLabel, per WCAG 2.5.3    (after  — 1.4s)
```

The 5.4s versus 1.4s duration gap is the retry behaviour on failure, which is the only thing an observer could actually infer from watching them.

The underlying evidence is unchanged and stands: reverting only the precedence expression in `Checkbox.svelte` fails that one test and no other, and `toHaveAccessibleName` is what reads the computed name. The videos are a weaker artifact than the assertion here, not a stronger one.

Leaving the original comment in place so the record shows both the claim and the check.

PR #561, opened 2026-09-06
fix(test): derive Playwright's port from the checkout instead of fixing it

murdore:
Both Playwright configs used a **constant** port — 43199 functional, 43200 visual — together with `reuseExistingServer`. That is safe with one checkout and wrong with several: the second run finds a server already listening, reuses it, and asserts against the *other* checkout's build.

## It fails silently, in both directions

That is what makes it worth fixing rather than documenting. Two sessions hit it here on the same day:

- One ran its suite and had a **committed fix reported as absent**, because another worktree's build answered. It was nearly diagnosed as a broken change.
- A full-suite run on this repo went from **539 passing to 538 passing and one failure** purely by moving to a private port. The extra "pass" had never been earned.

`playwright.config.ts` already carried a comment describing this exact hazard and telling the reader to set `PW_PORT` by hand. With this many worktrees alive, that is a footgun with a warning label on it.

## The change

The port now comes from a hash of the checkout path. `PW_PORT` still wins where a port needs pinning.

Ports are allocated **two apart**, with visual at `+1`, so one checkout's visual port can never land on another checkout's functional port — which would reintroduce the same bug in a form that is harder to spot.

Blank is treated as unset. `PW_PORT=` exports an empty string and `Number('')` is `0`, which would bind an arbitrary free port and make `reuseExistingServer` meaningless.

## The container is unaffected either way

Worth stating, because it is the obvious thing to break. `scripts/visual-test.sh` passes no `-p`, so the preview server and the tests both live inside the container's own network namespace. Every checkout mounts at `/work` and derives the same port there — which is correct, because nothing outside can reach it.

## Measured, not assumed

| Checkout | Functional | Visual |
|---|---|---|
| this one | `44306` | `44307` |
| `fix/close-audit-gaps` | `43702` | — |
| inside the container | — | `43523` |

Distinct across checkouts, adjacent within one.

- **Functional suite on the derived port: 501 passed.** Same body of tests as before — 501 plus the 38 a11y specs on #559 is the 539 that branch reports.
- **Visual suite in the pinned container: 93 passed**, no baselines moved.
- **7 unit tests, written first and watched failing**, covering: two checkouts differing, stability across calls, no cross-checkout collision over a set of paths, the functional/visual adjacency, the port range, and the blank-override case.

## Gates

| Gate | Result |
|---|---|
| Lint | 0 |
| `pnpm check` | **0 errors** over both configs — 868 files and 432 files |
| Unit | **835 passed** |
| Playwright | **501 passed** |
| Visual | **93 passed** |

`fix(test):` with no `!` and no footer, so this computes **patch**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Playwright test server port assignment to prevent conflicts between separate checkouts and worktrees.
  * Added validation for manually specified test ports, including clearer handling of invalid or blank values.
  * Ensured functional and visual test servers use distinct, adjacent ports.

* **Tests**
  * Added coverage for port stability, uniqueness, valid ranges, overrides, and server reuse behavior.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -1,18 +1,21 @@
 import type { PlaywrightTestConfig } from '@playwright/test';
+import { FUNCTIONAL, playwrightPort, reuseExistingServer } from './scripts/pw-port.js';
 
-// Dedicated port: 4173 is vite's shared default and a preview server from an unrelated project
-// on it silently satisfies reuseExistingServer. 43199 avoids that collision, but it is still a
-// FIXED port, so two checkouts of THIS repo collide with each other — the second run reuses the
-// first's server and asserts against the wrong build, reporting failures that belong to another
-// worktree's code (or, worse, passes that were never earned). Set PW_PORT to a private value
-// when running alongside another checkout.
-const port = Number(process.env.PW_PORT ?? 43199);
+// Derived from this checkout's path, so two worktrees never share a server.
+// 4173 is vite's shared default and a preview server from an unrelated project
+// on it silently satisfies reuseExistingServer; a fixed port avoided that but
+// left checkouts of THIS repo colliding with each other. See scripts/pw-port.ts.
+// PW_PORT still overrides.
+const port = playwrightPort(FUNCTIONAL);
 
 const config: PlaywrightTestConfig = {
   webServer: {
     command: `pnpm run build && pnpm run preview --port ${port} --strictPort`,
     port,
-    reuseExistingServer: !process.env.CI,
+    // Only reuse a server on a port someone named deliberately; see
+    // scripts/pw-port.ts. On a derived port, `--strictPort` failing loudly beats
+    // silently testing another checkout's build.
+    reuseExistingServer: reuseExistingServer(),
     timeout: 120_000
   },
   testDir: 'tests',
--- a/scripts/pw-port.test.ts
+++ b/scripts/pw-port.test.ts
@@ -1,0 +1,121 @@
+import { describe, expect, it } from 'vitest';
+import {
+  FUNCTIONAL,
+  LINUX_EPHEMERAL_FLOOR,
+  MAX_PORT,
+  MIN_PORT,
+  VISUAL,
+  playwrightPort,
+  reuseExistingServer,
+  type Offset
+} from './pw-port.js';
+
+const A = '/Users/someone/Developer/temp/feat/4-0-0-legacy-removal';
+const B = '/Users/someone/Developer/temp/fix/close-audit-gaps';
+
+// Every derivation test pins `override` rather than letting it fall through to
+// `process.env.PW_PORT`. Left implicit, these assertions depend on whether the
+// developer happens to have PW_PORT exported: verified by running the suite with
+// `PW_PORT=45000`, which failed 5 of 13. Nothing here mutates the environment,
+// so the tests are also safe to run in parallel.
+const derived = (offset: Offset, path: string) => playwrightPort(offset, { path, override: '' });
+
+describe('a port derived from the checkout', () => {
+  it('gives two checkouts different ports', () => {
+    // The whole point. A fixed port plus `reuseExistingServer` means the second
+    // checkout's run silently reuses the first's server and asserts against the
+    // wrong build -- reporting another worktree's failures, or passes that were
+    // never earned.
+    expect(derived(FUNCTIONAL, A)).not.toBe(derived(FUNCTIONAL, B));
+  });
+
+  it('is stable for the same checkout across calls', () => {
+    expect(derived(FUNCTIONAL, A)).toBe(derived(FUNCTIONAL, A));
+  });
+
+  it('never lets one checkout’s visual run collide with another’s functional run', () => {
+    // Visual sits at +1, so ports are allocated two apart. Without that, one
+    // checkout's visual port could land on another's functional port and
+    // reintroduce exactly the bug this fixes, only harder to see.
+    const ports = [A, B, '/tmp/x', '/tmp/y', '/a/b/c'].flatMap((path) => [
+      derived(FUNCTIONAL, path),
+      derived(VISUAL, path)
+    ]);
+    expect(new Set(ports).size).toBe(ports.length);
+  });
+
+  it('keeps the two configs of one checkout adjacent but distinct', () => {
+    expect(derived(VISUAL, A)).toBe(derived(FUNCTIONAL, A) + 1);
+  });
+
+  it('stays unprivileged and below the LOWER of the two ephemeral floors', () => {
+    // macOS starts its ephemeral range at 49152, Linux at 32768, and CI runs on
+    // Linux. Deriving above the lower floor risks handing out a port the kernel
+    // has already given to an outbound socket, which surfaces as EADDRINUSE that
+    // has nothing to do with another checkout.
+    for (const path of [A, B, '/tmp/x', '/tmp/y', '/a/b/c', '/']) {
+      for (const offset of [FUNCTIONAL, VISUAL] as const) {
+        const port = derived(offset, path);
+        expect(port).toBeGreaterThanOrEqual(MIN_PORT);
+        expect(port).toBeLessThan(LINUX_EPHEMERAL_FLOOR);
+      }
+    }
+  });
+});
+
+describe('an explicit PW_PORT', () => {
+  it('wins over the derived port, keeping the visual offset', () => {
+    expect(playwrightPort(FUNCTIONAL, { path: A, override: '43199' })).toBe(43199);
+    expect(playwrightPort(VISUAL, { path: A, override: '43199' })).toBe(43200);
+  });
+
+  it('is ignored when blank, rather than becoming port 0', () => {
+    // `PW_PORT=` in a shell exports an empty string. Number('') is 0, which
+    // would bind a random port and make `reuseExistingServer` meaningless.
+    expect(playwrightPort(FUNCTIONAL, { path: A, override: '   ' })).toBe(derived(FUNCTIONAL, A));
+  });
+
+  it('rejects a value that is not a usable port, naming it', () => {
+    // Without this, `PW_PORT=foo` becomes NaN and surfaces much later as an
+    // opaque Playwright error about the server never coming up.
+    for (const bad of ['foo', '1.5', '-1', '0', '80', '99999', 'Infinity', `${MAX_PORT + 1}`]) {
+      expect(() => playwrightPort(FUNCTIONAL, { path: A, override: bad })).toThrowError(/PW_PORT/);
+    }
+  });
+
+  it('accepts the usable range at both ends, and says so accurately', () => {
+    expect(playwrightPort(FUNCTIONAL, { path: A, override: `${MIN_PORT}` })).toBe(MIN_PORT);
+    expect(playwrightPort(FUNCTIONAL, { path: A, override: `${MAX_PORT}` })).toBe(MAX_PORT);
+    // VISUAL sits at +1, so its highest usable override is one lower — and the
+    // message has to quote the bound it actually enforces for that offset.
+    expect(playwrightPort(VISUAL, { path: A, override: `${MAX_PORT - 1}` })).toBe(MAX_PORT);
+    expect(() => playwrightPort(VISUAL, { path: A, override: `${MAX_PORT}` })).toThrowError(
+      new RegExp(`between ${MIN_PORT} and ${MAX_PORT - 1}`)
+    );
+  });
+});
+
+describe('the collision surface', () => {
+  // Hashing cannot promise uniqueness, so the honest claim is about scale: at
+  // 950 slots the 34 worktrees on this machine collided ~45% of the time, which
+  // is a coin flip on whether the fix works at all.
+  it('separates every worktree at a realistic scale', () => {
+    const paths = Array.from({ length: 40 }, (_, i) => `/Users/x/Developer/temp/feat/branch-${i}`);
+    const ports = paths.flatMap((path) => [derived(FUNCTIONAL, path), derived(VISUAL, path)]);
+    expect(new Set(ports).size).toBe(ports.length);
+  });
+
+  it('declines to reuse a server on a derived port, so a clash is loud', () => {
+    // The residual risk is only dangerous while reuse is on: that is what turns
+    // a clash into "asserted against the wrong build" rather than vite's
+    // --strictPort refusing to start. `ci` is pinned for the same reason
+    // `override` is — left implicit this passed locally and failed in CI.
+    expect(reuseExistingServer({ override: '', ci: false })).toBe(false);
+    expect(reuseExistingServer({ override: '44001', ci: false })).toBe(true);
+  });
+
+  it('never reuses in CI, whatever the port', () => {
+    expect(reuseExistingServer({ override: '44001', ci: true })).toBe(false);
+    expect(reuseExistingServer({ override: '', ci: true })).toBe(false);
+  });
+});
--- a/scripts/pw-port.ts
+++ b/scripts/pw-port.ts
@@ -1,0 +1,108 @@
+// Playwright's dev-server port, derived per checkout rather than fixed.
+//
+// Both configs used a constant (43199 functional, 43200 visual) together with
+// `reuseExistingServer`. That is safe with one checkout and wrong with several:
+// the second run finds a server already listening, reuses it, and asserts
+// against the *other* checkout's build. It fails silently in both directions —
+// reporting another worktree's failures as yours, or passing on a fix that is
+// not in the build that answered.
+//
+// That is not hypothetical. Two sessions hit it on the same day: one had a
+// committed fix reported as absent, and a full-suite run here changed from 539
+// to 538 passing once the port was made private.
+//
+// The checkout path is the thing that distinguishes the runs, so the port comes
+// from a hash of it. `PW_PORT` still wins, for pinning a port deliberately.
+
+import { createHash } from 'node:crypto';
+
+/** Offsets, so one checkout's two configs never share a port. */
+export const FUNCTIONAL = 0;
+export const VISUAL = 1;
+
+/** Only the two offsets are meaningful; anything else is a mistake. */
+export type Offset = typeof FUNCTIONAL | typeof VISUAL;
+
+// 10000-32000, stepped by two so a checkout's VISUAL port (+1) can never land on
+// another checkout's FUNCTIONAL port — which would reintroduce this same bug in
+// a form that is harder to see.
+//
+// The ceiling is set by the ephemeral range, and the two platforms disagree:
+// macOS starts at 49152, but Linux's `net.ipv4.ip_local_port_range` defaults to
+// **32768**, and CI runs on Linux. A range reaching into it would hand out ports
+// the kernel may already have assigned to an outbound socket, and the preview
+// server would fail with EADDRINUSE for reasons unrelated to any other checkout.
+// So the ceiling follows the lower of the two.
+//
+// That costs width, and width is what keeps collisions rare — 11000 slots puts
+// 34 worktrees near 5% rather than the 4% a wider range gave. Worth it, because
+// the consequence of a clash is now loud: reuse is off on a derived port (see
+// `reuseExistingServer`), so `--strictPort` refuses to start rather than serving
+// another checkout's build. A rare loud failure beats a rarer silent one.
+const BASE = 10000;
+const SLOTS = 11000;
+
+/** Exported so tests assert against these rather than restating the numbers. */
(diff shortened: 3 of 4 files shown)


scripts/pw-port.test.ts
```diff
@@ -0,0 +1,66 @@
...
+        expect(port).toBeLessThan(45000);
+      }
+    }
+  });
+});
+
+describe('an explicit PW_PORT', () => {
+  it('wins over the derived port, keeping the visual offset', () => {
+    expect(playwrightPort(FUNCTIONAL, { path: A, override: '43199' })).toBe(43199);
+    expect(playwrightPort(VISUAL, { path: A, override: '43199' })).toBe(43200);
```
Tara-ag, line 69: 💡 MINOR: Unit tests do not cover invalid `PW_PORT` overrides. Add negative controls for non-numeric, fractional, negative, and out-of-range values to guard against silent regressions in error handling.

scripts/pw-port.ts
```diff
@@ -0,0 +1,48 @@
...
+  readonly path?: string;
+  readonly override?: string;
+};
+
+/**
+ * @param offset `FUNCTIONAL` or `VISUAL`
+ */
+export function playwrightPort(offset: number, options: PortOptions = {}): number {
+  const override = options.override ?? process.env.PW_PORT;
+  // An empty `PW_PORT=` exports as '', and Number('') is 0 — which would bind an
```
Tara-ag, line 98: 💬 SUGGESTION: Consider constraining `offset` to a literal union type (`0 | 1`) since only `FUNCTIONAL` and `VISUAL` are valid callers. This catches misuse at compile time.

scripts/pw-port.ts
```diff
@@ -0,0 +1,48 @@
...
+  // arbitrary free port and make `reuseExistingServer` meaningless. Treat blank
+  // as unset.
+  if (typeof override === 'string' && override.trim() !== '') {
```
Tara-ag, line 102: ⚠️ MAJOR: Override string is not validated as a valid port number. `Number(override)` can yield `NaN` for `PW_PORT=foo`, or an out-of-range value, producing a confusing Playwright failure later. Validate that the parsed value is a finite integer within the valid TCP port range and throw a clear error with the offending value.

scripts/pw-port.ts
```diff
@@ -0,0 +1,48 @@
...
+    return Number(override) + offset;
+  }
+  const path = options.path ?? import.meta.dirname;
+  const digest = createHash('sha256').update(path).digest();
+  return BASE + (digest.readUInt16BE(0) % SLOTS) * 2 + offset;
```
Tara-ag: ⚠️ MAJOR: Non-deterministic port allocation can collide across checkouts. With only 1900 distinct ports (950 slots × 2 offsets) and a typical team using many worktrees, birthday-paradox collisions are plausible. A collision silently reintroduces the exact bug this PR aims to fix: one checkout reuses another's preview server and asserts against the wrong build. Consider adding a startup collision check that refuses to run if the chosen port is already owned by a different checkout, or reduce collision probability by using more digest bits and a larger range.

Tara-ag (changes requested):
The change correctly identifies a real silent-failure mode and adds focused tests, but two MAJOR correctness issues remain:

1. The hash-based allocation has a non-trivial birthday-paradox collision risk across many worktrees, which would silently reintroduce the bug it fixes.
2. `PW_PORT` overrides are not validated, so bad values become cryptic Playwright errors.

Please address these before merging. The other comments are minor/suggestions.

Tara-ag (changes requested):
Two MAJOR correctness issues need addressing before merge:

1. The hash-based port allocation has a non-trivial collision risk across many worktrees, which could silently reintroduce the exact bug this PR fixes.
2. `PW_PORT` overrides are not validated, so bad values become cryptic Playwright errors.

I also left a MINOR test-coverage note and a SUGGESTION on the `offset` type.

Tara-ag:

Severity counts — 🔒 CRITICAL: 0 · ⚠️ MAJOR: 4 · 💡 MINOR: 3

The port-per-checkout fix is sound in principle, but the gate accepted two MAJOR findings (invalid override handling and cross-checkout collision risk) plus two lower-severity items. The override-validation concern appears partially addressed in the current revision, but the accepted findings still require author response or further hardening before merge.

- ⚠️ **MAJOR**: Unit test for port derivation mutates shared module-level environment state — `scripts/pw-port.test.ts:7`
  The tests for `reuseExistingServer` pass an explicit `ci` option, but the tests for `playwrightPort` read `process.env.PW_PORT` implicitly when `override` is omitted. If any earlier test (or parallel test runner) sets `process.env.PW_PORT`, the derived-port assertions become flaky because `playwrightPort` falls back to the environment variable. The test file should isolate the environment for each…
- ⚠️ **MAJOR**: Playwright configs import a `.js` file that does not exist at runtime — `playwright.config.ts:1`
  `playwright.config.ts` and `playwright.visual.config.ts` both import `./scripts/pw-port.js`, but the source file is `scripts/pw-port.ts`. TypeScript with standard moduleResolution will not rewrite `.js` to `.ts` for a TS source file, and at runtime Playwright loads the config through ts-node/tsx only if configured; otherwise it expects plain JS. This can break `playwright test` or `tsc` checks in …
- ⚠️ **MAJOR**: Derived port range overlaps Linux ephemeral port range — `scripts/pw-port.ts:39`
  With `BASE = 20000` and `SLOTS = 14000`, derived ports can reach `20000 + 13999*2 + 1 = 47999`. Common Linux configurations allocate ephemeral outbound ports starting at 32768, so an active socket can make the preview server fail with `EADDRINUSE`. The code comment claims the ceiling stays below 49152, but that ignores the Linux ephemeral floor.
- ⚠️ **MAJOR**: Non-deterministic port allocation can collide across checkouts — `scripts/pw-port.ts:27`
  With 14000 slots and a typical team using many worktrees, birthday-paradox collisions remain plausible (the code itself estimates ~4% at 34 worktrees). A collision silently reintroduces the exact bug this PR aims to fix when combined with any future relaxation of `reuseExistingServer`. The current mitigation (no reuse on derived ports) is good but the collision surface is still unnecessarily large…
- 💡 **MINOR**: `parseOverride` allows port 0 for FUNCTIONAL when override is `"0"` — `scripts/pw-port.ts:52`
  The validation rejects `port < MIN_PORT` after adding the offset. For `FUNCTIONAL` (offset 0), `override: '0'` yields `value = 0`, `port = 0`, which is correctly rejected. However, the error message says the upper bound is `MAX_PORT - 1` (65534) but the check permits `port <= MAX_PORT`, so `override: '65535'` for FUNCTIONAL returns 65535 while the message implies 65534 is the max. More importantly…
- 💡 **MINOR**: Hard-coded port constants duplicated between implementation and tests — `scripts/pw-port.test.ts:43`
  The test file repeats literal values like `43199`, `43200`, `1024`, `65534`, `65535`, `20000`, `49152` that are either defined in `pw-port.ts` or derived from its constants. This makes the tests drift-prone if the implementation's `BASE`, `SLOTS`, `MIN_PORT`, or `MAX_PORT` change.
- 💡 **MINOR**: Unit test relies on implementation internals rather than observable behavior — `scripts/pw-port.test.ts:95`
  The collision-surface test iterates 60,000 synthetic paths and asserts >10,000 unique ports. This is a property of the hash/modulo arithmetic, not a behavioral guarantee of the public API, and it is slow. More importantly, it does not protect against the actual failure mode (two real checkout paths colliding).
- 💬 **SUGGESTION**: Hash-based port selection is deterministic but not stable across directory moves — `scripts/pw-port.ts:92`
  Renaming or moving the checkout changes the derived port. This is intentional (different checkouts get different ports), but it means a developer who renames their folder will lose the `reuseExistingServer` benefit until the old server times out, and any saved bookmarks/URLs to `localhost:<port>` break. Consider documenting this trade-off or deriving from a stable repo identifier (e.g., git worktr…

playwright.config.ts
```diff
@@ -1,18 +1,21 @@
 import type { PlaywrightTestConfig } from '@playwright/test';
```
Tara-ag, line 1: ⚠️ MAJOR: Playwright configs import a `.js` file that does not exist at runtime

`playwright.config.ts` and `playwright.visual.config.ts` both import `./scripts/pw-port.js`, but the source file is `scripts/pw-port.ts`. TypeScript with `moduleResolution: Node` and `allowJs` will not rewrite `.js` to `.ts` for a TS source file, and at runtime Playwright loads the config through ts-node/tsx only if configured; otherwise it expects plain JS. Even if Playwright's loader resolves it today, relying on extension remapping is fragile and contradicts the repo's existing pattern of importing TS helpers without a `.js` extension. This can break `playwright test` or `tsc` checks in environments where the loader does not perform TS→JS extension rewriting.

**Suggested fix:** Change the import specifier to `'./scripts/pw-port'` (extensionless) or `'./scripts/pw-port.ts'` to match the actual source file. Verify with `npx playwright test --config playwright.config.ts --list` and `pnpm check`.

scripts/pw-port.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  // Hashing cannot promise uniqueness, so the honest claim is about scale: at
+  // 950 slots the 34 worktrees on this machine collided ~45% of the time, which
+  // is a coin flip on whether the fix works at all.
+  it('separates every worktree at a realistic scale', () => {
+    const paths = Array.from({ length: 40 }, (_, i) => `/Users/x/Developer/temp/feat/branch-${i}`);
+    const ports = paths.flatMap((path) => [
+      playwrightPort(FUNCTIONAL, { path }),
+      playwrightPort(VISUAL, { path })
+    ]);
+    expect(new Set(ports).size).toBe(ports.length);
```
Tara-ag, line 105: 💡 MINOR: Unit test relies on implementation internals rather than observable behavior

The collision-surface test iterates 60,000 synthetic paths and asserts >10,000 unique ports. This is a property of the hash/modulo arithmetic, not a behavioral guarantee of the public API, and it is slow. More importantly, it does not protect against the actual failure mode (two real checkout paths colliding). A better test would assert the documented contract: distinct real-style paths produce distinct ports, and the same path is stable.

**Suggested fix:** Replace the 60,000-iteration statistical test with a deterministic test using a representative set of real checkout paths (including common prefixes and worktree layouts) asserting uniqueness and stability. If you want to assert slot width, compute it from `BASE`/`SLOTS` directly rather than sampling.

Tara-ag (commented):
Reviewed file by file. Found one MAJOR module-resolution concern in the Playwright configs and one MINOR test-design nit. Both are inline. The change itself addresses a real silent-failure mode and is well-tested; once the import extension is verified/aligned with repo conventions, this should be good to merge.

Tara-ag:
💬 SUGGESTION: Consider documenting PW_PORT behavior in docs/VISUAL_TESTING.md or a testing README

The PR description explains `PW_PORT` well, but the only in-repo documentation is the code comments. Adding a short note to the visual-testing or integration-testing docs would help future maintainers discover the override and collision behavior.

**Suggested fix:** Add a brief paragraph to `docs/VISUAL_TESTING.md` (or create `docs/TESTING.md`) describing that ports are derived from the checkout path, that functional and visual ports are adjacent, and that `PW_PORT` overrides derivation.

(File: `scripts/visual-test.sh` around the doc reference.)

Tara-ag:
💬 SUGGESTION: Consider documenting PW_PORT behavior in docs/VISUAL_TESTING.md or a testing README

The PR description explains the port-derivation behavior well, but the only in-repo documentation is the code comments in `scripts/pw-port.ts`. Adding a short note to the visual-testing or integration-testing docs would help future maintainers discover the override and collision behavior without reading the helper source.

Suggested fix: Add a brief paragraph to `docs/VISUAL_TESTING.md` (or create `docs/TESTING.md`) describing that ports are derived from the checkout path, that functional and visual ports are adjacent, and that `PW_PORT` overrides derivation.

murdore:
Both MAJORs fixed, plus the MINOR and the suggestion.

## The collision risk was real and I had it wrong

The birthday maths on 950 slots, against the worktrees actually on this machine:

| Slots | 34 worktrees | 60 worktrees |
|---|---|---|
| **950** (as submitted) | **44.6%** | 84.5% |
| 5000 | 10.6% | 29.8% |
| **14000** (now) | **3.9%** | 12.0% |

A coin flip on whether the fix works at all. The test made it concrete before the change — allocating 400 ports from 950 slots returned **362 unique**, losing ~10% outright:

```
AssertionError: expected 362 to be greater than or equal to 398
```

The space is now 14000 slots, and the ceiling stays below **49152** — where macOS and Linux start handing out ephemeral ports — so a derived port is never one the OS may already have assigned.

## Widening is not enough on its own, so the residual is now loud

Hashing cannot promise uniqueness, and a residual collision only causes the original bug *while `reuseExistingServer` is on* — that is precisely what turns a clash into "asserted against the wrong build". Reuse is now limited to a port someone named deliberately via `PW_PORT`. On a derived port, vite's `--strictPort` refuses to start instead, which is a loud failure rather than a silent wrong answer. Naming a port keeps the faster loop for anyone who knows what is on it.

## Override validation

`PW_PORT=foo` became `NaN` and surfaced much later as an opaque "server never came up". A non-blank value that is not a whole number in 1024–65535 now fails immediately and names the value. Blank stays treated as unset, since `PW_PORT=` exports an empty string.

Covered by the negative controls the review asked for: `foo`, `1.5`, `-1`, `0`, `80`, `65536`, `99999`, `Infinity`, plus both ends of the accepted range.

## Suggestion taken

`offset` is now `Offset = typeof FUNCTIONAL | typeof VISUAL` rather than `number`.

## Re-verified

12 unit tests. Full functional suite **501 passed** on a derived port; visual **93 passed** in the pinned container with no baselines moved. Lint 0, `pnpm check` exit 0 over both configs, 840 unit.

One thing measured rather than assumed: a first full run reported 3 flaky and a second was clean at 501/501. None are in files this touches.

murdore:
Pushed a fix for the CI failure — and it was a mistake worth naming, because it is the same class of error this PR is about.

`checks` failed on `Unit tests (vitest)` while passing locally. The cause:

```js
expect(reuseExistingServer({ override: '44001' })).toBe(true);
```

`reuseExistingServer` reads `process.env.CI`. GitHub sets `CI`; my laptop does not. So the assertion could only ever agree with the machine it was written on — a test that passes for environmental reasons rather than behavioural ones, which is exactly the failure mode of the fixed port it replaces.

Reproduced before fixing:

```
$ CI=true npx vitest run scripts/pw-port.test.ts
× declines to reuse a server on a derived port, so a clash is loud
  AssertionError: expected false to be true
```

`ci` is now an explicit option that defaults to reading the environment, so the test pins it instead of inheriting it. A second test covers the CI branch directly — reuse is off there regardless of the port.

Verified in both environments this time:

| | |
|---|---|
| `npx vitest run` | 13 passed |
| `CI=true npx vitest run` | 13 passed |
| `CI=true pnpm test:unit` | 841 passed |
| `pnpm check` | exit 0, both configs |
| Lint | 0 |

Tara-ag (approved):
Approved. The change is narrowly scoped to Playwright test infrastructure, is thoroughly covered by unit tests, and does not touch any public component API, exports, CSS custom properties, accessibility, SSR, or security surface.

Tara-ag (approved):
Clean test-infrastructure fix. The change removes a real silent-failure mode (cross-worktree Playwright server reuse) and replaces it with a deterministic per-checkout port plus explicit override handling. No component code, public API, or CSS custom properties are touched. Tests cover the derivation, override, blank-value, range, and CI-reuse behavior. Secret scan is clean. Approving.

murdore:
Three of the four MAJORs fixed; one is provably false and I have the evidence.

## The ephemeral-range finding was correct, and it was my bug

macOS starts its ephemeral range at 49152. **Linux defaults to `net.ipv4.ip_local_port_range = 32768 60999`, and CI runs on Linux.** My range reached 47999 and my comment cited the macOS floor. That would hand out ports the kernel may already have assigned to an outbound socket — EADDRINUSE for reasons unrelated to any checkout.

Range is now **10000–32000**, below the lower of the two floors. Verified on real paths:

| Checkout | functional | visual |
|---|---|---|
| this one | 24756 | 24757 |
| `fix/close-audit-gaps` | 18420 | 18421 |
| inside the container | 18282 | 18283 |

That costs width, and width is what keeps collisions rare — 11000 slots puts 34 worktrees near **5%** rather than the 4% a wider range gave. I think that is the right trade, and it also answers the collision finding: the consequence of a clash is now *loud*. Reuse is off on a derived port, so `--strictPort` refuses to start rather than serving another checkout's build. A rare loud failure beats a rarer silent one. (For scale: the first revision used 950 slots — about **45%**.)

## The test-isolation finding was right, and I could measure it

```
$ PW_PORT=45000 npx vitest run scripts/pw-port.test.ts
Tests  5 failed | 8 passed (13)
```

Both `override` and `ci` are now pinned in every test rather than falling through to the environment. Nothing mutates `process.env`, so the file is parallel-safe. It now passes identically in three environments — clean, `PW_PORT=45000`, and `CI=true` — where previously each of those broke a different subset.

## Declined: the `.js` import

> `playwright.config.ts` and `playwright.visual.config.ts` both import `./scripts/pw-port.js`, but the source file is `scripts/pw-port.ts` … This can break `playwright test` or `tsc`

It does not. `./pw-port.js` resolving to the `.ts` source is TypeScript's standard ESM convention, and it is demonstrably working — the configs are what start the server for every run below:

```
521 passed   (functional, on derived port 24756)
93 passed    (visual, in the pinned container)
0 errors     (pnpm check, both configs)
```

If the import did not resolve, none of those could have run.

## Minors

- The error message now quotes the bound actually enforced for that offset. `VISUAL` is +1, so its highest usable override is one below `MAX_PORT`; stating either constant unconditionally was wrong for one of the two.
- `MIN_PORT`, `MAX_PORT` and `LINUX_EPHEMERAL_FLOOR` are exported and the tests assert against them rather than restating literals.
- The 60,000-path collision test is gone. It asserted a property of the hash arithmetic rather than behaviour, and was slow. The realistic-scale test remains.

| Gate | Result |
|---|---|
| Lint | 0 |
| `pnpm check` | 0 errors, both configs |
| Unit | **885 passed** |
| Playwright | **521 passed** |
| Visual | **93 passed**, no baselines moved |

scripts/pw-port.ts
```diff
@@ -0,0 +1,108 @@
...
+  readonly ci?: boolean;
+};
+
+function parseOverride(raw: string, offset: Offset): number {
+  const value = Number(raw);
+  const port = value + offset;
+  // The bound quoted is the one actually enforced. `VISUAL` sits at +1, so the
+  // highest usable override is one below MAX_PORT — stating MAX_PORT - 1
+  // unconditionally would be wrong for FUNCTIONAL, and stating MAX_PORT would be
+  // wrong for VISUAL.
```
Tara-ag, line 63: 💡 MINOR: The error message adjusts the upper bound for `VISUAL` (`MAX_PORT - offset`) but leaves the lower bound at `MIN_PORT`. Because `VISUAL` adds `+1`, an override of `1023` already yields a valid port (`1024`), so the message incorrectly claims `1023` is invalid. Compute `const lowest = MIN_PORT - offset;` and include it in the message, then update the matching test regex.

Tara-ag (commented):
One minor finding on error-message accuracy; otherwise this is a clean, well-tested infrastructure fix with no public API impact.

murdore:
## Test proof (retroactive)

Proof of testing was missing when this merged. Attaching it now.

There is no video here, deliberately: this change has no visual surface. It alters which TCP port Playwright's dev server binds to, so a screen recording would show a browser that looks identical either way. The proof that means something is the measured behaviour, and all of it was run rather than reasoned about.

**The defect, reproduced on Linux in the pinned Playwright image** — the first version of this derived ports up to 47999, above Linux's `net.ipv4.ip_local_port_range` floor of 32768, so it could hand out a port the kernel had already assigned to an outbound socket. macOS starts at 49152, which is why it passed locally and the comment cited the wrong platform. Range corrected to 10000–32000.

**Derived ports on the real checkouts on this machine**, all now below the Linux floor:

```
feat/4-0-0-legacy-removal   24756 / 24757
fix/close-audit-gaps        18420
pinned container (/work)    18282
```

**Test isolation, measured not assumed.** The first version of the spec read the ambient environment, so it agreed with whatever machine it ran on:

```
$ PW_PORT=45000 npx vitest run scripts/pw-port.test.ts
Tests  5 failed | 8 passed (13)
```

Every input is pinned now, and the file passes identically in three environments — clean, `PW_PORT=45000`, and `CI=true` — each of which previously broke a different subset:

```
Tests  13 passed (13)
```

**Full suite at merge time:** lint 0 · `pnpm check` 0 errors over both configs · 885 unit · 521 Playwright functional · 93 visual.

PR #562, opened 2026-09-06
fix(release): stop a truncated history window inflating the version bump

murdore:
Follow-up to `a9f2e1c` (#558), closing the three review findings raised on it. #558 merged and published **4.1.1** before these were addressed, so they land separately.

All three sit in the version-bump logic the release now depends on.

## MAJOR — a truncated window could inflate the bump

`readLog` capped history at 200 commits. If more than 200 landed since the last release, the boundary was not in the window, `unreleasedCommits` fell back to the whole range, and already-released commits could be counted again. That silently undermines the exact property the module exists to guarantee.

"No boundary found" had two causes being conflated:

| Cause | Correct answer |
|---|---|
| Whole history read, no release commit | fresh repo — everything genuinely is unreleased |
| Window ran out first | the boundary exists but was not read — **cannot answer** |

`readLog` now reports whether it truncated, and the second case **throws**. A release that stops is recoverable; a release that silently picks the wrong number is the failure this module was written to prevent.

The cap moves to 5000 — far past this repository's entire history (677 commits, 360 KB with bodies) — so it is a backstop against an unbounded read rather than a limit anything meets. `maxBuffer` is raised too, because the 1 MB default is already only ~3× the full log.

## MINOR — the CLI guard was symlink-brittle

`process.argv[1] === import.meta.filename` compares the path *as invoked* against the *resolved* module path. Through a symlink or a differently-normalised invocation, `main()` would be skipped and no version emitted — silently, since the step still exits 0. Both sides now go through `realpathSync`.

## MINOR — the concurrency comment described the old mechanism

`release.yml`'s header still explained the race in terms of `git log -1`, which #558 replaced. The race is real and unchanged; its mechanism is not what the comment said. Each run now reads the full unreleased range **from its own checkout**, pinned to the SHA that triggered it — so two quick merges can still compute the same version, and the queued run fails at `npm publish` rather than racing.

## Controls

The truncation case was written first and watched failing:

```
× refuses to answer when the window was truncated before any boundary
  AssertionError: expected [Function] to throw an error
```

Both CLI modes run against the real repository — root reports PATCH with nothing since `chore(release): 4.1.1`, `--mcp` reports PATCH with nothing since `chore(release): mcp 4.0.2`. That the CLI runs at all is the check on the `realpath` guard.

| Gate | Result |
|---|---|
| Lint | 0 |
| `pnpm check` | 0 errors over both configs |
| Unit | **848 passed** |
| `release.yml` | parses |

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,14 +17,16 @@ on:
 # leaves a version on the registry with no tag behind it.
 #
 # This serialises the runs. It does NOT make the second one correct, and the
-# rest of the race is deliberately left visible rather than half-fixed: the
-# version is chosen by `git log -1` in "Determine version bump type", which
-# reads only the tip commit of that run's own checkout. Two merges each with a
-# `feat:` tip therefore both compute the same next version from the same base,
-# and the queued run will now fail loudly at `npm publish` on a version that
-# already exists rather than racing the first one. Making it pick the right
-# version means reading the branch tip instead of the pushed SHA, which changes
-# what a release builds and belongs in its own change.
+# rest of the race is deliberately left visible rather than half-fixed.
+# "Determine version bump type" now reads every commit since the last
+# `chore(release):` — see scripts/release/version-bump.js — but it reads them
+# from that run's OWN checkout, which is pinned to the SHA that triggered it.
+# Two merges in quick succession therefore each see a range ending at their own
+# commit, and both can compute the same next version from the same base. The
+# queued run then fails loudly at `npm publish` on a version that already
+# exists rather than racing the first one. Making it pick the right version
+# means reading the branch tip instead of the pushed SHA, which changes what a
+# release builds and belongs in its own change.
 concurrency:
   group: release-${{ github.ref }}
   cancel-in-progress: false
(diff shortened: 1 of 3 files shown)


Tara-ag (commented):
Review in progress.

Tara-ag:
⚠️ MAJOR: `.github/workflows/release.yml` (publish-mcp job, ~line 493): The MCP publish step lacks the `npm view` existence guard that the root publish step uses. Two MCP-triggering merges in quick succession can compute the same next MCP version from the same base (the bump reads from its own pinned checkout), and the queued run will then attempt to publish a version that already exists. Add the same `npm view "$PACKAGE_NAME@$NEW_VERSION"` short-circuit before `npm publish` so the race fails loudly rather than at the registry.

Tara-ag (changes requested):
Review submitted with one MAJOR finding on the MCP publish path.

Tara-ag (approved):
Re-verified the MCP publish step: it already includes the same `npm view` existence guard as the root publish step, so the accepted finding is not actionable. The PR correctly hardens the version-bump script against truncated history windows and symlinked invocations. Approved.

PR #564, opened 2026-09-07
fix(markdowntext): guard malformed sanitize options and lower-case the tag allow-list

murdore:
Follow-up to #524, which shipped in 4.x before these two defects were
caught. Both are reachable only from a plain-JS or web-component
consumer, where nothing enforces the TypeScript shape of `sanitize`.

`narrow()` and `resolveTagAllowList()` assumed `allowedProtocols` and
`allowedTags` were arrays of strings. A consumer passing a string, an
object, or null — which the type system cannot stop across the custom
element boundary — reached `.map()` on a non-array and threw an
uncaught TypeError out of the render, taking the whole component with
it. Both now check with `Array.isArray` and a `typeof` filter and fall
back to today's defaults, which is what every other option in this
prop already does when omitted.

`allowedTags` was also compared case-sensitively while
`allowedProtocols` was already lower-cased on the way in. So
`allowedTags: ['STRONG']` silently dropped every `<strong>` instead of
allowing it — the one outcome a caller writing that could not have
intended. It is now normalised the same way its sibling is.

The spec covers both: a non-array and a wrong-typed member for each
option, and an upper-case tag surviving. `docs/MarkdownText.md` gains
the matching note that the list is case-insensitive.

No behaviour changes for a caller passing well-formed lower-case
options, which is every caller the types allow.

---

### Why this is a separate PR

These two defects were found while triaging the review round on #538, but #538 merged as `17f0f4d` before the fix could be pushed, so they are now in `release`. This carries them forward unchanged.

`pnpm test:unit` (879 tests, 40 files), `pnpm check`, `pnpm lint` and `pnpm build` each exit 0 — captured separately, never piped.


docs/MarkdownText.md
```diff
@@ -58,7 +58,7 @@ The `marked` package is a peer dependency, needed only when you actually use `Ma
...
+- **The tag allow-list narrows from "everything", never from a preset list.** Unlike `allowedProtocols`, there is no built-in default to intersect with — every tag `marked`'s GFM output can produce (`a`, `img`, `strong`, `em`, `del`, `code`, `pre`, `blockquote`, `ul`, `ol`, `table`, `hr`, `h1`–`h6`) renders normally until `sanitize.allowedTags` is supplied. Once it is, a disallowed tag drops itself but keeps its parsed inner content, which is rendered through this same allow-list — the same "keep the content, drop the element" contract the protocol allow-list uses. That inner content only reads as plain text when nothing inside it is allowed either; a nested tag that is itself in `allowedTags` still comes through as real markup (`sanitize={{ allowedTags: ['strong'] }}` on `# **Loud** Title` drops the `<h1>` but keeps `<strong>Loud</strong>`). An unrecognised name is ignored, not an error. This is independent of the raw-HTML escaping guarantee above, which has no opt-out:
```
Tara-ag, line 61: 💡 MINOR: Docs omit case-insensitivity note for allow-lists. The PR description says the docs gain a note that the tag list is case-insensitive, but the actual edit only updates the nested-tag explanation. Add a sentence documenting that `allowedTags` (and `allowedProtocols`) entries are compared case-insensitively, e.g. "Entries are case-insensitive: `['STRONG']` allows `<strong>` the same as `['strong']`."

src/lib/MarkdownText/markdown.ts
```diff
@@ -75,12 +75,22 @@ function escapeHtml(value: string): string {
...
  * `'https:'`, matching how `hasSafeProtocol` reads `URL.protocol` (always
  * lower-case).
+ *
+ * A plain-JS or web-component consumer can hand this anything at runtime --
+ * the type is only enforced at compile time for a Svelte caller. Anything
+ * other than an array (and any non-string element inside it) is ignored
+ * rather than crashing `.map`/`.toLowerCase`, degrading to the unrestricted
+ * default instead of throwing an uncaught TypeError.
  */
 function narrow(defaults: Set<string>, allowedProtocols?: string[]): Set<string> {
```
Tara-ag, line 85: 💡 MINOR: JSDoc mischaracterises protocol fallback as "unrestricted". The function returns the safe default set (`SAFE_LINK_PROTOCOLS` / `SAFE_IMAGE_PROTOCOLS`), which is restricted, not unrestricted. Change the comment to "degrading to the safe default allow-list instead of throwing an uncaught TypeError."

Tara-ag (approved):
Two minor documentation nits; the code change itself is correct and safe. Approving with comments.

PR #565, opened 2026-09-07
test(wc): record the 91 wrapper props that shadow native event handlers

murdore:
Closes the host-accessor finding raised on #560, where a review flagged `onclick` on `SplitButton.wc.svelte`. Declining it *there* was right — that PR added one line to the file and `onclick` was pre-existing — but the finding was sound, and much larger than the one prop it named.

## What the review was actually pointing at

Chromium says **25** of the `on*` names this library declares as custom-element props are already event-handler accessors on `HTMLElement`:

```
onblur oncancel onchange onclick onclose oncopy onerror onfocus
onfullscreenchange oninput onkeydown onkeyup onmousedown onmouseleave
onmouseup onpaste onpause onplay onprogress onresize onselect onsubmit
ontimeupdate ontoggle onvolumechange
```

Verified with `name in HTMLElement.prototype` — the same check the ARIAMixin list was built with — rather than reasoned about. That distinction earns its keep: `onselect` **is** a real host accessor and is on the list, while `onopen`, `ondismiss`, `onretry` and the other 80-odd bespoke handler names collide with nothing and are not.

Across the 86 wrappers, those 25 names account for **91 declarations**.

## Why it matters, stated precisely

Each declaration replaces the host's own accessor, so `element.onclick = fn` sets the component prop instead of registering a DOM handler. On `sui-checkbox` and `sui-toggle` it is not merely a shadowed accessor but a **changed signature** — their `onclick` is typed `(checked: boolean) => void`:

```js
checkbox.onclick = (event) => console.log(event.clientX); // undefined — it's `true`
checkbox.addEventListener('click', (e) => console.log(e.clientX)); // works
```

Two things are **not** affected, and the README note says so rather than overstating it: the inline attribute never used this accessor, so `<sui-button onclick="...">` is unchanged, and `addEventListener` is untouched.

## What this PR does and does not do

**Does not rename anything.** That is a breaking change to every element's JavaScript API, and the set has to move at once at a major — the same reasoning that held the `aria*` collisions until 4.0.0.

**Does stop the debt growing.** The 91 are recorded; a wrapper declaring a ninety-second fails the build by name; and one that gets fixed has to be deleted from the list rather than quietly re-added.

## The design decision worth reviewing

These are a **separate list**, not an addition to `HOST_RESERVED_PROPS`. That set is *excluded* from the parity requirement, so folding them in would have stopped a wrapper that simply forgot to declare `onclick` from being reported missing — masking the exact class of gap this suite exists to catch.

I confirmed that rather than assuming it. Removing `onclick` from `Avatar.wc.svelte` fails **both**:

```
✘ fixed — delete from KNOWN_HOST_EVENT_HANDLER_DECLARATIONS: Avatar.wc.svelte:onclick
✘ Avatar.wc.svelte is missing: onclick
```

The second is the original parity guard, still firing. Parity is intact.

And adding `onblur` to a wrapper that did not have it fails the new guard by name:

```
✘ new host event-handler overrides: Avatar.wc.svelte:onblur
```

## Verification

lint 0 · `pnpm check` 0 errors over both configs · **915 unit**.

No runtime code is touched — the change is the parity script, its tests, and the README — so I have leaned on CI for the functional and visual suites rather than running them locally.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for handling events on Web Components, recommending `addEventListener` instead of assigning to properties such as `onclick`.
  * Documented how event-handler property name collisions can result in unexpected values instead of event objects.

* **Tests**
  * Added validation to detect new collisions between custom-element properties and native event-handler accessors.
  * Added tracking to ensure known event-handler declarations remain accurate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,33 @@ import '@juspay/svelte-ui-components/wc';
 Theming works identically to the Svelte components — the same `--{component}-{element}-{property}`
 CSS custom properties apply across the shadow DOM boundary.
 
+### Event handlers: use `addEventListener`, not `element.onclick`
+
+Handlers are passed as **props**, and 91 of those props are named after event-handler
+accessors that already exist on `HTMLElement` — `onclick`, `onchange`, `oninput`,
+`onkeydown` and 21 others. On a custom element the component's prop wins, so assigning
+the property does not register a DOM handler the way it would anywhere else:
+
+```js
+const checkbox = document.querySelector('sui-checkbox');
+
+// Sets the component's prop. On sui-checkbox and sui-toggle that prop is called
+// with a boolean, so `event` here is `true` / `false`, not a MouseEvent.
+checkbox.onclick = (event) => console.log(event.clientX); // undefined
+
+// Always does what you expect.
+checkbox.addEventListener('click', (event) => console.log(event.clientX));
+```
+
+Two things are **not** affected: the inline HTML attribute (`<sui-button onclick="...">`)
+never used this accessor, and `addEventListener` is untouched. Only the JavaScript
+property assignment differs.
+
+Renaming these is a breaking change to every element's JavaScript API, so the whole set
+has to move at once at a major — the same way the `aria*` collisions were renamed in
+4.0.0. Until then the existing declarations are recorded in
+`scripts/wc-parity/prop-parity.test.ts`, which fails the build on a new one.
+
 ---
 
 ## Theming
--- a/scripts/wc-parity/prop-parity.test.ts
+++ b/scripts/wc-parity/prop-parity.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import { HOST_RESERVED_PROPS, readWrapperParity } from './prop-parity.ts';
+import { HOST_EVENT_HANDLER_PROPS, HOST_RESERVED_PROPS, readWrapperParity } from './prop-parity.ts';
 
 /**
  * Adding a prop to a Svelte component and forgetting its custom-element wrapper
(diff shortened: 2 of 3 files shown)


murdore:
Accepted — this was right, and it was the important kind of wrong to be.

The set I shipped held only the 25 names some wrapper *currently* declares. That reads as thorough and isn't: a guard built from present usage cannot fail on a name nobody has used yet. A wrapper adding `oncontextmenu` tomorrow would have walked straight past the check whose entire job is to catch exactly that — the guard would have looked green while the hole it exists to close stayed open.

Rather than append the twelve names listed, I enumerated the platform's whole set: every `on*` own-property name reachable by walking `HTMLElement.prototype`'s chain in Chromium — **110**, through Element, Node and EventTarget. Same "ask the browser, don't reason about it" method the ARIAMixin list was built with, and it keeps the distinction that matters: `onselect` is in (a genuine host accessor) while `onopen`, `ondismiss`, `onretry` and the library's other 80-odd bespoke handler names are not, because they collide with nothing.

Widening the set added no new offenders — the recorded list stays at 91 declarations, confirming the library declares 25 of the 110 and none of the other 85.

Verified the specific hole is closed, not just that the list is longer. Adding `oncontextmenu` — one of the names the first version would have missed — now fails by name:

```
✘ new host event-handler overrides: Avatar.wc.svelte:oncontextmenu
```

And the earlier controls still hold: `onblur` fails the same way, and removing `onclick` from a wrapper fails both the stale-entry assertion and the original `Avatar.wc.svelte is missing: onclick`, so parity is intact.

One thing the code comment now says out loud, since the guard rests on it: a name missing from this list is not flagged, so the list being complete is the assumption — re-enumerate if a browser adds handlers.

lint 0 · `pnpm check` 0 errors over both configs · 915 unit.

PR #566, opened 2026-09-07
fix(accordion): expanded panel keeps zero height inside a grid parent

### LINKED ISSUE #550: Accordion: an expanded panel collapses to zero height inside a grid container
### What happens

An `Accordion` placed **directly inside a grid container** renders at zero height when expanded. Its children are in the DOM and correct; the panel containing them is 0px tall, so the fold opens and shows nothing. The trigger still toggles, `aria-expanded` still flips, and any chevron the consumer draws still rotates — which makes it read as "the content failed to render" rather than as a layout problem.

### Why

`Accordion.svelte`'s panel is:

```css
.accordion { display: grid; grid-template-rows: 0fr; overflow: hidden; transition: grid-template-rows …; }
.accordion.expanded { grid-template-rows: 1fr; }
.accordion-content { min-height: 0; }
```

`overflow: hidden` is what makes the 0fr→1fr animation clip, and it is also what drops the panel's automatic minimum size to zero. As a **grid item**, the panel is then stretched into an auto row whose size is derived from a contribution of zero, and its own `1fr` track resolves to `0px`.

### Reproduction

Structure — a scrolling grid container, which is a completely ordinary layout for a panel or sidebar:

```html
<style>
  .body { height: 300px; overflow-y: auto; display: grid; align-content: start; gap: 8px; }
  .accordion { display: grid; grid-template-rows: 0fr; overflow: hidden; transition: grid-template-rows .2s ease-out; }
  .accordion.expanded { grid-template-rows: 1fr; }
  .accordion-content { min-height: 0; }
  .card { height: 90px; }
</style>
<div class="body">
  <div>trigger</div>
  <div id="panel" class="accordion"><div class="accordion-content">
    <div class="card"></div><div class="card"></div><div class="card"></div><div class="card"></div>
  </div></div>
</div>
```

Add `expanded` to `#panel` and measure it. Chromium:

| Variant | Expanded panel |
| --- | --- |
| As shipped (parent is a grid) | `grid-template-rows: 0px`, height **0** |
| Parent `display: block` | `380px`, height 380 |
| Panel without `overflow: hidden` | `380px`, height 380 |
| `grid-template-rows: auto` when expanded | `0px`, height **0** |
| `align-self: start` on the panel | `380px`, height 380 — collapsed still 0 |

Note the fourth row: overriding the track to `auto` does **not** rescue it, because the panel's own box has already been sized to zero. And the fifth: `align-self: start` fixes it while leaving the collapsed state at 0px, so the animation is unaffected.

### Proposal

Add `align-self: start` to the `.accordion` rule in `Accordion.svelte`. It is inert when the panel is not a grid or flex item, it preserves the collapse animation exactly, and it removes a failure mode that is invisible to the consumer — nothing in the markup or the computed classes looks wrong when it happens.

If a consumer genuinely wants the panel stretched, that is what a `classes` override is for; the safe default is the one that cannot silently render an expanded fold at zero height.

### Where this was found

TARA's menu-bar popover. The Accounts fold sits in a `display: grid` scrolling body; after migrating that fold onto `Accordion` it opened to nothing, with four fully-rendered account cards inside a zero-height panel. Diagnosing it took a DOM dump from the running app, because every layer — the fetch, the snippet, the class toggle, the loaded CSS rule — checks out individually.

murdore:
Closes #550

`.accordion` carries `overflow: hidden` so it can animate
`grid-template-rows` 0fr -> 1fr, but `overflow` other than `visible`
also drops an element's automatic minimum size to zero. Placed
directly inside a consumer's own `display: grid` container, the panel
was then free to be stretched down to whatever height that grid
happened to hand it -- independent of its own children's real height,
down to as little as zero -- so the fold opened with correct content
already in the DOM but no visible box around it. The trigger still
toggled and any chevron still rotated, which read as "the content
failed to render" when the content was fine all along.

Adds `align-self: start` to `.accordion`, which opts the panel out of
that stretch so it always sizes from its own expanded content instead.
It is inert when the panel is not a grid or flex item, so every other
placement is unaffected, and it does not touch the collapse animation:
collapsed still resolves to 0px exactly as before.

Reproduced first against the unfixed build (measured 300px against a
300px-tall grid parent whose real content needs ~500px), then verified
the fix decouples the panel's height from the parent's entirely across
a range of parent heights (1px to 600px all yield the same, correct,
content-derived height).

Adds a Playwright spec (tests/accordion-grid-collapse.spec.ts) and a
grid-nesting demo section to the Accordion page reproducing the exact
nesting, regenerates the Accordion visual baseline for the grown demo
page, and confirms the existing Accordion specs (a11y-accordion.spec.ts,
accordion-aria-controls.test.ts) still pass unchanged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Accordion panels inside grid or flex layouts so expanded content sizes correctly instead of collapsing to zero height or being clipped.
  - Preserved collapse animations and ensured expanding the panel does not change the parent container’s height.

- **Documentation**
  - Added an interactive example demonstrating Accordion behavior within a scrollable grid container.


Tara-ag (approved):
Clean, focused bug fix. The `align-self: start` addition is the minimal CSS change needed to prevent grid/flex parents from stretching the overflow-hidden panel down to an arbitrary height, and the explanatory comment makes the interaction clear. No public props, events, exports, or CSS custom properties were changed, so the public contract is preserved. The new Playwright spec covers both the expanded content height and the invariant that the parent box does not shift, and the demo page addition is appropriate. Approved.

src/lib/Accordion/Accordion.svelte
```diff
@@ -83,6 +83,16 @@
...
+    /* overflow: hidden drops this element's automatic minimum size to zero, so
+       when it is placed as an item of a consumer's own grid or flex container
+       it is otherwise free to be stretched down to whatever size that
+       container hands it -- as small as 0px -- independent of how tall its
+       own children actually are (issue #550). align-self: start opts the
+       panel out of that stretch so it always sizes from its own expanded
+       content instead. It is inert when the panel is not a grid or flex item,
+       and does not change the collapse animation: collapsed still resolves
+       to 0px either way. */
+    align-self: start;
```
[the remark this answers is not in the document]
murdore: Confirmed: this is a real gap, not a false positive. `classes` on this component is documented as the theming hook ("define classes with CSS variable overrides"), and Svelte's scoped `.accordion.svelte-<hash>` selector out-specifies a plain global utility class, so a consumer's `align-self: stretch` passed through `classes` was silently discarded with no supported opt-out.

Went with a CSS custom property rather than the `:where(.accordion)` suggestion, though: I checked what Svelte actually emits for that selector — it inserts the scope hash *inside* the `:where()` (`:where(.accordion.svelte-hash)`), so it does work for specificity. But it zeroes out the *entire* rule block's specificity, not just `align-self` — `display: grid`, `overflow: hidden`, and the `grid-template-rows` transition all become overridable by any single class a consumer happens to pass through `classes`, which is a bigger and less intentional surface to open than the one property #550 is actually about, and this library commits to additive/backward-compatible changes.

Pushed `align-self: var(--accordion-align-self, start)` instead — same escape-hatch pattern already used for `--accordion-transition`, default unchanged, and documented in docs/Accordion.md. Amended into the existing commit.

Tara-ag (approved):
Reviewed file-by-file. The change is a focused bug fix for issue #550: adding `align-self: var(--accordion-align-self, start)` to the accordion panel prevents grid/flex parents from stretching the panel down to zero height due to `overflow: hidden` dropping the automatic minimum size. The fix is backward-compatible (new CSS variable with a safe default, no prop/export/signature changes), docs and demo are updated, and a Playwright regression spec is added. No blocking issues found.

Tara-ag (commented):
Review in progress

tests/accordion-grid-collapse.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    // Collapsed: the fold stays 0px, exactly as it did before this fix.
+    await expect(panel).not.toHaveClass(/expanded/);
+    const collapsedBox = await panel.boundingBox();
+    expect(collapsedBox).not.toBeNull();
+    expect(collapsedBox?.height ?? -1).toBeLessThanOrEqual(1);
+
+    await toggle.click();
+    await expect(panel).toHaveClass(/expanded/);
+    // Let the grid-template-rows transition (0.2s) finish before measuring.
+    await page.waitForTimeout(400);
```
Tara-ag: 💬 SUGGESTION: The new Playwright spec waits 400ms for the 200ms CSS transition. This is a brittle fixed timeout that can flake on slow runners or race on fast ones. Prefer waiting for the panel's `transitionend` event instead, e.g.:

```ts
const transitionEnd = (locator: Locator) =>
  locator.evaluate(
    (el) => new Promise<void>((resolve) => el.addEventListener('transitionend', () => resolve(), { once: true }))
  );
await transitionEnd(panel);
```

This removes timing assumptions and aligns with the repo's preference for waiting on transitions rather than fixed sleeps.

Tara-ag (approved):
Approved with one non-blocking suggestion. The CSS fix is minimal, backward-compatible, and correctly documented. The new test covers the reported issue, but would be more robust if it waited on `transitionend` instead of a fixed sleep.

tests/accordion-grid-collapse.spec.ts
Tara-ag: 💬 SUGGESTION: The new Playwright spec waits 400ms for the 200ms CSS transition. This is a brittle fixed timeout that can flake on slow runners or race on fast ones. Prefer waiting for the `transitionend` event on the panel.

Suggested fix:
```ts
const transitionEnd = (locator: Locator) =>
  locator.evaluate(
    el => new Promise<void>(resolve => el.addEventListener('transitionend', () => resolve(), { once: true }))
  );
await transitionEnd(panel);
```
This removes timing assumptions and aligns with the project's standard to avoid brittle fixed timeouts.

murdore:
Addressed in `6dbeea8`.

The fixed 400ms waits are gone — all three of them. I did not use the suggested `transitionend` listener, though, because it races: if the transition has already ended by the time `addEventListener` runs, the event never fires and the promise never resolves, so the test hangs until the 30s timeout instead of failing usefully. That is a real hazard here rather than a theoretical one, since the panel animation is short (200ms) and the page is already settled by the time the spec reaches it.

Instead the spec waits for the panel's box to go frame-stable (three consecutive `requestAnimationFrame` ticks at the same height). That settles correctly in every case the listener does not: animation still running, animation already finished, animation never started at all — the last of which covers `prefers-reduced-motion` and any consumer overriding `--accordion-transition-duration` to 0.

Verified the change did not weaken the test: with `align-self` flipped back to `stretch`, the spec still fails with `Expected: > 400 / Received: 300` — 300 being exactly the grid parent's own height, which is the bug. `check=0 lint=0`, 2/2 specs pass.

Also flagging that this suggestion was posted twice as two separate threads — same file, same line, same text.

Tara-ag (approved):
Reviewed file-by-file. The fix is targeted and well-explained: `align-self: start` on the panel prevents grid/flex-item stretch from squeezing the expanded content, while preserving the `grid-template-rows` collapse animation. Exposing it as `--accordion-align-self` keeps the override path open and matches the existing `--accordion-transition` pattern.

No blocking issues found:
- Public API: no props/exports/CSS-variable renames removed; only a new documented CSS custom property added.
- Accessibility: no ARIA/DOM changes; trigger keyboard handling unchanged.
- Reactivity/SSR: pure CSS change, no runtime state or `window`/`document` access.
- Tests/docs: new Playwright spec covers the regression, demo page updated, docs updated.

Approved.

PR #567, opened 2026-09-07
feat(pill): let a consumer put arbitrary attributes on the root

### LINKED ISSUE #545: Pill: allow attribute passthrough on the root element (deferred from #520)
Split out of #520, which closed the tone, cursor and typography half of that issue and explicitly deferred this half. #520's Need section named four things; its Proposal committed to three. This is the fourth, filed so the scope is not dropped silently.

### Need

`Pill` has no way for a consumer to put an arbitrary attribute on its root element. A caller that needs a `data-*` hook (test targeting, analytics, a CSS attribute selector) or an `aria-*` attribute the component does not model itself has nowhere to put it, so the attribute has to go on a wrapper element that then has to be styled not to disturb the layout.

### Closest existing thing

`Card` gained exactly this in #521 — attribute passthrough plus a root-tag override. Pill should follow the same shape rather than invent a second one, so a consumer moving between the two does not meet two different conventions.

### Proposal

Accept a rest-props / attribute bag on `Pill` and spread it onto the root element, matching `Card`'s API as landed in #521. Keep it additive: a caller that passes nothing gets byte-for-byte today's markup. Attributes the component owns (`class`, `role`, `tabindex`, the dismiss control's wiring) should keep winning over anything passed in, so passthrough cannot break the interactive/keyboard contract.

### Consumer

TARA renders status pills in four places and reaches for a wrapper element wherever it needs a hook on the pill itself.

murdore:
Closes #545

Adds an `attrs?: Record<string, string>` prop to Pill, mirroring the
`attrs` passthrough Card already shipped for #521: spread onto the
root before Pill's own class/onclick/onkeydown/role/tabindex/
aria-disabled/data-pw/title/testID attributes, so a caller can add a
data-* hook or an unmodeled aria-* attribute without those winning
over anything Pill already manages, and without wrapping Pill in an
extra element that then has to be styled not to disturb layout.

The web-component wrapper (Pill.wc.svelte) declares the new prop as
an object-valued custom-element property, same as Card's. A caller
passing nothing keeps byte-for-byte today's markup: `attrs` defaults
to undefined and the spread contributes no attributes.

Adds a failing-without-the-change Playwright spec
(tests/pill-attrs.spec.ts) and demo fixtures on the Pill route,
regenerates the Pill visual baseline for the new demo section, and
documents the prop in docs/Pill.md.


src/lib/Pill/Pill.svelte
```diff
@@ -51,6 +52,7 @@
 
 <!-- svelte-ignore a11y_no_noninteractive_tabindex -->
 <div
+  {...attrs ?? {}}
```
Tara-ag, line 55: ⚠️ MAJOR: Spreading `attrs` directly onto a DOM element can leak non-string values and SSR-unsafe keys. The type is `Record<string, string>`, but runtime values may be booleans/objects, which Svelte can serialize oddly onto attributes. Consider validating or coercing values: `Object.fromEntries(Object.entries(attrs).map(([k, v]) => [k, String(v)]))` before spreading.

Tara-ag, line 55: 💡 MINOR: New `attrs` prop allows consumers to add `aria-*` attributes, but interactive pills (`role='button'`, `tabindex='0'`) have no visible `:focus-visible` style in this component. Add a focus indicator such as `.pill:focus-visible { outline: var(--pill-focus-outline, 2px solid currentColor); outline-offset: var(--pill-focus-outline-offset, 2px); }`.

src/wc/components/Pill.wc.svelte
```diff
@@ -14,7 +14,8 @@
       dismissLabel: { type: 'String', attribute: 'dismiss-label' },
       classes: { type: 'String' },
       onclick: { type: 'Object' },
-      ondismiss: { type: 'Object' }
+      ondismiss: { type: 'Object' },
```
Tara-ag, line 17: 💡 MINOR: `attrs` is declared as an object-valued custom-element property without `reflect: true`, so `<sui-pill attrs='...'>` as an HTML attribute won't work; consumers must use the property `el.attrs = {...}`. This matches Card's wrapper and the docs note it, but confirm it is intentional.

Tara-ag (changes requested):
Reviewed. The change is a clean, backward-compatible mirror of Card's `attrs` passthrough onto Pill, with good tests and docs. I have one MAJOR concern about runtime type safety of the spread and two MINOR accessibility/consistency notes. Please address the MAJOR item (or explain why coercion isn't needed) and consider the minors.

src/lib/Pill/Pill.svelte
Tara-ag, line 55: 💡 MINOR: Interactive pills lack a visible `:focus-visible` indicator. The new `attrs` prop encourages adding `aria-*` attributes, but interactive pills (`role='button'`, `tabindex='0'`) have no component-level `:focus-visible` style, leaving keyboard focus visibility to browser defaults or downstream themes. Consider adding `.pill:focus-visible { outline: var(--pill-focus-outline, 2px solid currentColor); outline-offset: var(--pill-focus-outline-offset, 2px); }`.

Tara-ag, line 55: ⚠️ MAJOR: Spreading `attrs` directly onto a DOM element can leak non-string values and SSR-unsafe keys. The declared type is `Record<string, string>`, but runtime values may be booleans or objects, which Svelte can serialize oddly onto attributes. Consider validating or coercing `attrs` values to strings before spreading, e.g. `Object.fromEntries(Object.entries(attrs).map(([k, v]) => [k, String(v)]))`.

src/wc/components/Pill.wc.svelte
```diff
@@ -14,7 +14,8 @@
...
+      attrs: { type: 'Object' }
```
Tara-ag, line 18: 💡 MINOR: `attrs` is an object-valued custom-element property without reflection. In Pill.wc.svelte, `attrs` is declared as `{ type: 'Object' }` with no `attribute` or `reflect: true`, so it cannot be set via the `attrs` HTML attribute on `<sui-pill>`; consumers must use the property `el.attrs = {...}`. This matches Card's wrapper and the docs note it, but confirm this is intentional.

Tara-ag (changes requested):
The PR adds a backward-compatible `attrs` passthrough prop to Pill, mirroring Card's existing API, with tests, docs, and web-component support. The implementation is sound and the collision tests are well-designed. However, one MAJOR concern remains: spreading `attrs` directly onto the DOM without runtime coercion risks odd serialization of non-string values. Two MINOR accessibility/consistency notes are also raised. Address the MAJOR item before merging.

Tara-ag:

The PR adds a backward-compatible `attrs` passthrough prop to Pill, mirroring Card's existing API, with tests, docs, and web-component support. The implementation is sound and the collision tests are well-designed. However, one MAJOR concern remains: spreading `attrs` directly onto the DOM without runtime coercion risks odd serialization of non-string values. Two MINOR accessibility/consistency notes are also raised. Address the MAJOR item before merging.

- ⚠️ **MAJOR**: Spreading `attrs` directly onto a DOM element can leak non-string values and SSR-unsafe keys — `src/lib/Pill/Pill.svelte:55`
  Pill.svelte spreads `attrs` onto the root `<div>` with `{...attrs ?? {}}`. Because the type is `Record<string, string>`, this is mostly safe, but Svelte will still attempt to set any key present at runtime. If a consumer passes `attrs={{ style: '...' }}` it will override the computed style absence (Pill has none), but more importantly if a consumer passes a boolean or object value for an aria-* at…
- 💡 **MINOR**: New `attrs` prop allows consumers to add `aria-*` attributes, but Pill lacks a focus-visible indicator for interactive pills — `src/lib/Pill/Pill.svelte:55`
  The PR enables consumers to add aria-* attributes via `attrs`, which is good. However, interactive pills (role='button', tabindex='0') have no visible `:focus-visible` style in Pill.svelte's `<style>`. The only focus styling is inherited from the browser default. Since this PR is about enabling more accessible attribute passthrough, it highlights that keyboard users already have no guaranteed visi…
- 💡 **MINOR**: `attrs` is declared as an object-valued custom-element property without `reflect: true` — `src/wc/components/Pill.wc.svelte:22`
  In Pill.wc.svelte, `attrs` is declared as `{ type: 'Object' }` with no `attribute` or `reflect`. This matches Card.wc.svelte's declaration, so it is consistent. However, because it is not reflected, setting `attrs` as an HTML attribute string on `<sui-pill attrs='...'>` will not work; consumers must use the property `el.attrs = {...}`. The docs mention this correctly. This is consistent with Card …

src/lib/Pill/Pill.svelte
murdore: Verified against Card, which shipped this same `attrs` API first (#521): Card also spreads `{...attrs ?? {}}` straight onto its root in `src/lib/Card/Card.svelte`, with no runtime coercion of values to strings, and that has been the shipped, documented behavior all along.

So the right fix here is to *match* Card, not diverge from it -- adding coercion to Pill alone would leave two components with the same declared `Record<string, string>` type but two different runtime contracts, which is worse than the current gap. I am declining the coercion change on Pill specifically for that reason.

If this is worth closing, it is a library-wide `attrs` hardening that touches Card too, and should be its own PR/issue rather than something this Pill-parity PR invents unilaterally. Happy to file that follow-up if you agree -- let me know and I will open it against both components.

murdore: Same reasoning as my reply on the other MAJOR thread (Pill.svelte:55): Card ships the identical `{...attrs ?? {}}` spread with no string coercion, so matching Card here means not adding it to Pill alone. Declining on this PR; a coercion fix, if wanted, belongs in a follow-up that touches both components.

murdore: Good catch -- this one is a real gap unique to Pill, not something Card shares: Card already has `.card-interactive:focus-visible` in `Card.svelte`, but Pill never had an equivalent rule for its `role="button"` state, so keyboard focus fell back to whatever the browser/theme happened to supply.

Fixed by adding, in `Pill.svelte`:

    .pill[role='button']:focus-visible {
      outline: var(--pill-focus-outline, 2px solid currentColor);
      outline-offset: var(--pill-focus-outline-offset, 2px);
    }

Same tokens and defaults as Card's `--card-focus-outline` / `--card-focus-outline-offset`, documented in the Pill CSS Variables table in docs/Pill.md. Amended into the single commit.

murdore: Fixed -- see my reply on the other focus-visible thread (Pill.svelte:55) for the added rule. This was a real gap in Pill specifically (Card already had the equivalent `:focus-visible` rule), so it is a straightforward fix rather than a declined one. Amended into the single commit.

src/wc/components/Pill.wc.svelte
murdore: Confirmed intentional, not an oversight: `Card.wc.svelte` declares its `attrs` property the exact same way -- `attrs: { type: 'Object' }` with no `attribute`/`reflect` -- and `docs/Card.md` already documents the consequence ("attrs is object-valued... like cssVars"). Pill's doc mirrors that same note ("attrs is object-valued (set as a property...), like Card's").

Reflection is genuinely awkward to add for an object-valued prop -- there is no sane string encoding of an arbitrary attribute bag for `<sui-pill attrs="...">` to parse (unlike a scalar `title` or `as`), so every object-valued prop in this library (`cssVars`, snippets, `attrs`) is property-only by the same convention, not something particular to this PR. I am declining this one as inherent to the prop's shape rather than a gap this PR introduced.

src/wc/components/Pill.wc.svelte
murdore: Same as my reply on the other `attrs`-reflection thread (Pill.wc.svelte): this matches Card.wc.svelte's identical declaration, is already documented in both docs/Card.md and docs/Pill.md as intentional, and reflection has no sane string encoding for an arbitrary object anyway. Declining -- not a Pill-specific gap.

Tara-ag (approved):
Approving. The previous review raised three concerns; the author has addressed the actionable one and provided a reasonable parity argument for the other two.

- ✅ Fixed: interactive pills now have a visible `:focus-visible` outline via `--pill-focus-outline` / `--pill-focus-outline-offset`, matching Card's `.card-interactive:focus-visible` rule.
- ✅ Accepted as established convention: `attrs` is spread directly without runtime string coercion, and `Pill.wc.svelte` declares `attrs` as an object-valued property without reflection — both mirror Card's existing, documented behavior. Introducing Pill-only divergence would be worse for consumers who use both components.
- ✅ Public API is purely additive (`attrs` prop + two new CSS variables); no exports, prop names, event signatures, or relied-upon CSS custom properties were changed.

The implementation is correct: `{...attrs ?? {}}` is placed before Pill's own attributes, so `data-pw`, `class`, `role`, `tabindex`, etc. always win. The collision tests verify this. Docs and web-component parity are updated consistently.

Tara-ag (approved):
## 🛡️ Yama Final Review Verdict: APPROVED

All gate-accepted findings already have inline comments on the pull request. The author has addressed the actionable accessibility gap by adding a `:focus-visible` rule and CSS variables for interactive pills, and has provided reasonable parity-based responses for the items that match Card's existing `attrs` contract. No blocking criteria are triggered: no secrets, no XSS, no breaking public API changes, and no removed exports or CSS variables.

src/lib/Pill/properties.ts
```diff
@@ -40,6 +40,19 @@ export type OptionalPillProperties = {
    */
   leadingIcon?: Snippet;
   classes?: string;
+  /**
+   * Arbitrary attributes spread onto the pill root — `data-*`/`aria-*` (or any
+   * other) attribute names mapped to string values. For consumers that key off
+   * attribute-selector CSS (`[data-state="waiting"]`, `[data-density]`) the way
+   * the rest of their app does, instead of wrapping Pill in an extra element
```
Tara-ag, line 47: 💬 SUGGESTION: Consider narrowing `attrs` to omit keys Pill already manages (e.g. `class`, `role`, `tabindex`, `data-pw`, `title`, `testID`, `onclick`, `onkeydown`, `aria-disabled`). Card currently uses the wider `Record<string, string>`, so matching Card keeps parity, but a narrower `Omit<HTMLAttributes<HTMLDivElement>, ...>` would let TypeScript catch attempts to override managed attributes at compile time.

Tara-ag (approved):
Approved with one suggestion. The change is additive, mirrors Card's existing `attrs` API, preserves backward compatibility, and includes tests + docs. No blocking issues.

src/lib/Pill/properties.ts
Tara-ag, line 47: 💬 SUGGESTION: Consider narrowing `attrs` type to omit keys Pill already manages. Card's `attrs` is typed as `Record<string, string>` and Pill mirrors it. Keeping parity with Card is reasonable, but a narrower type (e.g. `Omit<HTMLAttributes<HTMLDivElement>, 'class' | 'role' | 'tabindex' | 'data-pw' | 'title' | 'testID' | 'onclick' | 'onkeydown' | 'aria-disabled'>`) would let TypeScript catch attempts to override managed attributes at compile time rather than relying solely on runtime spread ordering.

Tara-ag (approved):
Reviewed file-by-file. The change is additive and backward-compatible: Pill gains an `attrs` passthrough prop matching Card's existing API, with managed attributes spread afterward so they cannot be overridden. Web-component parity, tests, demo fixtures, visual baseline, and docs are included. No blocking issues found; one optional type-narrowing suggestion left inline.

murdore:
Declining this one, with reasoning rather than silence — and noting it arrived as two threads with the same content.

The suggestion narrows `attrs` to `Omit<HTMLAttributes<HTMLDivElement>, 'class' | 'role' | ...>` so TypeScript rejects managed keys at compile time. The comment itself grants that "Card currently uses the wider `Record<string, string>`, so matching Card keeps parity", and parity is the reason to stop here:

1. **Divergence is the real cost.** `Card.attrs` is `Record<string, string>` today. Making Pill's stricter means two components with the same prop name and different contracts, so what compiles against Card fails against Pill for no reason a consumer can infer from the API. If this type should be narrower, it should be narrower for both, in one change that can be reviewed as a deliberate break — not introduced asymmetrically through a Pill feature PR.
2. **It would be a breaking change on a `feat`.** `HTMLAttributes` types event handlers as functions, so `Record<string, string>` and the proposed `Omit<...>` are not compatible in either direction. Consumers already passing `attrs` would need code changes. That belongs in a major, not here.
3. **Runtime ordering is not "relying solely on" anything.** Pill spreads `attrs` first and its managed attributes after, so a managed key is overwritten regardless of what the type permits. Compile-time rejection would be a better error message for a case that is already handled correctly.

Filed as a follow-up rather than dropped, so the idea survives: it is a real improvement, it just wants to land across Card and Pill together in a major.

murdore:
Follow-up filed: https://github.com/juspay/svelte-ui-components/issues/576 — narrowing `attrs` across Card and Pill together in a major.

PR #568, opened 2026-09-07
docs(card): stop documenting figcaption inside as="figure" as valid

murdore:
figcaption must be a direct child of figure per the HTML spec. Card
always wraps `children` in an inner .card-content div (and
`title`/`headerRight` in .card-header, `footer` in a <footer>), so a
figcaption placed in any of those slots — as the previous "Aurora"
example did — lands one level too deep and is never exposed as the
figure's accessible caption by any browser or assistive technology.

Replace that example and the matching properties.ts JSDoc with an
accurate description: as="figure" gives the root a real <figure> tag,
nothing more. Show the correct way to get a captioned figure today —
wrap a plain Card in your own <figure>/<figcaption> pair — and note a
first-class caption slot as a known limitation, not something being
added here. No API change; Card.svelte itself is untouched.

## Summary by CodeRabbit

- **Documentation**
  - Clarified that using `as="figure"` on Card does not support a direct `<figcaption>` caption.
  - Updated examples to show wrapping Card with an outer `<figure>` and `<figcaption>` when a caption is needed.
  - Added the same guidance to the Card property documentation.

### DIFF
--- a/docs/Card.md
+++ b/docs/Card.md
@@ -180,8 +180,8 @@ This pattern gives full layout control to the consumer (any number of zones, any
 consumer that already keys state off attribute-selector CSS (`[data-state="waiting"]`,
 `[data-density="compact"]`) elsewhere in their app, instead of encoding the same state as
 a second `classes` modifier. `as` overrides the rendered tag independent of `href`, for a
-card that needs real `<figure>` semantics (so a `<figcaption>` inside it is valid markup)
-without a separate wrapper element around an appearance-only Card.
+card that needs the root element itself to carry real `<figure>` semantics instead of a
+`<div>`.
 
 ```svelte
 <script>
@@ -193,10 +193,9 @@ without a separate wrapper element around an appearance-only Card.
   <p>Waiting for input…</p>
 </Card>
 
-<!-- Card's root IS the <figure> -- no wrapper element needed for the semantics -->
+<!-- Card's root IS the <figure> — no wrapper element needed for the tag itself -->
 <Card as="figure" title="Aurora">
   <div class="stage">…</div>
-  <figcaption>Aurora backdrop, 40% opacity</figcaption>
 </Card>
 ```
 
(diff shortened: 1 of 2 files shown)


Tara-ag (approved):
Documentation-only correction. Verified that Card.svelte wraps `children` in `.card-content`, `title`/`headerRight` in `.card-header`, and `footer` in `<footer>`, so a `<figcaption>` passed through any slot cannot be a direct child of a `<figure>` root. The corrected guidance and workaround example are accurate; no public API, type, export, or CSS custom-property changes are introduced. Approved.

Tara-ag (approved):
Approved. This is a documentation-only correction: it removes the inaccurate claim that `<figcaption>` inside a Card with `as="figure"` produces valid figure caption semantics, and explains the correct pattern (wrapping Card in an outer `<figure>`/`<figcaption>`). No component code, props, exports, or CSS custom properties are changed, so there is no public API or backward-compatibility impact.

Tara-ag (approved):
Documentation-only change that correctly removes a misleading `as="figure"` + `<figcaption>` example and explains why it violates the HTML spec (figcaption must be a direct child of figure). Verified against `Card.svelte`: children are indeed wrapped in `.card-content`, so the old example would not have produced valid figure/figcaption semantics. No component code, props, exports, or CSS custom properties are changed; no breaking changes; no security or accessibility regressions. Approved.

murdore:
Good catch, and correct — fixed in `7a2e90f`.

The sentence said "Do not put a `<figcaption>` inside a Card, at any nesting", which over-reaches. The defect being documented is that Card wraps `children` in `.card-content`, so a `<figcaption>` passed through a slot is never a *direct* child of the `as="figure"` root and is therefore not exposed as that figure's caption. That is a statement about captioning **Card's own root**, not about the element.

A `<figure>` nested inside a Card is a different element, and a `<figcaption>` that is its direct child is valid and correctly exposed. The old wording told readers to avoid a perfectly good pattern.

Reworded to prohibit only the root case, with the nested case called out explicitly so the distinction is not left to inference. `build=0 check=0 lint=0`.

Tara-ag (approved):
Approved. This is a documentation-only correction: it accurately explains that `as="figure"` does not support a direct `<figcaption>` because Card wraps slot content in inner elements, and it provides the correct workaround (wrap Card in an outer `<figure>`/`<figcaption>`). No component code, public API, exports, or CSS custom properties are changed.

PR #569, opened 2026-09-07
fix(sheet): name the dismissible overlay and restore the panel's focus ring

murdore:
Two accessibility defects have shipped on every Sheet side (left, right, top,
bottom, and center) since 4.0.0, found while reviewing an unrelated feature PR
and deliberately not fixed there so that additive PR wouldn't silently change
already-shipped behavior. This is that dedicated fix.

What changes:

1. The overlay carried role="button" unconditionally, with no accessible
   name -- a role with no name is announced as an unlabelled button. It now
   only carries role="button" (plus an aria-label, default "Close sheet",
   overridable via the new `overlayAriaLabel` prop) while it is actually
   dismissible (`dismissOnOutsideClick`). When dismissOnOutsideClick is
   false, clicking the overlay is already a no-op, so it now carries no role
   at all instead of announcing a button that does nothing -- the same fix
   already applied to Modal's overlay for the identical problem.

2. `.sheet-panel` set `outline: none` unconditionally. The panel is the
   sheet's initial focus target on open (and its only one on the "Raw"
   variant, which has no focusable content), so this left keyboard users
   with no visible sign that focus had moved into the sheet at all. A
   `.sheet-panel:focus-visible` rule (new `--sheet-panel-focus-outline` /
   `--sheet-panel-focus-outline-offset` tokens) restores a ring for
   keyboard/programmatic focus without adding one back for a mouse-opened
   sheet, matching the :focus-visible pattern already used elsewhere in this
   library (Card, StatCard, Draggable, FileInput, etc.).

What stays the same: every existing prop, CSS variable, and default
(dismissOnOutsideClick=false keeps its prior fully-unlabelled, roleless
overlay markup shape apart from losing role="button"; a mouse-opened sheet
still shows no ring). `overlayAriaLabel` is new and optional; the web
component wrapper gains the matching `overlay-aria-label` attribute for
parity.

Adds tests/sheet-overlay-a11y.test.ts, covering both fixes across the
right, left, and bottom sides (role/name for dismissible, non-dismissible,
overridden, and invisible-but-dismissible overlays; focus ring present on
keyboard-triggered opens and absent on mouse-triggered opens). Existing
Sheet specs (sheet-anchor-and-overlay, sheet-centered-variant) still pass
unmodified.


Tara-ag (approved):
Approved. The changes are focused, additive accessibility fixes with no breaking changes to the public API, CSS custom properties, or default behavior. Tests and docs are updated accordingly.

Tara-ag (commented):

src/lib/Sheet/Sheet.svelte
```diff
@@ -132,7 +142,8 @@
       : 'overlay-noninteractive'} {classes ?? ''}"
```
Tara-ag, line 164: ⚠️ MAJOR: The dismissible overlay now correctly receives `role="button"` and `aria-label`, but it only handles `onclick`. A `role="button"` element must be activatable with Enter/Space (WCAG 2.1 SC 2.1.1). Even though `tabindex="-1"` keeps it out of the Tab sequence, assistive-tech users navigating in virtual/browse mode can still reach it and expect activation to work. Modal's overlay solves the same problem with a dedicated `handleOverlayKeyDown` that fires on Enter/Space when `event.target === overlayDiv`.

Suggested fix: add an overlay-scoped keydown handler (or extend the existing handler with a target guard) that calls `close()` on Enter/Space when `dismissOnOutsideClick` is true, keeping it separate from the window-level Escape/Tab trap so it doesn't double-fire.

tests/sheet-overlay-a11y.test.ts
```diff
@@ -0,0 +1,122 @@
...
+//    (dismissOnOutsideClick), mirroring how Modal's overlay already handles
+//    the same shape of problem.
+// 2. `.sheet-panel` set `outline: none` unconditionally, which left a
+//    keyboard user with no visible sign that focus had moved into the panel
+//    on open (the panel is the initial focus target, and the only one on the
+//    "Raw" variant). A `:focus-visible` rule restores the ring for
+//    keyboard/programmatic focus without adding one for a mouse-opened sheet.
+test.describe('Sheet overlay accessible name', () => {
+  test('a dismissible overlay (side="right", default dismissOnOutsideClick) is a named button', async ({
+    page
```
Tara-ag, line 17: 💡 MINOR: The new specs select triggers with `page.getByText('Open right', ...)` and `page.getByRole('button', { name: 'Open right' })`. This couples the tests to the demo page's copy and layout; changing the gallery button text or reordering examples will break unrelated accessibility assertions.

Suggested fix: add stable `data-pw` attributes to the demo trigger buttons (e.g. `data-pw="open-sheet-right"`) and select them with `page.getByTestId(...)`. Alternatively, create an isolated fixture route under `/tests/fixtures/sheet-a11y` that renders only the variants under test.

tests/sheet-overlay-a11y.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    // keyboard input modality -- the precondition :focus-visible checks before
+    // showing a ring for the panel's subsequent programmatic focus() call.
+    // getByRole (not getByText) targets the actual <button> element -- Enter
+    // only triggers a synthetic click on the real interactive element, not on
+    // the text node inside it.
+    await page.getByRole('button', { name: 'Open right', exact: true }).press('Enter');
+    const panel = page.getByTestId('sheet-right-panel');
+    await expect(panel).toBeVisible();
+    await page.waitForTimeout(400);
+
```
Tara-ag, line 88: 💡 MINOR: The focus-ring tests use `await page.waitForTimeout(400)` to wait for the sheet open transition. Fixed timeouts slow the suite and flake on slower CI runners. The component already fires `onafteropen` once the intro transition ends, and the panel emits `transition:fly|global` events; either can be awaited deterministically.

Suggested fix: replace the fixed timeout with an explicit wait for the transition to finish (e.g. wait for `transitionend` on the panel, or for `onafteropen` to fire) before reading computed styles.

src/wc/components/Sheet.wc.svelte
```diff
@@ -7,6 +7,7 @@
       side: { type: 'String', reflect: true },
       sheetTitle: { type: 'String', reflect: true, attribute: 'title' },
       showOverlay: { type: 'Boolean', reflect: true, attribute: 'show-overlay' },
+      overlayAriaLabel: { type: 'String', attribute: 'overlay-aria-label' },
```
Tara-ag: 💡 MINOR: `overlayAriaLabel` is declared with `attribute: 'overlay-aria-label'` but omits `reflect: true`. Every other scalar Sheet prop (`side`, `sheetTitle`, `showOverlay`, `showCloseButton`, `headingLevel`) reflects to the host attribute. Reflection lets light-DOM consumers target the element via CSS attribute selectors and keeps the host attribute in sync when the property is set imperatively. The PR description itself says the wrapper gains the matching attribute "for parity".

Suggested fix: add `reflect: true`:
```js
overlayAriaLabel: { type: 'String', reflect: true, attribute: 'overlay-aria-label' }
```

Tara-ag (changes requested):
One MAJOR accessibility issue and three MINOR quality notes. The MAJOR issue blocks merge: the newly-labelled dismissible overlay is a `role="button"` but has no Enter/Space keyboard activation, failing WCAG keyboard requirements. The MINOR items are web-component reflection parity and test robustness/coupling improvements.

Tara-ag:

## 🤖 Yama Review Summary

⚠️ The review loop ended early (`time-limit`, 20 steps) — this verdict is built strictly from gate-verified findings.

murdore:
All four findings addressed in `f21d6cf` (rebased onto `69958f7`).

**MAJOR — Enter/Space on the `role="button"` overlay.** Fixed, and the finding was right that the obligation exists. One correction to the reasoning: the overlay `onkeydown` is bound only to the overlay element, not to window, so there was no window-level trap to double-fire against — extending the existing handler with a target guard was safe and is what I did. The guard is `event.target === overlayDiv`, mirroring what `handleOverlayClick` already does for the click case.

Worth noting the Modal parallel is not exact: Modal's overlay is `tabindex={overlayDismissible ? 0 : null}`, so it is genuinely in the Tab sequence. Sheet's is `tabindex="-1"` unconditionally, because it doubles as the focus trap's fallback target. I kept `-1` — putting it in the tab order would change Tab behaviour for every existing consumer. The handler is still reachable (the trap parks focus there, and browse-mode AT reaches it regardless), so it is not dead code. Two specs cover it, and the activation one fails without the handler — verified by reverting it.

**MINOR — `reflect: true` on `overlayAriaLabel`.** Fixed. This one was better than it looked: `docs/Sheet.md` already described the attribute as "reflecting `overlayAriaLabel`", so the doc was asserting something the wrapper did not do. It does now.

**MINOR — test selectors coupled to demo copy.** Fixed. `data-pw` hooks on the five triggers the specs drive, selected via `getByTestId`. This is the house convention rather than a preference — 110 spec files use `getByTestId` against `testIdAttribute: 'data-pw'`.

**MINOR — `waitForTimeout(400)`.** Fixed. Replaced with `await expect(panel).toHaveCSS('transform', 'none')`. `fly` animates an inline transform and Svelte drops that inline style when the intro ends; left/right and top/bottom panels have no resting transform, so the settled value is a real completion signal Playwright retries on, rather than an approximation.

Gates after rebase: `build=0 check=0 lint=0`, 9/9 specs pass.

One thing I did **not** do: the `visual` baseline mismatches locally (1923px vs the committed 1934px) — but identically on a clean tree with my changes stashed, so it is a local-vs-CI rendering difference, not a regression. CI's `visual` check passes against the committed baseline, so regenerating it locally would have broken CI.

Tara-ag (commented):
Review in progress

src/lib/Sheet/Sheet.svelte
```diff
@@ -132,7 +159,8 @@
...
     onclick={handleOverlayClick}
     onkeydown={handleKeyDown}
-    role="button"
+    role={dismissOnOutsideClick ? 'button' : null}
+    aria-label={dismissOnOutsideClick ? (overlayAriaLabel ?? 'Close sheet') : null}
```
Tara-ag: 💡 MINOR: Empty `overlayAriaLabel` renders a blank `aria-label` instead of falling back to the default. When a consumer passes `overlayAriaLabel=""` (or whitespace-only), `??` only skips `null`/`undefined`, so the overlay exposes `aria-label=""` and loses its accessible name despite the documented default. Treat empty/whitespace-only values as absent:

```svelte
aria-label={dismissOnOutsideClick ? (overlayAriaLabel?.trim() || 'Close sheet') : null}
```

Tara-ag (commented):
Approved with one minor robustness note on empty `overlayAriaLabel` values.

Tara-ag (approved):
Approved with one minor robustness note on empty `overlayAriaLabel` values.

Tara-ag (commented):
Review in progress.

Tara-ag:
💡 MINOR: Empty `overlayAriaLabel` renders a blank `aria-label` instead of falling back to the default

**File:** `src/lib/Sheet/Sheet.svelte` (line 163)

When a consumer passes `overlayAriaLabel=""` (or whitespace-only), the expression `overlayAriaLabel ?? 'Close sheet'` evaluates to the empty string because `??` only skips `null`/`undefined`. The overlay then exposes `aria-label=""`, leaving the dismissible overlay with no accessible name despite the documented default.

**Suggested fix:** Treat empty/whitespace-only values as absent so the default accessible name is preserved:

```svelte
aria-label={dismissOnOutsideClick ? (overlayAriaLabel?.trim() || 'Close sheet') : null}
```

murdore:
Second round addressed in `b4237d4`. The empty-`overlayAriaLabel` finding was correct and is now fixed — `overlayAriaLabel?.trim() || 'Close sheet'`.

Worth writing down how this was verified, because my first two attempts to prove it were both wrong and passing tests nearly hid that:

1. I first covered it through `<sui-sheet overlay-aria-label="">`. Three specs, all green. **They proved nothing** — reverting the fix to `??` left them green too. An empty attribute reaches Sheet as `undefined` through the custom-element layer, so the wc path cannot express "explicitly passed an empty string" at all. Those specs are deleted, not kept as filler.
2. I then wrote a component spec and ran the same check — still green under `??`. That one was my own error: prettier had reformatted the line to drop a pair of parentheses, so the `sed` that was supposed to revert the fix silently matched nothing and I was testing the fixed code against itself.

With the revert actually applied, the component spec fails exactly as the finding describes:

```
AssertionError: expected '' to be 'Close sheet'
AssertionError: expected '   ' to be 'Close sheet'
```

and the two control cases (`'Fermer'`, no override) pass either way — so the spec is sensitive to this defect and only this defect.

It lives in `src/lib/Sheet/Sheet.svelte.test.ts` rather than the demo page deliberately: a new demo variant would grow the Sheet gallery and move the visual baseline for an edge case. It needs a small `Element.prototype.animate` stub because Svelte 5 drives transitions through the Web Animations API and jsdom has none — kept local to the file rather than added to `vitest-setup.ts`, so it cannot mask an animation fault in some other suite.

Gates: `build=0 check=0 lint=0`, 9/9 Playwright, 4/4 unit.

Tara-ag (approved):
Approved. Independently verified the current revision:

- The dismissible overlay now carries `role="button"` and `aria-label` only when `dismissOnOutsideClick` is true, with a trimmed fallback to "Close sheet".
- Enter/Space keyboard activation is implemented with a target guard matching the click handler.
- The web-component wrapper reflects `overlay-aria-label` to the host attribute.
- Panel focus ring is restored via `:focus-visible` without affecting mouse-triggered opens.
- Unit and Playwright tests cover the new behavior, including the empty/whitespace `overlayAriaLabel` edge case.
- No breaking changes to public props, exports, event signatures, or CSS custom-property names.

All previous review findings have been addressed and CI is green.

PR #570, opened 2026-09-07
feat(modal): forward aria-label and role to sui-modal's content panel

murdore:
role and aria-label set on <sui-modal> land on the custom-element host --
the ARIAMixin accessors every element already has -- not on
.modal-content, the panel two levels inside the shadow tree that
actually carries aria-modal and the accessible name. The host wraps the
whole full-screen overlay, so the attribute looks wired and isn't:
assistive tech finds an unnamed, unroled panel.

Add modalAriaLabel/modalRole (attributes modal-aria-label/modal-role),
the same prefixed-alias pattern Toggle.wc.svelte already uses for
inputAriaLabel, forwarded to the inner Modal's ariaLabel/role. Purely
additive: a consumer setting neither alias sees the panel exactly as
before, and the host's own role/aria-label are untouched.

Separate from other work on this branch because it's an unrelated
web-component wrapper fix, not part of the same change.

### DIFF
--- a/docs/Modal.md
+++ b/docs/Modal.md
@@ -37,6 +37,11 @@ A dialog overlay component that renders on top of the page with configurable siz
 | ariaLabel                | `string`                                                                                                                                                  | No       | `-`             | Accessible name for the modal, rendered as `aria-label` on the content panel (the element carrying `role`/`aria-modal` when `role` is set). Without it a screen reader announces an unnamed dialog.                                                                                                                                                                                             |
 | role                     | `ModalRole = 'dialog' \| 'alertdialog'`                                                                                                                   | No       | `-` (unset)     | ARIA role for the content panel. Unset by default — the panel renders no `role` and no `aria-modal`, matching behavior before this prop existed, so an existing consumer who never touches `role` sees no DOM/ARIA change. Set explicitly to opt in: the panel then also carries `aria-modal="true"`. Use `'alertdialog'` for a dialog that interrupts to demand an immediate response (e.g. a destructive confirmation).                                                                                                                                                                                             |
 
+`<sui-modal>` adds two props that do not exist on the Svelte component: `modalRole` and
+`modalAriaLabel` (attributes `modal-role` / `modal-aria-label`). They are forwarding aliases
+for the `role` and `ariaLabel` rows above, and exist only because `role` and `aria-label` are
+reserved accessors on every custom element. See the Web Component section below.
+
 ## Snippets
 
 Svelte 5 Snippet props — pass content blocks to the component.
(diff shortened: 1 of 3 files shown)


Tara-ag (approved):
Approved. The change is purely additive: it introduces `modalAriaLabel`/`modalRole` aliases on the `sui-modal` web-component wrapper so consumers can name the inner `.modal-content` panel without colliding with the host's native `aria-label`/`role` accessors. Existing behavior is preserved when the aliases are unset, the pattern matches the established `inputAriaLabel` approach in `Toggle.wc.svelte`, and the new Playwright tests plus docs cover the behavior.

docs/Modal.md
```diff
@@ -248,7 +248,15 @@ Tag: `<sui-modal>`
 </sui-modal>
```
Tara-ag, line 253: 💬 SUGGESTION: `modalRole`/`modalAriaLabel` are web-component-only aliases, so they're correctly explained here, but a reader scanning the props table above will miss them. Consider adding a note or dedicated row indicating these are `<sui-modal>`-only props.

tests/modal-wc-aria-forwarding.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+    page
+  }) => {
+    await loadBundle(page);
+
+    await page.evaluate(() => {
+      const modal = document.createElement('sui-modal');
+      modal.setAttribute('data-pw', 'wc-modal-bare');
+      document.body.append(modal);
+    });
+
```
Tara-ag, line 84: 💡 MINOR: This assertion encodes the pre-existing default where the panel is unnamed/unroled, but it doesn't prove the additive guarantee the PR claims for consumers who previously set `role`/`aria-label` directly on the host. Consider adding a test that shows host-level `role`/`aria-label` still leave the panel unnamed, or rename this test to clarify it verifies the alias default-off state rather than "prior behavior".

Tara-ag (approved):
Approved with minor notes. The change is purely additive: it introduces `modalRole`/`modalAriaLabel` aliases on `<sui-modal>` that forward to the inner panel's `role`/`aria-label`, fixing the host-reserved accessor problem without touching existing props, exports, or CSS custom properties. Tests and docs are included. Two non-blocking notes left inline.

tests/modal-wc-aria-forwarding.spec.ts
Tara-ag, line 84: 💡 MINOR: Bare panel test asserts absence of aria-modal

The third Playwright test asserts that when neither modalRole nor modalAriaLabel is set, the panel has no role, no aria-label, and no aria-modal. However, Modal.svelte gates aria-modal on `role != null`, so if role is absent aria-modal is also absent. The assertion is technically correct today, but it encodes the pre-existing (and accessibility-wise undesirable) default where the panel is unnamed and unroled. More importantly, the test does not verify the additive/no-op guarantee for consumers who previously set role/aria-label directly on the host: those consumers still get a broken experience. This is fine as a regression guard, but the test name claims it matches prior behavior and the assertion on aria-modal is redundant given the role check.

Suggestion: Keep the regression assertion but consider adding a fourth test that proves setting role/aria-label directly on the host still leaves the panel unnamed (documenting the limitation the PR fixes), or rename the test to clarify it verifies the alias default-off state rather than 'prior behavior'.

docs/Modal.md
Tara-ag, line 253: 💬 SUGGESTION: Docs do not list modalRole/modalAriaLabel in props table

The PR adds modalRole and modalAriaLabel as web-component-only forwarding aliases, but the docs/Modal.md props table (which lists Svelte props) does not mention them. The web component section explains them, which is good, but a reader scanning the props table will miss them. Since these are wrapper-only props, this is a minor documentation inconsistency rather than a blocking issue.

Suggestion: Add a note or separate row in the props table indicating these are sui-modal-only aliases, or add a dedicated 'Web Component Props' subsection.

Tara-ag (approved):
Approved with two non-blocking notes. The change is purely additive, preserves the existing Svelte and web-component public API, and follows the established host-reserved-accessor workaround pattern.

murdore:
Both addressed in `c4c0cd3`.

**Props table.** Fair — a reader scanning the table would not learn these exist. Added a note under it naming `modalRole`/`modalAriaLabel` as `<sui-modal>`-only forwarding aliases for the `role`/`ariaLabel` rows, pointing at the Web Component section.

Note I wrote that pointer as a `[Web Component](#web-component)` anchor first and `pnpm build` rejected it — SvelteKit prerenders these docs as `/components/modal` and validates internal anchors, and no such id exists there. Plain prose instead. Flagging it because `check` and `lint` both pass on that link; only `build` catches it.

**Test naming / additive guarantee.** The stronger half of this is right: the third test only pinned the alias default-off state, and the PR's additive claim was stated but never asserted. Renamed it to say what it actually covers ("the shipped default"), and added a fourth test that pins the claim directly — `role` and `aria-label` set on the host stay on the host, and still do not reach `.modal-content`. It asserts both halves, including that the attributes really are present on the host, since that is exactly what makes the trap convincing.

One thing I did not adopt: the framing that those consumers "still get a broken experience". They get precisely what they got before this PR — that is what makes it additive. This PR adds a spelling that works; it does not regress the one that never did. The new test is the evidence.

Gates: `build=0 check=0 lint=0`, 4/4 specs. This suggestion also arrived twice as two threads.

Tara-ag:

Review could not be completed: no PR diff or repository files were fetched before the time budget was exhausted. Without examining the changed files, cross-referencing callers, or gating findings through submit_review, I cannot verify whether the PR meets the project's public-API, accessibility, reactivity, SSR, security, or testing standards. The PR should be re-reviewed with sufficient time to inspect each changed file.

PR #577, opened 2026-09-07
fix(release): abort a release whose build is already stale, and clear the MCP advisories

murdore:
Two changes in one commit, at the repository owner's direction. They ride together because the MCP half is what exercises the publish path the first half protects.

## 4.10.4 shipped without a commit it claims to contain

The root release job builds from its own checkout — pinned to the SHA that triggered it — and then rebases the release commit onto whatever has landed since, before pushing. A commit arriving in that window ends up **under** a landmark for a version built without it.

That is not hypothetical. Five commits merged in under two minutes today; the run triggered by the first built after four of them had landed, rebased its release commit above all five, and published:

| commit | in the published 4.10.4 tarball |
|---|---|
| `4ba6957` fix(accordion) | ✅ |
| `d4c8a91` feat(pill) · `c6f0382` docs(card) | ✅ |
| `6162f5f` fix(sheet) — `overlayAriaLabel` | ✅ |
| `754694c` **feat(modal)** — `modalAriaLabel` | ❌ |

`chore(release): 4.10.4` sits above all of them. So `feat(modal)` is recorded as released, is absent from the registry, and — because the landmark has moved past it — its `feat` can never reach a later bump. It will ship inside whatever patch comes next.

That is the 3.5.1 failure again, a version understating what is in it, reached by a different route: not the version derivation, which is correct, but the gap between what was built and where the landmark landed.

## The fix

Both jobs now check for commits that landed during the run and abort **before** writing anything — the last point at which stopping is free.

`publish-mcp` got this in #571. The root job has the identical window and I left it alone, which was an omission rather than a decision: I argued its git-writes-first *ordering* was correct — it is — and did not notice that the rebase window is a separate defect present in both.

The root check excludes `mcp/` and `.github/` for the same reason `detect-changes` does: neither is part of this package's build, so neither makes its artifacts stale.

### Verified against the real history, not assumed

```
git rev-list b941104..8df18d4 -- . ':(exclude)mcp/' ':(exclude).github/'   → 6
git rev-list 1811375^..1811375 -- . ':(exclude)mcp/' ':(exclude).github/'  → 0
```

The first is the 4.10.4 range: the guard would have fired. The second is #571's workflow-only merge: a release that only changes CI is not blocked by its own commit.

## The MCP half

`npm audit fix` on the MCP lockfile, clearing **5 advisories — 2 high, 3 moderate** — in `hono`, `ip-address` and `qs`. `npm audit` now reports 0, and `npm run build` passes.

It also resyncs the lockfile's own version, which said `4.0.2` while `mcp/package.json` said `4.0.3`: the failed release run bumped the manifest, committed it, and never updated the lockfile beside it.

## Why this is the run to watch

Trusted publishing was configured for `@juspay/svelte-ui-components-mcp` today. Until now that job had **never** published — 4.0.1 and 4.0.2 were both published by hand, which is what hid the `ENEEDAUTH` failure. This should be the first MCP release that publishes from CI, and it carries out the `body-parser` CVE fix that #419 merged and never shipped.

Touching only `.github/` and `mcp/` means `detect-changes` reports `root_changed=false`, so **no root version is cut by this**.

## Scope

Two files: `.github/workflows/release.yml` and `mcp/package-lock.json`. No runtime, test or source code is touched, so I have leaned on CI for the suites rather than claiming local runs that would prove nothing about this change.

## Summary by CodeRabbit

- **Chores**
  - Release publishing now checks for new root-package changes that arrive during the release process.
  - If relevant changes are detected, the release is stopped before creating or publishing a release commit.
  - Changes limited to MCP-related or workflow configuration areas are excluded from this check.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -366,14 +366,53 @@ jobs:
       # the registry stays clean. The reverse failure (pushed, then publish fails)
       # leaves a version commit for an unpublished version, which the existing
       # `npm view ... && exit 0` guard below makes safe to resolve by re-running.
+      # The artifacts were built from this run's checkout, which is pinned to the
+      # SHA that triggered it, but the push below rebases the release commit onto
+      # whatever has landed since. A commit arriving in that window ends up UNDER
+      # a landmark for a version built without it: the repository records it as
+      # released, the tarball does not contain it, and because the landmark has
+      # moved past it its `feat` or `fix` can never reach a later bump.
+      #
+      # 4.10.4 is that, already shipped. Five commits merged in under two minutes;
+      # the run triggered by the first built after four of them had landed, then
+      # rebased its release commit above all five and published. The tarball has
+      # `overlayAriaLabel` from the fourth commit and not `modalAriaLabel` from
+      # the fifth, while `chore(release): 4.10.4` sits on top of both -- so a
+      # `feat(modal)` is recorded as released, is absent from the registry, and
+      # will ship inside whatever patch comes next.
+      #
+      # The check lives inside the push step, between its fetch and its rebase,
+      # rather than in a step of its own. A separate step would have to do its
+      # own fetch, and this one would then fetch again -- so a commit landing in
+      # between would pass the guard and still be rebased over, which is the bug
+      # with extra ceremony. Adjacent to the fetch it depends on, the window is
+      # the two lines below.
+      #
+      # `mcp/` and `.github/` are excluded for the same reason detect-changes
+      # excludes them: neither is part of this package's build, so neither makes
+      # these artifacts stale, and a CI-only release must not block on its own
+      # commit.
       - name: Push changes (without tags)
         run: |
           # Pull latest changes to avoid conflicts
           git fetch origin release
 
-          # Check if we're behind
+          STALE_AGAINST=". :(exclude)mcp/ :(exclude).github/"
+          LANDED="$(git rev-list HEAD..origin/release -- $STALE_AGAINST)"
+          if [ -n "$LANDED" ]; then
+            echo "::error::Commits landed on release while this run was building:"
+            git log --oneline HEAD..origin/release -- $STALE_AGAINST
+            echo "They are not in the artifacts this run built, and rebasing onto them"
+            echo "would put the release commit above code it does not contain."
+            echo "Aborting before anything is written, so the next run publishes a"
+            echo "version that contains them and derives its bump from them."
+            exit 1
+          fi
+
+          # Only refs this build is unaffected by can be below us here, so the
+          # rebase can no longer move the release commit over unbuilt code.
           if [ $(git rev-list --count HEAD..origin/release) -gt 0 ]; then
-            echo "Local branch is behind remote. Attempting to rebase..."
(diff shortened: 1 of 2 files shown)


Tara-ag (approved):
Approved. The change is workflow-only and lockfile maintenance with no runtime/source impact. The stale-build guard correctly closes the 4.10.4 class of failure by aborting before any remote git writes, and the exclusion of `mcp/` and `.github/` matches the existing `detect-changes` semantics. The remaining race between the freshness check and the subsequent push is acknowledged and is an acceptable best-effort limitation for this fix. The `npm audit fix` updates are standard transitive dependency patches with no source changes.

murdore:
Accepted, and it is the sharper version of the point. A guard in its own step still leaves the window it exists to close.

The push step does its own `git fetch origin release` before rebasing. So a commit landing between my separate check and that fetch would pass the guard and be rebased over regardless — the same bug, with a step of ceremony in front of it. The check is only worth anything adjacent to the fetch it depends on.

It now lives inside the push step, between the fetch and the rebase:

```yaml
git fetch origin release

STALE_AGAINST=". :(exclude)mcp/ :(exclude).github/"
LANDED="$(git rev-list HEAD..origin/release -- $STALE_AGAINST)"
if [ -n "$LANDED" ]; then
  echo "::error::Commits landed on release while this run was building:"
  git log --oneline HEAD..origin/release -- $STALE_AGAINST
  exit 1
fi

# Only refs this build is unaffected by can be below us here.
if [ $(git rev-list --count HEAD..origin/release) -gt 0 ]; then
  git rebase origin/release
fi
```

The window is now the two lines between the fetch and the rebase, rather than a step boundary with a build and a network round-trip in it.

## Same treatment for the MCP job

The identical second-look problem is in `publish-mcp`'s commit-and-tag step, which fetches again *after* the publish. A commit arriving between the pre-publish guard and that fetch would be rebased under a landmark for a version published without it. It now fails there instead of rebasing.

Failing at that point is the recoverable outcome, deliberately: the version is on the registry with no landmark, and the `npm view … && exit 0` guard already in the publish step lets a re-run write the landmark rather than republish. The unrecoverable outcome is the one being avoided — a landmark sitting above code the tarball does not contain.

## Verified against real history

Three controls on the actual branch, using the variable exactly as the step expands it:

```
git rev-list b941104..8df18d4   -- $STALE_AGAINST → 6   4.10.4 range — fires
git rev-list 1811375^..1811375  -- $STALE_AGAINST → 0   #571, CI-only — does not
git rev-list 5a42c2c^..5a42c2c  -- $STALE_AGAINST → 0   #419, mcp-only — does not
```

The first is the range that produced the stale 4.10.4, so the guard would have stopped it. The other two are the cases that must *not* block: a release that only changes CI, and a root run while an MCP-only commit lands.

## On the zizmor note

`excessive-permissions` on the workflow having no top-level `permissions:` block is a real observation, but it is pre-existing and repository-wide rather than something this change introduces — every job here already declares its own `permissions`, and narrowing the file-level default is a separate change with its own blast radius across five jobs.

Tara-ag (approved):
No issues found. The stale-build guard correctly closes the rebase-window race for both the root and MCP release jobs, and the lockfile audit fix restores the MCP lockfile version sync.

PR #579, opened 2026-09-08
fix(typewriter): honour prefers-reduced-motion, matching ChatController

### LINKED ISSUE #465: TypewriterText / ChatController: two independent char-reveal implementations, no cross-reference
## Context
Two components implement the same idea — reveal text progressively, the way a streaming model answer reads — with separate, hand-written timers and no shared code:
- `ChatController.feed()`/`.tick()`/`.drain()`/`.stopReveal()` (`src/lib/Chat/controller.svelte.ts:170-213`, plus constants at 9-10, fields at 41-42, `prefersReducedMotion()` at 12-18): buffers incoming transport chunks and drains 2 chars per 22ms `setInterval` tick into `msg.content` (~91 chars/sec), bypassed entirely under `prefers-reduced-motion: reduce`. ~54 lines.
- `TypewriterText.svelte` (`src/lib/TypewriterText/TypewriterText.svelte`, state at 14-19, `typeNextCharacter` at 21-27, `$effect`s at 33-61/64-72, lifecycle at 74-85): types 1 char per `speed` ms (default 15ms, ~67 chars/sec) via a recursive `setTimeout` chain, with continuation-vs-replacement detection and an immediate snap-to-full-text when `isStreaming` turns `false`. No `prefers-reduced-motion` handling at all. ~60 lines.

~114 LOC of independently-written reveal logic, with materially different behavior on cadence, reduced-motion, append-vs-replace handling, buffer scope (shared per-controller vs. per-instance), and end-of-stream flush.

`TypewriterText` was added four days after `ChatController`'s reveal loop already existed (`aa5f30a`, "feat: add four chat primitives — ThinkingIndicator, TypewriterText, AttachmentChipRow, HITL", 2026-08-24, vs. `a18c497`, "feat: add chat components", 2026-08-20), and its own doc tagline is "Reveals text one character at a time, the way a streaming model answer reads" (`docs/TypewriterText.md:3`) — written for exactly the use case `ChatController` already covers. Nothing wires it in: `ChatMessage.svelte` renders `msg.content` as plain text with no `TypewriterText` import anywhere under `src/lib/Chat*`, and `docs/Chat.md` documents `ChatControllerOptions.typewriter` (line 74, used in the flagship example at line 34) with zero mention of `TypewriterText` in either direction.

**Concrete failure mode if combined today:** set `typewriter: true` on the controller *and* pass a `messageBody` snippet wrapping `msg.content` in `<TypewriterText text={msg.content} isStreaming={msg.streaming} />` — a documented, supported pattern (`docs/Chat.md:80-90`). `msg.content` is already growing character-by-character from the controller's own reveal; `TypewriterText`'s continuation branch re-types every already-revealed increment on its own cadence on top of that — a compounding double-reveal, not just wasted work.

## Consolidation options considered
1. **Extract a shared headless reveal primitive** (push/tick/flush/cancel) used by both. Low-medium implementation risk but real design risk: the two call sites differ on cadence model, replace-vs-append semantics, reduced-motion handling, and lifecycle (Svelte `$effect`/`onMount` vs. a plain runes class with no component lifecycle) — a wrong abstraction can silently change either side's shipped, documented behavior, and this repo has no unit-test tier to catch it (only Playwright E2E).
2. **Delete `ChatController`'s built-in reveal; standardize on composing `TypewriterText` via `messageBody`.** Low implementation risk, high breaking-change risk: `ChatControllerOptions.typewriter` is public, documented, and used in the primary copy-pasteable example (`docs/Chat.md:34`) — removing/no-opping it silently drops the reveal effect for any existing consumer who set it without a custom `messageBody`. It would also promote `TypewriterText` to the canonical path while it still lacks `ChatController`'s reduced-motion handling — an accessibility regression unless fixed first.
3. **Docs-only: cross-link the two, document the divergence, warn against combining them.** Zero regression risk, ships immediately, but leaves the ~114 LOC duplicated and drifting — the cadence and reduced-motion gaps above already prove they drift when only one side gets fixed.

## Recommendation
Short term, do (3) to stop the silent double-reveal failure mode and document which primitive to reach for. Track (1) or (2) as a follow-up once there's a documented reduced-motion requirement to hold both implementations to.

murdore:
Closes #465.

Two components in this library reveal text progressively, and only one of them answered the reader's motion preference.

`ChatController` bypasses its drain loop entirely when `prefers-reduced-motion: reduce` is set. `TypewriterText` — which exists to do the same job, and whose own tagline is *"reveals text one character at a time, the way a streaming model answer reads"* — had **no such handling at all**. The same OS setting silenced the reveal in a chat bubble and left it running in a message body rendered inside that very bubble, which is a supported composition: `Chat`'s `messageBody` snippet.

## What changed

Typing *is* the animation here, so honouring the setting means suppressing the reveal itself rather than a transition around it. Under `reduce`, text is now disclosed as it arrives; `speed`, `variableDelay` and `resolveDelay` are not consulted. `onprogress` still fires, once per disclosure instead of once per character — the same contract the existing `isStreaming: false` bulk path already had.

The predicate moves to `src/lib/utils.ts` and **both** implementations import it, rather than a second hand-written copy. That is the point: a duplicated reduced-motion check is exactly how the two reveals drifted apart in the first place. `lockBodyScroll` set the precedent for a shared browser-level concern. It is not exported from the public index — this is a fix, not new API.

`revealRemainingText()` is extracted from the `isStreaming: false` effect, which already did precisely this, so the reduced-motion path and the end-of-stream path are now one implementation rather than two that must be kept in agreement.

## Evidence

A real red-green cycle, not a test written after the fact:

```
guard disabled -> reduce test FAILS (38 mutations, one per character)
                  control test PASSES
guard enabled  -> both PASS
```

The control is deliberate. Without it, `mutations === 1` would also pass if the reveal were broken outright or the observer never attached, so it proves the measurement distinguishes the two states rather than always reporting "instant".

**One trap worth recording, since it cost a wrong diagnosis.** The idiomatic `test.use({ reducedMotion: 'reduce' })` does **not** reach the page under this config — `matchMedia(...).matches` stays `false`, so the test renders an un-emulated page and fails for a reason unrelated to the component. Verified both ways:

```
PROBE test.use      = false
PROBE emulateMedia  = true
```

The test uses `page.emulateMedia()` and asserts the emulation actually landed before asserting anything about the reveal. There is a comment saying not to "simplify" it back.

## Docs

Both pages now warn against the double-reveal that combining `typewriter: true` with a `TypewriterText` `messageBody` produces — the controller is already growing `msg.content` character by character, so `TypewriterText` re-types each increment on its own cadence.

Cross-links use plain backticked names, not relative `.md` links: those resolve to `/components/<name>.md` and **fail the prerender build outright**, which is how I found out.

## Verification

| check | result |
| --- | --- |
| `pnpm lint` | 0 |
| `pnpm check` | 0 |
| Playwright — typewriter pacing, resolveDelay, reduced motion, ChatMessage typewriter | 18 passed |
| Playwright — full functional suite | 605 passed, 1 skipped |

## Scope

This closes the behavioural divergence and the documented failure mode. The ~114 LOC of independently-written reveal logic that #465 also describes is **not** consolidated here — that needs an API decision (the issue lays out three options, all with regression risk), and this change deliberately does not pre-empt it.

## Summary by CodeRabbit

- **New Features**
  - Added improved reduced-motion support for text reveal. When enabled, text appears immediately instead of animating, while progress callbacks continue to work appropriately.
  - Normal motion continues to reveal streamed text incrementally.

- **Documentation**
  - Clarified that controller-managed typewriter output and the standalone text-reveal component are alternative approaches.
  - Added configuration guidance to prevent doubled or compounded reveal effects when using custom message rendering.

### DIFF
--- a/src/lib/TypewriterText/TypewriterText.svelte
+++ b/src/lib/TypewriterText/TypewriterText.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import { onMount, onDestroy, untrack } from 'svelte';
+  import { prefersReducedMotion } from '../utils';
   import type { TypewriterTextProperties, TypewriterCharacterDelayRange } from './properties';
 
   let {
@@ -75,7 +76,34 @@
     return speed;
   };
 
+  // Disclose everything still pending at once. The whitespace count has to be rebuilt
+  // from the text itself because the per-character path that normally maintains it is
+  // skipped entirely, and a later continuation would otherwise resume from a count that
+  // never saw the revealed suffix.
+  const revealRemainingText = (): void => {
+    if (timeoutId !== null) {
+      clearTimeout(timeoutId);
+      timeoutId = null;
+    }
+    const hadRemainingText = currentIndex < text.length;
+    displayedText = text;
+    currentIndex = text.length;
+    revealedWordCount = countWhitespace(text);
+    if (hadRemainingText) {
+      onprogress?.({ index: currentIndex, total: text.length, displayedText });
+    }
+  };
+
   const typeNextCharacter = (): void => {
+    // Typing IS the animation here, so a reduced-motion request has to suppress the
+    // reveal itself rather than a transition around it: disclose the text instead.
+    // `ChatController` already bypasses its own drain loop on this setting, and the two
+    // reveal implementations have to answer it identically — a consumer composing this
+    // component through `Chat`'s `messageBody` snippet is running both at once.
+    if (prefersReducedMotion()) {
+      revealRemainingText();
+      return;
+    }
     if (currentIndex < text.length) {
       const revealedCharacter = text[currentIndex];
       const revealedCharacterIndex = currentIndex;
@@ -122,6 +150,7 @@
         if (currentIndex >= previousTextLength) {
           if (timeoutId !== null) {
             clearTimeout(timeoutId);
+            timeoutId = null;
           }
           typeNextCharacter();
         }
@@ -133,16 +162,7 @@
   // eslint-disable-next-line no-restricted-syntax
   $effect(() => {
     if (!isStreaming && text.length > 0) {
-      if (timeoutId !== null) {
-        clearTimeout(timeoutId);
-      }
-      const hadRemainingText = currentIndex < text.length;
-      displayedText = text;
-      currentIndex = text.length;
-      revealedWordCount = countWhitespace(text);
-      if (hadRemainingText) {
-        onprogress?.({ index: currentIndex, total: text.length, displayedText });
-      }
+      revealRemainingText();
     }
   });
 
--- a/tests/typewriter-text-reduced-motion.test.ts
+++ b/tests/typewriter-text-reduced-motion.test.ts
@@ -1,0 +1,159 @@
+import { expect, test, type Page } from '@playwright/test';
+
+// `ChatController` bypasses its character-by-character reveal entirely when the user
+// asks for reduced motion (`prefersReducedMotion()`, consulted by `feed()`), but
+// `TypewriterText` — which exists to do the same job, and whose own tagline is "reveals
+// text one character at a time, the way a streaming model answer reads" — had no such
+// handling. The same OS setting therefore silenced one reveal implementation and not the
+// other, so a consumer composing `TypewriterText` through `Chat`'s `messageBody` snippet
+// kept the animation the setting exists to suppress. See #465.
+//
+// Emulation is applied with `page.emulateMedia()` rather than the more idiomatic
+// `test.use({ reducedMotion: 'reduce' })`. That is deliberate and load-bearing: under
+// this config `test.use` does NOT reach the page — `matchMedia('(prefers-reduced-motion:
+// reduce)').matches` stays `false`, so the "reduce" test would render an un-emulated page
+// and fail for a reason that has nothing to do with the component. Verified both ways
+// before writing this. Do not "simplify" it back to `test.use`.
+//
+// Timing is measured from the FIRST text mutation to the LAST via a MutationObserver
+// installed before navigation, never from an expect() poll — matching the pattern the
+// component's existing pacing tests already establish.
+
+const BASELINE_TEXT = 'Baseline pacing check text stays fast.';
+const BASELINE_TEST_ID = 'typewriter-default-pacing';
+
+// Longer and slower (74 characters at speed=20) than the baseline fixture, which leaves room
+// to flip the preference while typing is genuinely still in flight.
+const PROGRESS_TEXT = 'Progress reporting keeps a scroll container pinned to the newest character.';
+const PROGRESS_TEST_ID = 'typewriter-progress-demo';
+
+type RevealTiming = { firstAt: number | null; lastAt: number | null; mutations: number };
+
+const installRevealObserver = async (
+  page: Page,
+  testId: string,
+  windowKey: string
+): Promise<void> => {
+  await page.addInitScript(
+    ({ testId, windowKey }: { testId: string; windowKey: string }) => {
+      const timing: RevealTiming = { firstAt: null, lastAt: null, mutations: 0 };
+      (window as unknown as Record<string, RevealTiming>)[windowKey] = timing;
+      const attach = () => {
+        const element = document.querySelector(`[data-pw="${testId}"]`);
+        if (!element) {
+          requestAnimationFrame(attach);
+          return;
+        }
+        new MutationObserver(() => {
+          if ((element.textContent || '').length === 0) {
+            return;
+          }
+          if (timing.firstAt === null) {
+            timing.firstAt = performance.now();
+          }
+          timing.lastAt = performance.now();
+          timing.mutations += 1;
+        }).observe(element, { childList: true, subtree: true, characterData: true });
+      };
+      requestAnimationFrame(attach);
+    },
+    { testId, windowKey }
+  );
+};
+
+const readTiming = async (
+  page: Page,
+  windowKey: string
+): Promise<{ mutations: number; elapsedMs: number }> => {
+  return page.evaluate((key: string) => {
+    const timing = (window as unknown as Record<string, RevealTiming>)[key];
+    const elapsedMs =
+      timing && timing.firstAt !== null && timing.lastAt !== null
+        ? Math.round(timing.lastAt - timing.firstAt)
+        : -1;
+    return { mutations: timing ? timing.mutations : -1, elapsedMs };
+  }, windowKey);
+};
+
+test('reduced motion reveals the whole string at once instead of typing it', async ({ page }) => {
+  await page.emulateMedia({ reducedMotion: 'reduce' });
+  await installRevealObserver(page, BASELINE_TEST_ID, '__reducedMotionTiming');
+  await page.goto('/components/typewriter-text');
+
+  // The emulation is the premise of the assertion below, so prove it landed rather than
+  // trusting it — an un-emulated page would type, and the failure would look like a
+  // component bug.
+  const emulationActive = await page.evaluate(
+    () => window.matchMedia('(prefers-reduced-motion: reduce)').matches
+  );
+  expect(emulationActive).toBe(true);
+
+  await expect(page.locator(`[data-pw="${BASELINE_TEST_ID}"]`)).toHaveText(BASELINE_TEXT);
+
+  const { mutations, elapsedMs } = await readTiming(page, '__reducedMotionTiming');
+
+  // One mutation carrying the entire string — not one per character. The length is what
+  // makes this meaningful: typing this text at the default 15ms cadence takes well over
+  // half a second and mutates the node 38 times.
+  expect(mutations).toBe(1);
+  expect(elapsedMs).toBeLessThan(50);
+});
+
+// Raised in review as a possible defect: because `prefersReducedMotion()` is a one-shot
+// call rather than a subscription, does a reveal already in flight keep typing until the
+// text changes or completes?
+//
+// It does not, and this is the proof rather than an argument. `typeNextCharacter()` re-enters
+// through its own `setTimeout` chain and re-reads the preference at the top of every call, so
+// the switch is picked up on the very next character — within one `speed` interval, 20ms on
+// this fixture. A subscription would buy nothing here and would add a listener to unregister.
+test('a mid-stream switch to reduced motion stops the typing immediately', async ({ page }) => {
+  await page.emulateMedia({ reducedMotion: 'no-preference' });
+  await installRevealObserver(page, PROGRESS_TEST_ID, '__midStreamTiming');
+  await page.goto('/components/typewriter-text');
+
+  const typewriter = page.locator(`[data-pw="${PROGRESS_TEST_ID}"]`);
+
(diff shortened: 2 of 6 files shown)


src/lib/TypewriterText/TypewriterText.svelte
```diff
@@ -133,16 +161,7 @@
...
-        clearTimeout(timeoutId);
-      }
-      const hadRemainingText = currentIndex < text.length;
-      displayedText = text;
-      currentIndex = text.length;
-      revealedWordCount = countWhitespace(text);
-      if (hadRemainingText) {
-        onprogress?.({ index: currentIndex, total: text.length, displayedText });
-      }
+      revealRemainingText();
```
Tara-ag, line 165: 💡 MINOR: `timeoutId` is not reset to `null` after clearing in the streaming `$effect`

`revealRemainingText()` clears and resets `timeoutId`, but the `$effect` that reacts to `text` changes calls `clearTimeout(timeoutId)` directly without resetting it. While `typeNextCharacter()` will overwrite it, this is inconsistent with `revealRemainingText()` and `onDestroy()`, and can mislead readers about whether a timer is active.

Suggested fix: refactor the `$effect` branch to also call `revealRemainingText()` when `currentIndex >= previousTextLength`, or add `timeoutId = null;` after the direct `clearTimeout`.

Tara-ag (changes requested):
One MAJOR reactivity issue and two smaller notes. The reduced-motion check should be reactive to mid-stream preference changes; otherwise the fix is solid and well-tested.

Tara-ag (commented):
Review comments for accepted findings.

tests/typewriter-text-reduced-motion.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    return { mutations: timing ? timing.mutations : -1, elapsedMs };
+  }, windowKey);
+};
+
+test('reduced motion reveals the whole string at once instead of typing it', async ({ page }) => {
+  await page.emulateMedia({ reducedMotion: 'reduce' });
+  await installRevealObserver(page, BASELINE_TEST_ID, '__reducedMotionTiming');
+  await page.goto('/components/typewriter-text');
+
+  // The emulation is the premise of the assertion below, so prove it landed rather than
```
Tara-ag, line 83: 💬 SUGGESTION: Add coverage for a mid-stream reduced-motion toggle

The current tests verify the initial reduced-motion state only. Consider adding a test that starts in no-preference, begins observing mutations, then calls `page.emulateMedia({ reducedMotion: 'reduce' })` after a few characters have appeared, and asserts the remainder of the string is disclosed in a single mutation. This guards against regressions if the one-shot `prefersReducedMotion()` check is ever refactored incorrectly.

src/lib/TypewriterText/TypewriterText.svelte
```diff
@@ -75,7 +76,34 @@
...
+    currentIndex = text.length;
+    revealedWordCount = countWhitespace(text);
+    if (hadRemainingText) {
+      onprogress?.({ index: currentIndex, total: text.length, displayedText });
+    }
+  };
+
   const typeNextCharacter = (): void => {
+    // Typing IS the animation here, so a reduced-motion request has to suppress the
+    // reveal itself rather than a transition around it: disclose the text instead.
```
Tara-ag, line 99: ⚠️ MAJOR: Reduced-motion check is not reactive to media-query changes

`prefersReducedMotion()` is called synchronously inside `typeNextCharacter()` and reads `window.matchMedia(...).matches` once per timer tick. If a user toggles the OS reduced-motion setting while a reveal is already running, the component keeps typing until the next text change or completion. Subscribe to the media query and let an effect re-run when the preference changes so the reveal responds immediately.

Suggested fix: Create a reactive reduced-motion signal (e.g. in `utils.ts`) using `matchMedia('(prefers-reduced-motion: reduce)').addEventListener('change', ...)` and consult that signal here instead of the one-shot function call.

Tara-ag (commented):
Review comments for accepted findings.

Tara-ag:

The PR fixes the documented reduced-motion divergence and adds good tests, but the reduced-motion check is one-shot rather than reactive. A user toggling the OS preference mid-reveal will not see the remaining text disclosed instantly, which is the core accessibility promise of this change. Fix that reactivity gap, address the minor timeout-handle cleanup inconsistency, and optionally add a mid-stream toggle test.

- ⚠️ **MAJOR**: Reduced-motion check is not reactive to media-query changes — `src/lib/TypewriterText/TypewriterText.svelte:99`
  `prefersReducedMotion()` is called synchronously inside `typeNextCharacter()` and `ChatController.feed()`. It reads `window.matchMedia(...).matches` once per timer tick, so if the user toggles the OS reduced-motion setting while a reveal is already running, the component keeps typing until the next text change or until completion. The correct pattern is to subscribe to the media query and let the …
- 💡 **MINOR**: `timeoutId = null` is missing after clearing in the streaming $effect — `src/lib/TypewriterText/TypewriterText.svelte:144`
  In the `$effect` that reacts to `text` changes, `clearTimeout(timeoutId)` is called but `timeoutId` is not reset to `null`. Although the subsequent `typeNextCharacter()` will overwrite it, leaving the stale handle can mislead readers and is inconsistent with the cleanup elsewhere (`revealRemainingText`, `onDestroy`).
- 💬 **SUGGESTION**: Reduced-motion test does not cover mid-stream preference toggle — `tests/typewriter-text-reduced-motion.test.ts:78`
  The current tests verify the initial reduced-motion state only. Consider adding a test that starts in no-preference, begins observing mutations, then calls page.emulateMedia({ reducedMotion: 'reduce' }) after a few characters have appeared, and asserts the remainder of the string is disclosed in a single mutation. This guards against regressions if the one-shot prefersReducedMotion() check is ever…

murdore:
Thanks — all three addressed. One of them turned out not to be a defect, and your own second suggestion is what proved it, so I want to show the measurement rather than just assert a disagreement.

## ⚠️ MAJOR — reduced-motion check not reactive to media-query changes

**This one I'm pushing back on, with evidence.** The concern was that a reveal already in flight keeps typing until the next text change or completion, and that a `matchMedia` subscription is needed.

`prefersReducedMotion()` isn't called once at mount — `typeNextCharacter()` re-enters through **its own `setTimeout` chain**, and the check sits at the top of every call:

```js
timeoutId = setTimeout(typeNextCharacter, resolveTypingDelay(...));
```

So the preference is re-read once per character, and a mid-stream switch is picked up on the very next one — 20ms on the fixture below.

I wrote the mid-stream test you suggested in the next comment, which is exactly the right instrument, and it settles it:

| implementation | mutations after the toggle |
| --- | --- |
| per-call check (shipped) | **3** |
| mount-only check (`currentIndex === 0 && …`) | **27** — typing continues |

Both other tests in the file pass under *either* implementation, so this new test covers precisely the gap you were worried about — it just demonstrates the current code already closes it. A subscription would buy nothing here and would add a listener to unregister on destroy.

If you still think a subscription is warranted for a reason the test doesn't capture, say so and I'll take it — but I didn't want to add a listener on the strength of a hypothesis the measurement contradicts.

## 💬 SUGGESTION — mid-stream toggle coverage

Taken, and it earned its place immediately (see above). Two revisions were needed before it was trustworthy rather than merely green, which is worth recording:

1. A first bound of `<= 2` was pinned to an observed number and failed at 3.
2. A scaled bound (`remaining / 4`) then failed its own meaningfulness guard — the default poll cadence let **57 of 74** characters type before it first reported, leaving too little to distinguish the behaviours.

The bound is now **absolute**, because the two behaviours differ in kind: disclosing costs one mutation plus whatever types during the CDP round-trip — a small constant, independent of how much text remains — whereas continuing to type costs one mutation per remaining character. The poll uses a 10ms interval so the toggle lands early, and there's an explicit guard that enough text remains for the assertion to mean anything.

Stable across 6 repeats.

## 💡 MINOR — `timeoutId` not reset to `null`

Correct, fixed. The growth `$effect` cleared the timer without nulling it, which is inconsistent with `revealRemainingText()` and `onDestroy()` and can mislead a reader about whether a timer is live.

## Verification

lint 0 · check 0 · 919 unit · 19 Playwright across reduced-motion, pacing, resolveDelay and ChatMessage typewriter.

One transparency note: during this work a single Playwright failure appeared once and did not reproduce in 66 subsequent runs (18 on the new file at 6 repeats, 48 on the pre-existing typewriter/chat files at 3 repeats). I could not attribute it to a specific test. Flagging it rather than quietly reporting a clean sweep.

src/lib/TypewriterText/TypewriterText.svelte
murdore: Correct, fixed. The growth `$effect` cleared the timer without nulling it, which is inconsistent with `revealRemainingText()` and `onDestroy()` and can mislead a reader about whether a timer is live. Now `timeoutId = null` alongside the `clearTimeout`.

tests/typewriter-text-reduced-motion.test.ts
murdore: Added, and it earned its place immediately — it's what settled the MAJOR on the sibling thread.

Two revisions were needed before it was trustworthy rather than merely green: a first bound of `<= 2` was pinned to an observed number and failed at 3; a scaled bound then failed its own meaningfulness guard, because the default poll cadence let 57 of 74 characters type before it first reported.

The bound is now absolute, because the two behaviours differ in kind — disclosing costs one mutation plus the CDP round-trip regardless of how much text remains, whereas typing costs one per remaining character. The poll uses a 10ms interval so the toggle lands early, plus an explicit guard that enough text remains for the assertion to mean anything. Stable across 6 repeats.

src/lib/TypewriterText/TypewriterText.svelte
murdore: Pushing back on this one, with a measurement rather than an argument.

`prefersReducedMotion()` is not read once at mount. `typeNextCharacter()` re-enters through **its own `setTimeout` chain** — `timeoutId = setTimeout(typeNextCharacter, resolveTypingDelay(...))` — and the check sits at the top of every call. So the preference is re-read once per character, and a mid-stream switch is picked up on the very next one: 20ms on the test fixture.

I implemented your other suggestion (the mid-stream toggle test), which is exactly the right instrument, and it settles it:

| implementation | mutations after the toggle |
| --- | --- |
| per-call check (shipped) | **3** |
| mount-only check (`currentIndex === 0 && …`) | **27** — typing continues |

Both other tests in that file pass under *either* implementation, so the new test covers precisely the gap you identified — it just shows the current code already closes it. A subscription would add a listener to register and unregister on destroy for no behavioural change.

Resolving because the concern was acted on: the coverage you asked for now exists and guards this permanently. If you think a subscription is still warranted for a reason the measurement doesn't capture, please reopen and say what it is — I'd rather add it than leave a real gap.

Tara-ag (approved):
Verified the changes file by file. The reduced-motion utility is safely shared between `ChatController` and `TypewriterText`, the new `revealRemainingText()` path keeps the bulk-disclosure and reduced-motion behaviors in one place, and the added Playwright tests (including the mid-stream toggle case) prove the behavior. The previous `timeoutId` cleanup nit is also fixed. No public API, export, prop, or CSS-variable contract is broken. Safe to merge.

PR #580, opened 2026-09-08
docs(pill): document the dark-theme recipe, and stop the demo silently killing tone

### LINKED ISSUE #464: Pill: no documented dark-theme recipe
## Context
`Pill.svelte`'s CSS-variable fallbacks are light-only literals (`src/lib/Pill/Pill.svelte`): `--pill-background: #e0e0e0`, `--pill-color: #333333`, `--pill-hover-background: var(--pill-background, #d0d0d0)`, `--pill-hover-color`, `--pill-disabled-opacity: 0.4`, `--pill-border: none`, etc. — 19 variables, all documented in `docs/Pill.md`'s "CSS Variables" table. `docs/Pill.md` has no "dark" section at all.

Two places already carry a dark recipe for Pill, but neither is attached to Pill's own docs:
1. `src/routes/components/demo.css` lines 160-164, inside a `[data-theme='dark']` block explicitly commented `/* Pill — chat suggestion chips and composer attachment pills */` (part of the larger `/* Chat family — dark theme */` section, lines 69-165). It overrides only 4 of Pill's 19 variables (`--pill-background: #252535`, `--pill-color: #d1d5db`, `--pill-hover-background: #2f2f45`, `--pill-hover-color: #e5e7eb`) and exists because Pill happens to be reused inside the chat surfaces, not because Pill's own dark story was documented.
2. The four theming variant classes shown in `docs/Pill.md`'s "Theming with Classes" section (`pill-success/warning/error/info`) are reimplemented locally in two demo pages — `src/routes/components/pill/+page.svelte` (lines 88-98) and `src/routes/components/table/+page.svelte` (lines 1056-1075) — as light-only literals with zero dark overrides anywhere.

So today: the bare default `<Pill>` gets an incidental dark palette; every themed variant gets none — those pastel light literals render byte-identical whether `[data-theme='dark']` is set or not.

## What a documented recipe would look like
`docs/StatCard.md` already has the pattern to copy — a `### Themed Card` subsection under `## Usage` (lines 59-74) showing a wrapper class overriding the component's CSS custom properties with a full dark palette:
```svelte
<div class="dark-stat">
  <StatCard title="GMV" value="₹8.4Cr" delta="+5.1%" subtitle="7-day rolling" />
</div>
<style>
  .dark-stat { --statcard-background: #1a1a2e; ... }
</style>
```
`docs/Pill.md` needs the equivalent: a `### Dark Theme` subsection covering at minimum `--pill-background`, `--pill-color`, `--pill-hover-background`, `--pill-hover-color`, `--pill-border`, `--pill-dismiss-color` — sourced from the values already proven out in `demo.css`'s incidental block — plus a decision on whether the `pill-success/warning/error/info` recipes get dark counterparts or are documented as intentionally theme-invariant.

## Recommendation
Add the dark subsection to `docs/Pill.md` (mirroring `StatCard.md`'s pattern), reusing the 4 values already live in `demo.css`, and either give the 4 status-variant classes a dark counterpart or explicitly document that they're intentionally theme-invariant chips.

murdore:
Closes #464.

Every fallback baked into Pill is a light literal — `--pill-background` to `#e0e0e0`, `--pill-color` to `#333333`, and each of the five tones to a pastel pair — and `docs/Pill.md` had no dark section at all. The only dark recipe in the repo was an incidental four-variable block in `src/routes/components/demo.css`, which exists because Pill is reused inside the chat surfaces, not because Pill's own dark story was ever written down.

Writing that section turned up a real bug in the demo, so this fixes it too.

## A global `--pill-background` silently disables `tone`

Pill resolves colour as `var(--pill-background, var(--_pill-tone-background, …))`, so an explicit base value wins and the tone layer is never consulted. There is no error and no visual hint that the prop stopped working.

`demo.css` set exactly that pair on `[data-theme='dark']` itself. Measured on the real elements on `/components/pill`, before the fix:

| tone | light | dark (before) |
| --- | --- | --- |
| accent | `rgb(209,236,241)` | `rgb(37,37,53)` |
| ok | `rgb(212,237,218)` | `rgb(37,37,53)` |
| warn | `rgb(255,243,205)` | `rgb(37,37,53)` |
| danger | `rgb(248,215,218)` | `rgb(37,37,53)` |
| muted | `rgb(241,241,241)` | `rgb(37,37,53)` |

All five byte-identical. In dark mode the Pill demo rendered five indistinguishable grey chips labelled Accent / Ok / Warn / Danger / Muted, on this library's own documentation site.

After:

```
accent rgb(17,38,46)  ok rgb(16,40,26)  warn rgb(46,36,16)
danger rgb(45,20,22)  muted rgb(37,37,53)
```

## Why the fix is scoped by tone, not by container

The override is not dropped — the untoned chat suggestion chips and composer attachment pills genuinely need it. It moves to a rule that excludes the five tone classes, so untoned pills keep the neutral chip and toned ones reach the tone layer.

Scoping by tone rather than by container matters: **six** components use Pill — `ChatSuggestions`, `ChatComposer`, `ThinkingIndicator`, `ToolCallLog`, `Combobox` and `Table`'s `BuiltinCell` — and a container-scoped rule would have missed some of them.

## Contrast was computed, not eyeballed

Every dark pair clears WCAG AA for normal text — **5.93:1** at the lowest (`muted`), the rest between **9.05:1** and **10.58:1**.

The failure mode is quantified too: the light `ok` text `#155724` on the neutral dark chip `#252535` is **1.74:1**, well under the 4.5:1 floor, which is why the light defaults cannot simply carry over.

## On the question the issue raised

Whether the `pill-success` / `warning` / `error` / `info` class recipes should get dark counterparts or be declared theme-invariant: **neither, as written.** They are example CSS a consumer owns once pasted, not something Pill ships. The docs say to give them dark counterparts if you use them, and point at `tone` as the reason not to need them.

## Verification

| check | result |
| --- | --- |
| `pnpm lint` | 0 |
| `pnpm check` | 0 |
| `pnpm build` | 0 |
| Playwright — new dark-theme tone tests | 2 passed |

Red-green confirmed by restoring the unscoped selector and watching the distinct-tone count drop from **5 to 1** while the untoned assertion held — so the test discriminates rather than failing everything.

No component file is touched.

### DIFF
--- a/docs/Pill.md
+++ b/docs/Pill.md
@@ -84,6 +84,75 @@ once per theme, then pass `tone` at each call site:
 directly, or via `classes`) always wins over the tone default, so a one-off recolor doesn't
 need to fork the tone.
 
+### Dark Theme
+
+Every fallback baked into Pill is a light value (`--pill-background` falls back to `#e0e0e0`,
+`--pill-color` to `#333333`, and each `tone` to a pastel pair), so a pill dropped onto a dark
+page keeps its light chip until you override the variables.
+
+Theme the **tone** variables at the theme root. These are always safe to set globally — each one
+only reaches the pills carrying that tone:
+
+```css
+/* app.css */
+[data-theme='dark'] {
+  --pill-tone-accent-background: #11262e;
+  --pill-tone-accent-color: #7dd3fc;
+  --pill-tone-ok-background: #10281a;
+  --pill-tone-ok-color: #6ee7a8;
+  --pill-tone-warn-background: #2e2410;
+  --pill-tone-warn-color: #fcd34d;
+  --pill-tone-danger-background: #2d1416;
+  --pill-tone-danger-color: #fca5a5;
+  --pill-tone-muted-background: #252535;
+  --pill-tone-muted-color: #9ca3af;
+}
+```
+
+Those pairs all clear WCAG AA for normal text (5.9:1 at the lowest, `muted`; the rest sit
+between 9:1 and 10.6:1). Reusing the light defaults on a dark ground is what fails — `#155724`
+on `#252535` is 1.74:1, well under the 4.5:1 floor.
+
+For pills that carry **no** tone, set the base pair — but scope it to the containers those pills
+live in, rather than the theme root:
+
+```css
+/* app.css — the filter bar and tag row hold untoned pills */
+[data-theme='dark'] .filter-bar,
+[data-theme='dark'] .tag-row {
+  --pill-background: #252535;
+  --pill-color: #d1d5db;
+  --pill-hover-background: #2f2f45;
+  --pill-hover-color: #e5e7eb;
+  --pill-dismiss-color: #9ca3af;
+  --pill-dismiss-hover-color: #e5e7eb;
+}
+```
+
+**The scoping is the whole point, and skipping it is easy to do: a global `--pill-background` /
+`--pill-color` silently disables `tone`.** Pill resolves colour as
+`var(--pill-background, var(--_pill-tone-background, …))`, so an explicit value wins and the tone
+layer is never consulted — every `tone="ok"` and `tone="danger"` chip renders in the same neutral
+grey as an untoned one, with no error and no visual hint that the prop stopped working. Measured
+on a real `tone="ok"` pill:
+
+| | background | color |
+| --- | --- | --- |
+| tone alone | `#d4edda` | `#155724` |
+| tone + a global `--pill-background`/`--pill-color` | `#252535` | `#d1d5db` |
+
+This library's own demo site had exactly that bug: all five tones rendered byte-identical in dark
+mode. If you have no convenient container to scope to, exclude the tone classes instead —
+`[data-theme='dark'] .pill:not(.tone-accent, .tone-ok, .tone-warn, .tone-danger, .tone-muted)` —
+though that couples your stylesheet to Pill's internal class names, so prefer your own containers
+where you have them.
+
+The `pill-success` / `pill-warning` / `pill-error` / `pill-info` class recipes in **Theming with
+Classes** above are light-only by design: they are example CSS you own once you paste them into
+your app, not something Pill ships. If you use them and support a dark theme, give them dark
+counterparts the same way — or prefer `tone`, which exists precisely so the colour pair lives in
+one themeable place instead of being repeated per call site.
+
 ### Attribute Passthrough
 
 `attrs` spreads arbitrary `data-*`/`aria-*` attributes onto the pill root — for a consumer
--- a/tests/pill-dark-theme-tones.test.ts
+++ b/tests/pill-dark-theme-tones.test.ts
@@ -1,0 +1,56 @@
+import { expect, test } from '@playwright/test';
+
+// #464. The demo shell used to set `--pill-background` / `--pill-color` on
+// `[data-theme='dark']` itself, for the untoned chat suggestion chips and composer
+// attachment pills. But Pill resolves colour as
+// `var(--pill-background, var(--_pill-tone-background, …))`, so a bare value on the theme
+// root wins over the tone layer by design — which silently flattened every `tone="…"` chip
+// to that same grey. All five tones rendered byte-identical, so the Pill demo showed five
+// indistinguishable chips labelled Accent/Ok/Warn/Danger/Muted in dark mode.
+//
+// This guards both halves of the fix: tones stay distinct, and the untoned pills keep the
+// neutral chip the override existed to give them.
+
+const TONES = ['accent', 'ok', 'warn', 'danger', 'muted'] as const;
+
+const NEUTRAL_DARK_BACKGROUND = 'rgb(37, 37, 53)';
+const NEUTRAL_DARK_COLOR = 'rgb(209, 213, 219)';
+
+const paintOf = (page: import('@playwright/test').Page, testId: string) =>
+  page.locator(`[data-pw="${testId}"]`).evaluate((el) => {
+    const s = getComputedStyle(el);
+    return `${s.backgroundColor} / ${s.color}`;
+  });
+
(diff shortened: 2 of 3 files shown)


tests/pill-dark-theme-tones.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+test.beforeEach(async ({ page }) => {
+  await page.goto('/components/pill');
+  await page.evaluate(() => document.documentElement.setAttribute('data-theme', 'dark'));
+});
+
+test('every tone stays visually distinct in the dark theme', async ({ page }) => {
+  const paints = await Promise.all(TONES.map((tone) => paintOf(page, `pill-tone-${tone}`)));
+
+  expect(new Set(paints).size).toBe(TONES.length);
```
Tara-ag, line 42: 💬 SUGGESTION: The theme is set synchronously here, but if the demo shell ever adds CSS transitions or async theme loading, the computed-style assertions below could flake. Wait for a deterministic dark-theme signal (e.g. `--doc-bg` resolves to `#0f0f17`) before the tests run.

docs/Pill.md
```diff
@@ -84,6 +84,64 @@ once per theme, then pass `tone` at each call site:
...
+### Dark Theme
+
+Every fallback baked into Pill is a light value (`--pill-background` falls back to `#e0e0e0`,
+`--pill-color` to `#333333`, and each `tone` to a pastel pair), so a pill dropped onto a dark
+page keeps its light chip until you override the variables. Set them under whatever selector
+your app already uses to mark the dark theme:
+
+```css
+/* app.css */
+[data-theme='dark'] {
```
Tara-ag: 💬 SUGGESTION: This example shows the exact pattern the next paragraph warns against — a global `--pill-background` / `--pill-color` silently disables `tone`. A skimming reader may copy this block and miss the caveat. Consider leading with the tone-variable recipe or adding a warning comment inside this block.

Tara-ag (approved):
Docs-and-demo-only change that fixes a real dark-mode regression in the Pill demo. No component source is modified, so the public API, CSS custom property names, and web-component registry are unchanged. The new Playwright test discriminates correctly (distinct tones + preserved untoned chip). Two minor suggestions only; nothing blocking.

docs/Pill.md
Tara-ag: 💬 SUGGESTION: Dark-theme doc section uses a global `--pill-background` example right after warning against it. In docs/Pill.md the new Dark Theme subsection first shows a `[data-theme='dark']` block setting `--pill-background` / `--pill-color` globally, then explains in the next paragraph that this exact pattern silently disables `tone`. Because the warning follows the example, a reader skimming for a copy-paste recipe may grab the first block and miss the caveat. Consider either leading with the tone-first recommendation or adding a prominent warning comment inside the first code block itself.

tests/pill-dark-theme-tones.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    return `${s.backgroundColor} / ${s.color}`;
+  });
+
+test.beforeEach(async ({ page }) => {
+  await page.goto('/components/pill');
+  await page.evaluate(() => document.documentElement.setAttribute('data-theme', 'dark'));
+});
+
+test('every tone stays visually distinct in the dark theme', async ({ page }) => {
+  const paints = await Promise.all(TONES.map((tone) => paintOf(page, `pill-tone-${tone}`)));
```
Tara-ag, line 40: 💬 SUGGESTION: New Playwright test relies on demo-page copy without waiting for theme transition. tests/pill-dark-theme-tones.test.ts sets `data-theme='dark'` via page.evaluate and immediately reads computed styles. If the demo shell applies a CSS transition or if future theme changes become asynchronous, the assertion may flake. The test also selects by `data-pw` which is good, but it does not wait for any deterministic signal that the dark styles have applied. After setting the dark theme, wait for a deterministic signal such as `await page.waitForFunction(() => getComputedStyle(document.documentElement).getPropertyValue('--doc-bg') === '#0f0f17')` before asserting pill colors.

Tara-ag (approved):
Docs-and-demo-only change that fixes a real dark-mode regression in the Pill demo. No component source is modified, so the public API, CSS custom property names, and web-component registry are unchanged. The new Playwright test discriminates correctly (distinct tones + preserved untoned chip). Two minor suggestions only; nothing blocking.

tests/pill-dark-theme-tones.test.ts
murdore: Done. The test now waits on a web-first assertion (`toHaveCSS`) in `beforeEach`, which retries until the style resolves, so the one-shot reads below can't run against an un-applied theme.

Deliberately not the suggested `waitForFunction` on `--doc-bg`: `getPropertyValue` returns custom properties verbatim, leading whitespace included, so an equality check there would be its own flake.

Confirmed the barrier doesn't mask what the test guards — with the unscoped selector restored, the barrier still passes and the distinct-tone count still drops from 5 to 1.

docs/Pill.md
murdore: Agreed and fixed — the section led with the exact anti-pattern it exists to warn about, so the first copy-pasteable block was the trap.

Restructured: it now leads with the tone variables (always safe to set globally, since each only reaches pills carrying that tone), then shows the base pair already scoped to the consumer's own containers. The trap and its measurements follow as the rationale rather than preceding as the recipe.

murdore: Same fix as the sibling thread — the Dark Theme section now leads with the tone-variable recipe and shows the base pair pre-scoped to consumer containers, so a reader grabbing the first block gets something that works. The warning and its measured before/after table follow as the rationale.

tests/pill-dark-theme-tones.test.ts
murdore: Same fix as the sibling thread — a `toHaveCSS` barrier in `beforeEach` now retries until the dark theme resolves, before any one-shot computed-style read runs.

PR #581, opened 2026-09-08
feat(pill): render a real button root for interactive chips

### LINKED ISSUE #469: Pill: interactive variant (real button semantics) for chip-like controls
## Gap

Pill's root is a non-interactive `<div>` (with an `onclick` prop). Anything chip-shaped that needs real control semantics — `aria-expanded` disclosure chips, `aria-pressed` toggle chips, keyboard focus/activation — cannot be built on Pill today without wrapping or forking it.

This was hit twice in #468's review-driven rework:
- **ToolCallLog** kept bespoke `<button>` chips (states + `aria-expanded` detail popovers) instead of composing Pill — recorded in its docs as a deliberate deviation.
- ThinkingIndicator's search-kind **query chip** could use Pill only because it is purely presentational.

## Proposal

An opt-in interactive form of Pill — e.g. `as: 'button'` (or automatically when `onclick` is present, behind a major/minor decision) rendering a real `<button type="button">` root with:
- forwarded `ariaExpanded` / `ariaPressed` props,
- visible focus treatment via tokens,
- unchanged div rendering for today's consumers (no breaking change).

Then ToolCallLog's chips (and future chip-shaped controls) can compose Pill instead of re-implementing its look.

Context: https://github.com/juspay/svelte-ui-components/pull/468

murdore:
Closes #469.

Pill's root is a non-interactive `<div>`, with a synthetic `role="button"` / `tabindex="0"` / keydown shim bolted on when `onclick` is supplied. That covers a plain click target, but it cannot express a chip that is a real control — a disclosure chip owning `aria-expanded`, or a toggle chip owning `aria-pressed`. This is why `ToolCallLog` kept bespoke `<button>` chips instead of composing Pill, recorded in its own docs as a deliberate deviation.

## What changed

`as="button"` renders a real `<button type="button">` root, so focus, activation and announcement come from the platform. `ariaExpanded` and `ariaPressed` are forwarded, and applied only when the pill actually *is* a control — a button root, or a div with `onclick` — since both attributes are meaningless on a label.

With a button root Pill deliberately emits **no** `role`, **no** `tabindex`, and registers no keydown handler. Adding them on top of native semantics would duplicate the platform's own activation: Enter would fire the handler twice.

`as` defaults to `'div'`, so every existing consumer renders byte-for-byte as before. Naming follows `Card`, which already types its root override as `as?: 'div' | 'a' | 'figure'`.

## Three things worth stating rather than burying

**A button may not contain another button.** A dismissible pill is inherently two controls, so `as="button"` with `dismissible` cannot be honoured. Rather than emit invalid markup it degrades to the div root — the same graceful-degradation rule `Card` applies to `as="a"` with no `href`. Asserted, including that the resulting subtree contains exactly one button.

**`disabled` stays `aria-disabled`, not the native attribute.** A natively disabled button leaves the tab order entirely, so a keyboard user never reaches it and never learns it is disabled. Keeping `aria-disabled` matches the div root and keeps the chip discoverable while the handlers stay inert.

**Opting into better semantics must not restyle the chip — and my first attempt did.** A native button arrives with user-agent styling a div never had, so I neutralised it with `font: inherit`. That shorthand *also* resets `line-height`, and `button.pill` outranks `.pill` on specificity, so the result was a button chip **30px** tall against the identical div's **25px**. The visual-parity test caught it. The rule now touches only `margin`, `font-family` and `appearance`, and carries a comment saying why `font: inherit` must not come back.

## Web component

The WC layer renames the two new props to `pillAriaExpanded` / `pillAriaPressed`, exactly as it already renames `title` to `pillTitle`: **ARIAMixin reflects `ariaExpanded` and `ariaPressed` as accessors on every `Element`**, so declaring them would replace the platform's own. The `prop-parity` guard caught this before it shipped, which is what it is for:

```
AssertionError: new host-accessor overrides:
  Pill.wc.svelte:ariaExpanded, Pill.wc.svelte:ariaPressed
```

The attributes are unchanged — `<sui-pill as="button" aria-expanded="true">` works as written.

## Visual baseline

`tests/visual/__screenshots__/pill.png` is regenerated, because the demo page gained a section for these fixtures.

It was **not** updated blind. The failing comparison was inspected first — `980x1219` → `980x1406`, 3011 differing pixels — and the diff image confirms every pixel above the new section is unchanged, so the update records the addition rather than blessing a regression.

## Verification

| check | result |
| --- | --- |
| `pnpm lint` | 0 |
| `pnpm check` | 0 |
| `pnpm build` | 0 |
| unit (vitest, incl. wc prop-parity) | 919 passed |
| Playwright — full Pill surface | 31 passed |
| Playwright — full functional suite | 605 passed, 1 skipped |

Red-green confirmed by disabling the button root and watching the root-element assertion fail while the rest held.


src/lib/Pill/Pill.svelte
```diff
@@ -50,16 +68,19 @@
...
   class:disabled
+  type={isButtonRoot ? 'button' : null}
   onclick={interactive ? handleClick : null}
-  onkeydown={interactive ? handleKeydown : null}
-  role={interactive ? 'button' : null}
-  tabindex={interactive ? 0 : null}
+  onkeydown={needsInteractiveShim ? handleKeydown : null}
+  role={needsInteractiveShim ? 'button' : null}
+  tabindex={needsInteractiveShim ? 0 : null}
   aria-disabled={interactive && disabled ? true : null}
```
Tara-ag: ⚠️ MAJOR: Disabled button root without `onclick` is not announced as disabled. The PR deliberately avoids the native `disabled` attribute on the button root, but `aria-disabled` is still gated on `interactive` (i.e. `onclick` is a function) rather than on whether the root is actually a control. A `<Pill as="button" disabled text="X" />` without `onclick` renders a focusable native `<button>` with neither `disabled` nor `aria-disabled`, so screen-reader users cannot discover it is disabled. Change `aria-disabled={interactive && disabled ? true : null}` to `aria-disabled={exposesControlState && disabled ? true : null}` to match the gating used for `aria-expanded`/`aria-pressed`.

docs/Pill.md
```diff
@@ -84,6 +84,50 @@ once per theme, then pass `tone` at each call site:
...
+
+**`as="button"` and `dismissible` cannot be combined.** A `<button>` may not contain another
+button, and a dismissible pill is inherently two controls. Rather than emit invalid markup,
+that combination falls back to the `div` root and its shim — the same graceful-degradation rule
+`Card` applies to `as="a"` with no `href`. If you need both, render the dismiss control as a
+sibling of the pill rather than inside it.
+
+Opting in changes semantics only, not appearance: a button root is neutralised back to the same
+computed font, padding, margin, border and height as the equivalent `div`.
+
```
Tara-ag, line 201: 💡 MINOR: Docs say disabled button stays discoverable but omit the `onclick` dependency. The docs state that `disabled` uses `aria-disabled` so the button stays discoverable, but the implementation only emits `aria-disabled` when `onclick` is also provided. A reader following the docs literally may believe `as="button" disabled` alone is sufficient for accessible disabled state. Either clarify that `aria-disabled` is emitted only when the pill is interactive, or align the implementation to use `exposesControlState` for `aria-disabled`.

Tara-ag (changes requested):
Good change overall — the button-root path is well-reasoned, thoroughly tested, and preserves backward compatibility. I found one accessibility gap in the new `as="button"` path and a corresponding docs inaccuracy that should be addressed before merging.

Tara-ag (approved):
No blocking issues found. The change is an additive, backward-compatible enhancement: `as="button"`, `ariaExpanded`, and `ariaPressed` are new optional props, default behavior remains identical, and the web-component wrapper correctly avoids host-accessor collisions. Tests and docs are updated.

Tara-ag:

The PR is a well-scoped, backward-compatible enhancement that adds real button-root semantics to Pill. However, the gate accepted one MAJOR accessibility issue: a disabled button root without `onclick` fails to announce its disabled state because `aria-disabled` is gated on `interactive` rather than on `exposesControlState`. There is also a MINOR docs inconsistency. Both should be addressed before merge.

- ⚠️ **MAJOR**: Disabled button root without onclick is not announced as disabled — `src/lib/Pill/Pill.svelte:81`
  The PR deliberately avoids the native `disabled` attribute on the button root (`disabled` stays `aria-disabled, not the native attribute`). However, `aria-disabled` is currently gated on `interactive` (i.e. `onclick` being a function), not on whether the root is actually a control. A `<Pill as="button" disabled text="X" />` without `onclick` therefore renders a focusable native `<button>` that has…
- 💡 **MINOR**: Docs say disabled button stays discoverable but omit the onclick dependency — `docs/Pill.md:130`
  docs/Pill.md states that `disabled` uses `aria-disabled` so the button stays discoverable, but the implementation only emits `aria-disabled` when `onclick` is also provided. A reader following the docs literally may believe `as="button" disabled` alone is sufficient for accessible disabled state. The docs should clarify that `aria-disabled` is emitted only when the pill is interactive, or the impl…

murdore:
Both taken. The MAJOR is a real defect and this PR is what introduced it — thanks for catching it.

## ⚠️ MAJOR — disabled button root without `onclick` is not announced as disabled

Correct, and worse than a gap in polish: **this PR created the regression.** Before it, `<Pill disabled text="X" />` rendered a non-focusable `<div>` that nobody could tab to, so the missing `aria-disabled` cost nothing. `as="button"` makes it a real, focusable control — and I gated `aria-disabled` on `interactive` while correctly gating `aria-expanded` / `aria-pressed` on `exposesControlState`. So a keyboard user could land on a `<button>` with neither `disabled` nor `aria-disabled` and nothing announcing it was inert.

My own tests missed it because every `as="button"` fixture I wrote had an `onclick`. The one shape that breaks is the one I never rendered.

Fixed exactly as suggested — `aria-disabled={exposesControlState && disabled ? true : null}` — with a demo fixture carrying no handler and a test that fails without the fix:

```
before:  Expected "true", Received ""     ← focusable <button>, nothing announced
after:   aria-disabled="true", still focusable
```

The `exposesControlState` declaration now carries a comment saying why it must not be gated on `onclick`, so the next person doesn't re-tighten it.

## 💡 MINOR — docs omitted the `onclick` dependency

Same root cause, so the docs are now accurate rather than merely clarified: a button root emits `aria-disabled` whether or not an `onclick` was supplied (it is a real control either way), while a `div` root emits it only once an `onclick` makes it one. That distinction is stated explicitly.

## Verification

lint 0 · check 0 · build 0 · 919 unit (incl. the wc prop-parity guard) · 8 interactive-root Playwright tests · 24 pre-existing Pill tests (attrs, dismiss-colour, line-height, EmptyState) · visual regenerated.

The `pill.png` baseline moved again for the new fixture. As before it was not updated blind — the failing diff was inspected first (894 pixels, no dimension change) and the diff image confirms the only movement is the new chip being inserted into the last row and shifting the two chips after it. Nothing above changed.

src/lib/Pill/Pill.svelte
murdore: Correct, and this PR is what introduced it — thanks for catching it.

Before this change, `<Pill disabled text="X" />` rendered a non-focusable `<div>` nobody could tab to, so the missing `aria-disabled` cost nothing. `as="button"` makes it a real, focusable control, and I gated `aria-disabled` on `interactive` while correctly gating `aria-expanded`/`aria-pressed` on `exposesControlState`. So a keyboard user could land on a `<button>` with neither `disabled` nor `aria-disabled` and nothing announcing it was inert.

My own tests missed it because every `as="button"` fixture I wrote had an `onclick`. The one shape that breaks is the one I never rendered.

Fixed exactly as suggested, with a demo fixture carrying no handler and a test that fails without it:

```
before:  Expected "true", Received ""     ← focusable <button>, nothing announced
after:   aria-disabled="true", still focusable
```

The `exposesControlState` declaration now carries a comment saying why it must not be re-tightened onto `onclick`.

docs/Pill.md
murdore: Fixed at the source rather than clarified: since `aria-disabled` now uses `exposesControlState`, the docs statement is accurate as written. I also made the distinction explicit — a button root emits `aria-disabled` whether or not an `onclick` was supplied (it is a real control either way), while a div root emits it only once an `onclick` makes it one.

Tara-ag (approved):
Reviewed file-by-file. The change is additive (`as`, `ariaExpanded`, `ariaPressed`), defaults preserve the existing `<div>` root byte-for-byte, and the button-root path correctly avoids nested buttons when `dismissible` is set. Accessibility handling is sound: real button semantics when requested, `aria-disabled` kept for discoverability, and ARIA state attributes gated to interactive roots. WC prop renaming respects ARIAMixin. Tests and docs cover the new behavior. No blocking issues found.

PR #582, opened 2026-09-08
chore(deps): refresh vulnerable transitives, clearing every high advisory

murdore:
Clears **11 of the 13 open dependabot alerts, including all 10 highs**, by refreshing transitive resolutions inside their existing ranges. **No manifest change — the lockfile alone moves.**

| package | | | clears |
| --- | --- | --- | --- |
| `fast-uri` | 3.1.2 → | **3.1.7** | 6 high |
| `js-yaml` | 4.2.0 → | **4.3.2** | 2 high |
| `postcss` | 8.5.15 → | **8.5.28** | 1 high + 1 medium |
| `brace-expansion` | 5.0.6 → | **5.0.9** | 1 high |
| `postcss-selector-parser` | 7.1.1 → | **7.1.6** | 1 low |

(`nanoid` 3.3.12 → 3.3.18 comes along as a postcss transitive.)

## Context that changes how urgent this was

**Every one of the thirteen is `development` scope.** The published package declares **no runtime dependencies at all** — only peers (`lottie-web`, `marked`, `svelte`, `type-decoder`, `typescript`), none of which appear in any advisory:

```
$ npm view @juspay/svelte-ui-components@4.11.0 dependencies
(empty)
```

So nothing vulnerable has ever reached a consumer's `node_modules`. This is build-toolchain exposure, not shipped exposure. Still worth fixing — not worth alarm. All thirteen are transitive; no direct dependency is vulnerable.

## The two left open, deliberately

**`cookie` (low, CVE-2024-47764)** is pinned by SvelteKit, not by us. `@sveltejs/kit@2.70.3` is the current latest and *still* declares `cookie: ^0.6.0`:

```
$ npm view @sveltejs/kit@latest dependencies.cookie
^0.6.0
```

There is no version to upgrade to. Forcing a `pnpm.overrides` against kit's own declared range — to clear one low-severity dev-only advisory in a project that ships no server and reads no cookies — is a worse trade than the advisory. It resolves when upstream moves.

**`@sveltejs/kit` (medium, wants 2.69.1)** is deliberately not bundled here. I tried it: **2.70.3 breaks `pnpm check` with 28 errors** — `Cannot find name 'node:fs'`, `'process'`, and similar across `tests/` and `scripts/`. The cause is structural, not incidental: `tsconfig.json` extends `./.svelte-kit/tsconfig.json`, which kit *generates*, so the upgrade changes type resolution underneath the repo's own config.

That was isolated rather than assumed — with kit pinned back to 2.63.1 and all six transitive bumps kept, `check` returns **0**. The bumps are provably not the cause, so this change is safe to land alone. The kit upgrade needs its own PR that actually fixes the type resolution; folding a broken `check` into a security fix to make one number go down would be the wrong trade.

## Verification

| check | result |
| --- | --- |
| `pnpm lint` | 0 |
| `pnpm check` | 0 |
| `pnpm build` | 0 |
| Playwright functional | 629 passed, 1 skipped |

The visual run is not ceremony here: **postcss processes every stylesheet in the library**, and a CSS-processor bump is exactly the kind of change that moves rendering by a pixel without touching a line of source. It didn't.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -752,9 +752,9 @@ packages:
   bidi-js@1.0.3:
     resolution: {integrity: sha512-RKshQI1R3YQ+n9YJz2QQ147P66ELpa1FQEg20Dk8oW9t2KgLbpDLLp9aGZ7y8WHSshDknG0bknqGw5/tyCs5tw==}
 
-  brace-expansion@5.0.6:
-    resolution: {integrity: sha512-kLpxurY4Z4r9sgMsyG0Z9uzsBlgiU/EFKhj/h91/8yHu0edo7XuixOIH3VcJ8kkxs6/jPzoI6U9Vj3WqbMQ94g==}
-    engines: {node: 18 || 20 || >=22}
+  brace-expansion@5.0.9:
+    resolution: {integrity: sha512-ScQ4IuvIEF1TMlP7Zt+vjJ//9zlPb2SDcxWxM3bk8s6t6GGdJ7KO1dCcTidOPJKePW30LE/2cT7wCyPho9/Wxg==}
+    engines: {node: 20 || >=22}
 
   callsites@3.1.0:
     resolution: {integrity: sha512-P8BjAsXvZS+VIDUI11hHCQEv74YT67YUi5JJFNWIqL235sBmjX4+qx9Muvls5ivyNENctx46xQLQ3aTuE7ssaQ==}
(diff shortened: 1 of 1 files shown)


Tara-ag (approved):
Lockfile-only transitive dependency refresh within existing ranges. No manifest or source changes; no public API, accessibility, reactivity, SSR, or security surface to review. Verification results in the PR description cover the relevant mechanical checks. Approved.

Issue #415, opened 2026-07-17
Select: in-menu search box variant (DS keyboard contract)


murdore:
The DS dropdown sheet's keyboard-accessibility guidelines (Figma node 34901:7034) place the search box **inside the open menu** (searchbox access via Tab after opening), while Select's `searchable` puts the filter input in the trigger. An in-menu search variant would match the DS contract; the sheet also specs multi-select Clear-All (pinned row) + sticky 'Select (N)' apply, both currently consumer recipes via showSelectAll/bottomContent.

Issue #414, opened 2026-07-17
Select: size variants (sm/md/lg), error state, trigger clear button, pressed state


murdore:
The Breeze Dashboard DS dropdown sheet (Figma node 827-44480) specs a 130-variant Dropdown system the Select component partially covers. Missing pieces, in priority order:

1. **Sizes sm/md/lg** — one trigger size today; consumers hand-roll min-height overrides (Lighthouse carries a light-theme-only 36px `!important` in shopify.css as a workaround).
2. **Error state** — red-border + hint treatment (`State=Error` in the sheet).
3. **Trigger clear button (×)** — `Has Clear Button=Yes` variants; resets selection without opening the menu.
4. **Pressed state** — `#ededed`-tint active/pressed feedback on options and trigger.

Reference captures live in the Lighthouse repo under docs/analytics-figma-parity/evidence/figma-ds/ (dropdown-summary.png).

PR #585, opened 2026-09-08
feat(select): add sm/md/lg trigger size presets

murdore:
Item 1 of #414 — the one with a live consumer workaround to retire: Lighthouse carries a light-theme-only `36px !important` in `shopify.css` because Select had a single trigger size.

## What was actually missing

Worth being precise, because it changes the design. `--select-trigger-min-height`, `--select-trigger-padding` and `--select-font-size` have **always** been themeable individually. The gap was never "Select cannot be resized" — it was that resizing it *coherently* means knowing which three variables to set and keeping them in step. That is why a consumer reached for `!important` on the computed property instead.

So `size` names the combinations rather than adding new capability:

| size | min-height | padding | font-size |
| --- | --- | --- | --- |
| `sm` | 32px | 4px 10px | 13px |
| `md` (default) | 40px | 8px 12px | 14px |
| `lg` | 48px | 12px 14px | 15px |

## Built on the pattern this repo already has

- **`selectSize.ts`** exports a pure `selectSizeClass()`, unit-testable without a browser — vitest here runs no DOM, so a `.svelte` template's output cannot be asserted directly. Same reason `pillTone.ts` and `normalizeColumns.ts` exist.
- An internal **`--_select-size-*` layer sits behind** the public token in each `var()` chain, so an explicit `--select-trigger-min-height` still wins — and wins *alone*, leaving the rest of the preset in force.
- **`md` resolves to no class at all**, so a consumer who never sets `size` renders byte-identically to before this prop existed.

## Evidence

Red-green verified rather than assumed. With the `var()` chains reverted, **3 of the 5** new Playwright tests fail (heights, font sizes, precedence) while the two compatibility tests keep passing — correctly, since they assert the *absence* of change, which holds either way. Those two would not have caught a regression on their own, and saying so is more useful than counting them as coverage.

The wc prop-parity guard caught the missing `Select.wc.svelte` declaration before CI did:

```
AssertionError: Select.wc.svelte is missing: size
```

## One thing I could not fully explain

Adding the demo fixtures shifts a paragraph roughly 3000px **above** them by ~2px, in the pinned container only. The regenerated `select.png` therefore contains a small reflow that is not part of the new section, and I would rather flag that than let the baseline imply the diff is only my addition.

**What I did establish:**

- **The library CSS is innocent.** With the component change applied and the demo fixtures removed, `select.png` passes byte-identically. That bisect is the finding that matters.
- It is **deterministic**, not a font-loading race — repeated runs give an identical differing-pixel count (107,089).
- Below the divergence it is a **pure offset**: `mismatch=0` at `delta=+2`, with rows 0–910 byte-identical.
- It reproduces **only in the Linux container**, never on macOS — which is also the mistake I made early on, measuring the wrong renderer for several rounds before noticing.

**What I did not:** the mechanism. The best remaining hypothesis is a scrollbar-width interaction with the harness's viewport-fitting loop (Linux reserves scrollbar width, macOS overlay scrollbars do not), but `main.content` measures 980px wide in *both* captures, which that hypothesis does not account for.

## Scope

Closes nothing on its own. **#414 items 2–4** (error state, trigger clear button, pressed state) and **#415** remain open, each warranting its own PR — the error state in particular needs a decision on whether hint text is a prop or a snippet.

## Verification

| check | result |
| --- | --- |
| `pnpm lint` | 0 |
| `pnpm check` | 0 |
| `pnpm build` | 0 |
| unit (vitest, incl. wc prop-parity) | 924 passed |
| Playwright functional | 634 passed, 1 skipped |

Re-verified after rebasing onto `b929c18`, since `release` moved while this was in progress.

## Summary by CodeRabbit

* **New Features**
  * Added small, medium, and large size presets to the Select component.
  * Added support for configuring Select size through the web component’s `size` attribute.
  * Added CSS variable overrides for customizing preset dimensions.

* **Documentation**
  * Added Select sizing guidance and interactive examples, including override behavior.

* **Tests**
  * Added coverage for size variants, default compatibility, sizing order, and CSS overrides.


docs/Select.md
Tara-ag, line 1: 💡 MINOR: Web component `size` attribute not listed in attributes table

The PR adds `size` to `Select.wc.svelte` as a reflected String attribute, but the 'Attributes (Web Component)' table does not include it. Since `size` can be set directly via HTML attribute on `<sui-select>` (unlike array props), it should appear alongside `placeholder`, `multiple`, `searchable`, etc.

```markdown
| `size` | `size` | `'sm' \| 'md' \| 'lg'` | Trigger size preset. Defaults to `'md'`. |
```

Tara-ag (commented):
Approved with one minor documentation note.

The change is cleanly backward-compatible: `size` is optional, `md`/omitted emits no class, existing CSS custom property names are preserved, and explicit `--select-trigger-*` / `--select-font-size` tokens still win over the preset via the internal `--_select-size-*` layer. The pattern mirrors `Pill`'s `tone` implementation, and web-component parity is maintained.

One documentation gap: the web component attributes table in `docs/Select.md` should list the new reflected `size` attribute.

Tara-ag (commented):
Review in progress.

Tara-ag:
💡 MINOR: Web component `size` attribute not listed in attributes table

The PR adds `size` to `Select.wc.svelte` as a reflected String attribute, but the **Attributes (Web Component)** table in `docs/Select.md` does not include it. Since `size` can be set directly via HTML attribute on `<sui-select>` (unlike array props), it should appear alongside `placeholder`, `multiple`, `searchable`, etc.

**Suggested fix:** Add a row to the table in `docs/Select.md`:

```markdown
| `size` | `size` | `'sm' \| 'md' \| 'lg'` | Trigger size preset. Defaults to `'md'`. |
```

docs/Select.md
murdore: Correct, fixed. `size` is reflected on `<sui-select>`, so it is settable straight from HTML — unlike the array props — and leaving it out of the attributes table meant the one audience who needs that table most had no way to discover it.

Added alongside `placeholder` / `multiple` / `searchable`, noting that it is reflected and pointing at the **Sizing** section for the preset values.

Tara-ag (approved):
Approved. The change is a clean, backward-compatible addition: a new optional `size` prop with `'sm' | 'md' | 'lg'` presets, where `'md'`/omitted emits no class and preserves the existing rendering. Public CSS custom properties retain precedence over the internal `--_select-size-*` fallbacks, web-component parity is maintained, and tests/docs are updated. No blocking issues found.

murdore:
Closing this in favour of a documented class recipe. The API this added was not worth its weight.

The case against it is simple: **`--select-trigger-min-height`, `--select-trigger-padding` and `--select-font-size` are already themeable, so sizing works today on 4.11.2 with no library code at all.**

```css
/* app.css */
.select-sm {
  --select-trigger-min-height: 32px;
  --select-trigger-padding: 4px 10px;
  --select-font-size: 13px;
}
```

```svelte
<Select {items} classes="select-sm" />
```

Against that, this PR spent a `SelectSize` type, a `selectSize.ts` resolver, a unit-test file, a reflected web-component attribute, a prop-parity entry and an internal `--_select-size-*` CSS layer — to deliver three numbers a consumer can already set, and to lock those numbers to a release cycle rather than letting an app define its own scale.

Worth recording the counter-argument, since it is real and this repo's own history supports it: `docs/Pill.md` documents "Theming with Classes", and #520 later promoted that pattern to a `tone` prop precisely because the class recipe meant "a hand-rolled `tone-*` class repeated at every call site". So the library has gone class → prop before. The judgement here is that `tone` is design-system-governed in a way trigger geometry is not — a consumer wanting a 36px trigger should not need us to ship a `size` value for it.

Replacing with a docs-only recipe in `docs/Select.md`, carrying the same measured values. Nothing is lost from the shipped library, because nothing was shipped: this was never merged.

#414 item 1 is addressed by that recipe rather than by this API. Items 2–4 (error state, trigger clear button, pressed state) and #415 remain open and are unaffected — those are genuine behaviour, not token bundles, and will need real props.

PR #586, opened 2026-09-08
fix(img): preserve data-URI artwork in inlineSvg sanitization

rajarshi-pal:
## Problem

`Img`'s `inlineSvg` path adds fetched SVG markup to the live document and sanitizes it first: every descendant `href`/`xlink:href`/`src` attribute is removed unless the value starts with `#`. That hardening is right for remote/`javascript:` URLs — but it also removes `data:image/*` hrefs.

Brand/company marks (merchant-platform logos, payment marks, ad-platform logos, social icons) are very often not stroke/path art. They come out of Figma as:

```svg
<rect fill="url(#pattern0)"/>
<pattern id="pattern0"><use xlink:href="#art0"/></pattern>
<image id="art0" xlink:href="data:image/png;base64,…"/>
```

The base64 payload **is** the artwork; the vector part is just a fill instruction. Stripping the attribute yields a structurally valid, safe, **invisible** icon. Today this blanks every brand-mark icon rendered through the library's forced-inlining seams: `Modal`'s hardcoded `` on `header.leftImage`/`rightImage`, `Table.BuiltinCell` icon cells, `Icon`, `Menu`, `Tabs`, `Toast`, `Status`, `CommandMenu`, `ChatSuggestions`, `FileDropzoneTrigger`.

Consumer evidence: Lighthouse BZ-6079 / BZ-6080 — data-source logos missing across Manage Juspay / connection-modal headers, signal cards, and the signal detail modal.

## Fix

One allowance in the descendant URL rule: a value is adoptable when it is a same-document fragment (`#…`) **or a `data:image/*` payload**:

```ts
const trimmed = value.trim();
return trimmed.startsWith('#') || trimmed.startsWith('data:image/');
```

Why this is safe:

- **No fetch** — the payload carries its own bytes; nothing can dial out on adoption (the actual remote-URL concern the original guard encoded).
- **No script context** — inside `<image>`/`<use>`, a data-URL image payload is raster/vector data, never an execution context; `onerror`/`on*` handlers remain stripped by the separate `on*` pass.
- `javascript:` and genuine remote URLs are removed exactly as before — asserted by the new spec (hostile remote + javascript siblings on the same payload lose their hrefs while the artwork keeps its own).

## Tests

New spec `tests/img-inline-svg-data-uri-artwork.spec.ts`: intercepts the demo asset with a Figma-style pattern-filled brand mark + hostile `https://` and `javascript:` siblings; asserts art href + use-fragment survive, hostile hrefs don't, pattern structure intact. Negative control: the spec **fails against the pre-fix predicate** ("pattern art xlink:href was stripped").

Green locally: this spec + `img-inline-svg-attr-allowlist`, `img-inline-svg-src-change`, `img-review-findings`, `modal-left-image-interactivity` (**12 passed**).

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved valid embedded image data when displaying inline SVG artwork.
  - Continued removing unsafe remote and JavaScript-based image references from inline SVGs.

- **Tests**
  - Added coverage for inline SVGs containing pattern-based artwork and embedded image data.

### DIFF
--- a/src/lib/Img/Img.svelte
+++ b/src/lib/Img/Img.svelte
@@ -145,16 +145,27 @@
   const URL_ATTRIBUTES: ReadonlySet<string> = new Set(['href', 'xlink:href', 'src']);
 
   /**
-   * Whether a URL attribute points inside this same document (`#gradient`).
+   * Whether a URL attribute is safe to adopt into the live document.
    *
-   * Only a fragment survives. A `javascript:` URL is the classic vector, but
-   * an ordinary remote URL is not safe either: on `<image href>` or `<use
-   * href>` it fetches on adoption, which is a request the consumer's page
-   * never asked to make, from markup the consumer did not write. A same
-   * document reference can do neither.
+   * A fragment (`#gradient`) points inside this same document and is safe on
+   * any element. A `data:image/*` payload also carries its art inline and makes
+   * no network request on adoption — but it is only safe on image-rendering
+   * elements (`<image>`, `<feImage>`), where the payload is rasterized with no
+   * script context. On any other element (`<use>`, `<a>`, …) the allowance
+   * stays fragment-only. Media type comparison is case-insensitive because
+   * MIME types are. Everything else is unsafe: a `javascript:` URL is the
+   * classic vector, and an ordinary remote URL fetches on adoption, which is
+   * a request the consumer's page never asked to make, from markup the
+   * consumer did not write.
    */
-  function isSameDocumentReference(value: string): boolean {
-    return value.trim().startsWith('#');
+  function isSafeInlineUrlValue(element: Element, value: string): boolean {
+    const trimmed = value.trim();
+    if (trimmed.startsWith('#')) {
+      return true;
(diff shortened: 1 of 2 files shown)


dakshana-j (approved):

rajarshi-pal:
Both CodeRabbit findings addressed in `4d8cdfa` (amended -- PR history is a single fix commit):

1. **Element scoping (Major)**: `data:image/*` is now accepted **only on `<image>` and `<feImage>`**. `<use>` and `<a>` keep the fragment-only rule; `javascript:` and remote URLs stay stripped on every element.
2. **Case-insensitivity (Minor)**: the media-type prefix comparison uses `toLowerCase()`; the original attribute value is preserved untouched.

The spec gained three assertions proving both: uppercase `data:image/PNG` artwork survives inside `<image>`; a hostile `data:image/svg+xml` `<use>` reference and a `data:image/png` `<a href>` lose their hrefs. Full Img-focused suite green (12/12).

PR #588, opened 2026-09-08
fix(deps): declare node types so SvelteKit 2.70 can be adopted

murdore:
Supersedes #584. Adopts SvelteKit 2.70.3 and fixes the type-resolution break that made the bump fail, closing the repo's last open Dependabot advisory.

## The failure

Dependabot's bump fails `check` with 28 errors that name no cause:

```
Cannot find name 'node:fs'.
Cannot find name 'node:path'.
Cannot find name 'process'.
Property 'dirname' does not exist on type 'ImportMeta'.
```

`@types/node` is declared (`^22.20.1`), locked and installed throughout. It simply stops resolving — on a bump that touches no source file.

## The cause is a one-line diff in a file nobody wrote

`tsconfig.json` extends `./.svelte-kit/tsconfig.json`, which SvelteKit **generates**. Diffing that generated file across the two versions:

```diff
  "target": "esnext",
- "types": ["node"]        ← emitted by kit 2.63.1, not by 2.70.3
```

That is the entire difference.

## Why the fix is a convention, not a workaround

The root `tsconfig.json` was the **only** config in the repo relying on that injection:

| Config | `types` | Affected |
| --- | --- | --- |
| `tsconfig.json` | *inherited from generated* | **yes** |
| `tsconfig.wc.json` | `["vite/client"]` | no |
| `scripts/codemod/tsconfig.json` | `["node"]` | no |
| `scripts/migrate/tsconfig.json` | `["node"]` | no |
| `scripts/release/tsconfig.json` | `["node"]` | no |
| `scripts/wc-parity/tsconfig.json` | `["node"]` | no |

Four of six already declare it explicitly. This brings the outlier in line, and makes type resolution here independent of whatever a generated file happens to contain.

## Measured, not inferred

Controlled experiment at kit 2.70.3, single variable:

| `types: ["node"]` | svelte-check |
| --- | --- |
| absent | **28 errors**, exit 1 |
| present | **0 errors**, exit 0 |

## The regression test, and why it earns its place

The line looks redundant beside a generated config that used to supply it — which is exactly what makes it easy to delete, and deleting it costs an investigation. This one cost two: once here when kit was deliberately excluded from #582, and again on Dependabot's own PR.

`src/tsconfig-node-types.test.ts` turns that deletion into a failure that names the cause. Red-green verified rather than assumed:

```
AssertionError: tsconfig.json must declare compilerOptions.types
  — SvelteKit no longer injects it: expected null not to be null
```

Restored, it passes. So the guard fails for the right reason, not because of a typo in the test.

## Scope

2.70.3 rather than Dependabot's 2.70.2 — current patch, same advisory fix.

**Deliberately not included:** `svelte-check` also warns that `moduleResolution: "Node"` (node10) is deprecated and stops working in TypeScript 7. That is pre-existing, is not caused by this bump, and bundling it would blur what this change proves. Worth its own PR.

## Verification

| Gate | Exit |
| --- | --- |
| `lint` | 0 |
| `check` | 0 |
| `test:unit` | 0 |
| `build` | 0 |
| `test:integration` | 0 |

Closes #584

## Summary by CodeRabbit

- **Chores**
  - Updated the SvelteKit development dependency to a newer version.

- **Bug Fixes**
  - Ensured Node.js type definitions are explicitly available through the project’s TypeScript configuration.

- **Tests**
  - Added coverage verifying that Node.js types are declared directly in the root TypeScript configuration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "@playwright/test": "^1.60.0",
     "@sveltejs/adapter-auto": "^7.0.1",
     "@sveltejs/adapter-static": "^3.0.10",
-    "@sveltejs/kit": "^2.63.1",
+    "@sveltejs/kit": "^2.70.3",
     "@sveltejs/package": "^2.5.8",
     "@sveltejs/vite-plugin-svelte": "^7.1.2",
     "@testing-library/svelte": "^5.4.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -25,13 +25,13 @@ importers:
         version: 1.60.0
       '@sveltejs/adapter-auto':
         specifier: ^7.0.1
-        version: 7.0.1(@sveltejs/kit@2.63.1(@sveltejs/vite-plugin-svelte@7.1.2(svelte@5.56.3(@typescript-eslint/types@8.60.1))(vite@8.0.16(@types/node@22.20.1)(jiti@2.6.1)))(svelte@5.56.3(@typescript-eslint/types@8.60.1))(typescript@6.0.3)(vite@8.0.16(@types/node@22.20.1)(jiti@2.6.1)))
+        version: 7.0.1(@sveltejs/kit@2.70.3(@sveltejs/vite-plugin-svelte@7.1.2(svelte@5.56.3(@typescript-eslint/types@8.60.1))(vite@8.0.16(@types/node@22.20.1)(jiti@2.6.1)))(svelte@5.56.3(@typescript-eslint/types@8.60.1))(typescript@6.0.3)(vite@8.0.16(@types/node@22.20.1)(jiti@2.6.1)))
       '@sveltejs/adapter-static':
         specifier: ^3.0.10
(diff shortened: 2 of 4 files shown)


murdore:
## Testing and validation evidence

Final PR commit: `911e58494b27e32b1224849f4b53f9e637755ad1`; rebased merge: `14ff6027efab55bcc9f479267536b2ab5dcb3138`.

- [Exact-commit CI run](https://github.com/juspay/svelte-ui-components/actions/runs/34258430423)
- [Download the Playwright HTML report, videos and traces](https://github.com/juspay/svelte-ui-components/actions/runs/34258430423/artifacts/10069214763)
- [Successful release run](https://github.com/juspay/svelte-ui-components/actions/runs/34267186372); npm version `4.11.5` independently verified.

The report was downloaded and its embedded data archive inspected: **631 total, 630 passed, zero failed, zero flaky, one skipped**. Every referenced attachment exists and is non-empty. There are 629 video files and 631 trace archives. One passing static-source test (`no component still declares two spellings of the same event`) has no browser video because it does not open a page.

The one skip is explicitly NOT counted as tested: `sibling radio-mode Choiceboxes form a single-tab-stop group with arrow-key roving selection`.

Local report: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-svelte-ui-components/fa89880e-485f-4141-b463-d288ec6722af/scratchpad/evidence/588-ci/index.html`

Validation receipt: `/private/tmp/claude-501/-Users-sachinsharma-Developer-temp-svelte-ui-components/fa89880e-485f-4141-b463-d288ec6722af/scratchpad/evidence/588-proof-validation.json`

CI artifacts have finite retention. The local copy preserves the report. File/reference validation does not claim every video was manually viewed or fully decoded.

Security follow-up: the kit advisory is resolved, but GitHub currently still lists a **low-severity development-only cookie advisory**. This merge does not close all security debt.

PR #589, opened 2026-09-08
feat(markdown-text): let sanitize strip raw HTML instead of escaping it

### LINKED ISSUE #573: MarkdownText: sanitize cannot strip raw HTML, only escape it
### Need

`MarkdownText`'s `sanitize` prop can narrow which markdown-*generated* tags render, but it has no effect at all on **raw HTML typed literally into the source**. A consumer that wants raw HTML *removed* rather than *escaped* has to pre-process the source before it reaches the component, which defeats the point of the prop.

### Evidence

In `markdown.js`, the `html(token)` renderer unconditionally returns `escapeHtml(token.text)` with no `tagAllowed` gate — the behaviour is identical with and without `allowedTags: []`. So `allowedTags: []` does force every markdown-generated tag (bold, links, headings, lists, tables) to plain text, but a literal `<script>alert(1)</script>` in the source still renders as visible escaped tag text.

Escaping is the right *safety* default and nobody is asking to change it. The gap is that "escape it" and "drop it" are different products, and only one is expressible.

### Where this bit

TARA renders agent- and user-authored text in transcripts and approval cards. It runs `sanitiseMarkdownSource()` over the source first, which strips raw HTML entirely, so stray angle-bracket text never appears. It cannot retire that pre-processing in favour of `sanitize`, because doing so would newly render escaped tag text in transcripts that show nothing today — a visible regression in exactly the surface where a stray `<` is most likely to arrive.

So the workaround stays, and `sanitize` ends up narrowing protocols only.

### Proposal

Add a third option alongside `allowedProtocols` and `allowedTags`, for example:

```ts
sanitize?: {
  allowedProtocols?: string[]
  allowedTags?: string[]
  rawHtml?: 'escape' | 'strip'   // default 'escape' — today's behaviour
}
```

`'strip'` would drop raw HTML tokens instead of escaping them, leaving their text content. Purely additive; a consumer setting nothing gets today's rendering byte-for-byte.

murdore:
Adds `sanitize.rawHtml` so raw HTML in the source can be **stripped to its text** rather than only escaped.

## The gap

`sanitize` could narrow which markdown-**generated** tags render, but had no effect at all on raw HTML typed literally into the source. `markdown.ts` returned `escapeHtml(token.text)` with no gate, so `allowedTags: []` behaved identically with and without it, and a literal `<script>alert(1)</script>` still rendered as visible escaped tag text.

Escaping is the right safety default and **stays the default**. The gap is that "escape it" and "drop it" are different products, and only one was expressible — so a surface wanting raw HTML removed had to pre-mangle the source with its own regex before it reached the component. TARA does exactly that, and the issue records why it cannot retire the workaround in favour of `sanitize`.

```svelte
<MarkdownText {markdown} sanitize={{ rawHtml: 'strip' }} />
```

## I had the contract wrong, and fixed it

An earlier revision of this branch dropped **block-level** raw HTML whole, because `marked` hands the renderer the entire raw block as one string. #573 asked for tags to be dropped *"leaving their text content"*. That was my decision rather than the issue's, and the wrong one to make silently: `<div>hello</div>` losing the word "hello" is data loss a consumer cannot see happening.

Each raw token is now parsed with `parse5` — an inert, DOM-free parser, so SSR and browser render identically — and its text nodes are kept:

| Source | `'escape'` (default) | `'strip'` |
| --- | --- | --- |
| `a <b>bold</b> word` | `a &lt;b&gt;bold&lt;/b&gt; word` | `a bold word` |
| `<div>hello <em>world</em></div>` | escaped | `hello world` |
| `<div title="a > b">visible</div>` | escaped | `visible` |
| `<div>before<!-- hidden -->after</div>` | escaped | `beforeafter` |
| `<div>fish &amp;amp; chips &amp;#60;3</div>` | escaped | `fish &amp;amp; chips &amp;lt;3` |
| `<script>alert(1)</script>` | escaped | `alert(1)`, as text |
| `a < b and c > d` | unchanged | unchanged |

Quoted `>` inside an attribute, entity references, unclosed tags and raw-text elements are exactly the cases a regex gets wrong — which is why this is a real parser, not a substitution.

**Stripping never widens what renders.** Every retained text node is escaped on the way out, so encoded markup cannot re-enter as markup; a test asserts `<div>&lt;img src=x onerror=alert(1)&gt;</div>` yields `&lt;img` and no `` element. Script and style contents survive as visible text and are never executed.

The last row is the one worth pinning: `a < b` is a **text** token, never an HTML one, so neither setting touches it. A strip that reached it would silently eat arithmetic out of a sentence. Fenced code is untouched for the same reason — a code block is not a raw-HTML token.

## Two traps found while building it

**The instance cache.** `renderMarkdown` caches a `Marked` instance per resolved option set. The mode had to join that cache key — without it, the first render through a given source would have decided disposal for every later one. Covered by a test that renders both modes in one process.

**Unrecognised values.** A plain-JS or web-component caller can pass anything. `'nonsense'` degrades to escaping rather than throwing, matching how `narrow` and `resolveTagAllowList` already handle untyped input. An unrecognised value must never be the reason content disappears.

## Additive, and proved so

Four of the new unit tests assert byte-for-byte parity with the previous output, including an explicit `rawHtml: 'escape'` vs no-option comparison across five sources in both block and inline mode. **Those four pass on the pre-change code**, which is what makes them back-compat tests rather than restatements of the new behaviour.

## Docs corrected, not just extended

Every claim this makes false is fixed rather than left standing: the overview's "no way to opt out", the `allowedTags` note, the security-model bullet, the `markdown` prop description and the `rawHtml` docstring.

`parse5` becomes a runtime dependency of the pipeline. The web component needs no change — it already declares `sanitize` as an `Object` and passes it through whole.

## Verification

| Gate | Exit |
| --- | --- |
| lint | 0 |
| check | 0 |
| test:unit | 0 |
| build | 0 |
| test:integration | 0 |

**940 unit tests** (45 over the markdown pipeline, 10 new for these cases) and **635 functional tests** with 1 pre-existing skip (`a11y-choicebox.spec.ts:78`, skipped on the default branch too).

**Proof of testing.** 5 Playwright cases, each recorded and **reviewed frame-by-frame** rather than assumed to exist — an earlier round of these videos was playable but framed on the wrong part of the page, so the tests now scroll the examples into view first.

The `markdown-text.png` visual baseline is regenerated because the strip example renders more text than it used to. The before/after pair was reviewed, and the committed baseline then passed a second comparison run **without** `--update-snapshots`.

## Two review findings, both real

**`<template>` text was lost.** parse5 stores a template's children in a separate `content` fragment rather than in `childNodes`, so the recursion walked past them and `<template>hello</template>` came out empty. Fixed, with tests for a lone template and one nested inside a `<div>`.

**parse5 was pinned to 8.0.0 while the lockfile already carried 8.0.1** transitively — two copies of the same parser installed for no benefit. Now `^8.0.1`, deduplicated to one.

Closes #573


Tara-ag (approved):
Reviewed file by file. The change is additive, backward-compatible, and well-tested:

- `sanitize.rawHtml` defaults to `'escape'`, preserving existing behavior byte-for-byte.
- The new `'strip'` mode never parses raw HTML as markup; it only changes visible output.
- The `Marked` instance cache key correctly includes the new mode, preventing cross-configuration leakage.
- Unknown values degrade to escaping, matching existing runtime-defensive patterns.
- Unit tests include back-compat parity assertions; Playwright tests verify rendered text and DOM safety.
- Docs and demo page are updated consistently; web-component prop parity is unaffected because `sanitize` is already passed through as an object.

No blocking issues found.

murdore:
Both points taken, and the second one changed the feature rather than the prose.

**The contract now matches the issue.** #573 asked for `'strip'` to drop raw HTML tokens *"leaving their text content"*. The first implementation dropped a block-level token whole, because `marked` hands the renderer the entire raw block as one string and keeping its text would mean parsing HTML. That was my decision, not the issue's, and it was the wrong one to make silently — `<div>hello</div>` losing the word "hello" is data loss a consumer cannot see happening.

Strip now parses each raw token with `parse5` — an inert, DOM-free parser that runs identically under SSR — and keeps the text nodes, escaping every one on the way out. Tags, attributes, declarations and comments go; text stays:

| Source | `'strip'` |
| --- | --- |
| `a <b>bold</b> word` | `a bold word` |
| `<div>hello <em>world</em></div>` | `hello world` |
| `<div title="a > b">visible</div>` | `visible` |
| `<div>before<!-- hidden -->after</div>` | `beforeafter` |
| `<div>fish &amp; chips &#60;3</div>` | `fish &amp; chips &lt;3` |
| `<script>alert(1)</script>` | `alert(1)` as text — never a script element |

Quoted `>` inside an attribute, entity references, unclosed tags and raw-text elements are exactly the cases a regex gets wrong, which is why this is a real parser rather than a substitution. Escaping the extracted text is what keeps `&lt;img src=x onerror=...&gt;` from re-entering as markup — there is a test asserting the output contains `&lt;img` and no `<img`.

**The stale descriptions are fixed** — all three lines named, plus the two that said raw HTML is "always" escaped:

- `src/lib/MarkdownText/properties.ts:7` and the `rawHtml` docstring
- `docs/MarkdownText.md` overview, the `allowedTags` note and the security-model bullet

Fenced code is untouched (a code block is not a raw-HTML token) and prose containing bare angle brackets still escapes to a visible `<`, both pinned by tests. 45 unit tests over the markdown pipeline, 10 of them new for these cases.

murdore:
## Recorded proof — 3 of the 5 Playwright cases

Recorded at `e4e71e0`, the commit on this branch. Each clip was decoded end-to-end with `ffmpeg -xerror` before upload, and a frame was extracted and reviewed, because an earlier round of these recordings was playable but framed on the wrong part of the page.

**Stripping keeps the text an inline tag surrounded** — `a <b>bold</b> word` renders as `a bold word`, no tag text left behind.

https://github.com/user-attachments/assets/99b132c2-26ad-4404-8f39-d13c2fa516e8

**Stripping keeps the text inside a block-level tag** — this is the contract I had wrong: the earlier revision dropped `<div>A whole block of raw HTML.</div>` entirely, sentence and all. The clip shows the text surviving without its tags.

https://github.com/user-attachments/assets/8777aa91-b75f-469b-ba54-4ef8601f8095

**Escaping is unchanged** — the default still shows `<b>bold</b>` and the whole `<div>…</div>` as visible text, which is what makes this additive.

https://github.com/user-attachments/assets/50abafe6-72b9-446b-af0a-082058466041

A Playwright trace for the first case, if you want to step it frame by frame:

[strip-keeps-inner-text-trace.zip](https://github.com/user-attachments/files/31986060/strip-keeps-inner-text-trace.zip)

### What the recordings do not prove on their own

They show three of five browser cases. The unit suite is what covers the parser edge cases — quoted `>` inside an attribute, entity references, comments, unclosed tags, `<script>`/`<style>` text, and fenced code — and CI attaches the full report as the `playwright-report` artifact.

The remaining 1 skipped functional test (`a11y-choicebox.spec.ts:78`) is skipped on the default branch too, not by this change.

Tara-ag (commented):

src/lib/MarkdownText/markdown.ts
```diff
@@ -133,14 +139,38 @@ function tagAllowed(allowed: Set<string> | null, tag: string): boolean {
...
+  return rawHtml === 'strip' ? 'strip' : 'escape';
+}
+
+// An inert, server-compatible HTML parser handles quoted delimiters, entities
+// and raw-text elements. Escaping its text nodes prevents decoded entities from
+// becoming markup when the resulting string is rendered by Svelte.
+function rawHtmlText(node: DefaultTreeAdapterTypes.Node): string {
+  if ('value' in node) {
+    return escapeHtml(node.value);
+  }
```
Tara-ag, line 158: 💡 MINOR: Strip mode drops `<template>` content instead of retaining its text. The `rawHtmlText` helper recurses through `childNodes`, but parse5 stores the content of a `<template>` element in a `content` DocumentFragment property, not in `childNodes`. As a result, any raw `<template>` HTML in the markdown source will be silently dropped rather than stripped to its text, contradicting the documented contract that "tags ... are removed; text is retained".

Suggestion: handle template elements explicitly before the `childNodes` branch:
```ts
if ('content' in node && node.content) {
  return rawHtmlText(node.content);
}
```
Add a unit test for `<template>visible</template>` expecting `visible`.

package.json
```diff
@@ -201,5 +201,8 @@
   },
   "engines": {
     "node": ">=22.18"
+  },
+  "dependencies": {
+    "parse5": "8.0.0"
   }
```
Tara-ag, line 207: 💡 MINOR: parse5 pinned to 8.0.0 while 8.0.1 is already in the lockfile. The pnpm-lock diff shows parse5@8.0.1 is already present transitively, and this PR adds parse5@8.0.0 as a direct runtime dependency. This forces consumers to install two patch versions of the same package, increasing node_modules size and the bundled surface for no benefit. Aligning to the newer patch avoids duplication and picks up any bug fixes in 8.0.1.

Suggestion: bump the dependency to `"parse5": "8.0.1"` and regenerate the lockfile so the direct dependency deduplicates with the existing transitive one.

Tara-ag (commented):
Review complete. Two minor findings: one dependency deduplication nit and one edge-case correctness issue with `<template>` stripping. Both are non-blocking; the change is safe to merge as-is, though addressing them before merge would be ideal.

Tara-ag (commented):
Yama review in progress.

src/lib/MarkdownText/markdown.ts
Tara-ag, line 158: 💡 MINOR: Strip mode drops <template> content instead of retaining its text

The `rawHtmlText` helper recurses through `childNodes`, but parse5 stores the content of a `<template>` element in a `content` DocumentFragment property, not in `childNodes`. As a result, any raw `<template>` HTML in the markdown source will be silently dropped rather than stripped to its text, contradicting the documented contract that "tags ... are removed; text is retained".

**Suggestion:** Handle template elements explicitly before the childNodes branch:
```ts
if ('content' in node && node.content) {
  return rawHtmlText(node.content);
}
```
Add a unit test for `<template>visible</template>` expecting `visible`.

package.json
Tara-ag, line 207: 💡 MINOR: parse5 pinned to 8.0.0 while 8.0.1 is already in the lockfile

The pnpm-lock diff shows parse5@8.0.1 is already present transitively, and this PR adds parse5@8.0.0 as a direct runtime dependency. This forces consumers to install two patch versions of the same package, increasing node_modules size and the bundled surface for no benefit. Aligning to the newer patch avoids duplication and picks up any bug fixes in 8.0.1.

**Suggestion:** Bump the dependency to `"parse5": "8.0.1"` and regenerate the lockfile so the direct dependency deduplicates with the existing transitive one.

Tara-ag (commented):
Yama review complete. Two minor non-blocking findings; the change is backward-compatible and safe to merge.

src/lib/MarkdownText/markdown.ts
Tara-ag, line 158: 💡 MINOR: `rawHtmlText` drops text inside `<template>` elements. parse5 stores template children in a separate `.content` DocumentFragment, not in `childNodes`, so `<template>hello</template>` would lose "hello" in strip mode. Add a template branch that recurses into `(node as DefaultTreeAdapterTypes.Template).content` before falling through to `childNodes`.

Tara-ag (commented):
Approved with one minor edge-case note. The additive `sanitize.rawHtml` feature is well-tested, defaults preserve existing behavior, and the public API/contracts are unchanged. Consider handling `<template>` content in `rawHtmlText` so strip mode does not silently drop text inside template elements.

murdore:
## `<template>` fix, with proof at `cbf504f`

Decoded end-to-end and frame-reviewed before upload. This clip shows the block-level case the contract turns on — the tags go, the sentence stays:

https://github.com/user-attachments/assets/10e34921-f79f-4179-978d-d847e80a53fe

The `<template>` case itself is a unit test rather than a browser one, because a template's contents never render — the whole point is that its text was being dropped from the *output string*, which only `renderMarkdown` can show:

```
'<template>hidden template text</template>'          → '<p>hidden template text</p>'
'<div>before<template>inside</template>after</div>'  → 'beforeinsideafter'
```

Full gates on this commit: lint 0, check 0, unit 0, build 0, integration 0.

src/lib/MarkdownText/markdown.ts
murdore: Taken — real, and the mechanism is exactly as described. parse5 puts a `<template>`'s children in a separate `content` DocumentFragment, so recursing through `childNodes` alone walked straight past them and the text was lost.

```ts
if ('content' in node) {
  return rawHtmlText(node.content);
}
```

Two tests, both failing before the fix:

```
'<template>hidden template text</template>'          → '<p>hidden template text</p>'
'<div>before<template>inside</template>after</div>'  → 'beforeinsideafter'
```

The first keeps its paragraph wrapper because marked treats a lone `<template>` as inline rather than block — worth pinning explicitly rather than trimming away, since the two token kinds are the whole contract here.

murdore: Taken — real, and the mechanism is exactly as described. parse5 puts a `<template>`'s children in a separate `content` DocumentFragment, so recursing through `childNodes` alone walked straight past them and the text was lost.

```ts
if ('content' in node) {
  return rawHtmlText(node.content);
}
```

Two tests, both failing before the fix:

```
'<template>hidden template text</template>'          → '<p>hidden template text</p>'
'<div>before<template>inside</template>after</div>'  → 'beforeinsideafter'
```

The first keeps its paragraph wrapper because marked treats a lone `<template>` as inline rather than block — worth pinning explicitly rather than trimming away, since the two token kinds are the whole contract here.

murdore: Taken — real, and the mechanism is exactly as described. parse5 puts a `<template>`'s children in a separate `content` DocumentFragment, so recursing through `childNodes` alone walked straight past them and the text was lost.

```ts
if ('content' in node) {
  return rawHtmlText(node.content);
}
```

Two tests, both failing before the fix:

```
'<template>hidden template text</template>'          → '<p>hidden template text</p>'
'<div>before<template>inside</template>after</div>'  → 'beforeinsideafter'
```

The first keeps its paragraph wrapper because marked treats a lone `<template>` as inline rather than block — worth pinning explicitly rather than trimming away, since the two token kinds are the whole contract here.

package.json
murdore: Taken, and the duplication was real — the lockfile carried both:

```
parse5@8.0.0:   ← this PR's direct dependency
parse5@8.0.1:   ← already present transitively
```

Now `"parse5": "^8.0.1"`, which deduplicates to the single copy already being installed. The caret also matches how the rest of this package declares dependencies, rather than pinning one exact patch of a parser.

murdore: Taken, and the duplication was real — the lockfile carried both:

```
parse5@8.0.0:   ← this PR's direct dependency
parse5@8.0.1:   ← already present transitively
```

Now `"parse5": "^8.0.1"`, which deduplicates to the single copy already being installed. The caret also matches how the rest of this package declares dependencies, rather than pinning one exact patch of a parser.

Tara-ag (approved):
Approved. The change is additive, backward-compatible, and thoroughly tested. Default escaping behavior is unchanged; the new `sanitize.rawHtml: 'strip'` mode safely extracts text via parse5 and re-escapes it before output. No public API breakage, SSR concerns, or security issues identified.

PR #590, opened 2026-09-08
feat(snippet): extract the copy affordance as createCopyState

### LINKED ISSUE #574: Snippet: the copy affordance cannot be used without the code-block presentation
### Need

`Snippet` couples two separable things: the **copy-and-flash affordance** (button, copied label, reset timer, clipboard handling, unmount-safe cleanup) and the **presentation** (a `<code>` box with a `$`-style prompt). There is no way to take the first without the second.

So a consumer who wants "copy this value" beside arbitrary content — a hostname, a token, a URL in a settings row — cannot use `Snippet` and ends up hand-rolling the affordance over `Button`, including the copied-label state and its reset timer. That is precisely the logic #530 made configurable inside `Snippet`, now duplicated outside it.

### Where this bit

TARA's `CopyButton.svelte` is exactly that hand-rolled duplicate: its own `copiedLabel`, its own 1.5s timer, its own effect cleanup. Its `/machines` call site places an independently styled, truncating `<code>` element *next to* the button rather than inside a prompt box, and the `/share` and `/settings` cases want a bare inline copy action beside other content. None of them wants a code block.

It stays hand-rolled — correctly, since `Snippet` would change what those surfaces look like — which means #530's `copiedLabel`/`copyResetMs` cannot reach the one component in TARA that most wants them.

### Proposal

Either:

1. **A presentation opt-out on `Snippet`** — something like `bare` / `showBox={false}` that renders only the copy control and lets the caller supply the display; or
2. **Extract the affordance** — export the copy-state logic (label swap, reset timer, clipboard, cleanup) as an action or a small headless helper that both `Snippet` and a consumer's own button can use.

Option 2 composes better and matches the precedent already set by `Tooltip/tooltip-action.ts`, which keeps DOM timing outside the component. Either is additive; `Snippet`'s current rendering stays the default.

murdore:
Extracts `Snippet`'s copy-and-flash behaviour as a headless factory, so a consumer can take the affordance without the `<code>` box.

## The gap

`Snippet` coupled two separable things: the copy state machine (clipboard call, `copied` flag, reset timer, unmount cleanup) and the presentation (a `<code>` box with a `$`-style prompt). A consumer who wants "copy this value" beside arbitrary content — a hostname in a settings row, a token, a URL — could not take the first without the second, so the logic got hand-rolled outside the library. That is exactly how #530's timer-safety work fails to reach the surfaces that need it most.

```svelte
<script>
  import { createCopyState, Button } from '@juspay/svelte-ui-components';

  const copy = createCopyState({ copyResetMs: 1500 });
  onDestroy(copy.destroy);
</script>

<code>{host}</code>
<Button onclick={() => void copy.copy(host)}>
  <span aria-live="polite">{copy.copied ? 'Copied!' : 'Copy'}</span>
</Button>
```

**Option 2 from the issue, not a `bare`/`showBox` prop.** Extraction composes with any presentation; a flag only removes one. `Snippet` now runs on the same state machine it exports, so the two cannot drift.

Options are read at copy time rather than captured at creation, so reactive callers pass getters — which is how `Snippet` forwards its own props.

## Seven defects, each reproduced before it was fixed

- **A clipboard promise settling after `destroy()`** set state on a torn-down owner, fired `oncopy`, and armed a timer that outlived the component.
- **`copy()` after `destroy()`** still wrote to the clipboard.
- **A throwing `oncopy`** rejected the returned promise and skipped the feedback reset, so the control stuck on "Copied!" forever.
- **Overlapping writes** needed a defined contract: acknowledged in completion order, each successful completion restarts the single timer, and a failed write neither starts nor extends feedback an earlier success is still showing.
- **A non-numeric `copyResetMs` reached `setTimeout`.** `?? DEFAULT` only substitutes for `null`/`undefined`, so `NaN` arrived as `NaN` — coerced to `0`, firing immediately, so the feedback vanished before anyone saw it. The type says `number`, but plain-JS and web-component callers are not bound by it.
- **A throwing `copyResetMs` getter left `copied` stuck true forever** — the throw escaped after the flag was set but before the timer was armed.
- **The live region was inside the button and conditionally rendered.** A live region only announces when it is already in the DOM and its *text* changes. A persistent, visually hidden `role="status"` sibling now carries the announcement.

The documented contract now matches the implementation, including the SSR path (creation is safe without browser globals; only the write needs a browser).

## The demo had a real accessibility bug

Caught in review: `ariaLabel="Copy host"` overrode the visible label, so a screen-reader user heard "Copy host" while the button read "Copied!" (WCAG 2.5.3, Label in Name), and the state change announced nothing.

The example now lets the visible text be the accessible name, and the docs name the caveat that a live region **inside** a button is not announced reliably by every screen-reader/browser pair, with the sibling `role="status"` fallback spelled out. `createCopyState` still injects no DOM — a state machine that quietly added a live region would reintroduce the coupling this extraction removes.

The `./wc` entry point question raised twice in review is now documented rather than only argued: that entry registers custom elements and exports **no** utilities (`build:wc` ends with `echo 'export {};' > dist-wc/index.d.ts`), so this is imported from the main entry, as `renderMarkdown` already is.

## Verification

| Gate | Exit |
| --- | --- |
| lint | 0 |
| check | 0 |
| test:unit | 0 |
| build | 0 |
| test:integration | 0 |

**25 copy-state cases** and **6 Playwright cases**. The browser tests assert the accessible name changes from "Copy" to "Copied!" rather than that a class toggled — an assertion the pre-fix demo could not satisfy.

## Two more, from the same review round

**Snippet's button kept `ariaLabel="Copy to clipboard"` in the copied state**, so the accessible name and the visible "Copied!" disagreed — WCAG 2.5.3, Label in Name. The label is now supplied only while the button is icon-only.

**The demo modelled the pattern the docs warn against**, with `aria-live` inside the button. It now uses the persistent sibling `role="status"` region the docs recommend and the component itself uses.

Closes #574


Tara-ag (commented):
Reviewed file-by-file. Two minor findings accepted; no blocking issues remain.

src/routes/components/snippet/+page.svelte
```diff
@@ -28,3 +38,27 @@
     copyResetMs={300}
   />
 </div>
+
+<h2>The affordance without the box</h2>
+<p class="demo-note">
+  <code>createCopyState</code> is the same state machine <code>Snippet</code> runs on — clipboard write,
```
Tara-ag, line 44: 💡 MINOR: Demo copy button relies solely on `ariaLabel` and swaps visible label without preserving accessible name

The headless demo button uses `ariaLabel="Copy host"` while its visible text swaps between "Copy" and "Copied!". The visible label text must be included in the accessible name; an `aria-label` prop should not override the visible label. Screen-reader users hear "Copy host" while sighted users see "Copy"/"Copied!", and the state change is not announced because `aria-live` lives inside `Snippet.svelte`, not this demo.

Remove `ariaLabel` and let the visible text be the accessible name. Wrap the dynamic label in a span with `aria-live="polite"`:

```svelte
<Button onclick={() => void hostCopy.copy(host)}>
  <span aria-live="polite">{hostCopy.copied ? 'Copied!' : 'Copy'}</span>
</Button>
```

Or keep `ariaLabel` only when the button has no visible text.

Tara-ag (commented):
Reviewed PR #590. The extraction of `createCopyState` is clean and preserves Snippet's existing behavior. No blocking issues found. One MINOR accessibility finding on the demo page and one MINOR API-parity note about the web-component bundle (the latter could not be posted inline because `src/wc/index.ts` is unchanged in this PR).

murdore:
Both findings addressed — one taken, one respectfully disputed with evidence.

## Accessibility finding — taken, and it was a real bug

You're right, and the reasoning is exactly right: `ariaLabel="Copy host"` overrode the visible label, so a screen-reader user would hear "Copy host" while the button read "Copied!" (WCAG 2.5.3, Label in Name), and the state change announced nothing because the `aria-live` region lives inside `Snippet.svelte` — which this demo deliberately does not render.

Applied your suggested shape:

```svelte
<Button onclick={() => void hostCopy.copy(host)}>
  <span aria-live="polite">{hostCopy.copied ? 'Copied!' : 'Copy'}</span>
</Button>
```

Two things worth adding beyond the fix:

**The live region belongs here, not in the helper.** The whole point of `createCopyState` is that the caller owns the presentation — and the announcement is part of the presentation. Putting `aria-live` inside the factory would hand back a state machine that quietly injects DOM, which is the coupling this PR exists to remove. The docs now show the live region in the usage example so the next consumer copies the accessible version.

**A test now pins it**, so it cannot regress:

```ts
await expect(button).toHaveAccessibleName('Copy');
await button.click();
await expect(button).toHaveAccessibleName('Copied!');
```

That assertion also forced the locators off `getByRole('button', { name: 'Copy host' })` — which is the finding restated: the old test could only find the button by a name no sighted user ever saw.

## Web-component parity note — the premise doesn't hold

> One MINOR API-parity note about the web-component bundle (the latter could not be posted inline because `src/wc/index.ts` is unchanged in this PR).

`src/wc/index.ts` is unchanged because there is nothing there to change. It contains only side-effecting registrations and **exports nothing at all**:

```ts
import './components/Avatar.wc.svelte';
import './components/Badge.wc.svelte';
…
```

The build makes that explicit — `build:wc` ends with:

```
echo 'export {};' > dist-wc/index.d.ts
```

So the `./wc` entry point has no value-export surface for `createCopyState` to be absent from. A custom-element consumer reaches it the same way they reach `renderMarkdown` today: from the main `.` entry, which does export it (`src/lib/index.ts`). Adding the first-ever export to the WC bundle would be a new public API decision, not a parity fix, and it isn't one this PR should make silently.

Happy to be corrected if you meant something narrower.

src/routes/components/snippet/+page.svelte
murdore: Fixed in ebbb278: the demo and docs now render a persistent polite live region without overriding the visible name. Tests assert both Copy and Copied! accessible names. Five targeted browser tests, lint and pinned-container visual comparison pass.

src/lib/Snippet/copyState.test.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  it('honours a custom reset delay', async () => {
+    stubClipboard(resolves);
+    const state = createCopyState({ copyResetMs: 500 });
+
+    await state.copy('x');
+    vi.advanceTimersByTime(500);
+
+    expect(state.copied).toBe(false);
+  });
```
Tara-ag, line 95: 💡 MINOR: Unit test does not assert getter-based option reactivity

The PR's core claim is that options are read at copy time via getters so reactive values are not frozen at mount. This test covers a static custom delay, but no test passes a getter that changes between two `copy()` calls and verifies the second call reads the new value. Without that, the getter pattern used in `Snippet.svelte` is not actually protected by tests.

Suggested addition:
```ts
it('reads copyResetMs from a getter on every copy', async () => {
  stubClipboard(resolves);
  let ms = 500;
  const state = createCopyState({ get copyResetMs() { return ms; } });

  await state.copy('first');
  ms = 1000;
  await state.copy('second');

  vi.advanceTimersByTime(500);
  expect(state.copied).toBe(true);
  vi.advanceTimersByTime(500);
  expect(state.copied).toBe(false);
});
```

docs/Snippet.md
```diff
@@ -24,15 +24,65 @@ A copyable command-line code snippet with a prompt prefix symbol and an inline c
...
+  import { onDestroy } from 'svelte';
+  import { createCopyState, Button } from '@juspay/svelte-ui-components';
+
+  const host = 'ssh://runner-04.internal.example.com:2222';
+  const copy = createCopyState({ copyResetMs: 1500 });
+  onDestroy(copy.destroy);
+</script>
+
+<code class="host">{host}</code>
+<Button onclick={() => void copy.copy(host)}>
```
Tara-ag, line 52: 💬 SUGGESTION: Docs example could mention aria-live placement caveat

The guidance to wrap the swapping label in `aria-live="polite"` is correct, but placing the live region inside a `<button>` is not announced reliably by all browser/screen-reader combinations. Consider adding a short note that if the "Copied!" announcement is missed, the live region should be moved outside the button as a sibling element.

Tara-ag:
💡 MINOR: New public export `createCopyState` is not available from the web-component entry point

`createCopyState` is exported from `src/lib/index.ts` and its types are exported there, but `src/wc/index.ts` only registers Svelte components as custom elements and does not re-export the factory. Consumers using `@juspay/svelte-ui-components/wc` will not be able to import this utility, creating an inconsistency in the public surface across the two package entry points.

If the `./wc` entry is meant to expose the same public utilities as the Svelte entry point, consider adding:
```ts
export { createCopyState } from '../lib/Snippet/copyState.svelte';
```
Otherwise, document the limitation in `docs/Snippet.md`.

`src/wc/index.ts` (not part of this PR's diff, so posted as a top-level comment).

Tara-ag (commented):
Review complete. No blocking issues found. Two MINOR findings and one SUGGESTION are noted inline; the wc-entry-point inconsistency is commented separately since that file is not in the diff.

Tara-ag (commented):
Review comments inline.

Tara-ag:
💡 MINOR: New public export createCopyState is not registered for web-component build

`createCopyState` is exported from `src/lib/index.ts` and its types are exported, but `src/wc/index.ts` only registers Svelte components as custom elements. The PR description positions `createCopyState` as a public utility for consumers of `@juspay/svelte-ui-components`. The web-component bundle entry (`src/wc`) does not re-export it, so consumers using the `./wc` entry cannot import the factory. This is an inconsistency in the public surface across the two package entry points.

If the web-component entry point is meant to expose the same public utilities as the Svelte entry point, add `export { createCopyState } from '../lib/Snippet/copyState.svelte';` to `src/wc/index.ts`. If not, document the limitation.

src/lib/Snippet/copyState.test.ts
murdore: Taken, and it was a real gap: the getter pattern in `Snippet.svelte` had no test that would fail if the options were captured at creation instead of read per copy.

Added `reads changed delay and callback getters for each successful write`, which changes **both** option getters between two copies and proves the second call read the new values — the delay through the timer boundary (still `copied` at 500ms, cleared at 1000ms) and the callback by asserting each of the two spies fired exactly once.

Three more lifecycle cases landed alongside it, each written against the current implementation first:

- **Overlapping writes** — two clipboard promises in flight, resolved out of order. Acknowledged in completion order, one live timer (`vi.getTimerCount()` is 1), `oncopy` fires twice.
- **A failed repeat write** — does not erase feedback an earlier successful copy is still showing, and does not extend it either.
- **SSR-shaped creation** — created and destroyed with no browser globals; `copy()` resolves `false` and arms no timer.

`copyState.test.ts` is 18 cases now, all green.

docs/Snippet.md
murdore: Correct, and worth being explicit about rather than leaving the example looking universally safe. A live region inside the button is not announced reliably across screen-reader/browser combinations, and the caller owns the presentation here.

`docs/Snippet.md` now says exactly that, and names the fallback shape: keep the button label as-is and put a separate, visually hidden `role="status"` sibling **outside** the button carrying the announcement text, testing it against the assistive technology the consumer actually supports.

`createCopyState` still injects no DOM of its own — a state machine that quietly added a live region would reintroduce the coupling this extraction exists to remove.

murdore:
Raised twice, so worth closing out properly — and I have now documented it rather than only arguing it.

`src/wc/index.ts` exports nothing at all. It is a list of side-effecting imports that call `customElements.define`, and the build makes that explicit — `build:wc` finishes with `echo 'export {};' > dist-wc/index.d.ts`. So the `./wc` entry has no value-export surface for `createCopyState` to be missing from; adding the first-ever export there would be a new public-API decision, not a parity fix, and not one to make silently inside this PR.

What was legitimately missing is that nothing said so. `docs/Snippet.md` now does:

> The factory is safe to create during SSR; clipboard writes require a browser. Import it from the package's main entry. The `./wc` entry registers custom elements and exports no utilities; it does not re-export this factory. This is the existing entry-point contract, not a missing custom-element registration.

A custom-element consumer reaches `createCopyState` the same way they reach `renderMarkdown` today: from the main `.` entry, which does export it.

murdore:
## Recorded proof — the affordance without the box

Recorded at `2f6bf62`, the commit on this branch. Each clip was decoded end-to-end with `ffmpeg -xerror` and a frame reviewed before upload.

**A real clipboard write from the headless factory**, against the browser's own clipboard rather than a stub.

https://github.com/user-attachments/assets/90f0eec1-4a2a-4780-aaf6-884e080217fc

**The copy control beside arbitrary content** — a truncating hostname with a plain `Button`, no `<code>` box and no `$` prompt, which is exactly what the issue said could not be done. The accessible name changes from `Copy` to `Copied!`, which is the assertion that would have failed against the demo's original `ariaLabel` override.

https://github.com/user-attachments/assets/3d716c59-4d5c-4481-97b9-e8d60597e33a

Two of five browser cases. The lifecycle contract — a clipboard promise settling after `destroy()`, calls after `destroy()`, a throwing `oncopy`, overlapping writes resolving out of order, and SSR-shaped creation — is covered by the 18 unit cases, none of which is observable in a recording.

Local gate log for this commit: lint 0, check 0, unit 0, build 0, integration 0 (934 unit, 633 functional, 1 pre-existing skip).

src/lib/Snippet/copyState.svelte.ts
```diff
@@ -0,0 +1,98 @@
...
+      copied = true;
+      resetTimer.arm(() => {
+        copied = false;
+      }, options.copyResetMs ?? DEFAULT_COPY_RESET_MS);
+      try {
+        options.oncopy?.();
+      } catch {
+        // A notification callback cannot undo the successful clipboard write.
+      }
+      return true;
```
Tara-ag, line 113: ⚠️ MAJOR: Throwing `copyResetMs` getter leaves `copied` stuck forever

The factory encourages callers to pass getters for reactive options, but the getter is evaluated only after `copied = true`. If `get copyResetMs()` throws, the successful clipboard write has already flipped the flag and the reset timer is never armed, so `copied` stays `true` permanently.

Guard the option read and fall back to the default delay so the copy-and-flash contract survives a misbehaving getter:

```ts
let copyResetMs = DEFAULT_COPY_RESET_MS;
try {
  copyResetMs = options.copyResetMs ?? DEFAULT_COPY_RESET_MS;
} catch {
  // Keep the successful copy state consistent with the public contract.
}
copied = true;
resetTimer.arm(() => {
  copied = false;
}, copyResetMs);
```

Tara-ag (changes requested):
One actionable issue found: a throwing `copyResetMs` getter can strand the `copied` flag. Please address it, then this is good to merge.

Tara-ag:

## 🤖 Yama Review Summary

⚠️ The review loop ended early (`time-limit`, 15 steps) — this verdict is built strictly from gate-verified findings.

**Verified findings (3):**

- **MINOR**: Docs example contradicts the accessibility guidance that follows it — `docs/Snippet.md:42`
- **MINOR**: Snippet status live region is not associated with the copy button — `src/lib/Snippet/Snippet.svelte:65`
- **SUGGESTION**: Consider exposing `createCopyState` from a stable subpath instead of only the barrel — `src/lib/index.ts:106`

- 💡 **MINOR**: Docs example contradicts the accessibility guidance that follows it — `docs/Snippet.md:42`
  The `createCopyState` example in docs/Snippet.md renders `<span aria-live="polite">` inside the Button, but the two accessibility notes immediately below explain that a live region must already be in the DOM before its text changes and recommend a persistent, visually-hidden `role="status"` sibling outside the button. A reader copying the example will implement the anti-pattern the docs then warn …
- 💡 **MINOR**: Snippet status live region is not associated with the copy button — `src/lib/Snippet/Snippet.svelte:65`
  The new `.snippet-status` live region is rendered as a sibling of `.snippet-copy` but has no `id` and the `<Button>` has no `aria-describedby` pointing to it, so screen-reader users may not perceive the status as related to the copy control. While the text itself is announced, the relationship is implicit at best.
- 💬 **SUGGESTION**: Consider exposing `createCopyState` from a stable subpath instead of only the barrel — `src/lib/index.ts:106`
  The new utility is exported only from the main barrel (`src/lib/index.ts`). Consumers using tree-shaking bundlers will still pull it from the barrel, but the PR description notes `./wc` intentionally does not export utilities. There is no mention of a dedicated subpath export (e.g. `@juspay/svelte-ui-components/snippet`), which would let consumers import the utility without importing the entire ba…

murdore:
All three MAJOR findings taken — each one was real, and each is now fixed with a test that fails against the previous revision.

## 1. A non-numeric `copyResetMs` reached `setTimeout`

Correct, and the consequence is worse than the finding states. `options.copyResetMs ?? DEFAULT` only substitutes for `null`/`undefined`, so `''` arrived as `''`, and `NaN` arrived as `NaN` — which `setTimeout` coerces to **0**, firing immediately. The feedback would vanish before anyone saw it, which reads as "copy is broken" rather than "the delay was invalid".

The type says `number`, but this factory is reachable from plain JS and from a web-component attribute, so the type is not the guard:

```ts
function resolveResetDelay(options: CopyStateOptions): number {
  let requested: unknown;
  try {
    requested = options.copyResetMs;
  } catch {
    return DEFAULT_COPY_RESET_MS;
  }
  if (typeof requested !== 'number' || !Number.isFinite(requested) || requested < 0) {
    return DEFAULT_COPY_RESET_MS;
  }
  return requested;
}
```

Covered by a table-driven test over `''`, `'   '`, `NaN`, `Infinity`, `-1` and `null`, each asserting the default 2000ms boundary is used.

## 2. A throwing `copyResetMs` getter left `copied` stuck

This one is the sharpest of the three, because the failure is permanent. The options are read at copy time by design, so a reactive getter can throw — and the throw escaped **after** `copied = true` but **before** the timer was armed. The control would read "Copied!" forever, with nothing scheduled to clear it.

The delay is now resolved *before* the flag is set, so there is no window where one is true without the other. The test installs a getter that throws and asserts the flag still clears at 2000ms.

## 3. The live region was inside the button and conditionally rendered

Also correct, and it goes further than the earlier docs note I wrote on this PR — documenting the caveat while `Snippet` itself still had the bug was not good enough.

A live region announces when it is **already in the DOM** and its text changes. The old markup inserted the element and its text together, inside the button, in a branch that swaps with the icon. Now:

```svelte
<span class="snippet-status" role="status" aria-live="polite">{copyState.copied ? copiedLabel : ''}</span>
```

Always rendered, outside the button, visually hidden, with only its text changing. The visible label still swaps inside the button, so nothing changes on screen and there is no duplicated text.

Pinned by a browser test asserting the region exists with empty text before the click, is **not** inside a `button`, and holds `Copied!` after.

## On the two MINOR/SUGGESTION items

- **Docs getter example omits `oncopy`** — the example is deliberately minimal, but the prose promised more than it showed, so the `copyResetMs` row now states the fallback behaviour explicitly and the accessibility note describes the persistent-region shape rather than just warning about the naive one.
- **Integration test uses a fixed timeout for the reset delay** — kept. The demo sets `copyResetMs=300` and the assertion allows 2000ms, so it proves the option is read rather than racing it; there is no event to await here, and the unit suite covers the timing precisely with fake timers.

`copyState.test.ts` is 25 cases now. Gates re-run on the new commit.

Tara-ag (commented):

src/routes/components/snippet/+page.svelte
```diff
@@ -28,3 +38,33 @@
...
+<div class="demo-row" data-pw="snippet-headless" style="align-items: center; gap: 8px;">
+  <code class="host-value">{host}</code>
+  <!-- No ariaLabel: it would override the visible text, so a screen-reader user
+       would hear "Copy host" while the button reads "Copied!" (WCAG 2.5.3,
+       Label in Name). The live region is here rather than in the component
+       because this demo deliberately renders no Snippet -- the whole point is
+       that the caller owns the presentation, and the announcement is part of
+       the presentation. -->
+  <Button onclick={() => void hostCopy.copy(host)}>
+    <span aria-live="polite">{hostCopy.copied ? 'Copied!' : 'Copy'}</span>
```
Tara-ag: 💡 MINOR: The headless demo places `aria-live="polite"` on the label span *inside* the Button, but the docs' `createCopyState` example recommends a persistent `role="status"` sibling *outside* the button because live regions inside buttons are not reliably announced. The demo models the less reliable pattern and contradicts the documented guidance.

Suggested fix: move the live region outside the Button as a persistent visually-hidden `role="status"` sibling, matching the docs' recommended pattern and the Snippet component's own markup.

src/lib/Snippet/Snippet.svelte
```diff
@@ -54,8 +50,8 @@
   {#if showCopyButton}
     <div class="snippet-copy">
       <Button onclick={handleCopy} ariaLabel="Copy to clipboard">
```
Tara-ag: ⚠️ MAJOR: The Button always receives `ariaLabel="Copy to clipboard"`. When `copyState.copied` is true, the visible label swaps to `"Copied!"` but the accessible name remains `"Copy to clipboard"`, violating WCAG 2.5.3 (Label in Name). Screen-reader users will not hear the visible feedback state.

Suggested fix: make `ariaLabel` conditional so the visible text becomes the accessible name during the copied state:
```svelte
<Button onclick={handleCopy} ariaLabel={copyState.copied ? undefined : 'Copy to clipboard'}>
```

Tara-ag (changes requested):
Review complete. One MAJOR accessibility issue in Snippet.svelte and one MINOR demo/docs inconsistency in the headless example. Both are actionable and should be addressed before merge.

murdore:
## Both accessibility findings fixed, with proof at `a9f5b4a`

Each clip was decoded end-to-end with `ffmpeg -xerror` and its frame reviewed before upload.

**The accessible name now follows the visible label.** The button is icon-only at rest, so it needs `ariaLabel`; keeping that label once the text becomes "Copied!" is what broke WCAG 2.5.3. The label is now supplied only in the icon state, asserted as `toHaveAccessibleName('Copy to clipboard')` → `toHaveAccessibleName('Copied!')`.

https://github.com/user-attachments/assets/47370654-9076-49d9-9dc4-79130302fa1f

**The demo no longer models the pattern the docs warn against.** Its live region is now a persistent `role="status"` sibling outside the button — the test asserts `el.closest('button') === null` and that no `[aria-live]` remains inside the button.

https://github.com/user-attachments/assets/18f8ff8b-7daf-4bba-8f63-b8abd784e6da

The visually hidden region is deliberately invisible in the recording; the assertions are what verify it, which is exactly why a video alone is not the proof here.

`copyState.test.ts` is 25 cases and the browser spec is 6. Full gates on this commit: lint 0, check 0, unit 0, build 0, integration 0.

---

> **Superseded.** The branch has moved on to `dd0818c` since this was posted. The clips above were accurate at `a9f5b4a` and the only change since is `docs/Snippet.md`, but the current recording is https://github.com/juspay/svelte-ui-components/pull/590#issuecomment-5601526154.

src/lib/Snippet/copyState.svelte.ts
```diff
@@ -0,0 +1,98 @@
...
+      }
+
+      // The clipboard promise can settle after its owner has unmounted.
+      if (destroyed) {
+        return false;
+      }
+      copied = true;
+      resetTimer.arm(() => {
+        copied = false;
+      }, options.copyResetMs ?? DEFAULT_COPY_RESET_MS);
```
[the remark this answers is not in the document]
murdore: Already fixed in the commit this was raised against — the finding and the fix agree, so this is a stale read rather than a disagreement.

The delay is resolved **before** the flag is set, so there is no window where `copied` is true with nothing armed:

```ts
const delay = resolveResetDelay(options);   // try/catch inside
copied = true;
resetTimer.arm(() => { copied = false; }, delay);
```

`resolveResetDelay` also rejects non-numeric values, which is the related hole: `?? DEFAULT` only substitutes for null/undefined, so `NaN` reached `setTimeout` and was coerced to 0 — the feedback vanished before anyone saw it. Covered by a table over `''`, `'   '`, `NaN`, `Infinity`, `-1`, `null`, plus a throwing-getter case. Six of those fail against the previous revision.

src/lib/Snippet/copyState.svelte.ts
murdore: Already fixed in the commit this was raised against — the finding and the fix agree, so this is a stale read rather than a disagreement.

The delay is resolved **before** the flag is set, so there is no window where `copied` is true with nothing armed:

```ts
const delay = resolveResetDelay(options);   // try/catch inside
copied = true;
resetTimer.arm(() => { copied = false; }, delay);
```

`resolveResetDelay` also rejects non-numeric values, which is the related hole: `?? DEFAULT` only substitutes for null/undefined, so `NaN` reached `setTimeout` and was coerced to 0 — the feedback vanished before anyone saw it. Covered by a table over `''`, `'   '`, `NaN`, `Infinity`, `-1`, `null`, plus a throwing-getter case. Six of those fail against the previous revision.

src/lib/Snippet/Snippet.svelte
murdore: Taken, and this one was a genuine WCAG 2.5.3 break that I had walked past twice.

The button is icon-only at rest, so it does need a label. Keeping that label once the visible text becomes "Copied!" is what created the mismatch. It is now supplied only in the icon state, using this repo's conditional-spread idiom (the lint rules rule out `undefined` here):

```svelte
<Button onclick={handleCopy} {...copyState.copied ? {} : { ariaLabel: 'Copy to clipboard' }}>
```

Asserted through the accessible name rather than the attribute:

```ts
await expect(button).toHaveAccessibleName('Copy to clipboard');
await button.click();
await expect(button).toHaveAccessibleName('Copied!');
```

src/routes/components/snippet/+page.svelte
murdore: Taken — and you are right that it was the demo contradicting its own documentation, which is worse than either being wrong alone.

The demo now uses the persistent sibling the docs recommend and `Snippet` itself uses: always rendered, outside the button, visually hidden, with only its text changing.

```svelte
<Button onclick={() => void hostCopy.copy(host)}>{hostCopy.copied ? 'Copied!' : 'Copy'}</Button>
<span class="visually-hidden" role="status" aria-live="polite">{hostCopy.copied ? 'Copied to clipboard' : ''}</span>
```

The test asserts the region is not inside a `button` and that no `[aria-live]` remains within one, so the demo cannot drift back to the pattern the docs warn about.

docs/Snippet.md
```diff
@@ -24,15 +24,67 @@ A copyable command-line code snippet with a prompt prefix symbol and an inline c
...
+| copiedLabel    | `string`  | No       | `Copied!` | Text shown in place of the copy icon after a successful copy.                                                                                                          |
+| copyResetMs    | `number`  | No       | `2000`    | Milliseconds before the copied feedback reverts to the copy icon.                                                                                                      |
+| classes        | `string`  | No       | `-`       | CSS class string applied to the component's top-level element. Useful for theming — define classes with CSS variable overrides and pass them to create variant styles. |
+
+## The copy affordance on its own — `createCopyState`
+
+`Snippet` couples two separable things: the copy-and-flash behaviour, and the `<code>` box with its `$`-style prompt. When you want the first without the second — a copy control beside a hostname in a settings row, a token, a URL — import the state machine directly instead of hand-rolling a lookalike:
+
+```svelte
+<script>
```
Tara-ag, line 42: 💡 MINOR: Docs example contradicts the accessibility guidance that follows it

The `createCopyState` example here renders `<span aria-live="polite">` inside the `Button`, but the two accessibility notes immediately below explain that a live region must already be in the DOM before its text changes and recommend a persistent, visually-hidden `role="status"` sibling **outside** the button. A reader copying the example will implement the anti-pattern the docs then warn against.

Suggested fix: update the example to match the recommended pattern:

```svelte
<Button onclick={() => void copy.copy(host)}>
  {copy.copied ? 'Copied!' : 'Copy'}
</Button>
<span class="visually-hidden" role="status" aria-live="polite">
  {copy.copied ? 'Copied to clipboard' : ''}
</span>

<style>
  .visually-hidden {
    position: absolute;
    width: 1px;
    height: 1px;
    margin: -1px;
    padding: 0;
    overflow: hidden;
    clip-path: inset(50%);
    white-space: nowrap;
    border: 0;
  }
</style>
```

src/lib/Snippet/Snippet.svelte
```diff
@@ -64,6 +64,13 @@
         {/if}
       </Button>
```
Tara-ag, line 65: 💡 MINOR: Snippet status live region is not associated with the copy button

The new `.snippet-status` live region is rendered as a sibling of `.snippet-copy` but has no `id` and the `<Button>` has no `aria-describedby` pointing to it, so screen-reader users may not perceive the status as related to the copy control. While the text itself is announced, the relationship is implicit at best.

Suggested fix: add a deterministic id to the status span and reference it from the Button via `aria-describedby`, e.g.:

```svelte
<Button
  onclick={handleCopy}
  {...copyState.copied ? {} : { ariaLabel: 'Copy to clipboard' }}
  ariaDescribedBy="{testId ? `${testId}-status` : 'snippet-copy-status'}"
>
  ...
</Button>
<span
  id="{testId ? `${testId}-status` : 'snippet-copy-status'}"
  class="snippet-status"
  role="status"
  aria-live="polite"
>{copyState.copied ? copiedLabel : ''}</span>
```

(If `Button` does not yet expose `ariaDescribedBy`, wire `aria-describedby` directly on the underlying element.)

src/lib/index.ts
```diff
@@ -103,6 +103,7 @@ export { default as Gallery } from './Gallery/Gallery.svelte';
 export { ChatController } from './Chat/controller.svelte';
 export { partyOf } from './Chat/roles';
 export { SpeechToTextController } from './SpeechToText/controller.svelte';
+export { createCopyState } from './Snippet/copyState.svelte';
```
Tara-ag, line 106: 💬 SUGGESTION: Consider exposing `createCopyState` from a stable subpath instead of only the barrel

The new utility is exported only from the main barrel. Consumers using tree-shaking bundlers will still pull it from the barrel, but the PR description notes `./wc` intentionally does not export utilities. There is no mention of a dedicated subpath export (e.g. `@juspay/svelte-ui-components/snippet`), which would let consumers import the utility without importing the entire barrel and would mirror how some larger libraries expose headless helpers.

Not blocking — consider adding a `package.json` export such as:

```json
"./snippet": {
  "types": "./dist/Snippet/copyState.svelte.d.ts",
  "svelte": "./dist/Snippet/copyState.svelte.js"
}
```

Tara-ag (commented):
Approving with minor inline comments. The extraction is clean, the state-machine behavior is well-tested, and the accessibility fixes (persistent live region, Label-in-Name compliance) are solid improvements. No blocking issues.

docs/Snippet.md
murdore: Correct — fixed in `dd0818c`.

The example really did demonstrate the anti-pattern that the bullet directly beneath it warns against: `<span aria-live="polite">` nested inside the `Button`, where it is created and populated in the same tick, which is exactly the case many screen-reader/browser pairs never announce. Anyone copying it got the broken version and read the correct advice immediately afterwards.

The example is now the pattern the prose describes — a persistent visually-hidden `role="status"` sibling outside the button, holding empty text until a copy succeeds — so the code and the guidance say the same thing.

Also retitled the section from "Two accessibility notes on that example" to "Why the example is shaped that way", since the bullets are now explaining the code rather than correcting it.

src/lib/Snippet/Snippet.svelte
murdore: Declining this one, for four reasons — the third is a bug in the suggested snippet itself.

**1. `aria-describedby` on a live region announces twice.** `role="status"` already announces autonomously when its text changes. A description is read *on focus*. After a copy, focus is still on the button, so the user hears "Copied!" from the live region; then every subsequent return of focus to that button re-reads it as a description, for as long as the flash lasts. The relationship would be explicit at the cost of the announcement being duplicated and then repeated out of context.

**2. It is empty exactly when a description would help.** The span holds `''` until a copy succeeds. `aria-describedby` pointing at an empty node contributes no accessible description at all, so in the default state the attribute does nothing. It only resolves to text inside the ~1.5s window where the live region has already spoken.

**3. The proposed id scheme collides.** `testId ? \`${testId}-status\` : 'snippet-copy-status'` gives every `Snippet` rendered without a `testId` the same literal id. Two snippets on a page — a docs page listing install commands, say — produce duplicate ids, and `aria-describedby` resolves to the first match, so the second button would describe the first button's status. `testId` is an optional Playwright hook, not an identity. If an id were needed here the repo already has `$props.id()` for it, which is what `PieChart` uses for its `aria-controls`.

**4. `Button` has no `ariaDescribedBy` prop.** Its ARIA surface is `ariaLabel` / `ariaExpanded` / `ariaControls` / `ariaHaspopup` / `ariaSelected` / `ariaBusy`. Wiring this would mean adding a public prop to a component used ~40 places, to carry an attribute that costs more than it pays.

The underlying goal — that a screen-reader user understands the status belongs to the copy control — is already met by proximity and timing: the region announces immediately on activation of that button, which is the association mechanism a status message is designed to use. This is the pattern the ARIA Authoring Practices' status/clipboard examples use, and the pattern `Snippet` already implements.

Happy to reopen if you can point at a screen-reader/browser pair where the standalone `role="status"` is missed here.

src/lib/index.ts
murdore: Not doing this here, but the finding points at something real — just not the thing the fix addresses.

**Why not in this PR.** The package has exactly two export conditions today, `.` and `./wc`. Every one of the ~50 components ships through the barrel; none has a subpath. Adding `./snippet` for one helper would invent a public API convention that nothing else in the library follows, and export maps are permanent — once `@juspay/svelte-ui-components/snippet` exists, removing it is a breaking change. That is a library-wide packaging decision, and it should be made deliberately for all of the surface at once, not introduced as a side effect of extracting one copy helper.

**The real finding underneath it.** If the concern is that consumers cannot avoid pulling the barrel, a subpath per helper is the expensive way to fix it. `package.json` has no `sideEffects` field, so bundlers must conservatively assume every module in the barrel has side effects and keep them. One field would make the whole barrel tree-shakeable, for every component, with no new public surface:

```json
"sideEffects": false
```

I am not slipping that into this PR either, and deliberately: Svelte components can carry genuine side effects through imported CSS, so the flag needs a real check across the library — a built bundle diffed against the current one, plus confirmation that component styles survive in a consuming app — before it is safe to claim. That is its own PR with its own evidence.

Filing it as a follow-up rather than doing it half-way here. If you would rather have the subpath as well after that, it is a much easier conversation once the barrel is provably shakeable, because at that point it is ergonomics rather than a workaround.

murdore:
## Re-recorded proof at `dd0818c`

The previous recording was taken at `a9f5b4a`. The only change since is `docs/Snippet.md`, so the runtime behaviour in that clip was still accurate — but the comment claimed a commit that is no longer the head, and a proof label that needs a caveat to be true is not a proof label. Re-recorded rather than annotated.

**`Snippet` announces through a persistent live region.** The region is in the DOM before the copy, holding empty text, and only its text changes — which is what makes the announcement land on screen-reader/browser pairs that miss a region inserted together with its content.

https://github.com/user-attachments/assets/05dc62cb-6e61-405b-a80c-438d1d7407b0

**The headless demo announces through a sibling region, not one inside the button.** This is the case behind the docs fix in this revision: the documented example previously nested `aria-live` inside the `Button`, which is the anti-pattern the guidance immediately below it warned against. The example now matches the guidance, and this test pins the shape.

https://github.com/user-attachments/assets/21efe94c-af32-47d9-8162-3d9f313e39dd

All 6 cases in `snippet-copy-state.spec.ts` pass at this commit. Unit **944 passed**, lint 0, check 0, build 0. Every clip decoded end-to-end with `ffmpeg -xerror`, frame reviewed before upload.

murdore:
## Runtime proof, re-recorded at the merged commit `756408b`

Re-recorded at the commit that actually landed, so every PR in this batch carries proof from its merged state rather than its branch head. 6/6 passing.

### `createCopyState` — the copy affordance without the presentation (#574)

| Behaviour | Recording |
| --- | --- |
| Copies the value to the **real** clipboard | [watch](https://github.com/user-attachments/assets/c40053d3-90cf-4ae6-bf6c-21994bcc5471) |
| Swaps the label on copy and reverts after the reset delay | [watch](https://github.com/user-attachments/assets/19449bde-5d00-4d3e-95b7-42de1f873c32) |
| Renders no Snippet presentation around it | [watch](https://github.com/user-attachments/assets/128c1fe1-d555-4485-b4d8-0faa5be9014c) |

### Announcement, which is the part that is easy to get wrong

| Behaviour | Recording |
| --- | --- |
| Snippet announces the copy through a **persistent** live region | [watch](https://github.com/user-attachments/assets/9dacf351-6fec-46cb-991f-835892d6a588) |
| The accessible name follows the visible label into the copied state | [watch](https://github.com/user-attachments/assets/a510e19a-2303-45b5-a69e-21d5e7b4c0b0) |
| The headless demo announces through a **sibling** region, not one inside the button | [watch](https://github.com/user-attachments/assets/d6d31a08-0cd0-4bb0-b2c4-c0c13469e5a2) |

Those three exist because the obvious implementations are each subtly wrong: a live region created at copy time is often not announced at all, since screen readers need it present beforehand; and putting the region inside the button folds its text into the button's own accessible name, so the control renames itself mid-interaction. Both are asserted, not assumed.

### On the review suggestions I declined

Two suggestions on this PR were declined with evidence rather than implemented — adding `aria-describedby` to the live region, and adding a subpath export for `createCopyState`. Both replies are in-thread. For the second I noted that `sideEffects: false` is the better fix for the underlying tree-shaking concern, which remains open as a follow-up rather than being quietly dropped.

### Verification

Release run #386 on `756408b`: `verify` success, `release` success, published as **4.13.0**.

PR #591, opened 2026-09-08
feat(sheet): let top/bottom panels be width-capped and centred

### LINKED ISSUE #572: Sheet: top/bottom sides cannot be width-capped or centred, unlike left/right
### Need

`Sheet`'s `top` and `bottom` sides have no width-capping mechanism. `left` and `right` expose `--sheet-width`, but top/bottom are always full-viewport-width with only a border on the entering edge.

That makes Sheet unusable for a very common shape: a **centred, width-capped card** that enters from the bottom — the layout most mobile-first "launch" or "action" sheets actually use above phone widths.

### Where this bit

TARA's `LaunchSheet.svelte` hand-rolls its own overlay and panel and cannot adopt `Sheet`. Its panel is `width: min(560px, 100vw)`, horizontally centred, fully bordered, at every width below its 900px breakpoint — not only on narrow phones. Reproducing that on `side="bottom"` today would mean faking the cap with hand-computed `calc()`-based `--sheet-left`/`--sheet-right` overrides plus a global border override for the edges Sheet deliberately omits — which relocates the consumer's centring maths into a global stylesheet rather than adopting the component.

Worth noting what is *not* the blocker: `side` is a perfectly usable reactive prop, so a media-query-driven drawer/dialog switch works fine. The gap is purely the width and border treatment of the top/bottom sides.

### Proposal

Give top/bottom the same treatment left/right already have:

- `--sheet-width` (or `--sheet-max-width`) honoured on `top`/`bottom`, centring the panel horizontally when it is narrower than the viewport.
- A way to get the full border on a capped panel, since a floating card needs all four edges rather than only the entering one.

Both are additive: a consumer setting neither keeps today's full-bleed rendering exactly.

murdore:
Gives `top`/`bottom` sheets the width cap and full border that `left`/`right` already have.

## The gap

`left` and `right` expose `--sheet-width`; `top` and `bottom` were always full-viewport-width with a border only on the entering edge. That made the shape most mobile-first action sheets actually use above phone widths — a centred, width-capped, fully bordered card entering from the bottom — impossible without faking it through hand-computed `--sheet-left`/`--sheet-right` `calc()` offsets plus a global border override, which relocates the consumer's centring maths into a stylesheet rather than adopting the component.

```css
.launch-sheet {
  --sheet-band-width: min(560px, 100vw);
  --sheet-band-box-sizing: border-box;
  --sheet-band-border: 1px solid #d4d4d8;
}
```

| Token | Default | What it does |
| --- | --- | --- |
| `--sheet-band-width` | `auto` | Caps the panel and centres it between the insets |
| `--sheet-band-max-width` | `none` | Caps responsively without fixing the width |
| `--sheet-band-border` | `none` | Draws **all four** edges, which a floating card needs |
| `--sheet-band-box-sizing` | `content-box` | Opt into `border-box` so borders fit inside the cap |

**Deliberately not `--sheet-width`/`--sheet-max-width`**, even though the issue proposed those names. Those already mean "the width of a left/right panel", so a consumer who themes them for their side sheets would find their bottom sheets silently reshaped by an upgrade. `side="center"` avoids the same collision the same way with its own `--sheet-center-*` tokens. A regression test sets `--sheet-width` on a bottom sheet and asserts it stays edge-to-edge.

## Two defects the acceptance tests found

**The documented recipe overflowed a phone.** A 560px cap plus 1px borders measures 562px, and the panel is content-box like `side="center"` is — so at 320px the card ran past the viewport. Changing that rule outright would shift every existing top/bottom sheet that sets `--sheet-border` by a pixel, so box-sizing became an opt-in token and the documented recipe uses `min(560px, 100vw)` with `border-box`. Verified at 320px, where the panel now fits exactly.

**Escape did not return focus to the trigger.** The scroll-lock action moved focus into the panel on open but restored nothing on close, so dismissing from the keyboard dropped focus to `<body>`. It now restores the previously focused element — and only when focus is still inside the sheet or on `<body>`, so a consumer who deliberately moves focus elsewhere while closing is not overridden.

## Verification

| Gate | Exit |
| --- | --- |
| lint | 0 |
| check | 0 |
| test:unit | 0 |
| build | 0 |
| test:integration | 0 |

**11 Playwright cases**: the cap, real centring (equal **and** non-zero side gaps — equal alone passes on a full-bleed panel, where both are zero), bottom-edge anchoring, all four borders, the 320px fit, asymmetric top insets with a responsive max-width, focus trap, body scroll lock, Escape focus restoration, outside-click dismissal, and two back-compat guards that an unconfigured sheet is unchanged.

Transitions are awaited through the Web Animations API (`element.getAnimations()`) rather than a fixed sleep, so the tests do not race the fly transition.

Closes #572


Tara-ag (commented):
Review in progress.

tests/sheet-band-width.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  // The panel slides in over 300ms; measuring mid-transition reads a
+  // transform that is still moving.
+  await page.waitForTimeout(400);
+  return panel;
+};
+
+const boxOf = async (locator: import('@playwright/test').Locator) => {
+  const box = await locator.boundingBox();
+  if (box === null) {
+    throw new Error('panel boundingBox is null');
```
Tara-ag, line 36: 💡 MINOR: Fixed timeout waits for transition instead of deterministic event

`page.waitForTimeout(400)` is brittle under CI load (may finish before the transition on slow runners, or waste time when it finishes early) and contradicts the repo's preferred pattern of waiting for `transitionend` or `onafteropen`.

Replace with a deterministic wait for the panel's `transitionend` event, e.g.:

```ts
const settle = async (locator: Locator) => {
  await locator.evaluate(
    (el) => new Promise<void>((resolve) => {
      const handler = () => {
        el.removeEventListener('transitionend', handler);
        resolve();
      };
      el.addEventListener('transitionend', handler);
    })
  );
};
```

Then call `await settle(panel);` after `await expect(panel).toBeVisible();`.

tests/sheet-band-width.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  const panel = page.getByTestId(panelId);
+  await page.getByTestId(triggerId).click();
+  await expect(panel).toBeVisible();
+  // The panel slides in over 300ms; measuring mid-transition reads a
+  // transform that is still moving.
+  await page.waitForTimeout(400);
+  return panel;
+};
+
+const boxOf = async (locator: import('@playwright/test').Locator) => {
```
Tara-ag, line 33: 💡 MINOR: Fixed timeout in transition settle helper. `page.waitForTimeout(400)` waits for the 300ms fly transition, but the project's test standards prefer deterministic waits (e.g. `transitionend`, `onafteropen`, or web-first assertions) over fixed sleeps. Existing sheet tests already carry this same pattern, so this is a consistency/maintenance nit rather than a blocker, but consider replacing it with a wait for the panel's `transitionend` event or a settled-state marker.

Tara-ag (approved):
Approved with one minor testing nit. The change is a clean, additive feature: new `--sheet-band-width` and `--sheet-band-border` tokens for top/bottom sheets, with solid back-compat guards and documentation. No public props, events, exports, or CSS variable names were changed.

Tara-ag (approved):
Approved with one minor testing nit already noted inline. The change is purely additive: new `--sheet-band-width`, `--sheet-band-max-width`, and `--sheet-band-border` tokens for top/bottom sheets, with no changes to existing props, events, exports, or CSS variable names. Docs and regression/back-compat tests are included.

tests/sheet-band-width.test.ts
murdore: Replaced the fixed delay with element.getAnimations() and Animation.finished, matching Svelte fly transitions. Unlike attaching a transitionend listener after visibility, it also handles an animation that has already finished. All seven targeted tests passed. Pushed in c259496 with the reviewed visual baseline.

tests/sheet-band-width.test.ts
murdore: Addressed in c259496: the helper now waits for the panel Web Animations finished promises instead of sleeping. Seven targeted tests pass, and the updated screenshot passes pinned-container comparison.

Tara-ag (approved):
Approved. The change is purely additive: it introduces three new CSS custom properties (`--sheet-band-width`, `--sheet-band-max-width`, `--sheet-band-border`) for `top`/`bottom` sheets while preserving the existing `--sheet-width`/`--sheet-border` semantics for side sheets. Default values keep prior rendering identical, and the new tokens follow the same naming precedent already established by `side="center"` (`--sheet-center-*`).

Verified:
- No props, exports, or event signatures changed.
- No browser-global access added; change is CSS-only.
- Docs and Playwright tests cover the new behavior and include explicit back-compat guards.
- CSS cascade (`border` shorthand followed by per-side override) correctly gives capped panels four edges while falling back to `--sheet-border` on the entering edge when the band token is unset.

murdore:
## Recorded proof — the two defects this branch fixed

Recorded at `9589251`, the commit on this branch. Every clip was decoded end-to-end with `ffmpeg -xerror` before upload, and a frame was extracted and reviewed, so these are checked recordings rather than files that merely exist.

**All four borders on a capped panel.** The centred `Launch` card with `--sheet-band-width` + `--sheet-band-border`, which is the shape the issue asked for and the one `top`/`bottom` could not produce.

https://github.com/user-attachments/assets/e08fad21-100c-4406-9977-9e9e80e7943c

**The 320px fit — the defect the acceptance pass found.** The documented 560px recipe plus 1px borders measured 562px on a phone, because the panel is content-box. With `--sheet-band-box-sizing: border-box` and `min(560px, 100vw)` it fits exactly, asserted as `box.x >= 0` and `box.x + box.width <= 320`.

https://github.com/user-attachments/assets/aa9688d1-f6e1-47ef-8b1f-a71c8bcf01ca

Two of eleven cases. The rest — real centring with non-zero side gaps, bottom-edge anchoring, asymmetric insets with a responsive max-width, the focus trap, body scroll lock, Escape focus restoration, outside-click dismissal and the back-compat guards — are in the same spec, and CI attaches the full report as the `playwright-report` artifact.

Local gate log for this commit: lint 0, check 0, unit 0, build 0, integration 0 (920 unit, 637 functional, 1 pre-existing skip).

Tara-ag (commented):
Review in progress.

tests/sheet-band-width.test.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  test('a bordered cap fits a phone viewport without horizontal overflow', async ({ page }) => {
+    await page.setViewportSize({ width: 320, height: 700 });
+    const panel = await openAndSettle(
+      page,
+      'sheet-capped-bottom-trigger',
+      'sheet-capped-bottom-panel'
+    );
+    const box = await boxOf(panel);
+    expect(box.x).toBeGreaterThanOrEqual(0);
```
Tara-ag, line 150: 💡 MINOR: This test couples to the demo page's button label text. If the demo copy is reworded, this assertion breaks for a reason unrelated to the width-capping feature. The rest of the file uses `testId` selectors; add `testId="sheet-top-trigger"` to the 'Open top' trigger in `src/routes/components/sheet/+page.svelte` and select it here with `page.getByTestId('sheet-top-trigger')`.

Tara-ag (approved):
Approved with one minor testing nit. The change is purely additive: new `--sheet-band-width`, `--sheet-band-max-width`, and `--sheet-band-border` tokens for `top`/`bottom` sheets, with no changes to existing props, events, exports, or CSS variable names. Docs and regression/back-compat tests are included.

tests/sheet-band-width.test.ts
Tara-ag, line 150: 💡 MINOR: Test couples to demo-page button text. Selecting the trigger via `getByRole('button', { name: 'Open top', exact: true })` ties this integration test to the demo page's copy; rewording the demo heading or button text would break this test for a reason unrelated to the width-capping feature. The rest of this file already uses `testId` selectors.

Suggested fix: add `testId="sheet-top-trigger"` to the 'Open top' `Button` in `src/routes/components/sheet/+page.svelte` and select it here with `page.getByTestId('sheet-top-trigger')`, matching the pattern used for the other sheet variants in this file.

Tara-ag (approved):
Approved with one minor testing nit. The change is additive and backward-compatible: it introduces new CSS custom properties for top/bottom sheets without renaming or removing existing props, events, exports, or CSS variables. Docs and regression/back-compat Playwright tests are included.

murdore:
## Test-selector fix, with proof at `5bef30e`

The top-sheet case now selects its trigger by test id rather than the demo's button copy. Same assertions, no coupling to prose:

https://github.com/user-attachments/assets/e28b7a2b-40ec-4910-8ca1-583e8634ffd5

Full gates on this commit: lint 0, check 0, unit 0, build 0, integration 0 — 11 Playwright cases.

tests/sheet-band-width.test.ts
murdore: Taken. The demo's button copy is prose and free to change without this test being wrong, so coupling to it makes the test brittle for no benefit.

Both the trigger and the panel now have test ids, and the case selects `sheet-top-trigger` / `sheet-top-panel`. Assertions are unchanged.

murdore: Taken. The demo's button copy is prose and free to change without this test being wrong, so coupling to it makes the test brittle for no benefit.

Both the trigger and the panel now have test ids, and the case selects `sheet-top-trigger` / `sheet-top-panel`. Assertions are unchanged.

Tara-ag:

Review could not be completed because the time budget was exhausted before any files or the PR diff could be inspected. No findings were verified, so the change cannot be considered safe to merge.

murdore:
## Runtime proof, re-recorded at the merged commit `21f3975`

Re-recorded at the commit that actually landed, so every PR in this batch carries proof from its merged state rather than its branch head. 11/11 passing.

### Nothing changes unless you opt in

| Behaviour | Recording |
| --- | --- |
| An unconfigured bottom sheet is still edge-to-edge | [watch](https://github.com/user-attachments/assets/9c82091c-21b5-4b97-baa1-cc8a90f4feb4) |
| An unconfigured bottom sheet draws no border at all | [watch](https://github.com/user-attachments/assets/1cd4dc9f-5f60-4f24-bf50-8402c7fe7c84) |
| `--sheet-width` does not leak into a bottom sheet | [watch](https://github.com/user-attachments/assets/1460bab7-4f93-4802-a8aa-4b8224f618bb) |

### Width-capped top/bottom panels (#572)

| Behaviour | Recording |
| --- | --- |
| `--sheet-band-width` caps the panel below the viewport width | [watch](https://github.com/user-attachments/assets/d9d12758-0402-4aa1-a464-1713b74a8c5d) |
| A capped panel is centred horizontally | [watch](https://github.com/user-attachments/assets/9106bd40-2bff-4ce0-a48e-c820691a1a81) |
| A capped panel still meets the bottom edge it enters from | [watch](https://github.com/user-attachments/assets/127f2ff0-00c8-4963-a223-e433dae1a1eb) |
| `--sheet-band-border` gives a capped panel all four edges | [watch](https://github.com/user-attachments/assets/44bc2e74-dcba-4380-9838-545665561412) |
| A top cap honours asymmetric insets and `max-width` | [watch](https://github.com/user-attachments/assets/ac364879-d489-42bd-a1ef-120c8d9b7913) |
| A bordered cap fits a phone viewport with no horizontal overflow | [watch](https://github.com/user-attachments/assets/61ae7541-e06b-40d2-8d6a-6f00ba49566a) |

### Dialog behaviour is preserved when capped

| Behaviour | Recording |
| --- | --- |
| Capped sheets trap focus, lock scrolling and restore focus after Escape | [watch](https://github.com/user-attachments/assets/c65f2c17-31b5-4845-a6f2-1c5389e4cd1c) |
| Clicking outside a capped sheet dismisses it | [watch](https://github.com/user-attachments/assets/a7882f93-90d7-45a6-8525-0480e5ff2c08) |

That last pair is the part worth scrutinising: capping the width changes the panel's box, which is exactly the kind of change that quietly breaks a focus trap or an outside-click target. Both are asserted rather than assumed.

### Verification

Release run #387 on `21f3975`: `verify` success, `release` success, published as **4.14.0**.

PR #592, opened 2026-09-08
feat(pie-chart): add a right-side value legend with a capped +N more expander

### LINKED ISSUE #417: PieChart: right-side value legend with maxItems/'+N more' expander
The DS pie sheet (Figma node 38527-7313) shows the donut legend as a right-side column (dot + label + right-aligned %) that truncates at 5 items with an in-card '+2 more ˅' expander. PieChart has `legendShowValues` (below-chart list) but no right-position layout, no item cap, no expander — so Lighthouse's analytics page hand-rolls a bespoke legend `<ul>` with a view-more modal (third parallel legend implementation). A `legendPosition: 'right'` + `legendMaxItems`/`onLegendMore` API would let consumers delete the bespoke ones.

murdore:
Adds the right-side value legend, item cap and `+N more` expander the DS pie sheet specifies.

## The gap

PieChart had `legendShowValues` — a below-chart list — but no right placement, no item cap and no expander. So Lighthouse's analytics page hand-rolls a bespoke legend `<ul>` plus a view-more modal: the third parallel legend implementation in that codebase.

```svelte
<PieChart {data} showLegend legendShowValues legendPosition="right" legendMaxItems={5} />
```

| Prop | Default | What it does |
| --- | --- | --- |
| `legendPosition` | `'bottom'` | `'right'` places the values legend beside the chart |
| `legendMaxItems` | `-` | Caps the visible rows and shows `+N more` |
| `onlegendmore` | `-` | Hands the overflow action to the consumer |

All three are opt-in — without them the legend renders exactly where and how it did, which four tests assert. Only the values legend moves; the simple top legend is mutually exclusive with it and is untouched.

## Three defects the regression tests found

- **A fractional cap rendered `+1.2999999999999998 more`.** The cap is floored and clamped to at least one row.
- **Expansion survived a data change.** Expand, shrink the data below the cap, restore it — and the legend stayed expanded. It now resets when the data length or the cap changes.
- **The expander had no themeable focus ring.** Worth being precise: a native `<button>` already draws the user-agent ring, so it was not invisible — it was un-themeable and inconsistent with the rest of the library. `--piechart-legend-more-focus-outline` now drives it, and the test sets `3px` and asserts the computed width, so it fails against the browser default of `1px` rather than passing on it.

## `aria-expanded` is present only where expansion is real

The built-in expander controls expansion, so it carries `aria-expanded` and toggles it.

When `onlegendmore` is supplied the component expands **nothing** and cannot know what the consumer's handler does — Lighthouse opens a modal, another caller might navigate. Claiming `aria-expanded="false"` would announce a collapsed disclosure this button never expands, and `aria-haspopup="dialog"` would promise a dialog the component cannot guarantee. Both are assertions about someone else's UI, so the callback variant renders a plain action button and the docs say the ARIA state is the consumer's to own. A test pins it.

## A docs claim that was simply false

The docs presented these props as `sui-pie-chart` attributes. **There is no PieChart web-component wrapper at all.** That example promised an API that does not exist; it is removed and the Svelte-only availability stated plainly. Building a first wrapper is a separate change, not something to slip into a legend feature.

## Verification

| Gate | Exit |
| --- | --- |
| lint | 0 |
| check | 0 |
| test:unit | 0 |
| build | 0 |
| test:integration | 0 |

**931 unit tests** and **642 functional tests** with 1 pre-existing skip. 12 Playwright cases and 12 DOM cases cover placement, the cap, the computed hidden count, expand/collapse, callback override, keyboard activation, the focus token, and the boundaries: `0`, `-1`, `NaN`, `Infinity`, `0.5` and `5.7` caps, a cap at or above the item count, empty data, and 57-character labels in a 280px card staying inside the chart's own box.

## Review findings

**`aria-controls` on the built-in expander** — added, pointing at the legend list, and omitted in callback mode where the button controls nothing in this component.

**The empty state no longer takes the beside-chart row layout**, which had turned the empty message into a flex item rather than the full-width block it is documented to be.

**On the `sui-pie-chart` finding, I disagree and have not made the change.** There is no `PieChart.wc.svelte` and no `customElements.define('sui-pie-chart', …)` anywhere in `src/wc/` — the element never registered, so no consumer could depend on it. Marking this `feat!` would announce a break that cannot have happened. The docs now carry an explicit correction instead. Reasoning in the thread.

Closes #417

## Summary by CodeRabbit

* **New Features**
  * Added support for displaying PieChart legends beside the chart.
  * Added optional legend item limits with a “+N more” expansion control.
  * Added callback support for handling legend expansion externally.
  * Added accessible keyboard interaction and focus styling for the legend control.

* **Documentation**
  * Updated PieChart properties, legend behavior, styling variables, and accessibility guidance.
  * Clarified that the previously documented web-component wrapper is unavailable.


tests/piechart-legend-right.spec.ts
```diff
@@ -0,0 +1,113 @@
+import { expect, test } from '@playwright/test';
+
+// Closes #417: PieChart had `legendShowValues` (a below-chart list) but no
+// right-position layout, no item cap and no expander, so Lighthouse's
+// analytics page hand-rolls a bespoke legend <ul> plus a view-more modal --
+// the third parallel legend implementation in that codebase.
+//
+// All three additions are opt-in: without legendPosition/legendMaxItems the
+// legend renders exactly where and how it did.
```
Tara-ag, line 9: 💡 MINOR: The spec navigates the shared demo page and asserts counts based on its hard-coded data arrays. This couples the tests to demo-page content that may change independently. Prefer an isolated fixture route or document that the demo data is part of the test contract.

src/lib/PieChart/PieChart.svelte
```diff
@@ -469,4 +514,55 @@
...
+    cursor: var(--cursor, pointer);
+    font-size: var(--piechart-legend-more-font-size, 12px);
+    font-family: inherit;
+    color: var(--piechart-legend-more-color, #2563eb);
+  }
+
+  .pie-legend-more:hover {
+    text-decoration: underline;
+  }
 </style>
```
Tara-ag, line 597: ⚠️ MAJOR: The `.pie-legend-more` button only styles `:hover`. It has no `:focus-visible` outline or other visible focus indicator, so keyboard users cannot see when the control is focused. Add a focus-visible style, e.g.:
```css
.pie-legend-more:focus-visible {
  outline: 2px solid currentColor;
  outline-offset: 2px;
  border-radius: 2px;
}
```
or expose `--piechart-legend-more-focus-outline`.

src/lib/PieChart/PieChart.svelte
```diff
@@ -358,17 +391,29 @@
...
+              </span>
+            </li>
+          {/each}
+        </ul>
+        {#if legendHiddenCount > 0}
+          <button
+            type="button"
+            class="pie-legend-more"
+            aria-expanded={typeof onlegendmore === 'function' ? null : legendExpanded}
+            onclick={handleLegendMore}
```
Tara-ag, line 435: ⚠️ MAJOR: When `onlegendmore` is provided, `aria-expanded` is set to `null` and the attribute is omitted. The button still triggers an expand-like action (opening the consumer's modal), but screen-reader users lose the expanded/collapsed state. Keep `aria-expanded` on the button in both modes — default it to `false` when the consumer manages the overflow, or use `aria-haspopup="dialog"` and document that the consumer must supply expansion state.

src/lib/PieChart/PieChart.svelte
```diff
@@ -40,6 +40,9 @@
     classes,
     semiCircle = false,
     legendShowValues = false,
+    legendPosition = 'bottom',
```
Tara-ag, line 43: 🔒 CRITICAL: New public props (`legendPosition`, `legendMaxItems`, `onlegendmore`) are added here but there is no `src/wc/components/PieChart.wc.svelte` wrapper, and `src/wc/index.ts` does not register `sui-pie-chart`. The docs nevertheless claim these map to kebab-case attributes (`legend-position`, `legend-max-items`), so web-component consumers are told they can use props that have no implementation. Add a `PieChart.wc.svelte` wrapper that reflects scalar props to host attributes and forwards callbacks, then register it in `src/wc/index.ts`; or remove the new attributes from the web-component docs until the wrapper exists.

src/lib/PieChart/PieChart.svelte
```diff
@@ -49,6 +52,7 @@
 
   // ── State ──────────────────────────────────────────────────────
 
+  let legendExpanded = $state(false);
```
Tara-ag, line 55: 💡 MINOR: `legendExpanded` is toggled by clicks but never reset when `legendCap` or `data.length` changes. If the data shrinks so the cap is no longer exceeded, the expander disappears but the internal expanded state stays true. Add a `$effect` to reset it when the cap or data changes:
```ts
$effect(() => {
  void legendCap;
  void data.length;
  legendExpanded = false;
});
```

Tara-ag (changes requested):
Blocking: the new public props are documented as web-component attributes but the `sui-pie-chart` wrapper does not exist, so the web-component API contract is broken. Additionally, the new expander button needs a visible focus indicator and correct ARIA expansion state. Please address the CRITICAL and MAJOR findings.

src/lib/PieChart/PieChart.svelte
Tara-ag, line 435: ⚠️ MAJOR: `aria-expanded` removed when `onlegendmore` is provided. When a consumer passes `onlegendmore`, the button's `aria-expanded` is set to `null`, so the attribute is omitted. The button still toggles/expands something (the consumer's modal), but the ARIA state that communicates expansion is gone. Screen-reader users lose the expanded/collapsed announcement for the consumer-managed overflow. Keep `aria-expanded` on the button in both modes — default it to `false` (not expanded) when `onlegendmore` is provided, rather than removing the attribute entirely. If the intent is that the button becomes a plain action button, consider using `aria-haspopup="dialog"` and document that the consumer must supply expansion state.

tests/piechart-legend-right.spec.ts
Tara-ag, line 9: 💡 MINOR: Integration tests rely on shared demo page state. The spec navigates to `/components/pie-chart` and asserts counts based on the demo page's hard-coded data arrays. While the test IDs isolate instances, the assertions are coupled to the demo page content; if the demo page is edited independently, the tests will break. Consider adding a dedicated isolated route or fixture for these assertions, or at minimum document that the test data in `src/routes/components/pie-chart/+page.svelte` is part of the test contract.

src/lib/PieChart/PieChart.svelte
Tara-ag, line 597: ⚠️ MAJOR: Expander button lacks visible focus indicator. `.pie-legend-more` only has a `:hover` underline style and no `:focus-visible` outline or other visible focus indicator. Keyboard users tabbing to the control will not see where focus is, violating WCAG 2.4.7 Focus Visible. Add a `:focus-visible` style, e.g.:
```css
.pie-legend-more:focus-visible {
  outline: 2px solid currentColor;
  outline-offset: 2px;
  border-radius: 2px;
}
```
or expose a `--piechart-legend-more-focus-outline` CSS variable.

src/lib/PieChart/PieChart.svelte
Tara-ag, line 43: 🔒 CRITICAL: New public props not exposed on web-component wrapper. PieChart has a web-component build (`sui-pie-chart`) but there is no `src/wc/components/PieChart.wc.svelte` wrapper at all. This PR adds three new public props (`legendPosition`, `legendMaxItems`, `onlegendmore`) and several new CSS custom properties. Because the wrapper is missing, web-component consumers cannot set these props/attributes. The docs already document scalar props mapping to kebab-case attributes, but the wrapper does not exist to honor that contract. Add `src/wc/components/PieChart.wc.svelte` that mirrors the Svelte component's scalar/string prop surface and register it in `src/wc/index.ts`, or remove/qualify the web-component documentation that claims these attributes work.

src/lib/PieChart/PieChart.svelte
Tara-ag, line 55: 💡 MINOR: `legendExpanded` state persists when data or cap changes. `legendExpanded` is toggled by user interaction but is never reset when `data.length` or `legendMaxItems` changes. If a user expands a capped legend and the data is then reduced so the cap is no longer exceeded, the expander disappears but `legendExpanded` remains `true`. Add a `$effect` that resets `legendExpanded` to `false` when `legendCap` or `data.length` changes:
```ts
$effect(() => {
  void legendCap;
  void data.length;
  legendExpanded = false;
});
```

Tara-ag (changes requested):
The PR adds useful opt-in legend features and good test coverage, but it cannot merge yet because the new public props are not available to web-component consumers (no `PieChart.wc.svelte` wrapper exists). There are also accessibility gaps on the expander button. Please address the CRITICAL and MAJOR findings.

Tara-ag:

Severity counts — 🔒 CRITICAL: 1 · ⚠️ MAJOR: 2 · 💡 MINOR: 0

Three findings were accepted by the review gate this run: one CRITICAL backward-compatibility issue (documented web-component API removed without major bump/migration note) and two MAJOR issues ($effect fragility around legendExpanded reset, and missing accessible name on the expander button). All three already have inline comments posted. Per the blocking criteria, a CRITICAL public-contract breaking change blocks merge.

- 🔒 **CRITICAL**: Docs removed a documented web-component mapping without a major-version bump — `docs/PieChart.md:290`
  The PR deletes the documented `sui-pie-chart` web-component usage and its scalar prop-to-attribute mapping from docs/PieChart.md, replacing it with a note that PieChart is Svelte-only. This is a breaking change to the documented public contract of a published library component. Even though no actual `PieChart.wc.svelte` wrapper or `sui-pie-chart` registration exists on base, the docs presented thi…
- ⚠️ **MAJOR**: $effect resets legendExpanded on every render due to void expressions — `src/lib/PieChart/PieChart.svelte:173`
  The `$effect` that resets `legendExpanded` when `legendCap` or `data.length` changes uses `void legendCap; void data.length;` to force dependency tracking. In Svelte 5 runes mode, `void` expressions do track reactivity, but this pattern is fragile and non-idiomatic. More importantly, because `data` is a prop (likely an array reference), `data.length` only tracks length changes if the array referen…
- ⚠️ **MAJOR**: Expander button lacks an accessible name when no visible label is present — `src/lib/PieChart/PieChart.svelte:414`
  The `+N more` / `Show less` button text provides a visible label, which is good. However, when `onlegendmore` is provided the button is rendered as an action button without `aria-expanded` (correct), but it also has no `aria-haspopup` or other hint about what it opens. More importantly, the button relies solely on its text content for an accessible name; if a consumer ever passes an empty/whitespa…

tests/piechart-legend-right.spec.ts
murdore: The channels fixture now explicitly documents the seven-items/five-visible contract shared with this spec. The existing demo-based test approach is retained.

src/lib/PieChart/PieChart.svelte
murdore: Confirmed the documentation was wrong: no PieChart custom-element wrapper is registered. Removed the nonfunctional sui-pie-chart example and attribute mapping, replacing them with an explicit Svelte-only availability note rather than silently introducing an entire new wrapper.

src/lib/PieChart/PieChart.svelte
murdore: Fixed: expansion resets on normalized cap or data length changes. A DOM regression test failed before the fix (7 rows instead of 5 after shrink/regrow) and passes afterward. Also fixed a fractional cap producing +1.2999999999999998 more; row slicing and hidden counts now use the same integer cap.

src/lib/PieChart/PieChart.svelte
murdore: Added a themeable focus-visible outline and a browser regression test. The original native button already had a 1px browser outline, so complete absence of focus was not reproduced; the new explicit rule is deterministic and themeable. RED received 1px rather than requested 3px; GREEN passes all 11 browser tests.

src/lib/PieChart/PieChart.svelte
murdore: Not changing aria-expanded in callback mode: this is deliberately a plain action, whose callback may navigate or open any UI. A constant false would announce false state after a consumer opens its dialog; aria-haspopup=dialog would also assert an unknown destination. Documented this contract and added a DOM test proving callback activation leaves the list capped and carries neither false expansion nor a fabricated dialog hint. Built-in expansion still has accurate aria-expanded.

tests/piechart-legend-right.spec.ts
murdore: Addressed with the matching earlier thread: the channels array now documents the exact seven-total/five-visible test contract. Pushed in f47551c.

src/lib/PieChart/PieChart.svelte
murdore: Addressed: removed the inaccurate custom-element example and mapping, explicitly documenting PieChart as Svelte-only. No wrapper exists in the bundle; the docs no longer claim otherwise.

src/lib/PieChart/PieChart.svelte
murdore: Fixed in f47551c with a failing-then-passing shrink/regrow regression test. Expansion resets when the normalized cap or data length changes.

src/lib/PieChart/PieChart.svelte
murdore: Added an explicit themeable focus-visible outline and verified it with a browser test. All 11 legend browser cases pass. The previous native outline existed, but the new rule no longer depends on browser styling.

src/lib/PieChart/PieChart.svelte
murdore: Kept the callback path as a plain action button and documented/tested that contract. Neither constant aria-expanded=false nor a dialog hint can truthfully represent an arbitrary callback. Built-in disclosure mode retains its actual expansion state. See the detailed reply on the matching earlier thread.

docs/PieChart.md
```diff
@@ -246,8 +290,4 @@ Thresholds: ≥ 1 Cr (10 million) → `"1.5Cr"`, ≥ 1 L (100 thousand) → `"2.
```
Tara-ag, line 294: 🔒 CRITICAL: Documented web-component API for PieChart is removed without major-version bump or migration note. The existing docs/PieChart.md documents `<sui-pie-chart>` as a usable web component with scalar kebab-case attributes and function props set via JS property assignment. This PR deletes that entire section and replaces it with a statement that `sui-pie-chart` is not registered. This withdraws a previously documented public usage path — a breaking change. The PR title/body is `feat(pie-chart)` with no `!` or major-version bump, and no migration note is provided.

Either restore the documented `<sui-pie-chart>` web-component usage (and add the missing wrapper if needed), or mark this as a breaking change with `feat(pie-chart)!:` and add a migration note explaining that consumers using `<sui-pie-chart>` must switch to the Svelte `PieChart` import.

murdore:
Following up on the expander findings with the evidence behind each, now that all three are settled.

**Visible focus — taken.** `.pie-legend-more:focus-visible` now draws `var(--piechart-legend-more-focus-outline, 2px solid currentColor)` at a 2px offset. One correction to the finding's premise, since it matters for what the test asserts: the control was not previously *unfocusable* — a native `<button>` already draws the user-agent ring — it was un-*themeable* and inconsistent with the rest of the library. So the test sets the token to `3px` and asserts the computed width, which fails against the browser default of `1px` rather than passing on it.

**`aria-expanded` in callback mode — respectfully still disputed, and now pinned by a test rather than an argument.** When `onlegendmore` is supplied, the component does not expand anything: it hands control to the consumer and the rows do not change. What that consumer does next is unknown to the component — Lighthouse opens a modal, another caller might navigate, or log, or fetch. Emitting `aria-expanded="false"` would announce a collapsed disclosure that this button never expands, and `aria-haspopup="dialog"` would promise a dialog the component cannot guarantee. Both are assertions about someone else's UI.

So the callback variant renders a plain action button, and the consumer owns the state their own UI has. `legend.test.ts` asserts exactly that:

```ts
expect(button.hasAttribute('aria-expanded')).toBe(false);
expect(button.hasAttribute('aria-haspopup')).toBe(false);
await fireEvent.click(button);
expect(calls).toBe(1);
expect(container.querySelectorAll('.pie-legend-row')).toHaveLength(5);
```

The built-in expander — the one that *does* control expansion — keeps `aria-expanded` and toggles it. `docs/PieChart.md` states the split so a consumer opening a dialog knows the ARIA state is theirs to set.

**Stale expansion — taken**, reset when `data.length` or the cap changes, covered both ways: expand → shrink data → regrow returns to the capped rows, and changing the cap while expanded collapses with `aria-expanded="false"`.

**Boundary cases the review prompted, all now covered:** `0`, `-1`, `NaN` and `Infinity` caps render every row with no expander; `0.5` clamps to one row and `+6 more`; `5.7` floors to 5 rows and `+2 more` (it produced `+1.2999999999999998 more` before the fix); a cap at or above the item count renders no expander; empty data renders no rows and no expander; and a 280px-wide card with 57-character labels keeps every legend row inside the chart's own box.

**Web component — the docs were wrong and are corrected, not papered over.** There is no `PieChart.wc.svelte`; `sui-pie-chart` does not exist. The attribute example promising these props to custom-element consumers has been removed and the Svelte-only availability stated plainly. Building a first wrapper for this component is a separate change, not something to slip into a legend feature.

murdore:
## Recorded proof — expander and layout

Recorded at `03a4c24`, the commit on this branch. Each clip was decoded end-to-end with `ffmpeg -xerror` and a frame reviewed before upload.

**The capped legend and its expander.** Five rows beside the chart, `+2 more`, then all seven in place with `Show less` — the DS shape this issue asked for.

https://github.com/user-attachments/assets/1b352e4a-9e1a-4cde-8792-081a121b7870

**Long labels in a narrow card.** A 280px-wide chart at a 375px viewport with 57-character labels: every legend row's right edge is asserted to stay within the chart's own box, so the right-hand placement cannot silently overflow the card it sits in.

https://github.com/user-attachments/assets/ee866e0c-4365-4d41-a7fc-70b89f798b2d

Two of twelve browser cases; the numeric boundaries (`0`, `-1`, `NaN`, `Infinity`, `0.5`, `5.7`, a cap at or above the item count, empty data) are DOM tests, because a screenshot of a legend cannot show that `5.7` floors to five rather than rendering `+1.2999999999999998 more` — which is what it did before the fix.

Local gate log for this commit: lint 0, check 0, unit 0, build 0, integration 0 (931 unit, 642 functional, 1 pre-existing skip).

src/lib/PieChart/PieChart.svelte
```diff
@@ -358,17 +398,29 @@
...
+            </li>
+          {/each}
+        </ul>
+        {#if legendHiddenCount > 0}
+          <button
+            type="button"
+            class="pie-legend-more"
+            aria-expanded={typeof onlegendmore === 'function' ? null : legendExpanded}
+            onclick={handleLegendMore}
+          >
```
Tara-ag, line 436: 💬 SUGGESTION: The built-in expander button exposes `aria-expanded` but no `aria-controls`. Adding it improves screen-reader discoverability of the expanded/collapsed region. Use a component-scoped id on the `<ul>` and forward it to the button only in built-in mode (omit it in callback mode, which already correctly omits `aria-expanded`).

src/lib/PieChart/PieChart.svelte
```diff
@@ -153,6 +157,41 @@
...
+    legendExpanded = false;
+  });
+  const legendHiddenCount = $derived(legendCap === null ? 0 : Math.max(0, data.length - legendCap));
+  const legendRows = $derived(
+    legendCap === null || legendExpanded ? data : data.slice(0, legendCap)
+  );
+
+  const legendBesideChart = $derived(showLegend && legendShowValues && legendPosition === 'right');
+
+  const handleLegendMore = (): void => {
```
Tara-ag, line 200: 💡 MINOR: When `data` is empty and `legendPosition="right"` is set, `.legend-right` is still applied because `legendBesideChart` does not factor in `isEmpty`. The empty-state div then becomes a flex item instead of spanning the container width, altering the documented centered empty-state layout.

Suggested fix: guard the right-layout derived with `!isEmpty`:
```ts
const legendBesideChart = $derived(
  !isEmpty && showLegend && legendShowValues && legendPosition === 'right'
);
```

Tara-ag (approved):
Approved with two minor notes: (1) empty-state layout under `legendPosition="right"` should not apply the flex row class, and (2) the built-in expander could add `aria-controls` for better screen-reader context. Neither blocks the PR.

Tara-ag (approved):
Approved. The two findings accepted by the review gate this run are minor quality/accessibility nits (empty-state layout under `legendPosition='right'` and `aria-controls` on the built-in expander) and do not block merge. All previously raised higher-severity concerns have been addressed by the author.

murdore:
## Review fixes, with proof at `88c35ae`

Decoded and frame-reviewed before upload. The expander, now carrying `aria-controls` pointing at the list it expands:

https://github.com/user-attachments/assets/257f6947-5d02-46f1-9214-5fcccba531e4

`aria-controls` and the empty-state layout are both asserted in `legend.test.ts` (14 cases), since neither an attribute nor an absent CSS class shows up in a recording.

Full gates on this commit: lint 0, check 0, unit 0, build 0, integration 0.

---

> **Superseded.** `PieChart.svelte` has changed since this was posted (`88c35ae` → `cc76ca8`), so this clip no longer depicts the shipping code. Current recording: https://github.com/juspay/svelte-ui-components/pull/592#issuecomment-5601529589.

src/lib/PieChart/PieChart.svelte
murdore: Taken. The built-in expander now points at the list it expands:

```svelte
aria-expanded={typeof onlegendmore === 'function' ? null : legendExpanded}
{...typeof onlegendmore === 'function' ? {} : { 'aria-controls': legendListId }}
```

Omitted in callback mode, for the same reason `aria-expanded` is: that button controls nothing inside this component, and pointing it at a list it does not change would be a false claim. The id comes from `$props.id()`, so two charts on one page do not collide. Asserted in `legend.test.ts` against the list's own id.

src/lib/PieChart/PieChart.svelte
murdore: Taken, and the consequence is exactly as described: with no rows the root still became a flex row, so the empty-state message turned into a flex item instead of the full-width block it is documented to be.

```ts
const legendBesideChart = $derived(
  !isEmpty && showLegend && legendShowValues && legendPosition === 'right'
);
```

Covered by a test asserting `.pie-chart.legend-right` is absent when `data` is empty and `legendPosition="right"` is set.

docs/PieChart.md
murdore: This one I'm pushing back on, and the evidence is checkable in three commands.

**There is no `sui-pie-chart`, and there never was.**

```
$ ls src/wc/components/ | grep -i pie      # no PieChart.wc.svelte
$ rg -n 'pie-chart|PieChart' src/wc/       # no registration, no import
```

`src/wc/index.ts` is a list of side-effecting imports that call `customElements.define`, one per wrapper file. PieChart has no wrapper file, so it is not imported there, so `customElements.define('sui-pie-chart', …)` never runs. `<sui-pie-chart>` in a consumer's HTML has always been an unknown element: no upgrade, no props, no render.

So the section I removed did not document a "previously documented public usage path" — it documented a component that does not exist. Following its instructions produced nothing on the page.

**Why I am not marking this `feat(pie-chart)!`.** A breaking change tells consumers that something they depend on has changed and they must act. Nobody can depend on an element that never registered. Publishing a major for this would send every consumer looking for a migration they cannot perform, and would put a false entry in the changelog — the version number would claim a break that did not happen.

**Why I am not restoring the docs either.** That is the option that actually breaks someone: a consumer reads the restored section, writes `<sui-pie-chart data="…">`, and gets a blank space with no error. Documentation that describes a component into existence is the defect here, not the fix for it.

**What I have done instead**, since "the docs silently changed" is a fair complaint even when the old text was wrong: `docs/PieChart.md` now states plainly that PieChart is Svelte-only, that `sui-pie-chart` is not registered, and that earlier revisions of this file described a wrapper that was never built. Anyone who tried the old snippet and saw nothing render now finds the explanation in the place they would look.

Building the wrapper is a reasonable thing to want, and it is a real piece of work — a `PieChart.wc.svelte`, a registration, attribute coercion for the scalar props, prop-parity tests, and its own visual baseline. That belongs in its own PR, not smuggled into a legend feature where it would ship untested.

Happy to be shown otherwise if you can point at a registration I have missed.

src/lib/PieChart/PieChart.svelte
```diff
@@ -153,6 +157,49 @@
...
+     hand this anything. */
+  const legendCap = $derived(
+    typeof legendMaxItems === 'number' && Number.isFinite(legendMaxItems) && legendMaxItems > 0
+      ? Math.max(1, Math.floor(legendMaxItems))
+      : null
+  );
+  // A changed list or cap starts a new collapsed view, not a stale expansion.
+  // eslint-disable-next-line no-restricted-syntax
+  $effect(() => {
+    void legendCap;
```
Tara-ag: ⚠️ MAJOR: `$effect` resets `legendExpanded` on every render due to `void` expressions. The `void legendCap; void data.length;` pattern is fragile and non-idiomatic. Worse, because `data` is a prop array, a parent passing a new array literal on every render will cause this effect to fire repeatedly and collapse the legend even when the semantic data has not changed. Prefer deriving the collapsed state from `legendCap` and `data.length` directly (e.g. `const shouldCollapse = $derived(...)` and use `$effect` only to sync `legendExpanded` when that derived value changes), or compare against a stable snapshot of the data identity/length.

docs/PieChart.md
```diff
@@ -246,8 +290,6 @@ Thresholds: ≥ 1 Cr (10 million) → `"1.5Cr"`, ≥ 1 L (100 thousand) → `"2.
```
Tara-ag, line 294: 🔒 CRITICAL: Docs removed a documented web-component mapping without a major-version bump. The prior `<sui-pie-chart>` example and kebab-case attribute mapping were part of the published public contract. Even though no wrapper was ever registered, the docs presented this API as available. Removing it without a major-version bump or migration note breaks backward compatibility for any consumer who followed the documented web-component path. Either restore the section with a deprecation notice, add a migration note, or bump the major version.

src/lib/PieChart/PieChart.svelte
```diff
@@ -358,17 +406,30 @@
...
-            </span>
-          </li>
-        {/each}
-      </ul>
+      <div class="pie-legend-column">
+        <ul class="pie-legend-values" id={legendListId}>
+          {#each legendRows as d, i (i)}
+            <li class="pie-legend-row">
+              <span class="pie-legend-swatch" style="background: {d.color ?? getColor(i)}"></span>
+              <span class="pie-legend-label">{d.label}</span>
```
Tara-ag, line 422: ⚠️ MAJOR: Expander button relies solely on visible text for its accessible name. In callback mode it is a plain action button with no `aria-expanded` (intentional) and no `aria-haspopup`, so screen-reader users get no hint about what the button does beyond the visible count. Add an explicit `aria-label` such as `aria-label={legendExpanded ? 'Show less legend items' : \`Show ${legendHiddenCount} more legend items\`}` so the button remains accessible even if the visible text is overridden or truncated.

Tara-ag (changes requested):
Yama review complete. One CRITICAL and two MAJOR findings remain; see inline comments.

murdore:
Three findings this run. One taken, one refused on a false premise, one already open above.

## `$effect` fragility — taken, and it taught me something

The concern about `void legendCap; void data.length;` is fair: `void` reads are easy to mistake for dead code, and a reader cannot tell they are load-bearing.

I first replaced the effect entirely with a derived key:

```ts
const legendKey = $derived(`${legendCap}:${data.length}`);
const legendExpanded = $derived(expandedFor === legendKey);
```

**The existing regression test rejected it**, which is exactly what it is for:

```
resets expansion when the data length changes
  expected 5 rows, got 7
```

Collapsing depends on the **transition**, not on the current values. A list that shrinks below the cap and then returns to its former length must stay collapsed — but a key built from `(cap, length)` matches its earlier self, so the stale expansion came back. That is a real behaviour, guarded by a real test, and no amount of derivation expresses it because the state is about *what changed*, not *what is*.

So the effect stays, and the `void` expressions are gone. The dependencies are now ordinary function arguments, which read as deliberate:

```ts
const collapseLegend = (_cap: number | null, _rowCount: number): void => {
  legendExpanded = false;
};
$effect(() => {
  collapseLegend(legendCap, data.length);
});
```

Same tracking, no `void`, and the comment now records why an effect is the right tool rather than leaving the next reader to relitigate it.

On the second half of that finding — that `data.length` only tracks when the array reference changes — this is worth stating precisely: an in-place `push` on a prop array is not reactive for *anything* in this component, including the chart itself, so the legend is not a special case. I have added a test for the case that **is** reachable, where the row count stays the same and the cap changes, and it passes.

## Accessible name on the expander — refused, the premise does not hold

> the button relies solely on its text content for an accessible name; if a consumer ever passes an empty/whitespace…

**There is no prop that sets this label.** The text is generated by the component:

```svelte
{legendExpanded ? 'Show less' : `+${legendHiddenCount} more`}
```

`legendHiddenCount` is a positive integer whenever the button renders at all — the button is inside `{#if legendHiddenCount > 0}` — so the label is never empty and no consumer input reaches it. There is no `legendMoreLabel` prop in `properties.ts` to pass whitespace through.

Adding `aria-label` here would *replace* a correct visible name with a duplicate, and adding `aria-haspopup` in callback mode would assert a popup this component cannot know about — the same reason `aria-expanded` is omitted there, which this review previously agreed with.

I have pinned the label instead, so a future change that made it blank fails a test rather than shipping.

## The `sui-pie-chart` CRITICAL

Unchanged, and the thread on `docs/PieChart.md` is left **open** for a human decision rather than resolved by me. This run's own wording concedes the substance — *"no actual `PieChart.wc.svelte` wrapper or `sui-pie-chart` registration exists on base"* — and my position is that a docs section describing a component that never registered cannot be a public contract that a major version break protects. Full reasoning in that thread.

src/lib/PieChart/legend.test.ts
```diff
@@ -0,0 +1,121 @@
...
+  it('has no rows or expander for empty data', () => {
+    const { container, queryByRole } = render(PieChart, { ...props, data: [] });
+    expect(container.querySelectorAll('.pie-legend-row')).toHaveLength(0);
+    expect(queryByRole('button')).toBeNull();
+  });
+
+  it('collapses when the cap changes after expansion', async () => {
+    const { container, getByRole, rerender } = render(PieChart, props);
+    await fireEvent.click(getByRole('button'));
+    await rerender({ ...props, legendMaxItems: 3 });
```
Tara-ag, line 68: 💡 MINOR: The test title and comment describe replacing the data array with a different same-length array, but the test body only changes `legendMaxItems` from 5 to 4. Rename this test to match what it actually exercises (e.g. "collapses when the cap decreases after expansion"), and add a separate test if you want to cover same-length data replacement.

Tara-ag (approved):
Approved with one minor note about a misleading test name/comment. The public API additions are opt-in and backward-compatible, docs and tests are thorough, and the accessibility handling of the built-in vs. callback expander is well-reasoned and documented.

src/lib/PieChart/legend.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    expect(container.querySelectorAll('.pie-legend-row')).toHaveLength(3);
+    expect(getByRole('button').getAttribute('aria-expanded')).toBe('false');
+  });
+
+  it('collapses when the same-length data is replaced with different rows', async () => {
+    // The old reset keyed off data.length, so swapping rows without changing
+    // the count left a stale expansion showing the new list fully expanded.
+    const { container, getByRole, rerender } = render(PieChart, props);
+    await fireEvent.click(getByRole('button'));
+    expect(container.querySelectorAll('.pie-legend-row')).toHaveLength(7);
```
Tara-ag, line 80: 💡 MINOR: The test name and comment describe a same-length data swap, but the test body only rerenders with a smaller `legendMaxItems`. Rename this test to describe the cap-change scenario it actually exercises, or add a true data-array swap if you want to cover that case.

Tara-ag (approved):
Approved with one minor note about a misleading test name/comment. The public API additions are backward-compatible, accessibility handling for the expander is sound, and the docs correction for the non-existent web-component wrapper is appropriate.

docs/PieChart.md
```diff
@@ -72,6 +72,36 @@ Use `percentDecimals` to control decimal places in the percentage column and on-
...
+By default that control expands the rest in place and toggles back to "Show less". If you already have somewhere better to put the overflow — a modal, a drawer, a linked report — pass `onlegendmore` and it takes over instead, so the legend does not also expand underneath your own UI:
+
+```svelte
+<PieChart
+  {data}
+  showLegend
+  legendShowValues
+  legendPosition="right"
+  legendMaxItems={5}
+  onlegendmore={() => (breakdownOpen = true)}
```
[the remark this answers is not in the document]
murdore: Correct — fixed in `cc76ca8`.

The snippet assigned to `breakdownOpen` without declaring it anywhere, so anyone pasting it got a `ReferenceError` on the first click (or a silent global in non-strict contexts, which is worse). Added the declaration so the example stands on its own:

```svelte
<script>
  let breakdownOpen = $state(false);
</script>

<PieChart
  {data}
  showLegend
  legendShowValues
  legendPosition="right"
  legendMaxItems={5}
  onlegendmore={() => (breakdownOpen = true)}
/>
```

src/lib/PieChart/PieChart.svelte
murdore: Taken, and already landed — the `void` expressions are gone as of `81f4105` (now `cc76ca8`).

The dependencies are ordinary function arguments now, which read as deliberate rather than looking like dead statements:

```ts
const collapseLegend = (_cap: number | null, _rowCount: number): void => {
  legendExpanded = false;
};
$effect(() => {
  collapseLegend(legendCap, data.length);
});
```

On the suggested alternative — deriving the collapsed state instead of using an effect — I tried that first and **the existing regression test rejected it**:

```
resets expansion when the data length changes
  expected 5 rows, got 7
```

Collapsing depends on the *transition*, not on the current values. A list that shrinks below the cap and then returns to its former length must stay collapsed, but a key built from `(cap, length)` matches its earlier self, so the stale expansion came back. The state here is about what changed, not what is, which is what an effect expresses and a derived value cannot.

On the second half — that a parent passing a new array literal every render collapses the legend: that is true and intended. A new array is a new dataset as far as the component can tell; there is no way to distinguish "same data, new reference" from "different data, same length" without deep comparison, and collapsing is the safe direction (it shows the capped view, never a stale expansion over rows the user never chose to reveal).

src/lib/PieChart/PieChart.svelte
murdore: Declining — the premise does not hold for this button.

> so the button remains accessible even if the visible text is overridden or truncated

**Neither is reachable.** The label is generated by the component, from a value the component computes:

```svelte
{legendExpanded ? 'Show less' : `+${legendHiddenCount} more`}
```

There is no `legendMoreLabel` prop in `properties.ts`, no snippet override for this control, and nothing consumer-supplied reaches that string — so there is no path by which it is "overridden". And the button only exists inside `{#if legendHiddenCount > 0}`, so the count is always a positive integer; `+0 more` cannot render.

Truncation is a separate confusion worth naming: CSS truncation (`text-overflow`, clipping, `overflow: hidden`) does not affect the accessible name. The name comes from the DOM text, not from what is painted. A visually clipped `+3 more` still exposes "+3 more" to the accessibility tree.

**Adding `aria-label` here would make it worse**, not safer. It would replace a correct visible name with a duplicate string, and the two would then be free to drift — which is the actual failure mode WCAG 2.5.3 (Label in Name) exists to prevent. `Snippet` in the sibling PR has the inverse case, a button with *no* visible text, and that is where `ariaLabel` belongs.

On `aria-haspopup` in callback mode: the component cannot know what `onlegendmore` does. It might open a modal, navigate to a report, or expand a panel elsewhere on the page. Asserting `haspopup` would state a fact the component has no access to, and be wrong for two of those three. This is the same reasoning that made `aria-expanded` conditional here — which this review agreed with in the previous round.

I have pinned the behaviour instead, so a future change that blanked the label fails a test rather than shipping:

```ts
it('always gives the expander a non-empty accessible name', () => {
  const { getByRole } = render(PieChart, props);
  expect(getByRole('button').textContent?.trim()).toBe('+2 more');
});
```

src/lib/PieChart/legend.test.ts
murdore: Correct, and this one is mine — thanks for catching it.

The test claimed a same-length data swap and then only lowered `legendMaxItems` from 5 to 4, which is the scenario the test three lines above it (`collapses when the cap changes after expansion`) already covers. So it was both mis-named *and* redundant: a duplicate wearing the name of a case nothing tested.

Fixed by making the body do what the name says, rather than renaming the body's actual behaviour into a second cap test:

```ts
const swapped = data.map((row) => ({ ...row, label: `Other ${row.label}` }));
const { container, getByRole, rerender } = render(PieChart, props);
await fireEvent.click(getByRole('button'));
expect(container.querySelectorAll('.pie-legend-row')).toHaveLength(7);

await rerender({ ...props, data: swapped });
expect(container.querySelectorAll('.pie-legend-row')).toHaveLength(5);
expect(getByRole('button').getAttribute('aria-expanded')).toBe('false');
```

Seven rows before, seven rows after — the count cannot be what triggers the collapse, so this now exercises the prop-identity path that nothing else did.

**I checked it can fail**, since a test that has only ever been green proves nothing. Replacing the tracked read with a constant that cannot distinguish the swap:

```
- collapseLegend(legendCap, data.length);
+ collapseLegend(legendCap, 7);
```

```
× resets expansion when the data length changes
× collapses when same-length data is replaced with different rows
Tests  2 failed | 14 passed (16)
```

Restored, and green again at 16/16.

src/lib/PieChart/legend.test.ts
murdore: Correct, and this one is mine — thanks for catching it.

The test claimed a same-length data swap and then only lowered `legendMaxItems` from 5 to 4, which is the scenario the test three lines above it (`collapses when the cap changes after expansion`) already covers. So it was both mis-named *and* redundant: a duplicate wearing the name of a case nothing tested.

Fixed by making the body do what the name says, rather than renaming the body's actual behaviour into a second cap test:

```ts
const swapped = data.map((row) => ({ ...row, label: `Other ${row.label}` }));
const { container, getByRole, rerender } = render(PieChart, props);
await fireEvent.click(getByRole('button'));
expect(container.querySelectorAll('.pie-legend-row')).toHaveLength(7);

await rerender({ ...props, data: swapped });
expect(container.querySelectorAll('.pie-legend-row')).toHaveLength(5);
expect(getByRole('button').getAttribute('aria-expanded')).toBe('false');
```

Seven rows before, seven rows after — the count cannot be what triggers the collapse, so this now exercises the prop-identity path that nothing else did.

**I checked it can fail**, since a test that has only ever been green proves nothing. Replacing the tracked read with a constant that cannot distinguish the swap:

```
- collapseLegend(legendCap, data.length);
+ collapseLegend(legendCap, 7);
```

```
× resets expansion when the data length changes
× collapses when same-length data is replaced with different rows
Tests  2 failed | 14 passed (16)
```

Restored, and green again at 16/16.

docs/PieChart.md
murdore: Settled, and in your favour on the outcome: **the wrapper is now built** — https://github.com/juspay/svelte-ui-components/pull/597.

I stand by the analysis that there was nothing to migrate *from* (no `PieChart.wc.svelte`, no `customElements.define('sui-pie-chart', …)`, so the element never upgraded), which is why #592 is not marked breaking and the changelog will not claim a break that no consumer could have experienced. But "the docs promised this element" was the right thing to press on, and the useful answer to a doc describing a component that does not exist is to make it exist rather than to delete the paragraph. So `docs/PieChart.md` in #597 documents a `sui-pie-chart` that really registers.

It is stacked on this branch rather than on `release`, deliberately: `scripts/wc-parity` requires a wrapper to declare *every* prop its component accepts, and this PR adds four legend props. A wrapper built on `release` would pass today and break `check:wc-parity` the moment this merges.

Building it also shook out a real bug that only a working wrapper could have found. Declaring the slot snippets at the top level of the template compiles them to module scope, where the `<slot>` inside still references `$$props` — which exists only inside the component function:

```js
const emptySlot = ($$anchor) => {
  $.slot(node_1, $$props, 'empty', {}, null);
};
```

That throws `$$props is not defined` at *render* time, which surfaces as a silently empty shadow root rather than a build error, and only when the snippet is actually rendered — so it looked fine with data present and failed only on an empty dataset with slotted content. Fixed by declaring the snippets inside `<PieChart>`, which is what every other wrapper already does.

Resolving this thread since the work it asked for exists; the review of the wrapper itself belongs on #597.

docs/PieChart.md
murdore: Settled, and in your favour on the outcome: **the wrapper is now built** — https://github.com/juspay/svelte-ui-components/pull/597.

I stand by the analysis that there was nothing to migrate *from* (no `PieChart.wc.svelte`, no `customElements.define('sui-pie-chart', …)`, so the element never upgraded), which is why #592 is not marked breaking and the changelog will not claim a break that no consumer could have experienced. But "the docs promised this element" was the right thing to press on, and the useful answer to a doc describing a component that does not exist is to make it exist rather than to delete the paragraph. So `docs/PieChart.md` in #597 documents a `sui-pie-chart` that really registers.

It is stacked on this branch rather than on `release`, deliberately: `scripts/wc-parity` requires a wrapper to declare *every* prop its component accepts, and this PR adds four legend props. A wrapper built on `release` would pass today and break `check:wc-parity` the moment this merges.

Building it also shook out a real bug that only a working wrapper could have found. Declaring the slot snippets at the top level of the template compiles them to module scope, where the `<slot>` inside still references `$$props` — which exists only inside the component function:

```js
const emptySlot = ($$anchor) => {
  $.slot(node_1, $$props, 'empty', {}, null);
};
```

That throws `$$props is not defined` at *render* time, which surfaces as a silently empty shadow root rather than a build error, and only when the snippet is actually rendered — so it looked fine with data present and failed only on an empty dataset with slotted content. Fixed by declaring the snippets inside `<PieChart>`, which is what every other wrapper already does.

Resolving this thread since the work it asked for exists; the review of the wrapper itself belongs on #597.

murdore:
## Re-recorded proof at `cc76ca8`

The previous recording was taken at `88c35ae`. `PieChart.svelte` has changed since then — the `$effect` dependencies moved from `void` expressions to real function arguments — so that clip no longer depicted the shipping code. Replaced rather than relabelled.

**The expander reveals the rest in place.** Seven rows behind a cap of five, expanded, with the control switching to "Show less".

https://github.com/user-attachments/assets/8d4f0f80-f1d9-4ea4-9a61-93370a2c4ea0

**`onlegendmore` takes over instead.** With a callback supplied, activating the control fires it and the legend stays capped — no expansion underneath the consumer's own drawer.

https://github.com/user-attachments/assets/0b237bcc-db91-41e8-83e9-40772c0a5ad3

All 12 cases in `piechart-legend-right.spec.ts` pass at this commit; these two are the pair that distinguish the built-in behaviour from the delegated one. Every clip was decoded end-to-end with `ffmpeg -xerror` and a frame reviewed before upload.

Unit at this commit: `legend.test.ts` **16 passed**, full unit suite **935 passed**, lint 0, check 0, build 0.

murdore:
## Runtime proof, re-recorded at the merged commit `5c9f40b`

Re-recorded at the commit that actually landed, so every PR in this batch carries proof from its merged state rather than its branch head. 12/12 passing.

### Default behaviour is untouched

| Behaviour | Recording |
| --- | --- |
| The default value legend is still **below** the chart | [watch](https://github.com/user-attachments/assets/a4192f6e-b9e8-4ec0-9cbb-11eb5ac593aa) |
| The default legend shows every item, uncapped | [watch](https://github.com/user-attachments/assets/0e45d696-ce83-40eb-bb33-dc8528b65b16) |
| No expander appears when the cap is not exceeded | [watch](https://github.com/user-attachments/assets/cc0370ce-1551-4dcf-a8c6-ed563bd86be9) |

### The right-side legend (#417)

| Behaviour | Recording |
| --- | --- |
| `legendPosition="right"` places the legend beside the chart | [watch](https://github.com/user-attachments/assets/22777912-e3c6-4689-9e5d-7aee885b53aa) |
| `legendMaxItems` caps the visible rows | [watch](https://github.com/user-attachments/assets/f540327b-3855-4090-a3e6-53a9ad993c74) |
| The expander names how many items are hidden | [watch](https://github.com/user-attachments/assets/c3b74d5d-dc93-4385-b36e-39998bbb8494) |
| Clicking the expander reveals the rest in place | [watch](https://github.com/user-attachments/assets/3ef0c135-cda4-483d-82f1-d8b014c0bfc5) |
| The expander collapses again once expanded | [watch](https://github.com/user-attachments/assets/8c935b1c-28c5-4eba-9a2d-dbaa9695dcb1) |
| `onlegendmore` takes over instead of expanding in place | [watch](https://github.com/user-attachments/assets/3ac511c1-d739-4b7a-bc75-bd2165017323) |

### Accessibility and layout

| Behaviour | Recording |
| --- | --- |
| The expander is a real `button`, reachable by keyboard | [watch](https://github.com/user-attachments/assets/00839a9d-eacc-4b06-ac8a-9546b566b2e3) |
| Keyboard focus has a themeable visible outline | [watch](https://github.com/user-attachments/assets/c8f66189-41cd-4cf8-8eb4-aa1e988169d3) |
| The right legend stays inside a narrow card with long labels | [watch](https://github.com/user-attachments/assets/8e718a37-a908-40eb-b344-b88b38791f44) |

### Verification

Release run #388 on `5c9f40b`: `verify` success, `release` success, published as **4.15.0**.

One thing worth recording about this PR's own tests. A case here was originally named as though it proved a same-length data swap collapses the legend, but asserted a different thing; it now performs the swap it claims — replacing 7 rows with 7 differently-labelled rows — so it can only pass if the reset tracks the prop itself, rather than comparing row counts. A count-comparing implementation would leave the new list stuck open, and this case is what catches that.

PR #593, opened 2026-09-08
feat(select): add error state, trigger clear button and pressed feedback

### LINKED ISSUE #414: Select: size variants (sm/md/lg), error state, trigger clear button, pressed state

murdore:
Closes items 2–4 of the Breeze DS dropdown parity list. Item 1 (sizes) was closed separately as a documented class recipe, since the trigger variables were already themeable.

| Addition | What it does |
| --- | --- |
| `error` / `errorMessage` | Red border that outranks hover **and** focus, plus `aria-invalid` and a `role="alert"` hint |
| `clearable` / `onclear` | A clear control that resets the value **without opening the menu** |
| pressed feedback | `#ededed` tint on the trigger and on options while the pointer is down |

All opt-in except the pressed tint, which is an `:active` rule on the existing elements. Four tests assert an unconfigured Select is unchanged.

The error border deliberately outranks the hover and focus/open colours, at equal-or-higher specificity rather than with `!important`. **A field that stops looking invalid the moment the user enters it is invalid exactly when nobody can see it.**

## Five defects the acceptance tests found

Each was reproduced before it was fixed.

**Keyboard activation of clear did the wrong thing.** Enter and Space reached the combobox's own key handler, which opened the menu or selected an option instead of clearing. The button now stops propagation of those keys without stopping their default, so native activation still works.

**The clear button was a focusable descendant of `role="combobox"`**, which expects a single focusable control — the MAJOR finding in review. It is now a **sibling** of the trigger inside a positioned wrapper, placed over the trigger's reserved right padding, so the control looks identical and the combobox is single-focus again. Pinned by an assertion that fails on the old structure:

```ts
expect(await clear.evaluate((el) => el.closest('[role="combobox"]') === null)).toBe(true);
```

**Clearing dropped focus.** The button unmounts the moment the value empties, so activating it left focus on `<body>`. Focus now returns to the trigger — or to the search input in a searchable Select — without opening the menu.

**A whitespace-only `errorMessage`** rendered an empty alert region and pointed `aria-describedby` at it, so a screen reader announced nothing. The check trims, and clearing `error` now removes `aria-invalid` and the description reference together rather than leaving a stale pointer.

**A searchable Select carried the error semantics on the trigger** while focus was in the input. `aria-invalid`/`aria-describedby` are now on whichever control actually takes focus, in both single and multi mode.

Two smaller corrections came with them: `--select-clear-focus-outline` for a themeable focus ring (defaulting to `currentColor`, so it stays visible in the ghost and dark recipes), and the listbox id now comes from `$props.id()` rather than `Math.random()` — stable across SSR hydration, and two instances on one page cannot collide.

## Verification

| Gate | Exit |
| --- | --- |
| lint | 0 |
| check | 0 |
| test:unit | 0 |
| build | 0 |
| test:integration | 0 |

**926 unit tests** including 7 new DOM cases for the error and clear contracts, and **22 Playwright cases** for this feature.

Pointer tests use `hover()` rather than a raw mouse move to a bounding box, and assert `:active` actually matched before reading a colour — a press that missed and a colour that is wrong are different bugs, and without that check the test cannot tell them apart.

**Visual baseline.** `select.png` is regenerated because the clear control moved out of the trigger into the wrapper — that is the whole diff. The before/after pair was reviewed, and the committed baseline then passed a second comparison run **without** `--update-snapshots`, so it is verified rather than merely accepted.

Closes #414


Tara-ag:

Review could not be performed: no PR diff, files, or codebase context was gathered before the time budget expired. A rigorous Yama review requires reading the changed files one at a time, verifying cross-file impact, and gating findings through submit_review. Without that inspection, I cannot certify that the change is safe to merge or identify actionable issues.

src/lib/Select/Select.svelte
```diff
@@ -684,6 +743,62 @@
...
+    height: var(--select-clear-size, 18px);
+    margin: 0;
+    padding: 0;
+    border: none;
+    border-radius: var(--select-clear-border-radius, var(--radius, 4px));
+    background: var(--select-clear-background, transparent);
+    color: var(--select-clear-color, #666666);
+    font-size: var(--select-clear-font-size, 16px);
+    line-height: 1;
+    cursor: pointer;
```
Tara-ag, line 824: 💡 MINOR: Clear button lacks visible focus indicator. The `.select-clear` styles define `:hover` and `:active` states but no `:focus-visible` outline. Keyboard users tabbing to the clear button may get an inconsistent or invisible focus ring. Add a visible focus indicator, e.g.:

```css
.select-clear:focus-visible {
  outline: 2px solid var(--select-clear-focus-outline-color, var(--select-trigger-focus-border-color, #2563eb));
  outline-offset: 2px;
}
```

tests/visual/__screenshots__/select.png
Tara-ag, line 1: 💡 MINOR: Visual baseline screenshot updated without explanation. Please add a note to the PR description describing what changed in this baseline and that the diff was inspected before regeneration, per the project's visual-baseline workflow expectations.

src/lib/Select/Select.svelte
```diff
@@ -56,6 +60,29 @@
...
+  /* Nothing selected means nothing to clear, so the control is not rendered at
+     all rather than rendered inert -- a disabled × that never enables reads as
+     broken. Same for `disabled`, where the whole field is inert already. */
+  const showClearButton = $derived(clearable && !disabled && value.length > 0);
+
+  function handleClear(event: MouseEvent): void {
+    /* The point of a trigger-level clear is that it does NOT open the menu.
+       The trigger's own click handler sits on the ancestor, so without this
+       the click would clear and then immediately open. */
+    event.stopPropagation();
```
Tara-ag, line 80: 💡 MINOR: Whitespace-only `errorMessage` renders an invisible alert and points `aria-describedby` to empty text. `hasErrorMessage` only checks that `errorMessage` is a string with length > 0. A whitespace-only string passes this check, causing an empty error-message div with `role="alert"` to render and `aria-describedby` to reference invisible text. Trim the message before testing, e.g.:

```ts
const hasErrorMessage = $derived(
  error && typeof errorMessage === 'string' && errorMessage.trim().length > 0
);
```

src/lib/Select/Select.svelte
```diff
@@ -482,10 +512,39 @@
         {displayText.length > 0 ? displayText : placeholder}
       </span>
     {/if}
+    {#if showClearButton}
+      <button
+        type="button"
+        class="select-clear"
```
Tara-ag, line 537: ⚠️ MAJOR: Focusable clear button nested inside combobox trigger violates ARIA combobox pattern. The trigger root has `role="combobox"` and contains a focusable `<button class="select-clear">`. ARIA comboboxes are expected to be single focusable controls; embedding another interactive widget inside the combobox can cause screen readers to announce the control inconsistently or fail to expose the clear action. Move the clear button outside the combobox trigger (e.g., as a sibling after the trigger) and wire it to clear the value. If it must stay inside for visual reasons, document that the clear action is primarily pointer-accessible and provide an equivalent keyboard path such as Backspace/Delete to clear the selection.

Tara-ag (changes requested):
Thanks for the PR. The new opt-in features are well-tested and documented, but the clear button placement inside the combobox trigger is an accessibility regression that needs to be addressed before merge. I've also noted two smaller items: whitespace-only error messages and a missing visible focus indicator for the clear button.

src/lib/Select/Select.svelte
murdore: Taken. `.select-clear` now carries a themeable focus ring:

```css
.select-clear:focus-visible {
  outline: var(--select-clear-focus-outline, 2px solid currentColor);
  outline-offset: 2px;
}
```

`currentColor` rather than a hard-coded blue, so it inherits the trigger's colour and stays visible in the ghost and dark recipes instead of matching one theme only. The token is documented in `docs/Select.md`.

The test sets the token to `3px` and asserts the computed width — a check that fails against the browser default (`1px`) rather than passing on it, which is what makes it a test of this rule rather than of the user agent's.

tests/visual/__screenshots__/select.png
murdore: Fair — a regenerated baseline with no explanation is indistinguishable from one accepted blindly, which is the failure mode the expectation exists to prevent.

The PR description now says what changed and why: **the clear control moved out of the trigger and into a positioned wrapper**, so it is no longer a focusable descendant of `role="combobox"` (the MAJOR finding on this PR). The pixels that moved are that control; nothing else on the Select page changed.

How it was regenerated, since "I looked at it" is not much of a claim either:

1. Rendered in the pinned `mcr.microsoft.com/playwright:v1.60.0-noble` container, so the baseline and CI measure the same renderer.
2. The before/after pair was inspected directly, not just accepted.
3. **The committed baseline then passed a second full run *without* `--update-snapshots`** — a comparison it can only pass if it reproduces a fresh render, which is the check that distinguishes a verified baseline from an accepted one.

One thing that pass caught, worth recording: `--update-snapshots` also rewrote `carousel.png` and `hitl.png`, two non-deterministic components this branch never touches. Both were reverted, and the verify run then passed against their committed originals. **`select.png` is the only baseline in this commit.**

src/lib/Select/Select.svelte
murdore: Taken — a whitespace-only message rendered an empty `role="alert"` region and pointed `aria-describedby` at it, so a screen-reader user got an announcement with nothing in it.

`hasErrorMessage` now trims before testing:

```ts
const hasErrorMessage = $derived(
  error && typeof errorMessage === 'string' && errorMessage.trim().length > 0
);
```

Covered by a unit test that renders `errorMessage=" \n "` and asserts there is no alert **and** no `aria-describedby` on the trigger — both halves, since leaving a dangling reference to a removed element is its own defect. The same suite pins that clearing `error` removes `aria-invalid` and the description reference together, rather than leaving a stale pointer behind.

src/lib/Select/Select.svelte
murdore: Taken — this was the right call and the fix is structural, not a documented caveat.

The clear button is no longer a descendant of the combobox. The trigger keeps `role="combobox"` and is now wrapped, with the button rendered as its **sibling** inside a positioned `.select-control` wrapper; it is placed over the trigger's reserved right padding, so the control looks identical while the combobox is once again a single focusable element.

Pinned by an assertion that fails on the old structure rather than by inspection:

```ts
expect(await clear.evaluate((el) => el.closest('[role="combobox"]') === null)).toBe(true);
```

Two behaviours the move made testable, both now covered:

- **Focus does not die with the button.** The clear button unmounts the moment the value empties, so activating it used to leave focus on `<body>`. Focus now returns to the trigger (or to the search input in a searchable Select), without opening the menu.
- **Keyboard activation still does not open the menu** — Enter and Space stop propagation to the combobox's own key handler, which would otherwise select or open.

The pointer path is unchanged, and `clearable` remains opt-in.

murdore:
## Recorded proof — the ARIA fix and the focus fix

Recorded at `203e90b`, the commit on this branch. Each clip was decoded end-to-end with `ffmpeg -xerror` and a frame reviewed before upload.

**The clear button is no longer inside the combobox.** This is the MAJOR finding: the assertion `el.closest('[role="combobox"]') === null` fails against the old structure and passes against this one, while the control looks identical.

https://github.com/user-attachments/assets/0ff130d2-b558-4d71-98c0-40d77c852d29

**Focus survives the button disappearing.** The clear control unmounts the instant the value empties, so activating it used to drop focus to `<body>`. The clip shows the field clearing and focus landing back on the trigger, with the menu still closed.

https://github.com/user-attachments/assets/faa239ca-3c2a-46ed-9834-cb5821615c3f

Two of twenty-two browser cases. The error semantics — whitespace-only messages, the description reference being removed with `error`, and `aria-invalid` following the actually-focused control in searchable variants — are DOM tests, since none of that is visible in a recording.

Local gate log for this commit: lint 0, check 0, unit 0, build 0, integration 0 (926 unit, 649 functional, 1 pre-existing skip).

murdore:
## Runtime proof, re-recorded at the merged commit `50835c2`

Re-recorded so every PR in this batch carries proof taken at the commit that actually landed, rather than at a branch head. The earlier proof on this PR was already valid; this replaces it with the merged-commit set for consistency. 22/22 passing.

### Error state (#414 item 2)

| Behaviour | Recording |
| --- | --- |
| Renders the error border on the trigger | [watch](https://github.com/user-attachments/assets/eee65473-0fd3-4c69-b8c9-382b188e91a5) |
| Keeps the error border while the field is focused | [watch](https://github.com/user-attachments/assets/58097256-bf1b-4417-bb89-34798d8e4a06) |
| Announces the message through an `alert` region | [watch](https://github.com/user-attachments/assets/e0f16976-eaec-44d8-9968-ec4ae00980bd) |
| Wires the message to the combobox with `aria-describedby` | [watch](https://github.com/user-attachments/assets/bfceed92-1d21-4615-b477-a2afb5f836a9) |
| Marks the combobox `aria-invalid` | [watch](https://github.com/user-attachments/assets/76b64b12-83a9-4772-90ef-64ebac085461) |
| Renders no message region when only `error` is set | [watch](https://github.com/user-attachments/assets/4ef034e9-f282-4b5b-9233-2a6f2a687038) |
| A select with no `error` prop keeps its normal border | [watch](https://github.com/user-attachments/assets/20d64cff-7a1d-402a-83d7-7d75aabd5bd9) |

### Trigger clear button (#414 item 3)

| Behaviour | Recording |
| --- | --- |
| Clears the selection | [watch](https://github.com/user-attachments/assets/55bec472-e82c-4ffa-95a7-936656d347a6) |
| Does **not** open the menu | [watch](https://github.com/user-attachments/assets/ed072625-35e7-49e9-ade4-162e970c8564) |
| Enter clears without opening the menu | [watch](https://github.com/user-attachments/assets/3e86a512-7591-4859-916d-93f1f0a16b73) |
| Space clears without opening the menu | [watch](https://github.com/user-attachments/assets/4f969aab-b0a4-4711-939e-3a4d265812bc) |
| Keeps the clear button outside the `combobox` role | [watch](https://github.com/user-attachments/assets/d077f45f-1c57-4d7a-9d49-2438c81f2634) |
| Returns keyboard focus to the trigger after clearing | [watch](https://github.com/user-attachments/assets/6a37b2cf-c3c9-49d6-bd0c-cf6760b1c321) |
| Supports an explicit focus outline token | [watch](https://github.com/user-attachments/assets/98c45906-3819-4ce8-9d56-f31dab02dac3) |
| Fires `onclear` | [watch](https://github.com/user-attachments/assets/9c997540-2b4c-4ddf-b411-cda0577b643c) |
| Disappears once the selection is empty | [watch](https://github.com/user-attachments/assets/2525a848-6a65-44b0-aa5a-d89739153fcc) |
| Absent when nothing is selected to begin with | [watch](https://github.com/user-attachments/assets/4d0a0bd4-5d0d-45c7-9298-72d61bac28d6) |
| Carries an accessible name | [watch](https://github.com/user-attachments/assets/a6d8394b-c837-4574-947a-975d9f7b4916) |
| Absent on a select that did not ask for it | [watch](https://github.com/user-attachments/assets/6a9a63ce-983f-4040-8790-ccde2c9c303f) |

### Pressed state (#414 item 4)

| Behaviour | Recording |
| --- | --- |
| The trigger tints while the pointer is down | [watch](https://github.com/user-attachments/assets/03000b15-2976-42d8-8b82-e16d5fc555d4) |
| The trigger returns to its normal background on release | [watch](https://github.com/user-attachments/assets/1ca4bb66-5883-44cf-b02b-f1996b5585ea) |
| An option tints while the pointer is down | [watch](https://github.com/user-attachments/assets/0a8b45a7-2be4-44ab-9bcf-deb7c82e48b0) |

### Verification

Release run #390 on `50835c2`: `verify` success, `release` success, published as **4.17.0**.

This commit is also the base #595 was replayed onto. Before resolving that PR's conflicts I confirmed `50835c2` was byte-identical to the branch #595 had been built against across every Select path, which is what made taking #595's side safe rather than a guess.

PR #595, opened 2026-09-08
feat(select): add an in-menu search placement

### LINKED ISSUE #415: Select: in-menu search box variant (DS keyboard contract)

murdore:
Adds the in-menu search placement the DS dropdown sheet's keyboard guidelines specify.

**Stacked on #593** (`feat/select-states`) rather than `release`, since both change `Select.svelte`. This PR is the in-menu delta alone — one commit.

## The gap

The DS guidelines put the search box **inside the open menu**, reached by Tab after opening. Select's `searchable` puts the filter input in the trigger and focuses it automatically. That difference is not cosmetic: with the input in the trigger there is no way to open the menu while keeping focus on the combobox.

```svelte
<Select {items} searchable searchPosition="menu" />
```

Defaults to `'trigger'`, so every existing searchable Select is untouched — asserted by a test that the default variant still renders its input in the trigger.

A listbox may only contain options, so the panel carries **no role** in this variant: the search input is a *sibling* of an inner `.select-menu-list` that owns `role="listbox"`, and `aria-controls` points at that list rather than the panel.

## Five defects the acceptance tests found

Each was reproduced before it was fixed.

**Tab always closed the menu**, so the search box could never be reached — the entire contract this variant exists to satisfy.

**A portaled panel is appended to `<body>`**, so it is not the trigger's DOM-order successor: native Tab skipped the whole widget, and Shift+Tab from the search landed outside it. Both directions now move focus explicitly, so the in-flow and portaled variants answer Tab identically.

**Pinned `bottomContent` actions rendered inside the listbox**, where they are invalid children and announce as options — and were unreachable anyway, because forward Tab from the search closed the menu. They now render outside the list, and Tab leaves the menu only once there is nothing left to reach.

**The scrollable option list was its own tab stop in Chromium**, swallowing the Tab that should reach those actions. It is `tabindex="-1"`; options are navigated from the search box via `aria-activedescendant`, so the list never needs to hold focus.

**Selecting an option left focus on `<body>`**, because the search box holding focus was removed with the menu. Focus returns to the trigger.

The panel's key handler is attached imperatively rather than in the template: a keydown handler written there would demand an ARIA role that a wrapper around a search box and a listbox must not claim.

## Verification

| Gate | Exit |
| --- | --- |
| lint | 0 |
| check | 0 |
| test:unit | 0 |
| build | 0 |
| test:integration | 0 |

**18 Playwright cases** for this variant plus the **22 inherited** from the parent branch, run together. They cover opening from the keyboard, Tab in and Shift+Tab back without closing, filtering, arrow navigation, Enter selection and focus return, Escape, the portaled tab cycle in both directions, pinned clear-all/apply actions, filtered select-all preserving values outside the filter, no-results recovery, and query reset on reopen.

**Visual baseline.** `select.png` is regenerated because this adds two demo sections (a portaled in-menu search and a multi-select with pinned actions). The before/after pair was reviewed, and the committed baseline then passed a second comparison run **without** `--update-snapshots`.

Closes #415

## Summary by CodeRabbit

* **New Features**
  * Added an option to place searchable Select inputs in the trigger or dropdown menu.
  * Added support for in-menu search with keyboard navigation, filtering, selection, and portal-compatible behavior.
  * Exposed the placement option for custom elements through the `search-position` attribute.
  * Added examples for in-menu search and multi-select scenarios.

* **Documentation**
  * Documented search placement options and styling variables for in-menu search.


murdore:
## Recorded proof — the keyboard contract

Recorded at `34223b9`, the commit on this branch. Each clip was decoded end-to-end with `ffmpeg -xerror` and a frame reviewed before upload.

**The portaled tab cycle.** A portaled panel lives at the end of `<body>`, so native Tab skipped the whole widget and Shift+Tab from the search landed outside it. The clip shows Tab into the search, Shift+Tab back to the trigger with the menu still open, then Tab out to the next field.

https://github.com/user-attachments/assets/9683fa08-40c0-47bb-a2d8-db5d30b538f1

**Pinned actions are reachable.** They rendered inside the listbox — invalid children, announced as options — and were unreachable anyway, because forward Tab from the search closed the menu. Two Tabs now land on `Clear all`, then `Select (N)`.

https://github.com/user-attachments/assets/e43598e9-2a63-488d-9e77-5e4cc2bb375a

The second one also covers the subtler bug behind it: the scrollable option list was its own tab stop in Chromium and swallowed the Tab that should have reached these buttons.

Two of eighteen cases for this variant, which run alongside the 22 inherited from #593. CI attaches the full report as the `playwright-report` artifact.

Local gate log for this commit: lint 0, check 0, unit 0, build 0, integration 0 (919 unit, 662 functional, 1 pre-existing skip).

murdore:
## Runtime proof, re-recorded at the merged commit `e6236d0`

**Replacing the earlier proof, which was stale in a way that mattered.** It cited `34223b9` — the pre-rebase commit — and unlike a plain rebase, **this branch's content actually changed** on the way in.

This PR was stacked on #593. When #593 merged, replaying this commit onto `release` produced four conflicts (`docs/Select.md`, `Select.svelte`, the demo route, and the visual baseline PNG). They were resolved by taking this branch's side, which is only safe because this branch's tree was a strict superset of #593's — verified two ways before resolving:

- #593's two test files were **byte-identical** here (`git rev-parse` on each blob).
- #593 landed on `release` byte-identically to the branch this was built against (`git diff 203e90b8 50835c2e` over every Select path: empty).

After resolution the delta came out at exactly the **639 insertions / 136 deletions** independently computed as 595-minus-593 beforehand. The proof that both features survived is that #593's suite and this one both pass together — 40 cases — which is what these recordings are.

### In-menu search behaviour, 18/18 at `e6236d0`

| Behaviour | Recording |
| --- | --- |
| Renders the search box inside the menu, not the trigger | [watch](https://github.com/user-attachments/assets/360823f5-e2c8-4120-8649-7dc0f7601c9e) |
| Leaves focus on the trigger when the menu opens | [watch](https://github.com/user-attachments/assets/dd3bae99-ff8c-49d5-ac7f-3b70acba6b3e) |
| Reaches the search box with Tab after opening | [watch](https://github.com/user-attachments/assets/e101959f-8570-490b-9bae-acea554cc5d3) |
| Filters options as the search box is typed into | [watch](https://github.com/user-attachments/assets/fa3fdb3f-887e-44b5-98ea-9bfea17c3272) |
| Keeps arrow-key navigation working from the search box | [watch](https://github.com/user-attachments/assets/d30ff55c-2dcd-40cd-be41-c24b122113d0) |
| Selects the highlighted option with Enter from the search box | [watch](https://github.com/user-attachments/assets/3d1d63fc-5c67-48f8-99e1-3782c857ec25) |
| Closes on Escape from the search box | [watch](https://github.com/user-attachments/assets/5d007475-1fce-4d86-9fbc-32565f221697) |
| Shift+Tab from search returns to the trigger without closing | [watch](https://github.com/user-attachments/assets/7fc57d8e-1284-4720-8f34-07d25bcb682d) |
| Shift+Tab from the trigger leaves and closes the menu | [watch](https://github.com/user-attachments/assets/c53bca6c-0c36-46c2-abfa-8fdaeeda7789) |
| Returns focus after selecting from menu search | [watch](https://github.com/user-attachments/assets/c23016f4-fd9b-45b1-aace-0e105d8d40b3) |
| Portaled menu supports forward and reverse Tab and exit | [watch](https://github.com/user-attachments/assets/b03d267e-6293-47d9-b110-b96fae7b2bbb) |

**ARIA structure** — the part that is easy to get wrong and invisible without assertions:

| Behaviour | Recording |
| --- | --- |
| Keeps the `listbox` role off the panel and on the option list | [watch](https://github.com/user-attachments/assets/afdc8e00-2547-4578-ad8f-3694db4de5d1) |
| The search box is **not** inside the `listbox` element | [watch](https://github.com/user-attachments/assets/b5267d5a-b96b-4107-9b76-da51ae36d137) |
| Bottom actions are outside the listbox and keyboard reachable | [watch](https://github.com/user-attachments/assets/ca784393-5a93-4481-b4af-ddecbd94397d) |

**Edge cases and no-regression:**

| Behaviour | Recording |
| --- | --- |
| Filtered select-all preserves other values; pinned actions apply/clear | [watch](https://github.com/user-attachments/assets/9b4244b4-283d-4170-9fa2-5d9e8770ea43) |
| Recovers from no results and resets the query on reopen | [watch](https://github.com/user-attachments/assets/0062815d-302b-4c62-adb7-c64a182415b5) |
| Preserves the dropdown stacking order | [watch](https://github.com/user-attachments/assets/4862b855-042d-46a2-9e73-ee32fb7e2d3b) |
| A default searchable Select still puts the input in the **trigger** | [watch](https://github.com/user-attachments/assets/b125d3ef-6211-445d-b2b7-3741eb99dbb8) |

### Verification at `e6236d0`

```
pnpm run check      exit 0     933 files, 0 errors
pnpm run lint       exit 0
pnpm run test:unit  exit 0     49 files, 993 tests passed
playwright          exit 0     40 passed across this suite and #593's
```

Release run #391 on the merged commit: `verify` success (which is what independently confirms the hand-resolved visual baseline), `release` success, published as **4.18.0**.

PR #597, opened 2026-09-09
feat(pie-chart): register sui-pie-chart as a web component

murdore:
Builds the `sui-pie-chart` wrapper that `docs/PieChart.md` has described for several releases without it ever existing. Split out of #592 at review request rather than smuggled into a legend feature.

**Stacked on `feat/piechart-legend-right` (#592), not `release`.** The parity checker requires a wrapper to declare *every* prop its component accepts, and #592 adds four legend props. A wrapper built on `release` would go green today and break `check:wc-parity` the moment #592 lands. Merge #592 first; the base retargets to `release` automatically.

## What was actually wrong

`src/wc/index.ts` is a list of side-effecting imports, one per wrapper file. PieChart had no wrapper file, so it was never imported, so `customElements.define('sui-pie-chart', …)` never ran. `<sui-pie-chart>` in a consumer's page was an unknown element: no upgrade, no props, no render. Anyone following the documented snippet got a blank space and no error.

## What is here

**All 30 props declared.** This is enforced, not asserted — `scripts/wc-parity/prop-parity.test.ts` fails a wrapper that misses one. Verified by removing a declaration:

```
× PieChart.wc.svelte declares every prop
AssertionError: PieChart.wc.svelte is missing: legendMaxItems
```

**Attribute coercion.** Scalars are kebab-case attributes with declared types, so `legend-max-items="3"` arrives as the number `3` rather than the string `"3"` — which matters, since the component floors a non-integer cap and ignores a non-finite one, so a string would surface as either every row or none. Booleans are presence-based and observed, so removing `show-legend` turns the legend off.

**Arrays, functions and snippets stay JS properties**, because they cannot cross the HTML-attribute boundary. `data`, `valueFormat`, `onsliceclick`, `onlegendmore`, `onchartready`, `onslicehover`.

**`center` and `empty` are slots**, since a plain-HTML consumer cannot construct a Svelte snippet — and they are passed *only when content is actually slotted*. PieChart gates its own defaults on `typeof snippet === 'function'`, so a wrapper that defined them unconditionally would replace the built-in empty state with a blank box and force an empty centre onto every donut.

`tooltipSnippet` deliberately gets no slot: it is called with `(slice, index)`, and a slot cannot receive those arguments.

## The bug this shook out

The first version declared the snippets at the top level of the template. That compiles to:

```js
const emptySlot = ($$anchor) => {          // hoisted to MODULE scope
  $.slot(node_1, $$props, 'empty', {}, null);   // $$props lives in the component fn
};
```

A top-level `{#snippet}` is hoisted, and the `<slot>` inside it references `$$props`, which only exists inside the component function. It threw `$$props is not defined` **at render time** — surfacing as a silently empty shadow root, not a build error, and only when the snippet was actually rendered. So it passed with data present and failed only on an empty dataset with slotted content.

Declaring each snippet inside `<PieChart>` keeps it in component scope, which is what every other wrapper in the directory already does. Confirmed by compiling: `hoisted snippets before component fn: 0`.

Worth flagging separately: the same trap catches anyone who writes a top-level slot snippet in a wrapper, and nothing in the repo would have caught it — the failure is silent and only at runtime.

## Verification

- `lint` 0 · `check` 0 · `build` 0 · unit **935 passed**
- `scripts/wc-parity` (source): **162 passed**, red/green verified
- Browser, against the built `dist-wc/index.js`: **19 passed** — the 9 new `sui-pie-chart` cases plus the existing `wc-prop-parity` and `wc-event-casing-parity` suites, whose `every wrapper in the directory is registered by the bundle` case now covers this tag.

The new cases cover registration, per-slice rendering from a JS-assigned array, kebab attribute coercion, boolean-attribute removal, in-place expansion, the empty-state default surviving an unslotted element, slotted `empty`, slotted `center` in a donut, and `onsliceclick` reaching the component.

## Not included: a pixel baseline

`tests/visual/` screenshots demo routes discovered from `src/routes/components/`, and no route in this repo renders a custom element — all 86 wrappers have zero pixel baselines. Adding one route for this one wrapper would invent a convention nothing else follows and add a page to the docs site as a side effect. The rendered output is the same `PieChart` already covered by `pie-chart.png`.

Happy to add a `wc` demo route and baseline as a follow-up if the coverage is wanted — as a deliberate decision for all the wrappers rather than a side effect of this one.

Refs #417

### DIFF
--- a/docs/PieChart.md
+++ b/docs/PieChart.md
@@ -294,6 +294,47 @@ Thresholds: ≥ 1 Cr (10 million) → `"1.5Cr"`, ≥ 1 L (100 thousand) → `"2.
 
 ## Web Component
 
-PieChart is currently available through the Svelte entry point only. The web-component bundle does not register `sui-pie-chart`; importing it will not create a functioning chart element. Use `PieChart` from `@juspay/svelte-ui-components` for the APIs documented above.
+`sui-pie-chart` is registered by the web-component bundle:
 
-> **Correction.** Earlier revisions of this page documented `<sui-pie-chart>` with kebab-case attributes and JS-assigned function props. **That wrapper was never built** — there is no `PieChart.wc.svelte` and no `customElements.define('sui-pie-chart', …)`, so the element never upgraded and the markup rendered nothing. If you followed that section and saw an empty space, this is why. Nothing was withdrawn here: the documentation described a component that did not exist, and now says so. Adding a real wrapper is tracked separately.
+```html
+<script type="module" src="@juspay/svelte-ui-components/wc"></script>
+
+<sui-pie-chart show-legend legend-show-values legend-position="right" legend-max-items="5">
+</sui-pie-chart>
+
+<script>
+  document.querySelector('sui-pie-chart').data = [
+    { label: 'UPI', value: 62 },
+    { label: 'Cards', value: 21 },
+    { label: 'Netbanking', value: 9 }
+  ];
+</script>
+```
+
+**Scalar props are kebab-case attributes**, coerced to their declared type: `inner-radius`, `pad-angle`, `show-labels`, `show-values`, `label-position`, `show-legend`, `start-angle`, `aspect-ratio`, `max-height`, `min-height`, `semi-circle`, `legend-show-values`, `legend-position`, `legend-max-items`, `percent-decimals`, `highlighted-index`, `change-percentage`, `change-invert-colors`, `test-id`, `classes`. Booleans are presence-based, so `show-legend` is on and removing the attribute turns it off.
+
+**Everything else is a JS property**, because arrays, objects and functions cannot cross the HTML-attribute boundary:
+
+```js
+const chart = document.querySelector('sui-pie-chart');
+chart.data = slices;
+chart.valueFormat = (value) => `₹${value}`;
+chart.onsliceclick = ({ index, slice }) => console.log(index, slice.label);
+chart.onlegendmore = () => openBreakdownDrawer();
+chart.onchartready = (api) => narrator.attach(api);
+```
+
+**`center` and `empty` are slots**, not properties — a plain-HTML consumer cannot construct a Svelte snippet:
+
+```html
+<sui-pie-chart inner-radius="0.6">
+  <div slot="center"><strong>₹4.2Cr</strong></div>
+  <span slot="empty">No transactions in this period</span>
+</sui-pie-chart>
+```
+
+Leaving a slot empty keeps the component's own default, so an unslotted element still renders the built-in empty state rather than a blank box.
+
+`tooltipSnippet` has no slot equivalent. It is called with `(slice, index)`, and a slot cannot receive those arguments, so it stays available only to Svelte consumers.
+
+> **Correction.** Revisions of this page before `4.12` documented `<sui-pie-chart>` while no wrapper existed — there was no `PieChart.wc.svelte` and no `customElements.define('sui-pie-chart', …)`, so the element never upgraded and the markup rendered nothing. If you followed that section and saw an empty space, this is why. The wrapper described above is the real one.
--- a/src/wc/components/PieChart.wc.svelte
+++ b/src/wc/components/PieChart.wc.svelte
@@ -1,0 +1,102 @@
+<svelte:options
+  customElement={{
+    tag: 'sui-pie-chart',
+    shadow: 'open',
+    props: {
+      // Complex props (arrays / objects / functions / snippets) cannot cross the
+      // HTML-attribute boundary, so they are exposed as JS properties only:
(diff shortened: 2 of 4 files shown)


murdore:
## Recorded proof at `0c3c668`

Nine browser cases run against the built `dist-wc/index.js`, so what is recorded is the shipped bundle registering and rendering the element — not the Svelte component standing in for it.

**Attribute coercion.** `legend-max-items="3"` with `legend-position="right"`: three rows and a `+3 more` control. If the string `"3"` reached the component uncoerced, the cap would be non-finite and every row would render.

https://github.com/user-attachments/assets/d730e659-d2ae-49af-8727-b3301c5e2567

**Slotted `center` in a donut.** `<strong slot="center">62%</strong>` projected into the centre of the ring.

https://github.com/user-attachments/assets/3f5a11c2-fc22-4b12-a57b-2a3258a59798

**Slotted `empty`.** The consumer's own empty message on an empty dataset — and the companion case proves the *unslotted* element still takes the component's built-in path rather than rendering a blank box.

https://github.com/user-attachments/assets/8917d749-fe60-4269-b6ae-a03b812c320d

### An audit correction worth recording

The first recording of this suite was worthless as proof and I nearly attached it. `loadBundle` navigates to `/` for a same-origin page and appended the element to `document.body` — which parks it below a full-length demo route, far outside the viewport. Every clip showed the top of the **Card** demo page while the assertions passed against an element nobody could see.

The tests were green and the videos decoded; they simply did not depict the thing they claimed to prove. Fixed by clearing the demo markup in `loadBundle` so the element under test is the only thing on the page, which also let the assertions be strengthened from `toHaveCount` to `toBeVisible` — the element is now required to actually render on screen, which is what the recording shows.

### Gates at this commit

`lint` 0 · `check` 0 · `build` 0 · unit **935 passed** · `scripts/wc-parity` **162 passed** · browser **19 passed** (the 9 above plus `wc-prop-parity` and `wc-event-casing-parity`, whose "every wrapper in the directory is registered by the bundle" case now covers `sui-pie-chart`).

Every clip was decoded end-to-end with `ffmpeg -xerror` and a frame extracted and reviewed before upload; traces are valid zips carrying a `.trace` entry.

murdore:
## Runtime proof, re-recorded at the merged commit `8e2bbc4`

**Replacing the earlier proof on this PR, which was stale.** It cited `0c3c668` — the pre-rebase commit. This PR was rebased onto `release` before merging (git dropped #592's commit as already-applied, leaving a single commit), so what actually landed is `8e2bbc4`. These recordings are from that commit, so the evidence and the merged code now match.

Every touch point of `<sui-pie-chart>`, 9/9 passing:

| What it proves | Recording |
| --- | --- |
| Registers as a custom element | [watch](https://github.com/user-attachments/assets/72478edd-ff11-490e-9060-f5295bcf6fcb) |
| Renders one arc per slice from a JS-assigned `data` array | [watch](https://github.com/user-attachments/assets/6eefa0ed-63c4-4bd5-8bcf-52e6958a4f9f) |
| Coerces kebab-case number attributes into typed props | [watch](https://github.com/user-attachments/assets/2b83094a-6152-47d3-9dab-6d2a1c2b79c6) |
| Observes a boolean attribute being **removed** | [watch](https://github.com/user-attachments/assets/5abbd4a1-07b2-4872-ae1f-c637111823ff) |
| Expands in place when the built-in control is activated | [watch](https://github.com/user-attachments/assets/382a8ab2-e90a-4e31-8322-5889f3f5a3ed) |
| Keeps the built-in empty state when nothing is slotted | [watch](https://github.com/user-attachments/assets/8e99b301-70f7-42cb-a46b-08432146ff13) |
| Uses slotted `empty` content when provided | [watch](https://github.com/user-attachments/assets/c6ebcec7-2689-41e3-aa0f-0e6ce4482e8b) |
| Renders slotted `center` content inside a donut | [watch](https://github.com/user-attachments/assets/0117362e-2adb-4eac-afde-916d35cb14d7) |
| Forwards the slice-click handler assigned as a property | [watch](https://github.com/user-attachments/assets/d32b0cc6-5950-443a-8b32-d53af0d2941f) |

### Shipped and verified on the registry

`sui-pie-chart` is present in the published `dist-wc/index.js` of **4.16.0** — checked by unpacking the tarball from npm, not inferred from the merge. Release run #389: `verify` success, `release` success.

Two defects in earlier revisions of this work, recorded so they are not repeated:

- The first videos showed an unrelated demo page, because the element under test was appended below a full-length route and sat outside the viewport — which made `toBeVisible` meaningless. `loadBundle` now clears the demo markup so the element under test is the only thing on the page.
- The empty-state case asserted only that `.chart-empty` was absent, which also passes when *nothing* renders. It now also asserts the `svg` is present, so it cannot pass on a blank shadow root.

Issue #576, opened 2026-09-07
attrs is Record<string, string> on both Card and Pill; narrowing it should happen once, in a major


murdore:
Raised as a review suggestion on #567 and deferred there deliberately, so it is not lost.

## Today

Both `Card` and `Pill` type their escape-hatch `attrs` prop as `Record<string, string>`:

- `src/lib/Card/properties.ts`
- `src/lib/Pill/properties.ts`

Both components spread `attrs` first and their own managed attributes after, so a consumer cannot actually clobber `class`, `role`, `tabindex`, `data-pw`, `title`, `testID`, `onclick`, `onkeydown` or `aria-disabled` at runtime — the spread order handles it.

## The suggestion

Type it so the compiler rejects those keys up front rather than silently ignoring them:

```ts
Omit<HTMLAttributes<HTMLDivElement>, 'class' | 'role' | 'tabindex' | 'data-pw' | 'title' | 'testID' | 'onclick' | 'onkeydown' | 'aria-disabled'>
```

The benefit is a real one: a consumer who writes `attrs={{ class: 'mine' }}` expecting it to win currently gets silence, and would instead get a compile error naming the problem.

## Why it is not in #567

1. **It must not land on one component only.** Two components sharing a prop name with different contracts is worse than one wide type — code that compiles against `Card` would fail against `Pill` for a reason the API does not reveal.
2. **It is breaking.** `HTMLAttributes` types event handlers as functions, so `Record<string, string>` and the narrowed type are incompatible in both directions. Any consumer already passing `attrs` needs code changes, which is a major, not a `feat`.

## Proposal

Do it as one change across `Card` and `Pill` (and any component that grows an `attrs` prop before then) in the next major, with a codemod entry if the shape of the break allows one.

murdore:
Scoping this for whenever the next major happens, because the shape of the change turns out to be meaningfully different from what this issue assumed — partly because of work that landed after it was filed.

## The suggested type no longer fits either component

This issue proposes:

```ts
Omit<HTMLAttributes<HTMLDivElement>, 'class' | 'role' | 'tabindex' | 'data-pw' | 'title' | 'testID' | 'onclick' | 'onkeydown' | 'aria-disabled'>
```

`HTMLDivElement` is wrong for **both** components, because neither has a `<div>`-only root any more:

| component | `as` union | root elements |
| --- | --- | --- |
[...]

PR #599, opened 2026-09-10
feat(card,pill): warn in dev when an attrs entry is discarded

murdore:
4.19.0 deprecated the `attrs` keys Card and Pill overwrite, which turns the
silent discard into a strikethrough for anyone reading TypeScript in an editor
— and does nothing at all for anyone else. A plain-JS consumer has no types, and
`<sui-card>`/`<sui-pill>` callers driving the web components have none either.
Those are exactly the callers the spread order was always protecting, so they
were the ones still getting silence.

Dev builds now log, once per mounted instance, which managed key was found in
`attrs` and what owns it:

  [svelte-ui-components] Card: the "class" entry in `attrs` is discarded,
  because Card writes its own "class" after spreading `attrs` — use the
  `classes` prop.

`src/lib/attrs-discard.ts` holds the two key→guidance maps and a pure
`discardedAttrWarnings`; the components decide whether to report. The maps
repeat what the `@deprecated` tags say, so the test pins each one against its
`*StrictAttrs` type — indexing the strict type by the runtime map's keys is
undefined-only exactly when every key is really managed, so a key that drifted
out of the type fails `pnpm check`. Verified by mutation: adding an unmanaged
key to the Card map does fail it.

Three repo rules shaped the implementation rather than the other way round:

- `$effect` is banned, so the check runs once at component init. That is also
  the right frequency: one warning per call site is what a consumer acts on, and
  re-warning on every reactive update would bury it.
- The built-in `Set` is banned in favour of `SvelteSet`. Warning at init removed
  the need to dedupe at all, so neither is used.
- The `undefined` keyword is banned, so `attrs` is the optional trailing
  parameter and an absent one is simply not passed.

Only own properties count: an object built on a prototype that happens to carry
`class` never asked for that attribute to be spread, and warning about it would
be noise a consumer cannot act on.

Guarded on `import.meta.env?.DEV === true`, read defensively because bundlers
that do not define `import.meta.env` should stay quiet rather than throw. It
compiles out of production builds, so nothing ships to end users.

Nothing about the rendered output changes; #576 is still open for the major that
makes `attrs` itself reject these keys.

Verified on this commit: check 0, lint 0, unit 0. 1021 unit tests, of which 14
are new, and 25 Card/Pill integration cases. The render tests assert the DOM
alongside the warning — a test that only checked the message could pass while
`attrs` had started winning instead — and were mutation-checked by stripping the
warning from Card, which fails them.

The full integration suite shows 3 failures locally on this branch and 3 on
`origin/release` with none of this applied, in different tests each run:
pre-existing flakiness in tooltip/menu placement and table cells, masked in CI
by `retries: process.env.CI ? 2 : 0`. Not introduced here, and not touched here.

Refs #576

## Summary by CodeRabbit

* **New Features**
  * Card and Pill now provide development-time warnings when managed attributes are passed through `attrs` and discarded.
  * Warnings identify the affected attribute and recommended alternative, helping plain JavaScript and web-component users resolve issues.
  * Warnings appear once per component instance and are omitted from production builds.

* **Documentation**
  * Updated Card and Pill documentation to describe the new development warnings.

### DIFF
--- a/docs/Card.md
+++ b/docs/Card.md
@@ -224,6 +224,8 @@ into per object:
 </script>
 ```
 
+Types only reach consumers who have them. In **dev builds**, Card also logs a `console.warn` naming any managed key it found in `attrs` and what owns it, so a plain-JS caller — or anyone driving `<sui-card>` as a web component, where there are no types at all — sees the discard too. It fires once per mounted instance and is compiled out of production builds.
+
 `CardStrictAttrs` also types native attributes properly — `onfocus` is a function, not a
 string — and rejects managed keys reached through a predeclared variable, not just an
 inline literal. It intentionally differs from `PillStrictAttrs`: Card writes no `aria-*`
--- a/docs/Pill.md
+++ b/docs/Pill.md
@@ -241,6 +241,8 @@ into per object:
 </script>
 ```
 
+Types only reach consumers who have them. In **dev builds**, Pill also logs a `console.warn` naming any managed key it found in `attrs` and what owns it, so a plain-JS caller — or anyone driving `<sui-pill>` as a web component, where there are no types at all — sees the discard too. It fires once per mounted instance and is compiled out of production builds.
+
 It intentionally differs from `CardStrictAttrs`: Pill writes no `style`, so `style` is
 usable on Pill and rejected on Card, while Card writes no `aria-*` state. Forbidding a key
 a component never touches would be a fake restriction.
--- a/src/lib/Card/Card.svelte
+++ b/src/lib/Card/Card.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import type { CardProperties } from './properties';
+  import { CARD_MANAGED_ATTRS, discardedAttrWarnings } from '../attrs-discard';
 
   let {
     children,
@@ -21,6 +22,27 @@
     attrs,
     as
   }: CardProperties = $props();
+
+  /*
+   * The runtime half of the `attrs` deprecation (#576). The `@deprecated`
+   * markers on the managed keys only reach someone reading TypeScript; a
+   * plain-JS caller, or anyone driving this as a web component, has no types
+   * and would otherwise still get silence.
+   *
+   * Dev builds only: this is an authoring aid, and warning on every production
+   * render would be noise nobody asked for. `import.meta.env` is absent under
+   * bundlers that do not define it, so the guard reads it defensively and stays
+   * quiet there rather than throwing.
+   *
+   * Deliberately at init rather than in an `$effect`, which this repo bans:
+   * once per mounted instance is what a consumer needs to act on, and it costs
+   * nothing on subsequent updates.
+   */
+  if (import.meta.env?.DEV === true) {
+    for (const message of discardedAttrWarnings('Card', CARD_MANAGED_ATTRS, attrs)) {
+      console.warn(message);
+    }
+  }
 
   // `as`, when given, picks the rendered tag outright. Omitted, the root keeps
   // its existing rule: an `<a>` exactly when `href` is set, a `<div>`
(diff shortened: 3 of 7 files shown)


murdore:
## Runtime proof at `0c1e4ba`

This PR adds no rendered output, so what needs proving is that the attributes it warns about are still the ones actually discarded, and that nothing about Card or Pill changed. 9/9 passing.

**The two that carry it** — the warning names exactly the keys these prove are overwritten:

| What it proves | Recording |
| --- | --- |
| `attrs` cannot override what **Card** manages (`data-pw`, `class`, `role`) | [watch](https://github.com/user-attachments/assets/c50179ec-bcc3-4d35-88bd-553726cf5237) |
| `attrs` cannot override what **Pill** manages (`data-pw`, `class`, `role`, `tabindex`) | [watch](https://github.com/user-attachments/assets/9bdce0e0-dccd-4508-b402-790230820340) |

**Nothing else moved:**

| Case | Recording |
| --- | --- |
| Card: `attrs` spreads arbitrary `data-*` onto the root | [watch](https://github.com/user-attachments/assets/5ba62fc7-978b-4983-b9f3-17ef1e0a9af0) |
| Pill: `attrs` spreads arbitrary `data-*` onto the root | [watch](https://github.com/user-attachments/assets/9da1a108-72fc-4004-b1e2-7220de1ee465) |
| Card: omitting `attrs`/`as` leaves the default `div` root unaffected | [watch](https://github.com/user-attachments/assets/c51cff46-9f12-41f1-b52f-8cd18ff3160d) |
| Pill: omitting `attrs` leaves the pill root unaffected | [watch](https://github.com/user-attachments/assets/a4b95c25-a71b-4d63-98e5-3484c19eca12) |
| Card: `as="figure"` renders a real `<figure>` | [watch](https://github.com/user-attachments/assets/5db631c0-188c-4752-a071-1e3cab9818d6) |
| Card: `as` overrides the href-derived default | [watch](https://github.com/user-attachments/assets/7dbda633-e707-47ab-9c6f-593c77511d56) |
| Card: `as="a"` without `href` keeps the keyboard shim | [watch](https://github.com/user-attachments/assets/36e2cb92-cbc9-464a-a75e-9e4e3779f7c3) |

### The unit tests are the real coverage here

The warning is a dev-build console message, which a Playwright video cannot show. Two suites cover it instead, and **both were mutation-checked rather than trusted**:

- `attrs-discard-runtime.test.ts` renders the real components and reads `console.warn`. It asserts the **DOM alongside the message** on purpose — a test checking only the message would keep passing if `attrs` ever started winning instead. Replacing Card's `console.warn` with a no-op fails it.
- `attrs-discard.test.ts` pins each runtime key→guidance map against its `*StrictAttrs` type. Indexing the strict type by the map's keys is undefined-only when every key is genuinely managed, so a key drifting out of the type fails `pnpm check`. Adding an unmanaged key (`hidden`) to the Card map does fail it.

### Verification at `0c1e4ba`

```
pnpm run check      exit 0
pnpm run lint       exit 0
pnpm run test:unit  exit 0     52 files, 1021 tests (14 new)
playwright          exit 0     25 Card/Pill integration cases
```

### One finding that is not about this PR

The **full** integration suite fails **3 of 69 locally on `origin/release` itself**, with none of this branch applied — and different tests each run (tooltip/menu placement, table interactive cells, status default icon). That is pre-existing flakiness, masked in CI by `retries: process.env.CI ? 2 : 0`. Not introduced here and not touched here; being investigated separately.

Tara-ag (approved):
Approved. The change is a clean, non-breaking follow-up to the 4.19.0 `attrs` deprecation: it adds a dev-only runtime warning for managed keys passed through `attrs`, covering plain-JS and web-component consumers who don't see TypeScript strikethroughs. The implementation correctly guards `import.meta.env` access, keeps warnings out of production, avoids `$effect` (which the repo bans), and the managed-key lists match what each component actually writes. Tests and docs are updated accordingly. No public API, export, prop, or CSS custom property changes were introduced.

PR #600, opened 2026-09-10
fix(tests): wait for hydration before driving demo pages

murdore:
The functional suite failed intermittently: a full local run at the default
worker count failed 8 of 713, a run of the same commit at `--workers=2` failed 0
of 713, and the failing tests moved from run to run. CI never showed it because
`retries: process.env.CI ? 2 : 0` re-ran them.

Root cause, confirmed rather than inferred. Every demo page is server-rendered,
so a control is present, visible, stable and enabled — passing every
actionability check Playwright makes — before hydration has attached its
handler. Playwright has no check for "a listener exists", so a click arriving in
that window is delivered to inert markup and silently does nothing; the
assertion after it then fails as a timeout naming the missing element, which
says nothing about the real cause. Recording `video` and `trace` on all 713
tests makes the window wide enough to hit whenever workers compete.

The confirmation was a probe that delays the module scripts, making the race
deterministic: the trigger passes `toBeVisible`, the click lands, and the
dropdown never appears. That probe is kept as
`tests/hydration-race.spec.ts`, alongside its inverse — the same delayed page
driven through the new helper, where the click works. Together they fail if
either half of this regresses.

`src/routes/+layout.svelte` now sets `document.documentElement.dataset.hydrated`
in `onMount`; a parent's `onMount` runs after its children have mounted, so the
root layout is the app as a whole reporting it is ready. `gotoHydrated` in
`tests/support/hydrated.ts` navigates and waits for it, and replaces
`page.goto` at 579 call sites across 149 spec files. The three remaining raw
`goto` calls are deliberate: the race probe needs an unhydrated page, the helper
itself, and the visual suite's templated navigation.

Separately, `typewriter-text-reduced-motion.test.ts` sampled its mutation count
from the test process on either side of an `emulateMedia` CDP round-trip, so
every character typed while that call was in flight counted against a bound of
6. The count is now recorded inside the page, when the media query actually
flips, which takes the round-trip out of the measured window and lets the bound
tighten to 2. Writing that exposed a second defect in my own first attempt:
`readTiming` projects a narrower shape, so the new field never reached the
assertion, and a `not.toBeNull()` guard passed on `undefined` while the test
measured the whole reveal. The guard now asserts a real number, which is what
catches a dropped projection.

Verified on this commit: check 0, lint 0, unit 0, integration 0. The full suite
at the default worker count that was failing 8 now passes 715 of 715.

The same test also sampled `textContent` from the test process to find the
moment typing had started. The whole string reveals in roughly 450ms and every
`expect.poll` iteration costs a round-trip, so on a loaded machine those spent
most of the window and the flip landed after typing had finished -- failing the
test's own `remaining > 12` guard, which exists to prove it caught the reveal
mid-stream. That wait now happens in the page via `waitForFunction`, so it ends
on the first revealed character and the only round-trip left in the critical
path is the `emulateMedia` call. The guard reads the in-page count at the flip
too, rather than a sample taken a round-trip earlier.

Fixing it in the demo app was tried first and abandoned: the site is
prerendered, so a query parameter cannot reach the server-rendered output, and
slowing the shared fixture outright would break the pacing assertions in
`typewriter-text-pacing-progress-render.test.ts` and change the visual baseline.

Verified on this commit: check 0, lint 0, unit 0, integration 0. The full suite
at the default worker count that was failing 8 now passes 715 of 715, and the
reduced-motion file passes 48 of 48 at `--repeat-each=16 --workers=14`, a load
at which it previously failed repeatedly.

### DIFF
--- a/tests/support/hydrated.ts
+++ b/tests/support/hydrated.ts
@@ -1,0 +1,35 @@
+import type { Page, Response } from '@playwright/test';
+
+/**
+ * Navigates and waits until the app is interactive.
+ *
+ * `page.goto` resolves on the navigation, not on hydration, and every demo page
+ * is server-rendered. A control is therefore present, visible and clickable
+ * while its handler does not yet exist -- and Playwright cannot check for a
+ * listener, so the click is delivered to inert markup and lost. The assertion
+ * that follows fails as an unexplained timeout.
+ *
+ * The marker is set in `src/routes/+layout.svelte`, whose `onMount` runs after
+ * its children have mounted.
+ *
+ * It does NOT mean custom elements are ready, and must not be used that way. A
+ * custom element has a second readiness step -- `customElements.define` runs,
+ * then matching elements upgrade -- and the marker is strictly earlier than it:
+ * measured on `/`, `data-hydrated` was set at ~755ms while
+ * `customElements.get('sui-status')` was still undefined and no element had
+ * upgraded. On top of that, `dist-wc` is a self-contained bundle rather than a
+ * route of the demo site, so a spec injects it with `addScriptTag` after
+ * navigating, and nothing is defined at marker time by construction.
+ *
+ * A spec driving a custom element therefore has to gate on the element itself,
+ * as every one of them already does:
+ *
+ *   await gotoHydrated(page, '/');
+ *   await page.addScriptTag({ path: 'dist-wc/index.js', type: 'module' });
+ *   await page.waitForFunction(() => Boolean(customElements.get('sui-status')));
+ */
+export const gotoHydrated = async (page: Page, path: string): Promise<Response | null> => {
+  const response = await page.goto(path);
+  await page.waitForFunction(() => document.documentElement.dataset.hydrated === 'true');
+  return response;
+};
--- a/tests/typewriter-text-reduced-motion.test.ts
+++ b/tests/typewriter-text-reduced-motion.test.ts
@@ -1,4 +1,5 @@
 import { expect, test, type Page } from '@playwright/test';
+import { gotoHydrated } from './support/hydrated';
 
 // `ChatController` bypasses its character-by-character reveal entirely when the user
 // asks for reduced motion (`prefersReducedMotion()`, consulted by `feed()`), but
@@ -27,7 +28,12 @@ const BASELINE_TEST_ID = 'typewriter-default-pacing';
 const PROGRESS_TEXT = 'Progress reporting keeps a scroll container pinned to the newest character.';
 const PROGRESS_TEST_ID = 'typewriter-progress-demo';
 
-type RevealTiming = { firstAt: number | null; lastAt: number | null; mutations: number };
+type RevealTiming = {
+  firstAt: number | null;
+  lastAt: number | null;
+  mutations: number;
+  mutationsAtReduceFlip: number | null;
+};
 
 const installRevealObserver = async (
   page: Page,
@@ -36,8 +42,27 @@ const installRevealObserver = async (
 ): Promise<void> => {
   await page.addInitScript(
     ({ testId, windowKey }: { testId: string; windowKey: string }) => {
-      const timing: RevealTiming = { firstAt: null, lastAt: null, mutations: 0 };
+      const timing: RevealTiming = {
+        firstAt: null,
+        lastAt: null,
+        mutations: 0,
+        mutationsAtReduceFlip: null
+      };
       (window as unknown as Record<string, RevealTiming>)[windowKey] = timing;
+      /*
+       * Recorded in the page, at the instant the preference actually flips.
+       * Sampling the count from the test process instead would put the
+       * `emulateMedia` CDP round-trip inside the measured window, and every
+       * character that types while that round-trip is in flight would count
+       * against the bound -- which is why this assertion failed under load and
+       * passed in isolation.
+       */
+      const reduceQuery = window.matchMedia('(prefers-reduced-motion: reduce)');
+      reduceQuery.addEventListener('change', (event) => {
+        if (event.matches && timing.mutationsAtReduceFlip === null) {
+          timing.mutationsAtReduceFlip = timing.mutations;
+        }
+      });
       const attach = () => {
         const element = document.querySelector(`[data-pw="${testId}"]`);
         if (!element) {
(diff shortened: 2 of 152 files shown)


murdore:
## Proof at `baf1320`

### The bug and the fix, side by side

These two are the whole argument. Both delay the module scripts, which turns a load-dependent race into a certainty:

| | Recording |
| --- | --- |
| **The bug.** The trigger passes `toBeVisible`, the click is delivered — and the dropdown never appears, because no handler exists yet | [watch](https://github.com/user-attachments/assets/7159d611-e499-4327-bf35-4cc85699827d) |
| **The fix.** Same delayed page, navigated with `gotoHydrated`; the click opens the menu | [watch](https://github.com/user-attachments/assets/a2ea08f6-3ca1-45b1-b409-461f976bf6fe) |

### The tests that were actually failing

`menu-placement` was one of the eight. All 5 now pass:

| Case | Recording |
| --- | --- |
| Auto placement flips to top-right at the viewport corner *(this was failing)* | [watch](https://github.com/user-attachments/assets/4c842e89-64e7-417f-96bb-5e4d2850a6e6) |
| Auto placement stays at the default corner when there is room *(this was failing)* | [watch](https://github.com/user-attachments/assets/4b9dffda-98a6-43ce-8f6d-d5e9fa8feed2) |
| Default placement keeps the panel left-aligned below the trigger | [watch](https://github.com/user-attachments/assets/0c81a792-4e44-4858-8ed1-2658bc0b2fd3) |
| `bottom-right` anchors the panel's right edge to the trigger | [watch](https://github.com/user-attachments/assets/5722324a-192a-4998-9901-352715bf9e06) |
| Existing consumers are unaffected | [watch](https://github.com/user-attachments/assets/58700dce-8141-4b19-bdd1-7f1265f4caaf) |

### The second, unrelated cause

`typewriter-text-reduced-motion` failed for its own reason — a measurement spanning an `emulateMedia` CDP round-trip, plus `expect.poll` round-trips eating a ~450ms window:

| Case | Recording |
| --- | --- |
| A mid-stream switch to reduced motion stops the typing immediately | [watch](https://github.com/user-attachments/assets/8975f691-c1a2-4853-aed7-f154ccece6b0) |
| The control: with motion allowed it still types character by character | [watch](https://github.com/user-attachments/assets/c0ff0853-225e-4034-a62f-128a981053bf) |
| Reduced motion reveals the whole string at once | [watch](https://github.com/user-attachments/assets/8421e84b-40f8-45f0-96d0-10b357acf805) |

### Numbers

| Condition | Result |
| --- | --- |
| Before, default workers | **8 failed** / 713 |
| Before, `--workers=2` | 0 failed / 713 |
| Before, reduced-motion at `--repeat-each=12 --workers=9` | failed repeatedly |
| **After, default workers** | **715 passed, 0 failed** |
| After, reduced-motion at `--repeat-each=16 --workers=14` | 48 / 48 |

```
pnpm run check      exit 0
pnpm run lint       exit 0
pnpm run test:unit  exit 0     1007 tests
playwright          exit 0     715 / 715
```

### Two things worth flagging for review

- **The diff is large but uniform.** 579 `page.goto` call sites across 149 files become `gotoHydrated(page, …)`. Three raw `goto` calls are deliberately left: the race probe needs an *unhydrated* page, the helper itself, and the visual suite's templated navigation.
- **A wrong turn worth recording.** I first tried widening the typing window with a query parameter on the demo page. That is unworkable here — the site is prerendered, so a parameter cannot reach the server-rendered output, and the build fails outright. Slowing the shared fixture instead would have broken the pacing assertions in `typewriter-text-pacing-progress-render.test.ts` and changed the visual baseline. Moving the wait into the page was the fix that needed no demo change at all.

Tara-ag (commented):
Reviewed the core changed files. Two minor suggestions; no blocking issues found.

tests/support/hydrated.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ *   await gotoHydrated(page, '/');
+ *   await page.addScriptTag({ path: 'dist-wc/index.js', type: 'module' });
+ *   await page.waitForFunction(() => Boolean(customElements.get('sui-status')));
+ */
+export const gotoHydrated = async (page: Page, path: string): Promise<Response | null> => {
+  const response = await page.goto(path);
+  await page.waitForFunction(() => document.documentElement.dataset.hydrated === 'true');
+  return response;
+};
```
Tara-ag, line 35: 💬 SUGGESTION: `gotoHydrated` could time out opaquely if the marker is never set. Consider adding a custom timeout message to `waitForFunction` so a future regression (e.g. the layout failing to mount or the marker being removed) produces a diagnostic error instead of Playwright's generic timeout. For example: `await page.waitForFunction(..., { timeout: 10000 });` and/or wrap with `expect.poll` and a descriptive message.

tests/typewriter-text-reduced-motion.test.ts
```diff
@@ -110,36 +139,53 @@ test('reduced motion reveals the whole string at once instead of typing it', asy
...
+  const flippedAt = timing.mutationsAtReduceFlip;
+  expect(typeof flippedAt).toBe('number');
+
+  // The flip has to land with more left to type than the bound below, or the
+  // assertion would pass for the wrong reason. Taken from the in-page count at
+  // the flip, so it measures where the reveal actually was rather than where a
+  // sample taken a round-trip earlier said it was.
+  expect(PROGRESS_TEXT.length - (flippedAt ?? 0)).toBeGreaterThan(12);
+
+  const mutationsAfterToggle = timing.mutations - (flippedAt ?? 0);
```
Tara-ag, line 183: 💬 SUGGESTION: The `<= 2` bound is tight and load-bearing. Consider documenting the demo fixture speed (e.g. `speed=20`) that makes this bound valid, so a future change to the demo speed does not silently invalidate the assertion.

Tara-ag (commented):
Review findings from Yama code review agent.

murdore:
## Control: the mechanism asserted directly at the failing sites

A methodological objection was raised about this PR's evidence, and it was a fair one, so here is the control it asked for.

**The objection.** "8 failures of 713 at default workers, 0 at `--workers=2`" varies parallelism, and parallelism varies contention, ordering and per-worker resource pressure at the same time. That comparison cannot separate "handlers not yet attached" from any other contention-sensitive flake — a slow route, a shared fixture, a state collision would all also improve at `--workers=2`. The deterministic probe in this PR proves the mechanism **exists and is reachable**; it does not prove those particular 8 failures **were** it.

**The control.** Restore the pre-fix navigation (raw `page.goto`, no hydration wait) against the current app, and record — inside the page, at the instant the first `pointerdown` is dispatched — whether hydration had completed. Nothing is read from the test process before the click, so the measurement cannot widen the window it is measuring:

```js
document.addEventListener('pointerdown', () => {
  if (store.__atClick === null) {
    store.__atClick = document.documentElement.dataset.hydrated === 'true';
  }
}, { capture: true });
```

Then correlate that against whether the interaction actually had its effect, at two of the eight failing sites (`menu-placement`, `table-interactive-cells`), under real suite load generated by a concurrent full run from a second checkout.

**Result — 60 samples:**

| hydrated at click | interaction had its effect | rows |
| --- | --- | --- |
| `false` | `false` | **50** |
| `true` | `true` | **10** |
| *disagreement* | | **0** |

Every click that landed before hydration was lost. Every click that landed after it worked. No row disagrees.

**What this changes.** The mechanism is now confirmed at the sites that were actually failing, rather than only in a contrived probe. The worker-count comparison becomes an illustration of *when* it surfaces and stops carrying evidentiary weight — which is how the commit message already framed it ("Root cause, confirmed rather than inferred"), but this makes it hold without leaning on that comparison at all.

**One claim that does not transfer.** The same review noted a sibling repository where CI is configured `workers: isCI ? 8 : 6` alongside `retries: isCI ? 2 : 0` — CI running *more* parallel while also carrying the retries that hide the result. **This repository sets no `workers` at all**, in either Playwright config, so only the `retries` asymmetry applies here. Worth stating plainly rather than letting the stronger version attach to this repo by association.

Probe deleted; nothing in this PR changed.

PR #601, opened 2026-09-11
feat(charts): add createNarrationHighlighter for the onChartReady hook

murdore:
`LineChart`, `BarChart` and `PieChart` have handed back a `ChartHighlightAPI`
from `onChartReady` since #465, described there as being "for imperative point
highlighting from external orchestrators (e.g. voice narration sync,
step-through animations)". The socket shipped; the plug did not, so the obvious
consumer -- an assistant speaking over a chart -- was written per app.

This is that orchestrator. `processText(transcriptSoFar)` highlights the
category the narration names, at most once per turn, clearing after a pause.

Ported from a Lighthouse implementation offered by the agentic-dashboard
session, at roughly a fifth the size, because most of what it carried the
library had already solved:

  - the 207-line chart registry is gone. `getCategories()` is the label list it
    duplicated and `highlight(null)` is the clear path it wrapped, so this holds
    one `ChartHighlightAPI` and nothing else
  - with it goes `getMostRecent()`, and the `chartId: 'current'` placeholder
    that stood in for a chart the orchestrator never actually identified
  - and the upsert branch, which existed only to tolerate `onChartReady` firing
    before a chart was registered -- a window that does not exist when the
    caller hands over the object it was just given
  - the telemetry, and with it an `@opentelemetry/api-logs` dependency. A
    consumer wanting instrumentation can wrap `processText`

Shaped as a factory returning a closure rather than a class with private mutable
fields, following `createCopyState` (#590): the same "behaviour without the
presentation" split, and the same convention of reading options at call time.

THE MATCHING RULE is the part worth review. The original matched
`text.includes(category) || text.includes(content)`, and the source it serves
reads categories straight off a model's tool output, so labels cannot be assumed
distinctive -- `Other` and `None` already appear in its fixtures. A substring
rule fires `Other` on `otherwise`, and because a category fires once per turn
that wrong slice then stays lit for the rest of the answer.

The default here matches whole words only, preferring the longest label when
several match, so `Q1 Europe` wins over `Q1`. Substring behaviour remains
available through `match`.

Whole-word narrows the collision surface; it does not close it, and the tests
pin that rather than implying otherwise. In "on the other hand", `other` is
genuinely a standalone word, so the label matches and no lexical rule could
decide differently without reading the sentence. That case is asserted as a
known limit, with the guidance -- pass a stricter `match`, and `reset()` between
turns -- in both the type and the docs.

Matching is a scan rather than a `RegExp`: a label is caller data that would
need escaping, and `\b` is defined over ASCII so it mismatches labels like
`Café` or an East Asian one. Lookbehind would fix that but is not available
everywhere this library runs.

A demo section on the LineChart page drives it from a scripted transcript fed a
word at a time, so the behaviour is visible rather than only asserted, and three
Playwright cases drive that demo through a mounted chart. The unit suite proves
the orchestrator against a recording stub; those prove the two ends are
connected -- that a category index it picks actually moves a chart. The
line-chart visual baseline is regenerated for the added section, in the pinned
container; the other 92 are untouched and the suite is 93/93.

Verified on this commit: check 0, lint 0, unit 0, integration 0, visual 0. 1042 unit tests, 21 of them
new, covering matching, once-per-turn, auto-clear and teardown, plus hostile
input from an untyped caller: `getCategories` throwing, returning a non-array,
and a non-string transcript. 718 integration cases.

The negative control was required by the offering session and earned its place.
A highlighter that never fires is indistinguishable from one whose categories
never matched, so both directions were mutation-tested: forcing the matcher to
never match fails 15 of 21, and swapping the default to substring matching fails
the case that exists to catch it.

That second mutation initially passed, which is why it is worth recording. The
test then had `Retail` alongside `Other` and asserted `Retail` won -- true under
substring matching too, because longest-match-wins picks the six-character label
over the five-character one. It was asserting the tie-break, not the matcher.
Isolating the category fixed it.

Refs #465

## Summary by CodeRabbit

- **New Features**
  - Added narration-driven highlighting for line charts, automatically emphasizing chart points as referenced categories appear in a transcript.
  - Added controls in the line-chart demo to play, monitor, reset, and replay narration highlighting.
  - Supports configurable matching, case sensitivity, automatic highlight clearing, and safe handling of invalid input.

- **Documentation**
  - Added usage guidance and configuration details for narration synchronization.

- **Tests**
  - Added unit and end-to-end coverage for matching, highlighting, replay, timing, and error handling.


murdore:
## Proof at `4645e31`

| What it shows | Recording |
| --- | --- |
| The transcript grows a word at a time and the chart highlights **Feb → Jun → Nov** as each is named | [watch](https://github.com/user-attachments/assets/9727afeb-ff79-4f30-959b-8c578e01c6c0) |
| Replaying resets, so the same months highlight again — without `reset()` the once-per-turn guard would block them forever | [watch](https://github.com/user-attachments/assets/5b30d88f-3b80-4b8e-922e-cc84e736c4ea) |
| The mounted chart reacts, not just the demo's own readout | [watch](https://github.com/user-attachments/assets/46d67eb1-02ea-4c2d-8e94-f0ac4a66faa2) |

These drive a real `LineChart` through the same `onchartready` API a consumer gets. The unit suite proves the orchestrator against a recording stub; these prove the two ends are connected — that an index it picks actually moves a chart.

### The negative control, which earned its place

The offering session asked for one, on the grounds that *a highlighter that never fires is indistinguishable from one whose categories never matched*. Both directions were mutation-tested:

| Mutation | Result |
| --- | --- |
| Matcher forced to never match | **15 of 21 fail** |
| Default swapped to substring matching | **1 fails** — the case that exists to catch it |

**The second mutation passed on the first attempt, and that is worth recording.** The test then had `Retail` alongside `Other` and asserted `Retail` won — which is true under substring matching too, because longest-match-wins picks the six-character label over the five-character one. It was asserting the tie-break, not the matcher. Isolating the category to just `Other` fixed it. A green harness that has never been seen fail proves nothing, which is exactly the point that was made to me.

### A correction to my own claim about the matcher

I said whole-word matching prevents `Other` firing on *"on the other hand"*. **That is wrong, and the tests now say so.** In that phrase `other` is genuinely a standalone word, so the label matches and no lexical rule could decide otherwise without reading the sentence.

What whole-word actually buys is `otherwise` and `others` — real collisions, since a substring rule fires on both and the once-per-turn guard then locks the wrong slice for the whole answer. The limit is asserted as its own case rather than left implied, with the mitigation (`match` for something stricter, `reset()` between turns) in both the type and the docs.

### Verification

```
pnpm run check        exit 0
pnpm run lint         exit 0
pnpm run test:unit    exit 0     1042 tests (21 new)
playwright            exit 0     718 integration
scripts/visual-test   exit 0     93 / 93
```

The line-chart visual baseline is regenerated in the pinned container for the added demo section; the other 92 are untouched.

### What did not come across

Ported from a Lighthouse implementation at roughly a fifth the size. The 207-line chart registry is gone — `getCategories()` is the label list it duplicated and `highlight(null)` is the clear path it wrapped — and with it `getMostRecent()`, the `chartId: 'current'` placeholder, the upsert branch that existed only to tolerate `onChartReady` firing before registration, and an `@opentelemetry/api-logs` dependency.

Tara-ag (commented):

tests/chart-narration-highlighter.test.ts
```diff
@@ -0,0 +1,65 @@
+import { expect, test } from '@playwright/test';
```
Tara-ag, line 1: ⚠️ MAJOR: Playwright spec imports from `@playwright/test` directly. The repo convention (and bootstrapped standards) expects specs to import from the local test fixture in `tests/support` so shared setup, base URL, and failure-screenshot behavior are preserved. Please switch to the repo fixture or confirm this is an intentional exception.

```ts
import { expect, test } from './support/<fixture>';
```

src/routes/components/line-chart/+page.svelte
```diff
@@ -80,6 +82,48 @@
...
+  };
+
+  const playNarration = (): void => {
+    stopNarration();
+    narrator?.reset();
+    narrationTranscript = '';
+    narrationIndex = null;
+
+    const words = NARRATION_SCRIPT.split(' ');
+    let spoken = 0;
```
Tara-ag, line 115: 💡 MINOR: `playNarration` arms a `setInterval` but the demo page never cleans it up on unmount. If a user navigates away mid-narration, the interval continues firing and mutates state on a destroyed component. Add an `onDestroy` (or Svelte 5 equivalent) teardown that calls `stopNarration()`.

```svelte
// ...
onDestroy(stopNarration);
```

Tara-ag (commented):
Review complete. Two verified findings: one MAJOR on test-fixture import convention, one MINOR on demo-page interval cleanup. No blocking public-API, security, accessibility, or SSR-safety issues found.

Tara-ag:

Severity counts — 🔒 CRITICAL: 0 · ⚠️ MAJOR: 2 · 💡 MINOR: 0

## 🤖 Yama Review Summary

**Verified findings (2):**

- **MAJOR**: Demo `onChartReady` destroys old narrator on remount but does not clear `narrationTimer` — `src/routes/components/line-chart/+page.svelte:87`
- **MAJOR**: Demo page calls destroy() on old highlighter but does not clear its visual highlight — `src/routes/components/line-chart/+page.svelte:90`

_2 additional claim(s) in the model verdict never passed verification and were quarantined (see ungatedIssues)._

- ⚠️ **MAJOR**: Demo `onChartReady` destroys old narrator on remount but does not clear `narrationTimer` — `src/routes/components/line-chart/+page.svelte:87`
  When `onChartReady` fires again (remount), it calls `narrator?.destroy()` and creates a new narrator. However, the old `narrationTimer` interval may still be running and calling `narrator?.processText()` and `narrator?.reset()` on the NEW narrator (because `narrator` is a plain `let` reassigned). Actually, `playNarration` captures `narrator` by variable reference, so after remount the interval wil…
- ⚠️ **MAJOR**: Demo page calls destroy() on old highlighter but does not clear its visual highlight — `src/routes/components/line-chart/+page.svelte:90`
  When onChartReady fires again (remount, hot reload, conditional rendering), the demo calls narrator?.destroy() and creates a new highlighter. destroy() only cancels the pending auto-clear timer; it does not call chart.highlight(null) on the old chart. The previous chart can therefore remain visually highlighted while the new highlighter starts fresh. Since the old chart instance is being replaced,…

murdore:
## Proof at `4645e31`

This PR had no recordings attached. Adding them, and being explicit about what they do and do not establish.

**The narration steps through the transcript and the chart follows**

https://github.com/user-attachments/assets/4b1ae28e-d0a0-4cab-8735-f0d84b30747f

The transcript grows word by word. As each month is named the readout moves `Feb` → `Jun` → `Nov`, and the crosshair and highlighted dot in the chart below move with it. Both the orchestrator's readout and the chart it drives are in frame together on purpose — a recording of the readout alone would prove only that a string changed.

**Replaying resets, so the same months highlight again**

https://github.com/user-attachments/assets/08f21c91-5cc0-47ad-82bb-90f13da90987

Without `reset()` the fired-set still holds `Feb` and it would never highlight a second time. This is that case, watchable.

### What I measured rather than assumed

Sampling ten frames across the first clip and diffing only the driven chart's region — deliberately excluding the readout, so a text change cannot be mistaken for a chart change:

```
f0→f1   1613 px changed    (the first highlight appearing)
f3→f4    101 px
f4→f5     39 px
f6→f7    195 px
f7→f8     84 px
```

The numbers are small because a crosshair plus one dot is not many pixels. They are not zero, which is the point: the chart itself responds, not just the demo's own state.

One correction worth recording. My first attempt at this measurement reported a crosshair sitting at a constant x-position in every frame, which would have suggested the highlight never moved. That detector was locking onto the chart's axis line, not the crosshair — the measurement was wrong, not the feature. I mention it because a plausible-looking number from a mis-aimed detector is exactly how a bad proof gets published, which already happened once on #602 today.

### What these recordings do NOT prove

They show one path through one demo. The matching rules — longest-match-wins, the once-per-turn guard, case sensitivity, the auto-clear timer, hostile input from an untyped caller — are covered by the 21 unit cases in `src/lib/_chart/narrationHighlighter.test.ts` against a recording stub, not by anything visible here.

In particular, the known limit is not visible in any video and is deliberately pinned as a test rather than hidden: a category label that is an ordinary English word still matches when used as one. `Other` fires on "on the other hand", because `other` genuinely is a standalone word there. Whole-word matching narrows the collision surface; it does not eliminate it, and no lexical rule could without understanding the sentence. A caller whose categories come from a model's output should pass `match` and require something stronger.

This PR remains held for API review — nothing here is a request to merge it.

murdore:
## Re-pushed at `50e02ba` — one finding fixed, one refuted

### MINOR (interval leak) — fixed

`src/routes/components/line-chart/+page.svelte` now calls `onDestroy(stopNarration)`. Navigating away mid-narration previously left the interval running and calling `processText` on a destroyed component. The finding was correct.

Worth noting for the reviewer: all three demo pages in this repo that use `setInterval` lacked this, so it is a pre-existing pattern gap rather than something this PR introduced. I fixed only the page this PR touches.

### MAJOR (Playwright import convention) — refuted

> Playwright spec imports from `@playwright/test` instead of repo fixture

There is no such fixture. Measured on this branch:

```
tests/support/          contains exactly one file: hydrated.ts
its exports             gotoHydrated  — and nothing else
specs importing test/expect from a local fixture     0
specs importing from '@playwright/test'            304
```

There is no `base.extend`, no exported `test`, no exported `expect` anywhere under `tests/support/`. The spec already imports `gotoHydrated` from `./support/hydrated`, which is the actual shared-setup convention this repo has — the one introduced by #600 to fix hydration races.

The verdict text itself concedes *"Other specs in this repo import from `@playwright/test` as well."* The standard it cites is not the standard this repository follows, and adopting it would require inventing a fixture that does not exist.

No change made. Happy to be shown a fixture I have missed.

### Verification at `50e02ba`

```
check   exit 0
lint    exit 0
vitest  1042 passed
```

docs/LineChart.md
```diff
@@ -130,6 +130,58 @@ Receive a `ChartHighlightAPI` after mount to drive point highlighting from an ex
...
+  const onTurnStart = () => highlighter?.reset();
+
+  $effect(() => () => highlighter?.destroy());
+</script>
+
+<LineChart {series} xAxisCategories={months} onchartready={onChartReady} />
+```
+
+It takes the `ChartHighlightAPI` directly — no registry of charts, because `getCategories()`
+is already the label list and `highlight(null)` is already the clear path. `processText`
```
Tara-ag, line 167: 💡 MINOR: Docs example names the parameter `chart` while the type reference calls it `ChartHighlightAPI`. Rename to `api` or `chartApi` to avoid implying the handler receives a component instance rather than the highlight API:

```svelte
const onChartReady = (api: ChartHighlightAPI) => {
  highlighter = createNarrationHighlighter({ chart: api });
};
```

src/routes/components/line-chart/+page.svelte
```diff
@@ -80,6 +83,52 @@
 
   const onChartReady = (api: ChartHighlightAPI): void => {
     highlightApi = api;
+    narrator = createNarrationHighlighter({ chart: api, autoClearMs: 1200 });
```
Tara-ag, line 90: ⚠️ MAJOR: Demo page creates a new highlighter on every `onChartReady` but never destroys the old one. If the `LineChart` remounts, the previous narrator is overwritten without `destroy()`, leaking its pending `setTimeout` and leaving it able to call `highlight(null)` on a stale chart. Destroy the previous instance before reassigning:

```ts
const onChartReady = (api: ChartHighlightAPI): void => {
  highlightApi = api;
  narrator?.destroy();
  narrator = createNarrationHighlighter({ chart: api, autoClearMs: 1200 });
};
```

Tara-ag (changes requested):
Solid, well-tested addition. One lifecycle issue in the demo page needs fixing before merge: `onChartReady` can recreate the narrator without destroying the previous one, leaking a pending auto-clear timer. The docs example could also be clearer about the parameter name. Once the demo teardown is addressed, this is good to go.

murdore:
## Review threads worked through at `cecb581`

7 unresolved threads: 5 fixed · 1 refuted · 1 already correct.


### **Fixed**

**`src/lib/_chart/narrationHighlighter.ts:141`** — readCategories filtered invalid entries with .filter(), shifting the array index passed to chart.highlight relative to the chart's own (unfiltered) categories array.

> Old code: `return categories.filter((category) => typeof category === 'string');` then processText's `categories.forEach((label, index) => ...)` used that post-filter index directly as `bestIndex` passed to `chart.highlight(bestIndex)`. For getCategories()=[null,'Revenue'], 'Revenue' would move from source index 1 to filtered index 0, highlighting the wrong slice. Fixed by adding a `ReadCategory {label, sourceIndex}` type and building it with `categories.flatMap((category, sourceIndex) => typeof category === 'string' ? [{label: category, sourceIndex}] : [])`, then using `sourceIndex` everywhere (fired.has/add, bestIndex, chart.highlight). Added regression test in narrationHighlighter.test.ts ('preserves each category's original index when an invalid entry is dropped') asserting getCategories()=[null,'Revenue'] + processText('revenue climbed') highlights index 1, not 0. Full vitest suite 

**`src/routes/components/line-chart/+page.svelte:127`** — The narration readout said 'highlighting Nov' even after createNarrationHighlighter's autoClearMs (1200ms) had already cleared the chart's actual highlight, misdescribing state that no longer held.

> narrationHighlighter.ts lines 201-205 clear the chart via `clearTimer = setTimeout(() => chart.highlight(null), autoClearMs)`, but the demo's `narrationIndex` state (+page.svelte line 99/117/126) is only reset by the next match or by reset() — it is never cleared by the timer, so the old readout text 'highlighting Nov' stayed on screen and stayed wrong indefinitely after 1200ms. Reworded the span to `last matched ${monthLabels[narrationIndex]}`, which remains true regardless of the auto-clear, and updated all six 'highlighting X' assertions in tests/chart-narration-highlighter.test.ts to 'last matched X'. Full playwright-adjacent build/typecheck/lint/vitest all pass with the new wording.

**`tests/chart-narration-highlighter.test.ts:50`** — The 'chart itself renders the highlight' test located the LAST demo-row containing an svg on the whole page (a chart unrelated to onChartReady) and asserted only toBeVisible(), so it could pass even if ChartHighlightAPI.highlight() did nothing.

> Read full +page.svelte: 16 `.demo-row` sections render a LineChart with an svg; only the 11th ('Highlight Hook — onChartReady...', line ~330) is wired to onChartReady, but 5 more chart demo-rows follow it (highlightedIndex, Legacy gradientFill, Shared Tooltip testId=line-shared-tooltip-chart, Dense Values testId=line-dense-values-chart, minHeight/maxHeight testId=line-height-bounds-chart). `.locator('.demo-row').filter({has: svg}).last()` therefore resolved to the minHeight/maxHeight chart, not the onChartReady chart, and `toBeVisible()` on its svg is trivially true regardless of highlighting. Fixed by adding `testId="line-highlight-hook-chart"` to the onChartReady-bound `<LineChart>` and rewriting the test to locate it via `page.getByTestId('line-highlight-hook-chart')`, asserting `circle.dot.highlighted` has count 0 before narration and count 1 once the readout shows 'last matched Feb'

**`docs/LineChart.md:163`** — The narration-sync doc example named its onChartReady callback parameter `chart` while the option it feeds (`NarrationHighlighterOptions.chart`) is typed `ChartHighlightAPI`, inconsistent with the doc's own earlier 'Highlight Hook' example.

> Read docs/LineChart.md: the narration-sync example at line 146 had `const onChartReady = (chart) => { highlighter = createNarrationHighlighter({ chart }); };`, while the earlier 'Highlight Hook — onChartReady' example (lines 113-116, same file) already names the same callback parameter `chartApi` / stores it as `api`, establishing the doc's own convention. Renamed the narration-sync example's parameter to `api` and updated the call to `createNarrationHighlighter({ chart: api })`.

**`src/routes/components/line-chart/+page.svelte:86`** — onChartReady created a new createNarrationHighlighter on every call without destroying the previous instance, so a remount would leak the old highlighter's pending auto-clear timer, which could call highlight(null) on a stale chart API.

> Before: `const onChartReady = (api) => { highlightApi = api; narrator = createNarrationHighlighter({ chart: api, autoClearMs: 1200 }); };` had no destroy call. narrationHighlighter.ts's `destroy` (returned as `cancelPendingClear`) is the only thing that cancels a pending setTimeout, and it is never called automatically, so an overwritten `narrator` variable was the last reference to an instance whose timer kept running. Added `narrator?.destroy();` immediately before reassignment, matching the pattern narrationHighlighter.ts's own doc comment describes for teardown.


### **Refuted**

**`tests/chart-narration-highlighter.test.ts:1`** — Reviewer claims the spec should import test/expect from a repo test fixture in tests/support rather than '@playwright/test' directly.

> `grep -rl "from '@playwright/test'" tests/ | wc -l` returns 153 — every other Playwright spec in the repo (e.g. tests/button-title-busy.test.ts:1, tests/carousel.test.ts:1) imports `{ expect, test }` from '@playwright/test' directly, exactly like the reviewed file. `tests/support/hydrated.ts` was read in full: it exports only `gotoHydrated` (a page.goto + hydration-wait helper) and does not re-export `test`/`expect` — there is no fixture-wrapping convention in this repo to switch to. The claimed convention does not exist; no change made.


### **Already correct**

**`src/routes/components/line-chart/+page.svelte:120`** — Reviewer claims playNarration's setInterval is never cleaned up on unmount, leaking an interval that would call processText on a destroyed component.

> Read +page.svelte lines 1-2, 100-111: `import { onDestroy } from 'svelte'` (line 2), `let narrationTimer: ReturnType<typeof setInterval> | null = null` (line 100), a `stopNarration` function that clears it (lines 102-107), and `onDestroy(stopNarration)` (line 111) with a comment already explaining exactly the leak the review describes: 'Navigating away mid-narration would otherwise leave the interval running and calling processText on a destroyed component.' The teardown the reviewer asks for is already present; no change made.


### Verification

```
check   exit 0
lint    exit 0
vitest  Test Files 53 passed (53); Tests 1043 passed (1043); Duration 22.44s
```

Tara-ag (commented):

src/routes/components/line-chart/+page.svelte
```diff
@@ -253,6 +306,24 @@
   />
 </div>
 
+<h3>Narration sync — createNarrationHighlighter</h3>
+<p>
+  The orchestrator the hook above exists for. <code>processText</code> receives the transcript as it
+  grows and highlights the month being named — <code>Feb</code>, then <code>Jun</code>, then
+  <code>Nov</code> — clearing itself after a pause.
```
Tara-ag, line 313: 💬 SUGGESTION: The narration demo updates transcript and highlight state visually, but screen-reader users get no announcement of the progressing narration. Add a visually-hidden `aria-live="polite"` region (or make this readout a live region) so the demo models accessible behavior:

```suggestion
  <span data-pw="narration-highlighted" class="narration-readout" aria-live="polite">
```

src/lib/index.ts
```diff
@@ -212,3 +213,4 @@ export type * from './Gallery/properties';
 export { createSoundKit } from './SoundKit/SoundKit';
 export { validateInput, lockBodyScroll, unlockBodyScroll } from './utils';
 export { formatNumberIndian } from './_chart/format';
+export { createNarrationHighlighter } from './_chart/narrationHighlighter';
```
Tara-ag, line 216: 💡 MINOR: Exporting a runtime factory from the component barrel is consistent with `createCopyState` / `createSoundKit`, but note that `createNarrationHighlighter` is only meaningful in a browser (it schedules `setTimeout`). The factory itself is SSR-safe, but consumers should instantiate it inside `onChartReady` / `onMount`. Consider a one-line note in `docs/LineChart.md` if you want to make that explicit.

src/routes/components/line-chart/+page.svelte
```diff
@@ -80,6 +83,56 @@
 
   const onChartReady = (api: ChartHighlightAPI): void => {
     highlightApi = api;
+    // A remount fires onChartReady again with a new api. Without this, the
+    // previous narrator's pending auto-clear timer survives and later calls
+    // highlight(null) on the stale api it closed over.
+    narrator?.destroy();
+    narrator = createNarrationHighlighter({ chart: api, autoClearMs: 1200 });
```
Tara-ag, line 90: ⚠️ MAJOR: `narrator?.destroy()` only cancels the pending auto-clear timer; it does not clear the highlight on the previous chart. On remount, the old chart can stay visually highlighted while the new highlighter starts fresh. Use `narrator?.reset()` here instead — it cancels the timer, clears the fired set, and calls `chart.highlight(null)` on the API that is about to be replaced.

```suggestion
    narrator?.reset();
    narrator = createNarrationHighlighter({ chart: api, autoClearMs: 1200 });
```

Tara-ag (commented):
Review complete. One MAJOR lifecycle issue in the demo page, one MINOR API-shape note, and one accessibility suggestion. The core library addition (`createNarrationHighlighter`) is well-tested and does not break the public contract.

murdore:
## Two applied, one refuted at `5c4df85`

### Applied — `aria-live` on the narration readout

Fair, and the demo is the place to model it. The readout is now `aria-live="polite"`, so a screen-reader user hears the narration progress rather than only seeing it.

### Applied — SSR note in the docs

`createNarrationHighlighter` is SSR-safe but the highlighter it returns schedules a `setTimeout`, so `docs/LineChart.md` now says to instantiate it inside `onchartready` or `onMount`, never at module scope.

### Refuted — `narrator?.destroy()` → `narrator?.reset()`

The described failure cannot occur, and the substitution would introduce a real one.

**`onchartready` fires once per chart instance.** `LineChart.svelte:124` calls `onchartready?.(api)` from inside `onMount` (line 103), which Svelte runs exactly once per instance:

```
103:  onMount(() => {
124:    onchartready?.(api);
```

So for the handler to run a second time, the previous `LineChart` must have been destroyed — and its highlight went with it. There is no live, highlighted old chart left to clear, which is the scenario the thread describes.

**And `reset()` would call into that destroyed chart.** `reset` ends with `chart.highlight(null)` at `narrationHighlighter.ts:226`, **unguarded** — the only `try`/`catch` in the module is in `readCategories` (147–155), added precisely because "a torn-down chart can throw or return junk". Swapping `destroy()` for `reset()` puts an unguarded call to a stale API inside the *new* chart's mount path.

That is exactly why `destroy` is documented as it is:

> Cancels a pending auto-clear. Call from the owner's teardown.
> Deliberately does not call into the chart: teardown ordering is the caller's, and the chart may already be gone.

The thread's reading of what each function does is correct. What does not hold is the situation it assumes.

```
check   exit 0
lint    exit 0
build   exit 0
vitest  1043 passed
playwright (narration spec)  3 passed
```

PR #602, opened 2026-09-11
feat(voice-orb): add an audio-reactive particle orb for voice surfaces

murdore:
Ported from the Lighthouse `AutomaticOrb` offered by the agentic-dashboard
session: a thousand seeded points on a rotating sphere, drawn to a canvas, with
a `listening` state that breathes.

Pass an `AnalyserNode` and `listening` is driven by the sound itself -- amplitude
moves both rotation speed and the size of the sphere. Without one it breathes on
a timer and is purely decorative.

The orb never opens a microphone. It does not call `getUserMedia`, does not
create an `AudioContext` and never closes one: permission prompts, context
lifecycle and the choice of source need application context a component does not
have, and one that grabbed a microphone because it was mounted would be a worse
component. The caller owns the node; the orb reads it. Any source works -- a
microphone, a media element, a synthesised tone.

The envelope is RMS over byte time-domain data, smoothed with a fast attack
(60ms) and a slower release (320ms). Speech is a burst followed by a gap, and
symmetric smoothing either chatters through every syllable or lags the onset
that makes the orb feel connected to the voice. Silence draws at 0.82 scale and
full level at 1.0, never above, so a shout cannot clip the sphere against the
canvas edge. `speedMultiplier` still applies on top, so `0` freezes an
audio-driven orb like any other.

`onlevel` hands the same smoothed envelope back to the caller, so a caption or
a meter stays in step with the sphere instead of measuring the analyser a second
time and drifting. It is rate-limited by change rather than by time -- it fires
when the level moves by at least 0.01, so silence is silent instead of sixty
calls a second, and it fires once more on the way back to zero when the audio
stops so a meter cannot stick at the last value it saw.

Degradation is deliberate and tested. No analyser means `listening` falls back
to the sine wave, so a consumer without an `AudioContext` keeps the animation.
Closing the context under a mounted orb is guarded: an exception thrown inside
the rAF callback would kill the loop permanently and leave a frozen sphere.

Particles draw in the inherited text colour. That is the whole theming contract:
legible on whatever ground a consumer puts it on, correct in both themes, no
configuration. `--sui-orb-particle-color` overrides it and accepts any CSS colour
-- `oklch()`, `color(display-p3 ...)`, named -- because the value is normalised
by assigning it to a canvas `fillStyle` and reading it back rather than parsed
here. It is resolved per frame against a string cache, so switching themes
recolours an orb that is already running.

Three app couplings had to go, not the one that was reported:

  - `session.store`, read for an RNG seed, becomes a `seed` prop taking a string
    or a number. A string is hashed with FNV-1a, so the same session id yields
    the same orb across navigations, which was the point of the coupling
  - `hexToRgb` from the app's utils is replaced by `parseCssColor`, which reads
    the two shapes a browser actually returns rather than hex alone
  - `$app/environment`'s `browser` is dropped. The library imports it nowhere,
    and `onMount` is client-only, so the guard was redundant here

`localStorage` is the substantive change. The original wrote `orb_rotationX`,
`orb_rotationY` and `orb_listeningTime` unconditionally under fixed global keys.
In a package 56 projects install that is not acceptable: two orbs on one page
would overwrite each other, and writing to a consumer's storage uninvited is not
a component's decision. It is now opt-in behind `persistKey`, namespaced under
it, and every read and write is guarded.

The geometry, seeding, envelopes and colour parsing live in `orbMath.ts` rather
than the component. This repo's vitest suite has no canvas -- `getContext` is
unimplemented in jsdom -- so anything left inside the component is unreachable
by a test. Same split as `Pill/pillTone.ts`.

A `speedMultiplier={0}` orb is on the demo page under "Held still". It is a real
state a consumer wants, and it is also the one arrangement whose pixels are
identical on every frame, which is what lets the visual suite baseline actual
orb pixels instead of masking the whole route.

Verified on this commit: check 0, lint 0, 1072 unit tests, 724 integration,
94/94 visual with the baseline stable across two container runs.

The audio path is covered end to end by driving a real oscillator through a real
`AnalyserNode` and measuring the painted sphere: loud is measurably wider than
quiet, which a component ignoring the analyser cannot fake. The maths is
mutation-tested -- reading the raw byte instead of its excursion from 128 makes
silence measure as full scale, swapping the attack and release constants loses
the fast onset, and dropping `speedMultiplier` from `audioSpeed` un-freezes a
frozen orb. Each fails its own case and no others.


murdore:
## Proof at `07b7a53`

| What it shows | Recording |
| --- | --- |
| The orb paints, and `onfirstframe` fires on the first painted frame — asserted against the canvas pixels, not the callback's word | [watch](https://github.com/user-attachments/assets/60976561-eedd-495f-b10e-3f07d4bca5c6) |
| The `idle` → `listening` switch reaches the running animation | [watch](https://github.com/user-attachments/assets/bd89e580-a90c-468d-a7f3-f53b3ec53e62) |
| Each instance gets its own canvas at its own height | [watch](https://github.com/user-attachments/assets/9fb350ad-c2bd-4a98-9726-c7173af17e21) |

The first case does not trust the callback. It samples the canvas through `getImageData` and fails unless some pixel has non-zero alpha — a callback that fired early on a blank canvas would pass a naive assertion and fails this one.

### Two of the repo's own guards caught real defects in this change

| Guard | What it caught |
| --- | --- |
| `prop-parity` | `onanimationstart` is a genuine `HTMLElement` accessor for the CSS animationstart event. A web-component prop of that name would **shadow the host's own**. Renamed `onfirstframe` — also the more honest name, since it fires on first paint, not mount. |
| `docs-index` | A documented component missing from `docs/_index.json`. |

Neither was something I reasoned my way to; both were the repo refusing the change.

### What changed from the Lighthouse original

**Three app couplings, not the one that was reported.** Alongside `session.store` there were `$app/environment` and `hexToRgb` from app utils. The library already exports its own `hexToRgb`, and it imports `$app/environment` nowhere, so that guard is dropped — `onMount` is client-only.

**`localStorage` is the substantive change.** The original wrote `orb_rotationX` / `orb_rotationY` / `orb_listeningTime` unconditionally under fixed global keys. In a package 56 projects install, two orbs on one page would overwrite each other's rotation, and writing to a consumer's storage uninvited is not a component's decision. It is now opt-in behind `persistKey`, namespaced under it, with every read and write guarded.

**One latent defect came across and is fixed.** The original called `hexToRgb(particleColor)` and used `.r` directly. This library's `hexToRgb` returns `Rgb | null`, and the colour comes from a CSS custom property a consumer controls — `rebeccapurple` or `oklch(...)` would have thrown on every frame.

### Testing a canvas in a suite that has no canvas

`getContext` is unimplemented in jsdom, so anything left inside the component is unreachable by a unit test. The geometry, seeding and listening envelope live in `orbMath.ts` — the same split as `Pill/pillTone.ts` and `Table/normalizeColumns.ts` — and are covered by 22 cases there, with the Playwright cases above covering what a pure test cannot.

The even-distribution claim is **mutation-tested**: sampling `phi` uniformly instead of `cos(phi)` crowds the poles and fails the case asserting it; removing the radius guard fails the case asserting NaN input still draws.

### Visual baselining

The canvas is rAF-driven, so `animations: 'disabled'` cannot pin it to a frame. It follows `lottie-player`: the two canvas regions are **masked** rather than the route excluded, so the page's headings and controls are still compared. Baseline checked twice in the pinned container, stable both times.

```
pnpm run check        exit 0
pnpm run lint         exit 0
pnpm run test:unit    exit 0     1045 tests (22 new)
playwright            exit 0     721 integration
scripts/visual-test   exit 0     94 / 94
```

murdore:
[602-01-voice-orb-VoiceOrb-paints--4605f-rst-frame-r-07b7a53.webm](https://github.com/user-attachments/assets/60976561-eedd-495f-b10e-3f07d4bca5c6)
[602-03-voice-orb-VoiceOrb-the-variant-switch-reache-07b7a53.webm](https://github.com/user-attachments/assets/bd89e580-a90c-468d-a7f3-f53b3ec53e62)
[602-02-voice-orb-VoiceOrb-renders-50be5-ach-sized-t-07b7a53.webm](https://github.com/user-attachments/assets/9fb350ad-c2bd-4a98-9726-c7173af17e21)

src/lib/VoiceOrb/VoiceOrb.svelte
```diff
@@ -0,0 +1,277 @@
...
+    listeningSpeed,
+    mulberry32,
+    orbDepthScale,
+    rotateOrbPoint,
+    type OrbParticle
+  } from './orbMath';
+  import type { VoiceOrbProperties } from './properties';
+
+  let {
+    variant = 'idle',
```
Tara-ag, line 25: ⚠️ MAJOR: These props are destructured into plain `let`s and then read inside the `draw()` rAF callback and the `onMount` closure without reactivity. Changing `variant` or `speedMultiplier` after mount will not affect the running animation because `draw()` closes over the initial values. The demo page toggles `variant` with a button, so this is observable broken behavior.

Suggested fix: keep the props reactive. Either do not destructure and read `props.<name>` everywhere, or declare e.g. `let variant = $derived(props.variant)` after destructuring. Values used inside `draw()` must be read through the reactive props object so Svelte's fine-grained updates propagate to the running animation.

src/wc/components/VoiceOrb.wc.svelte
```diff
@@ -0,0 +1,27 @@
+<svelte:options
+  customElement={{
+    tag: 'sui-voice-orb',
+    shadow: 'open',
+    props: {
+      variant: { type: 'String' },
+      speedMultiplier: { type: 'Number', attribute: 'speed-multiplier' },
+      seed: { type: 'String' },
```
Tara-ag, line 8: 💡 MINOR: The Svelte component accepts `seed: string | number`, but the web component declares `seed: { type: 'String' }`. A numeric attribute is coerced to string and then hashed, while the Svelte component uses a number directly. This is a behavioral divergence in the public contract across distribution channels.

Suggested fix: add a note in `docs/VoiceOrb.md` explaining that web-component consumers who need direct numeric seed behavior should set the property via JavaScript rather than the attribute.

tests/voice-orb.test.ts
```diff
@@ -0,0 +1,59 @@
+import { expect, test } from '@playwright/test';
```
Tara-ag: 💡 MINOR: Other animated components in this repo have reduced-motion tests (e.g. `tests/typewriter-text-reduced-motion.test.ts`). VoiceOrb adds a continuous ambient animation but has no corresponding test that animation pauses under reduced motion, leaving the accessibility gap unverified.

Suggested fix: add a Playwright test that emulates reduced motion, mounts the orb, captures two canvas samples a few hundred ms apart, and asserts the pixel buffer is unchanged (animation paused).

src/lib/VoiceOrb/VoiceOrb.svelte
```diff
@@ -0,0 +1,277 @@
...
+      particleColor = resolvedColor;
+    }
+
+    orb.angleX = readPersisted('angleX') ?? 0;
+    orb.angleY = readPersisted('angleY') ?? 0;
+    listeningElapsedMs = readPersisted('listeningMs') ?? 0;
+
+    particles = generateOrbParticles(particleCount, radius, mulberry32(resolveSeed()));
+    resize();
+
```
Tara-ag, line 326: 💬 SUGGESTION: The canvas is `aria-hidden`, which is appropriate because it carries no semantic information. However some a11y audits prefer a decorative canvas to also have `role='img'` and an explicit empty `aria-label` when it is the sole visual content. The current approach is defensible; this is optional.

src/lib/VoiceOrb/VoiceOrb.svelte
```diff
@@ -0,0 +1,277 @@
+<script lang="ts">
```
Tara-ag, line 1: ⚠️ MAJOR: VoiceOrb drives a continuous rAF-based rotation and breathing animation but does not query `prefersReducedMotion()`. The library has a shared utility for this (`src/lib/utils.ts`) and other components (`TypewriterText`, `Tabs`, `TaskList`, etc.) honor it. A decorative ambient animation is exactly the kind of motion users ask the OS to suppress; leaving it running violates the repo's accessibility pattern and may cause vestibular issues.

Suggested fix: import `prefersReducedMotion` from `'../utils'`. In `draw()`, when it returns `true`, skip the angle updates and `listeningElapsedMs` accumulation but still paint one static frame so the orb remains visible. Consider also exposing a `reducedMotion` prop override for test parity with other components.

Tara-ag (changes requested):
Good, careful port overall — especially the localStorage opt-in and the color-parse fallback. Two MAJOR issues need addressing before merge:

1. **Reactivity**: destructured props are read inside a long-lived rAF callback; changes to `variant`, `speedMultiplier`, etc. after mount won't reliably propagate, and `particleCount`/`radius`/`seed` are only consumed once in `onMount`. The demo's variant toggle is affected.
2. **Reduced motion**: a continuous ambient animation must honor `prefers-reduced-motion: reduce` using the existing `prefersReducedMotion()` utility, consistent with the rest of the library.

Also noted a MINOR web-component seed divergence and missing reduced-motion test coverage, plus an optional SUGGESTION on canvas role.

Tara-ag:

VoiceOrb is a solid addition — good pure-function split, reduced-motion handling, SSR guards, and thorough tests. However, the web-component distribution is currently broken because the new wrapper is not registered. I also could not fully verify from truncated output that the Svelte component itself is exported from src/lib/index.ts; please confirm `export { default as VoiceOrb } from './VoiceOrb/VoiceOrb.svelte';` is present.

- ⚠️ **MAJOR**: VoiceOrb web component is never registered — `src/wc/index.ts:1`
  src/wc/components/VoiceOrb.wc.svelte was added, but src/wc/index.ts does not import it. The web-component build relies on these side-effect imports to call customElements.define, so <sui-voice-orb> will be undefined for WC consumers even though the wrapper file exists.

murdore:
## Proof at `710275e` — replacing the previous comment, which was wrong

The proof I posted before was not proof. Three things were wrong with it, and the review that caught them was Sachin opening the PR and looking at it, not anything I ran.

**1. The screenshot was blank because I masked the whole component.** My `MASKS` entry was `'voice-orb': ['.demo-row', '.orb-grid']`, and those two selectors were *every* container on the demo page. The visual suite compared headings and paragraph text and nothing else. It reported 94/94 while looking at zero pixels of the orb.

**2. The orb really was invisible.** The default particle colour was `#f2f2f2` on a white page. `docs/VoiceOrb.md` said, in my own words, *"Set a dark value for light backgrounds, or the orb will be invisible on white"* — I documented the defect and then recorded the proof on a white page.

**3. The assertion I was proudest of was a false green.** I wrote that the test "does not trust the callback" because it sampled `getImageData` and required non-zero alpha. Alpha is not visibility: a near-white particle at alpha 0.8 passes it. It was the same false-green shape I have spent this week catching in other people's tests.

### Measured, in the orb region, before and after

| | before | after |
| --- | ---: | ---: |
| darkest pixel (background is 255) | **240** | **47** |
| max luminance change between frames | **10 / 255** | **200–249** |
| pixels visibly moving | **0** | **13.8k – 143k** |

The orb *was* rotating in the old recording — about 8° over the clip — at a contrast neither an eye nor a video codec could resolve. "There is no rotation" was the correct reading of what was on screen.

### The recordings

Each is 9–11s, recorded at `710275e`, and each was checked frame-by-frame before being attached.

**Rotation is visible and continuous**

https://github.com/user-attachments/assets/8910b73c-c276-4908-88ae-cafd16c95bcd

**`idle` → `listening`: the switch reaches the running animation, and the breathing is visible**

https://github.com/user-attachments/assets/7a4765f7-0da8-4a72-bdce-f16e33f00d52

**The orb follows the page theme** — dark particles on light, light on dark, resolved live rather than at mount

https://github.com/user-attachments/assets/e3fbb91f-340e-4566-967e-bc573be4937f

**Pointer repulsion**

https://github.com/user-attachments/assets/f2469171-c746-421f-bc04-ec19c51a7a54

**Seeded placement, and the held-still orb the visual baseline now compares**

https://github.com/user-attachments/assets/ff870dee-5e9f-4b9f-b444-00f9ad3498ca

### What changed in the code

**Particles draw in the inherited text colour.** That is the whole theming contract: legible on whatever ground a consumer puts it on, correct in both themes, no configuration. A fixed default cannot do that, which is what the first revision proved.

`--sui-orb-particle-color` still overrides, and now accepts **any** CSS colour — `oklch()`, `color(display-p3 ...)`, named colours — because the value is normalised by assigning it to a canvas `fillStyle` and reading it back, rather than parsed here. That removes the documented `oklch()` limitation instead of restating it. Unparseable values fall through to the inherited colour, then to a mid-grey.

The colour is resolved per frame against a string cache, so switching themes recolours an orb that is already running.

**The visual suite now compares real orb pixels.** A `speedMultiplier={0}` orb is on the demo page under "Held still" — a genuine state a consumer wants for a muted or paused surface, and the one arrangement whose pixels are identical on every frame. It is deliberately **not** masked. The moving orbs still are.

**One more defect the theme recording exposed:** the demo page's own button was hardcoded `#f9fafb` with no colour, so it rendered white-on-white the moment the page went dark. It is on the demo shell's tokens now.

### The tests that would have caught this

The visibility test was written first and **failed at 32 against a floor of 40** before any fix. It composites each painted pixel over the background a human actually sees behind it and measures the luminance difference, instead of asking whether alpha is non-zero.

Alongside it: rotation must move the pixels; the pinned orb must *not* move **and** must still be visible (a blank canvas is also perfectly stable); and the orb must recolour when the theme flips.

`parseCssColor` is mutation-tested — padding the short hex form instead of doubling its digits, and dropping the finite check on `rgb()` channels, each fail exactly one case and no others.

### Verification at `710275e`

```
pnpm run check        exit 0
pnpm run lint         exit 0
vitest                exit 0     1055 tests
playwright            exit 0     721 integration
scripts/visual-test   exit 0     94 / 94, baseline stable across two runs
```

### Still open

The name promises audio and the component ignores it. Sachin has asked for it to be made genuinely audio-reactive, so an optional analyser input driving the breathing envelope is the next commit on this branch. Until that lands, this is a decorative orb with a name that overpromises.

murdore:
## Audio-reactive at `d135a27`

Sachin's call on the previous round: the name promised audio and the component ignored it. It no longer does.

Pass an `AnalyserNode` and `listening` is driven by the sound itself. Without one it still breathes on a timer, so nobody loses the animation.

**The orb never opens a microphone.** No `getUserMedia`, no `AudioContext` created, none closed. Permission prompts, context lifecycle and the choice of source need application context a component does not have, and one that grabbed a microphone because it was mounted would be a worse component. The caller owns the node; the orb reads it. A microphone, a media element and a synthesised tone all work identically.

### The orb responds to the sound

The demo drives a real oscillator through a real `AnalyserNode`. The tone is routed to the destination through a **zero gain** — the graph stays alive, and the demo does not beep at you.

https://github.com/user-attachments/assets/2358c643-5cbb-440d-845f-55aee9bc7bea

I measured that recording rather than trusting it. The slider sequence in the clip is `1, 0, 1, 0, 0.5, 1, 0`, and the painted sphere width, frame by frame, is:

| tone level | 1 | 0 | 1 | 0 | **0.5** | 1 | 0 |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| sphere width (px) | 171 | 149 | 170 | 149 | **160** | 171 | 149 |

The half-level step landing between the two extremes is the part that matters: the orb is tracking the *level*, not toggling on a threshold. Range across the clip is 148–181 px, a 22% swing.

One correction to my own measurement: the first naive pass reported a 2.88× ratio. Two frames at the very start were taken before the page finished scrolling and caught a *different* orb higher up the page. Excluding them, the honest figure is the 1.22× above.

### The rest, re-recorded on this commit

**Rotation** — https://github.com/user-attachments/assets/413f0c44-d7c0-4ead-90f8-721b163e8d8e

**Theme** (particles follow the inherited text colour, resolved live) — https://github.com/user-attachments/assets/321fdbc4-8b7d-4709-8886-3667f46a0b89

**Pointer repulsion** — https://github.com/user-attachments/assets/6a38809d-e23e-4ed0-95f7-b4761325ab26

**Seeded placement and the held-still orb** — https://github.com/user-attachments/assets/bab8a0c0-22c8-4d55-b422-ae53a4d424a1

Every clip was put through a gate that fails it unless the component is both visible and moving, before being attached. That gate exists because the first proof on this PR was neither.

### How the envelope behaves

RMS over one buffer of byte time-domain data, smoothed with a **fast attack (60ms) and a slower release (320ms)**. Speech is a burst followed by a gap; symmetric smoothing either chatters through every syllable or lags the onset that makes the orb feel connected to the voice.

Silence draws the sphere at 0.82 scale, full level at 1.0, never above — so a shout cannot clip the sphere against the canvas edge. `speedMultiplier` still applies on top, so `speedMultiplier={0}` freezes an audio-driven orb exactly like any other.

### Degradation, because this is the part that bites later

- **No analyser** → `listening` falls back to the sine wave. Covered by a test that samples the painted width over time and requires it to vary.
- **Context closed under a mounted orb** → the read is guarded and the orb returns to the sine wave. Without the guard an exception inside the rAF callback kills the loop permanently and leaves a frozen sphere, which is why there is a test that closes the context mid-flight and then asserts the orb is still painting and still visible.

### Mutation-tested, not just covered

| mutation | what it breaks | caught by |
| --- | --- | --- |
| read the raw byte instead of its excursion from 128 | silence measures as full scale — the orb roars at nothing | 4 cases |
| swap the attack and release constants | loses the fast onset | 1 case |
| drop `speedMultiplier` from `audioSpeed` | un-freezes a deliberately frozen orb | 2 cases |

### A guard caught one more thing

Prerendering validates anchors in rendered docs, and my `[Audio-reactive](#audio-reactive)` link did not resolve to an id on the page — the build failed on it rather than shipping a dead link.

### Verification at `d135a27`

```
pnpm run check        exit 0
pnpm run lint         exit 0
vitest                exit 0     1072 tests
playwright            exit 0     724 integration
scripts/visual-test   exit 0     94 / 94, stable across two runs
```

One unrelated test — `a11y-context-menu` "Enter selects the focused item" — failed once during an intermediate full-suite run and passed 3/3 in isolation and on the final run. Reporting it rather than quietly re-running: it is load-dependent flake in that spec, not something this change touches.

murdore:
### `onlevel` added at `a6dd380`

Sachin asked for the level to be drivable into other UI. `onlevel` hands back the same smoothed 0..1 envelope the orb draws with, so a caption or a meter stays in step with the sphere rather than measuring the analyser a second time and drifting.

```svelte
<VoiceOrb {analyser} variant="listening" onlevel={(level) => (meter = level)} />
```

**Rate-limited by change, not by time.** It fires when the level moves by at least `0.01`, so steady silence produces no calls at all instead of sixty a second, and it fires once more on the way back to `0` when audio stops so a meter cannot stick at its last value.

That throttle is tested and mutation-tested: one case asserts the counter is still `0` after a second of nothing driving the orb, and asserts fewer than 60 further calls while a *steady* tone holds. Removing the threshold so it fires every frame fails that case and only that case.

I also caught a silent no-op while doing this, which is worth recording given how this PR started. My edit to the demo targeted a multi-line `<VoiceOrb .../>` that Prettier had already collapsed onto one line, so the replacement matched nothing and `onlevel` was never wired up — the tests failed on a `0.00` readout and found it. The earlier edits in this PR that went through without an assertion are exactly how the blank-screenshot defect got in; every edit script in this round asserts its anchor matched.

Re-verified at `a6dd380`:

```
pnpm run check        exit 0
pnpm run lint         exit 0
vitest                exit 0     1072 tests
playwright            exit 0     726 integration
scripts/visual-test   exit 0     94 / 94, stable across two runs
```

The recordings in the comment above are from `d135a27`; the only change since is `onlevel` and its docs, which alters nothing visible in them.

Tara-ag (changes requested):
Changes requested. Two MAJOR issues remain: destructured props lose reactivity inside the rAF loop, and the continuous ambient animation does not honor `prefers-reduced-motion`. Three MINOR documentation/test gaps and one accessibility suggestion are also noted.

murdore:
## Re-pushed at `376a5cc` — reduced motion added, and two findings answered rather than accepted

The verdict above reviewed an earlier commit and ended early (`time-limit`, 14 steps). Mapping each finding against `376a5cc`:

| Finding | State |
| --- | --- |
| MAJOR — ignores `prefers-reduced-motion` | **Fixed.** `prefersReducedMotion()` now gates the decorative motion, with `tests/voice-orb-reduced-motion.test.ts` covering it. |
| MINOR — no reduced-motion test coverage | **Fixed.** Same file, 138 lines. |
| MAJOR — props not reactive inside `onMount`/rAF | **Refuted, then changed anyway.** See below. |
| MAJOR — `localStorage` without guards | **Refuted.** `readPersisted`/`writePersisted` are called only from `onMount`, which Svelte runs in the browser only, and both already sit in try/catch for the storage-disabled case. |
| MINOR — wc callbacks typed `Object` | **Working as intended.** Function-valued props are assigned as properties, not attributes; that is inherent to custom elements. The repo's own `prop-parity` guard requires every prop be declared, which is why they are listed. |

### The reactivity finding is worth reading

I checked it by compiling `VoiceOrb.svelte` with the repo's own `svelte/compiler` and reading the output: destructured `$props()` become getter calls (`variant()`), so reads inside `draw()` are **already live**. The finding is wrong about Svelte 5.

I made `particles` a `$derived.by` keyed on `particleCount`/`radius`/`seed` regardless — not because the finding was right, but because regenerating them on a keyed derivation is clearer than the alternative.

### The design decision worth flagging

Reduced motion suppresses the **decorative** animation only. It does **not** suppress the audio-driven response, because that conveys state rather than decoration — a user who has asked for less motion still needs to see that the microphone is hearing them. Asserted directly:

```
'keeps following real audio under reduced motion -- it is state, not decoration'
```

### On the proof

The audio-reactivity claim is now carried by unit assertions as well as the clips: digital silence reads exactly `0`, a full-scale square wave `> 0.98`, a half-amplitude wave `≈ 0.5`, sign-insensitive, empty buffer `0` rather than `NaN`, and the envelope's attack more than twice its release without overshoot.

All ten recordings on this PR were re-downloaded today and put through a gate that fails a clip unless it is both non-blank and moving. All ten pass. That gate exists because the first proof I attached here was blank and I had not watched it.

### Verification at `376a5cc`

```
check       exit 0
lint        exit 0
vitest      1094 passed
functional  728 passed
```

One functional failure in that run was `typewriter-text-reduced-motion`, which this branch does not touch. Control: it fails 1 run in 9 here and **3 in 9** on a branch with no VoiceOrb in it. Pre-existing and load-induced, not a regression.

murdore:
## Review threads worked through at `cf1e5bf`

8 unresolved threads: 1 fixed · 3 refuted · 4 already correct.


### **Fixed**

**`docs/VoiceOrb.md:116`** — CodeRabbit: doc claims 'Any CSS colour works' for the particle-color custom property, overstating what a canvas 2D context actually accepts.

> Edited docs/VoiceOrb.md lines 116-119: 'Any CSS colour works' -> 'Any CSS colour your canvas accepts works', and 'An unreadable value falls through' -> 'A colour syntax the canvas does not recognise falls through', preserving the already-documented inherited-colour/mid-grey fallback. Verified with `git diff docs/VoiceOrb.md` after the edit and amended into the PR commit.


### **Refuted**

**`src/lib/VoiceOrb/VoiceOrb.svelte:24`** — Reviewer claimed destructured $props() (variant, speedMultiplier, etc.) are read as stale snapshots inside the draw() rAF closure because they are "plain lets".

> Compiled a minimal repro with the installed svelte@5.56.3 compiler (svelte.compile(..., {generate:'client'})): `let { variant = 'idle' } = $props()` lowers to `let variant = $.prop($$props, 'variant', 3, 'idle')`, and every read of `variant` anywhere in the component -- including inside a nested `function draw(){ requestAnimationFrame(draw); ... }` -- compiles to a getter call `variant()`. Each animation frame therefore re-reads the live prop via the getter; there is no captured JS value to go stale. Same mechanism applies to speedMultiplier/seed/etc. in the real VoiceOrb.svelte destructuring at lines 23-39.

**`src/wc/components/VoiceOrb.wc.svelte:8`** — Reviewer said the web component's `seed: {type:'String'}` diverges from the Svelte component's numeric seed handling and suggested documenting the limitation.

> src/lib/VoiceOrb/orbMath.ts:58-66 `resolveExplicitSeed` explicitly parses a numeric-looking string back into the same number (`Number(seed)` then `>>> 0`) before ever hashing, hashing only genuinely non-numeric strings. src/lib/VoiceOrb/orbMath.test.ts:60 asserts `expect(resolveExplicitSeed('42')).toBe(resolveExplicitSeed(42))`. So `<sui-voice-orb seed="42">` (string attribute) and `<VoiceOrb seed={42} />` (number prop) resolve to the identical seed -- no divergence to document.

**`src/wc/components/VoiceOrb.wc.svelte:30`** — CodeRabbit duplicate of the seed-type finding: numeric-looking seed attributes should be normalized to numbers before reaching resolveSeed.

> Identical to the earlier seed finding: orbMath.ts:58-66 resolveExplicitSeed already normalizes numeric-looking strings to the same integer as the numeric prop, and orbMath.test.ts:60 (`resolveExplicitSeed('42') === resolveExplicitSeed(42)`) proves it. No code change needed in the wc wrapper.


### **Already correct**

**`tests/voice-orb.test.ts:1`** — Reviewer said VoiceOrb has no test verifying the ambient animation pauses under prefers-reduced-motion, unlike TypewriterText's reduced-motion test.

> tests/voice-orb-reduced-motion.test.ts already exists in the worktree with 3 Playwright tests: 'freezes idle rotation', 'freezes the sine-wave listening breathing when there is no analyser', and 'does not silence audio-driven motion, since that is state rather than decoration'. It uses page.emulateMedia + two-sample pixel/alpha-signature comparison, the exact technique the reviewer suggested.

**`src/lib/VoiceOrb/VoiceOrb.svelte:333`** — Optional suggestion that a decorative canvas could also carry role='img' and an empty aria-label instead of just aria-hidden.

> The review comment is a non-blocking suggestion that explicitly says 'The current approach is defensible; this is optional.' Current markup at VoiceOrb.svelte:366 already has aria-hidden="true" on the canvas, matching the reviewer's own accepted baseline; no change requested to be applied.

**`src/lib/VoiceOrb/VoiceOrb.svelte:1`** — Reviewer said VoiceOrb never queries prefersReducedMotion() at all, unlike other components in the library.

> VoiceOrb.svelte:20 imports `prefersReducedMotion` from '../utils'; line 240 calls it once per frame inside draw() and threads the result into resolveMotionSpeed/resolveMotionScale. orbMath.ts:306-323 and :326-339 show resolveMotionSpeed returns 0 and resolveMotionScale returns 1 when reducedMotion is true (for the non-audio-driven paths), while clearRect/render still execute every frame (line 258) so a static frame stays visible.

**`src/lib/VoiceOrb/VoiceOrb.svelte:250`** — CodeRabbit duplicate of the reduced-motion finding, claiming draw() unconditionally advances angleX/angleY and listeningElapsedMs with prefersReducedMotion never imported.

> Same code path as the line:1 finding above -- VoiceOrb.svelte:20 and :240 import/call prefersReducedMotion, and orbMath.ts's resolveMotionSpeed (used at VoiceOrb.svelte:241-250 to compute `speed` before `orb.angleX += BASE_ROTATION_X * speed * normalizedDelta`) already zeroes speed under reduced motion for non-audio variants; the comment's own anchor line falls inside code that already depends on the reducedMotion-gated `speed` value.


### Verification

```
check   exit 0
lint    exit 0
vitest  Test Files  54 passed (54); Tests  1094 passed (1094)
```

Tara-ag (commented):
Review in progress — filing findings as they are verified.

src/wc/index.ts
Tara-ag, line 1: 🔒 CRITICAL: VoiceOrb web component is not registered here. `src/wc/components/VoiceOrb.wc.svelte` exists but is never imported, so `customElements.define('sui-voice-orb')` is never invoked and `<sui-voice-orb>` is undefined for web-component consumers. Add `import './components/VoiceOrb.wc.svelte';` alongside the other component imports.

Tara-ag (changes requested):
Blocking: the new VoiceOrb web component is not registered in `src/wc/index.ts`, so `<sui-voice-orb>` is undefined for web-component consumers. Please import the wrapper in `src/wc/index.ts` alongside the other components.

murdore:
## Refuted — `sui-voice-orb` is registered, and it works

> `src/wc/components/VoiceOrb.wc.svelte` exists but is never imported, so `customElements.define('sui-voice-orb')` is never invoked

The import is there, and this PR is what added it:

```
$ grep -n VoiceOrb src/wc/index.ts
29:import './components/VoiceOrb.wc.svelte';

$ git diff a88a8ff..HEAD -- src/wc/index.ts
+import './components/VoiceOrb.wc.svelte';
```

Rather than rest on the import line, I checked the built artefact and the runtime. Serving `dist-wc/index.js` into a blank page and driving it in Chromium:

```
sui-voice-orb defined:  true
shadow children:        3
canvas present:         true
page errors:            0
```

It registers, upgrades, and paints its canvas.

One note on my own method, since it nearly produced a wrong answer in the other direction: my first probe reported `shadow initialised: false`, which looks like a defect. It was not — the probe appended the element and read `shadowRoot.childNodes` in the same synchronous expression, before Svelte had upgraded it. Waiting for the upgrade gives the numbers above. A probe that measures too early reports a bug that is not there, exactly as a probe that measures nothing reports a clean run.

No change made.

murdore:
Yama BLOCKED this head with 2 CRITICAL + 1 MAJOR + 1 MINOR. I checked all four against the source and none of them hold — and the run log explains why, so leading with that:

> ⚠️ Review loop ended early (stopReason=time-limit) — result is PARTIAL.

The review did not finish reading the tree. Both CRITICALs are "X is missing", which is exactly the failure mode a truncated read produces: it never reached the files.

**CRITICAL — "VoiceOrb default export missing from `src/lib/index.ts`"** — it is there, twice:

```
$ grep -n "VoiceOrb" src/lib/index.ts
40:export { default as VoiceOrb } from './VoiceOrb/VoiceOrb.svelte';
195:export type * from './VoiceOrb/properties';
```

**CRITICAL — "Web component registration missing from `src/wc/index.ts`"** — also there, and the wrapper exists:

```
$ grep -n -i voice src/wc/index.ts
29:import './components/VoiceOrb.wc.svelte';
$ ls src/wc/components/ | grep -i voice
VoiceOrb.wc.svelte
```

**MAJOR — "Canvas state is invisible to assistive technology"** — the canvas is `aria-hidden="true"` deliberately, and `docs/VoiceOrb.md` has a section saying so and saying what a host must do instead:

> The canvas is `aria-hidden`, because it carries no information a screen reader could use. **The orb is not a status indicator.** If it is showing that the microphone is live, that state must also be announced somewhere a screen reader can reach it — a live region, or the label on the control that toggles it.

Putting an `aria-live` region inside a decorative visualiser would be the wrong fix: the component does not know whether the microphone is open, only what amplitude it was handed, and it would double-announce against the host's own status control.

**MINOR — "Non-standard `testID` attribute leaked to DOM"** — it is the repo-wide convention, not something introduced here:

```
$ grep -rl "testID={testId}" src/lib/*/*.svelte | wc -l
28
```

28 other components emit it alongside `data-pw`. Changing it in one component would break parity, not fix a leak.

Verified on this head (`655bd67`, rebased onto current release): `lint` exit 0, `check` exit 0, `vitest` 1189/1189 passing, `build` exit 0.

No code change. Requesting a re-run rather than a re-review of the same partial result — if a complete run still flags any of these, I will act on it.

Tara-ag (commented):

src/lib/VoiceOrb/VoiceOrb.svelte
```diff
@@ -0,0 +1,382 @@
...
+  onMount(() => {
+    orb.angleX = readPersisted('angleX') ?? 0;
+    orb.angleY = readPersisted('angleY') ?? 0;
+    listeningElapsedMs = readPersisted('listeningMs') ?? 0;
+
+    resize();
+
+    const observer =
+      typeof ResizeObserver === 'function' ? new ResizeObserver(() => resize()) : null;
+    if (observer !== null && container !== null) {
```
Tara-ag, line 341: 💡 MINOR: Non-standard testID attribute emitted on the root div. The root div sets both data-pw and testID; the capital-D testID is not a valid HTML attribute and can trigger framework warnings. data-pw already serves the test-selector purpose, so remove testID={...}.

src/lib/VoiceOrb/VoiceOrb.svelte
```diff
@@ -0,0 +1,382 @@
...
+      observer.observe(container);
+    }
+
+    frame = requestAnimationFrame(draw);
```
Tara-ag, line 357: ⚠️ MAJOR: Interactive canvas has no accessible name, role, or keyboard alternative. The <canvas> responds to pointer movement (particle repulsion) but is marked aria-hidden="true", making it unreachable to screen-reader and keyboard users. When interactive is true, expose the canvas with role="img" and an aria-label (e.g. "Voice activity visualizer") or move interaction behind a focusable element with keyboard handling.

src/lib/VoiceOrb/VoiceOrb.svelte
```diff
@@ -0,0 +1,382 @@
...
+    }
+    // Read once per frame and threaded through both calls below: two
+    // `matchMedia` reads per frame is wasted work `draw` already runs 60 times
+    // a second, and could observe the preference flip between them.
+    const reducedMotion = prefersReducedMotion();
+    const speed = resolveMotionSpeed(
+      speedMultiplier,
+      variant,
+      listeningElapsedMs,
+      level,
```
Tara-ag, line 241: ⚠️ MAJOR: prefers-reduced-motion is queried on every animation frame. Calling prefersReducedMotion() inside draw() invokes window.matchMedia('(prefers-reduced-motion: reduce)') at 60 Hz, which is unnecessary work and can cause style recalc churn. Cache the preference in a $state or module-level variable, update it via a matchMedia change listener inside onMount, and read the cached value in draw().

Tara-ag (commented):
Review complete. The PR introduces VoiceOrb with good test coverage and addresses prior review feedback (reactivity, reduced-motion support, web-component registration). I found three remaining quality issues: an interactive canvas that is aria-hidden and lacks keyboard access, a per-frame matchMedia query in the animation loop, and a non-standard `testID` attribute. None of these are blocking regressions to the public API or security, so I am approving with comments.

src/lib/VoiceOrb/VoiceOrb.svelte
murdore: Confirmed and fixed — you are right, and it was measurable rather than theoretical.

`prefersReducedMotion()` builds a fresh `MediaQueryList` on every call:

```ts
export function prefersReducedMotion(): boolean {          // src/lib/utils.ts:327
  return (
    typeof window !== 'undefined' &&
    typeof window.matchMedia === 'function' &&
    window.matchMedia('(prefers-reduced-motion: reduce)').matches
  );
}
```

and `draw` called it once per frame. I wrote the test before the fix, so the number is measured, not argued — four hand-driven frames, four calls:

```
AssertionError: expected 4 to be 1
```

**The fix.** The preference is read once and kept current by the query's own `change` event, in `onMount` (this repo's eslint bans `$effect`, so the listener lives with the other mount-scoped subscriptions and is torn down beside them):

```ts
let reducedMotion = prefersReducedMotion();
// in onMount:
const motionQuery = typeof window.matchMedia === 'function'
  ? window.matchMedia('(prefers-reduced-motion: reduce)') : null;
const onMotionChange = (event: MediaQueryListEvent): void => { reducedMotion = event.matches; };
if (motionQuery !== null) {
  reducedMotion = motionQuery.matches;
  motionQuery.addEventListener('change', onMotionChange);
}
// in the cleanup: motionQuery?.removeEventListener('change', onMotionChange);
```

**Two tests, because caching alone is the wrong fix.** The obvious over-correction is to cache the value and never update it, which would pass a call-count test while silently ignoring a user toggling the setting mid-session. So one test pins the cost and one pins the freshness, and I verified each fails for its own reason:

- `does not re-query matchMedia as frames go by` — fails on the old code (`expected 4 to be 1`).
- `still follows the preference when it changes mid-session` — I removed just the `addEventListener` line and re-ran: `expected false to be true`. It fails on a cache-without-listener.

`docs/VoiceOrb.md` said "the check happens inside the render loop itself", which this makes untrue, so that sentence is updated too.

Gate on the fix: lint 0, check 0, 1191/1191 (the two new tests), build 0.

src/lib/VoiceOrb/VoiceOrb.svelte
murdore: Refuting this one, though it is the finding I thought hardest about — the premise that the canvas is *interactive* in the accessibility sense does not hold.

`trackPointer` does exactly one thing:

```ts
const trackPointer = (event: PointerEvent): void => {
  if (canvas === null || !interactive) { return; }
  const rect = canvas.getBoundingClientRect();
  pointer.x = event.clientX - rect.left;
  pointer.y = event.clientY - rect.top;
};
```

and `pointer` is read in exactly one place — the particle loop, to push particles away from the cursor:

```ts
if (interactive && pointer.x !== null && pointer.y !== null) {   // VoiceOrb.svelte:274
```

No state changes, no action is performed, nothing is selected, no value is produced, and no callback fires. It is a decorative flourish on a decorative graphic. There is no functionality behind it for a keyboard alternative to reach — adding `tabindex` would put a stop on the tab order that does nothing when activated, which is a WCAG 2.4.3 problem rather than a fix.

`aria-hidden="true"` is therefore the correct treatment: the canvas is not focusable and contains nothing focusable, so there is no `aria-hidden-focus` violation, and nothing reachable is being hidden.

On the suggested `role="img"` + `aria-label="Voice activity visualizer"`: that would be actively worse, and the docs already commit to why:

> The canvas is `aria-hidden`, because it carries no information a screen reader could use. **The orb is not a status indicator.** If it is showing that the microphone is live, that state must also be announced somewhere a screen reader can reach it — a live region, or the label on the control that toggles it.
> — `docs/VoiceOrb.md`, Accessibility

A static label naming it a "visualizer" announces a thing that conveys nothing, and implies the orb reports microphone state — which is the exact misreading the docs warn callers away from. The orb is audio-*reactive*, so a truthful accessible name would have to be live-updating; the component instead hands that to the caller through `onlevel` (`properties.ts:108`), who owns the live region and knows what the app is actually doing.

Happy to reconsider if you think the decorative-graphic reading is wrong — but the fix would then be to make the orb genuinely convey status, not to label a decoration.

src/lib/VoiceOrb/VoiceOrb.svelte
murdore: Refuting this one — `testID` is a deliberate repo-wide convention, not something this PR introduced.

You are right that `testID` is not a standard HTML attribute. But it is emitted on purpose, by every component in the library, and removing it from VoiceOrb alone would make this the one component a consumer's test harness cannot find the same way as the other 97.

```
$ grep -rl "testID={" src/lib/ | wc -l
98
```

It was added deliberately, as its own fix:

```
$ git log --oneline --all --grep="testID" | head -1
b543c3c fix: emit a native testID alongside data-pw for every testId
```

and it is part of the documented public contract, not an internal detail:

```
docs/Carousel.md:27  | testId | `string` | No | `-` | `data-pw`/`testID` on the root element.
docs/Card.md:207     ... `testId` for `data-pw`/`testID` ...
```

VoiceOrb's line is character-for-character the shape every other component uses:

```svelte
testID={typeof testId === 'string' ? testId : null}
```

On the framework-warning concern specifically: it renders as a lowercase `testid` attribute and Svelte emits no warning for it — `npm run lint` and `npm run check` are both clean on this branch. If the convention should go, that is a 98-component change with a migration in `scripts/migrate/`, not a one-file edit here.

src/lib/VoiceOrb/VoiceOrb.svelte
murdore: Checked against the merged source, and this is half right — so I've fixed that half and am refuting the other, rather than doing both or neither.

**Right: the state was unreachable.** The orb exists to say a microphone is open, and no assistive technology could learn that. `docs/VoiceOrb.md` already told callers to build their own live region; there was no built-in way. Fixed in #619 with an optional `statusLabels` prop — the caller's words per variant, rendering a visually hidden `aria-live="polite"` region on the pattern Carousel already uses. Opt-in with no default, because the same animation means "listening" on one surface and "recording" on another, and a component that guessed would announce the wrong thing confidently.

**Refuting `role="img"` + `aria-label`:** that is worse here, not better. A static label on a *changing* state indicator announces once and never updates. A live region is what conveys a state change; `role="img"` would report the orb exists and then go silent exactly when the state moves.

**Refuting the keyboard alternative:** the pointer interaction is particle repulsion. It pushes points aside and accomplishes nothing else — no action, no information, no state. A focus stop for it would be a tab stop that leads nowhere, which is a worse outcome for a keyboard user than no stop at all. There is no functionality being withheld, which is precisely why the canvas can stay `aria-hidden`. That is now written down as a decision in both the prop docs and the Accessibility section, rather than left as an unexplained omission.

https://github.com/juspay/svelte-ui-components/pull/619

PR #603, opened 2026-09-11
feat(markdown-text): let a consumer add a class to the table wrapper

murdore:
`wrapTables` hardcoded `markdown-table-wrapper`, so a consumer whose own
stylesheet targets a different class silently lost all table styling when
they swapped to this renderer -- no error, no warning, just unstyled
tables. `tableWrapperClass` on `MarkdownTextProperties` and
`RenderMarkdownOptions` fixes that, following how the existing
`tableLabel` is typed, threaded and defaulted.

The class is ADDED, not substituted. That distinction is the whole
design. The wrapper is a `tabindex="0"` scroll container, and the three
things that make it one -- `overflow-x: auto`, the `:focus-visible` ring
and the table's `max-content` width -- are scoped to
`.markdown-table-wrapper` in the component's own styles. A Svelte scoped
selector cannot be templated to a class name supplied at runtime, so
substituting would leave a focusable element that cannot scroll and shows
no focus ring: a keyboard dead end, and a direct regression of 5cb4df4,
which made a wide table scroll in a focusable wrapper rather than
scrolling the table itself.

Adding costs the consumer nothing. Their selector matches the wrapper
exactly as it would have, and specificity still lets them override any
built-in declaration. `markdown-table-wrapper` is documented as a stable
public class besides, so theming against it directly remains an option
and the prop is not the only route.

The value is caller-supplied and lands inside a double-quoted HTML
attribute on a surface that is otherwise sanitizing-by-construction, so
it is escaped exactly as `tableLabel` is. A value of `" onmouseover=
"alert(1)` becomes `&quot;` rather than closing the attribute and
injecting a live event handler.

Omitting the prop, or passing an empty or whitespace-only string, leaves
the markup byte-identical to before this change.

Verified on this commit: check 0, lint 0, 1027 unit tests, 714
integration, build 0. Mutation-tested: substituting the class instead of
adding it fails 4 cases, and dropping the escape fails the injection
case and only that one.

## Summary by CodeRabbit

- **New Features**
  - Added an optional custom CSS class for wrappers around rendered Markdown tables.
  - The built-in `markdown-table-wrapper` class is always retained.
  - Custom classes are trimmed and safely escaped; empty values are ignored.
  - Exposed the option through the Markdown Text component and custom element attribute.

- **Documentation**
  - Added API documentation for the new table wrapper class option.

- **Tests**
  - Added coverage for custom, empty, combined, omitted, and safely escaped class values.

### DIFF
--- a/docs/MarkdownText.md
+++ b/docs/MarkdownText.md
@@ -36,14 +36,15 @@ The `marked` package is a peer dependency, needed only when you actually use `Ma
 
 ## Props
 
-| Prop       | Type                      | Required | Default | Description                                                                                                                                                                                                         |
-| ---------- | ------------------------- | -------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| markdown   | `string`                  | Yes      | `-`     | Markdown source. Raw HTML is escaped by default; strip mode removes tags and retains text; unsafe link/image protocols are stripped while their text stays.                                                         |
-| breaks     | `boolean`                 | No       | `false` | Render single newlines as `<br>` (GFM "breaks" mode).                                                                                                                                                               |
-| testId     | `string`                  | No       | `-`     | `data-pw` (and native `testID`) on the root element.                                                                                                                                                                |
-| classes    | `string`                  | No       | `-`     | Class string on the root element.                                                                                                                                                                                   |
-| tableLabel | `string`                  | No       | `-`     | Names the scrollable table region: adds `role="region"` and `aria-label` to the wrapper. Without it the wrapper is still focusable and scrollable but announces no landmark.                                        |
-| sanitize   | `MarkdownSanitizeOptions` | No       | `-`     | Narrows the link/image protocol allow-list, the set of tags allowed to render as themselves, whether task-list checkboxes render, and whether raw HTML is escaped or dropped. See the Security model section below. |
+| Prop              | Type                      | Required | Default | Description                                                                                                                                                                                                         |
+| ----------------- | ------------------------- | -------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| markdown          | `string`                  | Yes      | `-`     | Markdown source. Raw HTML is escaped by default; strip mode removes tags and retains text; unsafe link/image protocols are stripped while their text stays.                                                         |
+| breaks            | `boolean`                 | No       | `false` | Render single newlines as `<br>` (GFM "breaks" mode).                                                                                                                                                               |
+| testId            | `string`                  | No       | `-`     | `data-pw` (and native `testID`) on the root element.                                                                                                                                                                |
+| classes           | `string`                  | No       | `-`     | Class string on the root element.                                                                                                                                                                                   |
+| tableLabel        | `string`                  | No       | `-`     | Names the scrollable table region: adds `role="region"` and `aria-label` to the wrapper. Without it the wrapper is still focusable and scrollable but announces no landmark.                                        |
+| tableWrapperClass | `string`                  | No       | `-`     | An extra class on the scrollable table wrapper, added alongside the built-in `markdown-table-wrapper` rather than replacing it. See Wide tables below.                                                              |
+| sanitize          | `MarkdownSanitizeOptions` | No       | `-`     | Narrows the link/image protocol allow-list, the set of tags allowed to render as themselves, whether task-list checkboxes render, and whether raw HTML is escaped or dropped. See the Security model section below. |
 
 ## Security model
 
@@ -120,6 +121,39 @@ That adds `role="region"` and `aria-label`. Without it the wrapper stays
 keyboard-scrollable but announces no landmark, since a region the user cannot
 identify is worse than no region at all.
 
+`markdown-table-wrapper` is itself a stable public class name — style it
+directly in your own stylesheet if you would rather theme the wrapper that way
+than reach for CSS variables. If your app already has its own class for this
+(its own scroll/border/token treatment) and you want the rendered markup to
+carry that class instead, pass `tableWrapperClass`:
+
+```svelte
+<MarkdownText {markdown} tableWrapperClass="markdown-table-scroll" />
+```
+
+The wrapper then carries both: `class="markdown-table-wrapper markdown-table-scroll"`.
+
+**It is added, never substituted, and that is deliberate.** The wrapper is a
+`tabindex="0"` scroll container, and the three things that make it one —
+`overflow-x: auto`, the `:focus-visible` ring, and the table's `max-content`
+width — are scoped to `.markdown-table-wrapper` in the component's own styles.
+A Svelte scoped selector cannot be templated to a class name supplied at
+runtime, so replacing the class would leave a focusable element that cannot
+scroll and shows no focus ring: a keyboard dead end, and a regression of the
+fix that made wide tables scrollable in the first place.
+
+Adding costs you nothing. Your own selector matches the wrapper exactly as it
+would have, and you can still override any built-in declaration by specificity
+if you want different scroll or focus treatment.
+
+`markdown-table-wrapper` is a stable public class. Theme against it directly if
+you would rather not pass a prop at all.
+
+Omitting `tableWrapperClass`, or passing an empty or whitespace-only string,
+leaves the wrapper with just `markdown-table-wrapper`. The value is
+HTML-attribute-escaped the same way `tableLabel` is, so it cannot break out of
+the `class="..."` attribute it is placed in.
+
 ## CSS Variables
 
 | Variable                                  | Default                   | Description                    |
@@ -153,6 +187,7 @@ export type MarkdownTextProperties = {
   testId?: string;
   classes?: string;
   tableLabel?: string;
+  tableWrapperClass?: string;
   sanitize?: MarkdownSanitizeOptions;
 };
 
@@ -178,6 +213,7 @@ export type RenderMarkdownOptions = {
   breaks?: boolean;
   inline?: boolean;
   tableLabel?: string;
+  tableWrapperClass?: string;
   sanitize?: MarkdownSanitizeOptions;
 };
 ```
--- a/src/lib/MarkdownText/markdown.test.ts
+++ b/src/lib/MarkdownText/markdown.test.ts
@@ -226,4 +226,59 @@ describe('renderMarkdown — table wrapping', () => {
   it('leaves markdown without a table untouched', () => {
     expect(renderMarkdown('plain **text**')).not.toContain('markdown-table-wrapper');
   });
+
+  it('adds tableWrapperClass so a consumer selector matches the wrapper', () => {
+    const output = renderMarkdown(table, { tableWrapperClass: 'markdown-table-scroll' });
+    expect(output).toContain(
+      '<div class="markdown-table-wrapper markdown-table-scroll" tabindex="0"><table>'
+    );
+  });
+
+  it('falls back to the default wrapper class when tableWrapperClass is empty', () => {
+    const output = renderMarkdown(table, { tableWrapperClass: '' });
+    expect(output).toContain('<div class="markdown-table-wrapper" tabindex="0"><table>');
+  });
+
+  it('falls back to the default wrapper class when tableWrapperClass is whitespace-only', () => {
+    // Pins that a whitespace-only value is trimmed to empty rather than
+    // surviving as a meaningless custom class, matching the documented
+    // "empty string keeps the default" behaviour for the empty case above.
+    const output = renderMarkdown(table, { tableWrapperClass: '   ' });
+    expect(output).toContain('<div class="markdown-table-wrapper" tabindex="0"><table>');
+  });
+
+  it('combines a custom tableWrapperClass with tableLabel', () => {
+    const output = renderMarkdown(table, {
+      tableWrapperClass: 'markdown-table-scroll',
+      tableLabel: 'Recent orders'
+    });
+    expect(output).toContain(
+      '<div class="markdown-table-wrapper markdown-table-scroll" tabindex="0" role="region" aria-label="Recent orders"><table>'
+    );
+  });
+
+  it('keeps the built-in wrapper class alongside a custom one', () => {
+    // The wrapper is a `tabindex="0"` scroll container (5cb4df4). Its
+    // overflow-x, its focus ring and the table's max-content width are all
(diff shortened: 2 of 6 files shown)


Tara-ag (commented):
Reviewed file by file. The change is additive and backward-compatible: it introduces an optional `tableWrapperClass` prop/option that defaults to the existing `markdown-table-wrapper` class, mirrors the existing `tableLabel` escaping pattern, and is wired through the Svelte component, `renderMarkdown` utility, web-component wrapper, docs, and tests. Two minor notes inline.

docs/MarkdownText.md
```diff
@@ -36,14 +36,15 @@ The `marked` package is a peer dependency, needed only when you actually use `Ma
...
+| Prop               | Type                      | Required | Default | Description                                                                                                                                                                                                            |
+| ------------------ | ------------------------- | -------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| markdown           | `string`                  | Yes      | `-`     | Markdown source. Raw HTML is escaped by default; strip mode removes tags and retains text; unsafe link/image protocols are stripped while their text stays.                                                         |
+| breaks             | `boolean`                 | No       | `false` | Render single newlines as `<br>` (GFM "breaks" mode).                                                                                                                                                                 |
```
Tara-ag: 💡 MINOR: The Props table separator line is shorter than the widest Description cell (the `tableWrapperClass` row extends past the dashes). Malformed Markdown tables render with ragged columns in some viewers. Extend the Description separator dashes to match the longest cell in that column.

src/lib/MarkdownText/markdown.test.ts
```diff
@@ -226,4 +226,34 @@ describe('renderMarkdown — table wrapping', () => {
   it('leaves markdown without a table untouched', () => {
     expect(renderMarkdown('plain **text**')).not.toContain('markdown-table-wrapper');
   });
+
+  it('lets tableWrapperClass override the default wrapper class', () => {
```
Tara-ag: 💬 SUGGESTION: Consider adding a test for whitespace-only `tableWrapperClass` (e.g. `'   '`). The implementation currently uses `length > 0`, so spaces become a custom class and the default styling is lost, which may be surprising since empty string is documented to keep the default.

Tara-ag (commented):
Reviewed file by file. The change is additive and backward-compatible: it introduces an optional `tableWrapperClass` prop/option that defaults to the existing `markdown-table-wrapper` class, mirrors the existing `tableLabel` escaping pattern, and is wired through the Svelte component, `renderMarkdown` utility, web-component wrapper, docs, and tests. Two minor notes inline.

murdore:
## Verified independently — no recording, because there is nothing to watch

This change alters generated HTML, not anything on screen, so attaching a video would pad the PR rather than evidence it. What follows is measured instead.

### The escaping is real, proven by deleting it

The wrapper class is caller-supplied and lands inside a double-quoted HTML attribute on a surface that is otherwise sanitizing-by-construction. That is the one genuinely dangerous part of this change, and it is the part most likely to be skipped.

I removed `escapeHtml` from the class attribute and re-ran:

```
MUTATION: const classAttr = resolvedClass;   (escaping deleted)
  -> 1 failed | 33 passed

RESTORED
  -> 34 passed
```

Exactly one test died — the one asserting that `" onmouseover="alert(1)` is escaped to `&quot;` rather than closing `class="` early and injecting a live event handler. That is an assertion with teeth, not a comment claiming safety.

### The default does not move

`tableWrapperClass` defaults to `markdown-table-wrapper`, and an empty string falls back to it rather than emitting `class=""`. Existing consumers render byte-identically. `.markdown-table-wrapper` is now documented as a stable public contract, so a consumer can theme against it instead of overriding — both halves of the ask, so the reporting team needs no CSS change on their side either way.

### Verification I ran myself on `44e9a25`

```
npm run check     0 errors
npm run lint      clean
npx vitest run    1025 / 1025
npm run build     exit 0
```

One caveat stated rather than buried: the first `vitest` run on this branch reported 1 failure, and the rerun was clean at 1025/1025. Another session was running a Playwright container against a different worktree at the time, and the repo's `scripts/migrate/*` tests scan the whole tree under a 5s timeout — the same tests timed out under the same contention on a different branch earlier today and passed 75/75 in isolation. I am reporting the transient failure rather than only the clean rerun.

Tara-ag (commented):
Reviewed file-by-file. Two minor non-blocking findings.

src/lib/MarkdownText/markdown.test.ts
Tara-ag: 💬 SUGGESTION: Consider adding a test for whitespace-only `tableWrapperClass` (e.g. `'   '`). The implementation treats it as a custom class, while the docs say empty string keeps the default. A whitespace-only value is neither empty nor a meaningful class and could surprise consumers. Either trim before the length check, or add a test pinning the current behavior.

docs/MarkdownText.md
Tara-ag: 💡 MINOR: Docs table column alignment is off. The `Description` separator is shorter than the widest cell in that column, so the Markdown table renders with ragged columns in some viewers. Extend the dashes under `Description` to match the longest cell, e.g. `| ------------------ | ------------------------- | -------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------- |`.

Tara-ag (commented):
Reviewed file-by-file. The PR is a clean, backward-compatible addition of an optional `tableWrapperClass` prop/option. No blocking issues found. Two minor non-blocking notes posted inline.

murdore:
## Corrected at `edf35af` — the class is now added, not substituted

The agent that built this flagged a consequence in its report, which arrived after I had already opened this PR. It was right, and the consequence is worse than it described. Recording the correction rather than quietly force-pushing it.

### What was wrong

`tableWrapperClass` replaced the wrapper's class. The report noted that the component's built-in wrapper CSS is scoped to the literal `.markdown-table-wrapper`, so a consumer setting the prop "takes over all wrapper styling, not just colours."

That undersells it. The wrapper is emitted with `tabindex="0"` unconditionally, and the three declarations scoped to that class are:

```css
.markdown-text :global(.markdown-table-wrapper)               { overflow-x: auto; }
.markdown-text :global(.markdown-table-wrapper:focus-visible) { outline: 2px solid ...; }
.markdown-text :global(.markdown-table-wrapper table)         { width: max-content; }
```

Substituting the class leaves a **focusable element that cannot scroll and shows no focus ring**. The comment sitting above that focus rule in the source says it plainly — *"a tab stop with no visible ring is a keyboard user's dead end."*

This repo already fixed exactly that, in `5cb4df4 fix(MarkdownText): scroll a wide table in a focusable wrapper, not the table`. Shipping replace semantics would have handed anyone who used the new prop a silent regression of that fix — and the reporting team, whose whole reason for wanting the prop is wide tables, would have been first in line for it.

### What it does now

```
class="markdown-table-wrapper markdown-table-scroll"
```

The consumer's selector matches exactly as it would have, the scroll container and focus ring keep working, and specificity still lets them override any built-in declaration. Adding costs them nothing; substituting cost them keyboard access.

`markdown-table-wrapper` is documented as a stable public class either way, so theming against it directly remains an option and this prop is not the only route.

### Re-mutated, since the behaviour changed

```
MUTATION: substitute instead of add   -> 4 failed | 32 passed
MUTATION: drop escapeHtml on the class -> 1 failed | 35 passed
RESTORED                               -> 36 passed
```

The escaping is unchanged and still proven by deletion: `" onmouseover="alert(1)` escapes to `&quot;` instead of closing the attribute. The built-in class name is a literal and needs no escaping; only the caller's value is escaped.

### Re-verified on `edf35af`

```
npm run check     0 errors
npm run lint      clean
npx vitest run    1027 / 1027
npx playwright    714 passed, 1 skipped
npm run build     exit 0
```

One integration failure appeared in the full run — `status-default-icon.spec.ts` — and passed **25/25** when I re-ran it in isolation. It is load-dependent and unrelated to this change; another session was running Playwright containers against a different worktree throughout. Reporting it rather than only the isolated pass.

### Why this is in the PR rather than silently amended

I opened this PR having verified the escaping by deleting it, which was the risk I had told the agent to prioritise — and I missed this one, which was in its report and which I had not yet read. The fix was correct; the contract around it was not. Worth stating because "I verified the dangerous part" is not the same as "I verified the change."

src/lib/MarkdownText/properties.ts
```diff
@@ -23,6 +23,15 @@ export type OptionalMarkdownTextProperties = {
    * worse than none.
    */
   tableLabel?: string;
+  /**
```
Tara-ag, line 26: 💡 MINOR: JSDoc says class is 'in place of' the default wrapper class, but the implementation adds it alongside `markdown-table-wrapper`, never replacing it. This contradicts the actual behaviour and the docs/MarkdownText.md explanation. Please rewrite the JSDoc to say the class is added alongside the built-in class.

Tara-ag (commented):
One minor documentation-consistency finding. The implementation is otherwise sound: the new prop is purely additive, the default class is preserved, escaping is applied, web-component parity is maintained, and tests cover the new behaviour including the injection case.

src/lib/MarkdownText/properties.ts
Tara-ag, line 26: 💡 MINOR: JSDoc says class is 'in place of' the default wrapper class

The JSDoc claims `tableWrapperClass` is a class name 'in place of the library default markdown-table-wrapper', but the implementation adds the custom class alongside the default, never replacing it. This contradicts the actual behavior and the docs/MarkdownText.md explanation. It risks misleading consumers and future maintainers.

**Suggested fix:** Rewrite the JSDoc to say the class is added alongside `markdown-table-wrapper`, not in place of it. For example:
```ts
/**
 * Extra class name added to the scrollable region wrapping each table, alongside the
 * built-in `markdown-table-wrapper`. Lets a consumer whose own stylesheet targets a
 * different class keep that styling without losing the wrapper's scroll/focus behaviour.
 * Omitting it, or passing an empty string, keeps only `markdown-table-wrapper`.
 */
```

Tara-ag (approved):
Approved with one minor documentation nit. The change is additive and safe: the default `markdown-table-wrapper` class is preserved, the caller-supplied class is escaped before use, web-component parity is maintained, and tests/docs cover the new behaviour including an injection case.

murdore:
## Review threads worked through at `50b0bf4`

6 unresolved threads: 4 fixed · 2 refuted.


### **Fixed**

**`src/lib/MarkdownText/markdown.test.ts:230`** — [thread PRRT_kwDOKuH2cM6hhs_W] Reviewer says the impl's `length > 0` check on tableWrapperClass lets a whitespace-only value become a live custom class, losing default styling; asks for a test.

> src/lib/MarkdownText/markdown.ts:309 already does `wrapperClass.trim()` before the length check (`custom.length > 0` at line 315 tests the trimmed value), so whitespace-only input already falls back to the default class -- the claimed bug doesn't exist in current source. But no test pinned this, so I added `it('falls back to the default wrapper class when tableWrapperClass is whitespace-only', ...)` to markdown.test.ts asserting `tableWrapperClass: '   '` renders only `markdown-table-wrapper`. Ran `npx vitest run src/lib/MarkdownText/markdown.test.ts -t whitespace-only` -> 1 passed; full suite `npx vitest run --testTimeout=20000` -> 1028 passed (52 files), 0 failed.

**`src/lib/MarkdownText/markdown.test.ts:230`** — [thread PRRT_kwDOKuH2cM6hh18m] Duplicate suggestion: add a test for whitespace-only tableWrapperClass, since it is neither empty nor a meaningful class.

> Same underlying gap as the other test.ts:230 thread -- addressed by the single new test added at markdown.test.ts (whitespace-only tableWrapperClass falls back to markdown-table-wrapper). Verified passing in the same vitest run (1028/1028 passed).

**`src/lib/MarkdownText/properties.ts:26`** — [thread PRRT_kwDOKuH2cM6hiX8H] JSDoc for tableWrapperClass says the class is used 'in place of' the default markdown-table-wrapper, contradicting the actual add-alongside behavior.

> Read properties.ts:27 before the fix: '...in place of the library default `markdown-table-wrapper`...'. Cross-checked src/lib/MarkdownText/markdown.ts:309-317, which builds `classAttr` as `${DEFAULT_TABLE_WRAPPER_CLASS} ${escapeHtml(custom)}` -- concatenation, never a replacement -- and docs/MarkdownText.md's own Props table already says 'added alongside the built-in markdown-table-wrapper rather than replacing it'. Rewrote the JSDoc block in properties.ts to say the class is 'added ... alongside the built-in markdown-table-wrapper rather than replacing it' and to note whitespace-only strings also keep only the default. Verified with `npm run check` (exit 0, 0 errors) and `npm run lint` (exit 0).

**`src/lib/MarkdownText/properties.ts:26`** — [thread PRRT_kwDOKuH2cM6hiZhy] Duplicate, more detailed: same 'in place of' vs. add-alongside contradiction, with a suggested replacement JSDoc text.

> Same edit as the other properties.ts:26 thread resolves this one too -- JSDoc now reads 'Extra class name added to the scrollable region wrapping each table, alongside the built-in `markdown-table-wrapper` rather than replacing it...'. Diff visible via `git diff` on properties.ts lines 26-33 (pre-amend), now folded into the amended commit. Verified with `npm run check` (exit 0) and `npm run lint` (exit 0).


### **Refuted**

**`docs/MarkdownText.md:42`** — [thread PRRT_kwDOKuH2cM6hhs--] Reviewer claims the Props table's Description separator dashes are shorter than the widest Description cell (the tableWrapperClass row), so the table renders ragged.

> python3 measured docs/MarkdownText.md lines 39-47 (header, separator, and all 7 data rows including tableWrapperClass): every line is exactly 284 characters and splits on '|' into identical column widths [19,27,10,9,213] -- the separator's 211-dash Description column already matches the widest Description cell exactly. `awk 'NR==39,NR==47{print length}'` confirmed uniform 284-char lines. No misalignment exists in the current source.

**`docs/MarkdownText.md:42`** — [thread PRRT_kwDOKuH2cM6hh19J] Duplicate: Description column separator allegedly shorter than the widest cell; reviewer supplies a specific longer dash line as the fix.

> Same measurement as the other docs/MarkdownText.md:42 thread: lines 39-47 are all 284 chars with matching column widths (separator's Description dashes = 213 chars, same as every data row's Description cell width). The reviewer's suggested replacement separator is not longer than what is already there -- current source is already correctly aligned.


### Verification

```
check   exit 0
lint    exit 0
vitest  Test Files  52 passed (52) / Tests  1028 passed (1028)
```

PR #604, opened 2026-09-11
feat(speech): add SpeechSynthesisController

murdore:
Symmetric counterpart to SpeechToTextController, wrapping the browser
speechSynthesis API: voice enumeration with the voiceschanged dance,
speak()/stop() with rate/pitch/volume/voice, toggle-to-stop
orchestration, and a self-hiding error toast. All state is per-instance
(no module-level singletons, unlike the clientSpeech.ts reference this
promotes), voice selection is a caller-supplied hook instead of a
hardcoded name list, and the dead cloud-TTS stub was dropped since
there is no cloud path.

## Summary by CodeRabbit

* **New Features**
  * Added a headless text-to-speech controller using the browser’s speech synthesis capabilities.
  * Supports voice selection, speech rate, pitch, volume, start/stop controls, and toggle-to-stop behavior.
  * Provides support detection, speaking status, voice enumeration, and user-facing error messages.
  * Added public configuration and type definitions for customizing speech behavior and messages.

* **Documentation**
  * Added usage documentation and indexed the speech synthesis component.


murdore:
## Verified independently — no recording, because this is a headless primitive

`SpeechSynthesisController` has no UI. There is deliberately no demo route: it is a controller, the mirror of `SpeechToTextController`, and building a page for it would have been inventing a component that was not asked for. So there is nothing to film, and the evidence below is measured.

### The load-bearing claim, proven by breaking it

The implementation this was ported from held `availableVoices`, `preferredVoice`, `audioPlayer` and `isSpeakingValue` as bare module-level `let`s. In a bundle every consumer would share one TTS session, and two concurrent surfaces would collide on the same player. That is the same defect class the `VoiceOrb` port was rejected for (fixed global `localStorage` keys), so "state is per-instance" is the claim this PR stands on.

I hoisted the engine back to module scope — the exact rejected pattern — and re-ran:

```
MUTATION: let sharedEngine = null   (engine hoisted out of the class)
  -> 1 failed | 24 passed

RESTORED
  -> 25 passed
```

Exactly one test died: the one driving two controller+engine pairs and asserting that `speak()` / `fireStart()` / `stop()` interleaved across both never leaks `speaking` or call counts between them.

### What was deliberately not ported

- **The dead cloud stub.** `synthesizeSpeechWithCloud` claimed OpenAI-via-server in its docstring and its body logged one event and unconditionally returned `null`. The live path "tried cloud first, fell back to browser", which means the fallback was the only path that ever ran. It is not here.
- **The hardcoded voice list** (`Samantha`, `Victoria`, `Moira`, `Karen`, `Tessa`). A content choice, not architecture. Voice selection is now a caller-supplied `selectVoice` hook with a neutral default: platform-default among language matches, then first language match, then platform default.

### Two things worth knowing before review

- **`SpeechToTextController` itself has zero unit tests.** This controller has 25. "Mirror STT as closely as makes sense" was not extended to mirroring its lack of coverage.
- **A pre-existing docs/code inconsistency was matched, not fixed.** The repo's `scripts/migrate/rename-doc-usages.ts` forces STT's own doc to display `onerror` lowercase even though the field in `SpeechToText/types.ts` is `onError`. The new doc matches the established spelling rather than diverging; fixing the underlying inconsistency is out of scope for this PR and touches a file this change does not own.

### Verification I ran myself on `e4b447d`

```
npm run check       0 errors
npm run lint        clean
npx vitest run      1046 / 1046
npx playwright      715 passed, 1 skipped
npm run build       exit 0
```

Tara-ag (changes requested):
Reviewed file-by-file. Four MAJOR issues found in the new SpeechSynthesisController; see inline comments for details and suggested fixes. None are blocking per the published criteria, but they affect correctness, multi-instance isolation, runtime robustness, and documentation accuracy. Requesting changes to address them.

src/lib/SpeechSynthesis/controller.svelte.ts
```diff
@@ -0,0 +1,312 @@
...
+      utterance.onstart = () => callbacks.onstart();
+      utterance.onend = () => callbacks.onend();
+      utterance.onerror = (event) => callbacks.onerror(event.error);
+      synth.speak(utterance);
+    }
+  };
+}
+
+function defaultSynthesisEngine(): SpeechSynthesisEngineLike | null {
+  if (typeof window === 'undefined' || !('speechSynthesis' in window)) {
```
Tara-ag, line 80: ⚠️ MAJOR: Multiple default-engine controllers clobber `window.speechSynthesis.onvoiceschanged`. `createBrowserEngine()` assigns a fresh closure to the singleton `synth.onvoiceschanged` every time it is called. Because `window.speechSynthesis` is a single browser instance, initializing a second `SpeechSynthesisController` (or re-initializing after destroy) overwrites the previous handler, so earlier controllers never refresh their voices list when the platform fires `voiceschanged`. This contradicts the PR's claim that multiple instances do not cross-talk.

**Suggested fix:** Use `addEventListener/removeEventListener` per engine wrapper, or extend `SpeechSynthesisEngineLike` with a destroy/cleanup seam so each controller can register and unregister its own `voiceschanged` handler without overwriting siblings on the shared global.

src/lib/SpeechSynthesis/controller.svelte.ts
```diff
@@ -0,0 +1,312 @@
...
+        if (isCurrent()) {
+          this.handleStart();
+        }
+      },
+      onend: () => {
+        if (isCurrent()) {
+          this.handleEnd();
+        }
+      },
+      onerror: (code) => {
```
Tara-ag, line 204: ⚠️ MAJOR: Caller-supplied `selectVoice` hook can crash `speak()`. The controller guards `getSynthesisEngine` with `try/catch` but does not guard the caller-supplied `selectVoice` hook. A buggy or malicious hook that throws will propagate out of `speak()` instead of being treated as a voice-selection failure and surfaced through the normal error path.

**Suggested fix:** Wrap the `selectVoice` call in `try/catch`:
```ts
let voice: SpeechSynthesisVoice | null = null;
try {
  voice = selectVoice(this.voices, lang);
} catch (e) {
  this.diagnose('selectVoiceThrew', { error: errorDetail(e) });
}
```
Then use `voice` in the request.

src/lib/SpeechSynthesis/controller.svelte.ts
```diff
@@ -0,0 +1,312 @@
...
+    const isCurrent = (): boolean => generation === this.speechGeneration;
+    const callbacks: SpeechUtteranceCallbacks = {
+      onstart: () => {
+        if (isCurrent()) {
+          this.handleStart();
+        }
+      },
+      onend: () => {
+        if (isCurrent()) {
+          this.handleEnd();
```
Tara-ag, line 201: ⚠️ MAJOR: Numeric speech options are not validated before reaching the engine. `rate`, `pitch`, `volume` and `errorTimeoutMs` are taken from caller options and passed straight into `SpeechSynthesisUtterance` / `setTimeout` without clamping or finiteness checks. Out-of-range or `NaN` values can cause the synthesis engine to throw or behave unpredictably, and a negative `errorTimeoutMs` produces a confusing timer.

**Suggested fix:** Add a small normalizer that clamps `rate` to `[0.1, 10]`, `pitch` to `[0, 2]`, `volume` to `[0, 1]`, ensures all three are finite, and caps `errorTimeoutMs` at a finite non-negative value (e.g. `Math.max(0, Number.isFinite(v) ? v : DEFAULT_ERROR_TIMEOUT_MS)`).

docs/SpeechSynthesis.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Options
+
+| Option               | Type                                                                                    | Default        | Description                                                                                                                    |
+| --------------------- | ---------------------------------------------------------------------------------------- | -------------- | -------------------------------------------------------------------------------------------------------------------------------- |
+| `lang`                | `string`                                                                                  | `''`           | Utterance language; overridable per `speak()` call.                                                                              |
+| `rate`                | `number`                                                                                  | `1`            | Utterance rate; overridable per `speak()` call.                                                                                  |
```
Tara-ag, line 48: ⚠️ MAJOR: Documentation lists callback as `onerror` but implementation uses `onError`. The `SpeechSynthesis.md` options table documents the error callback as `onerror` (all lowercase), but the actual option name in `types.ts` and `controller.svelte.ts` is `onError` (camelCase). JavaScript consumers following the docs will pass a callback that is silently ignored.

**Suggested fix:** Rename the documented option to `onError` to match the implementation and TypeScript types.

Tara-ag (changes requested):
Yama review: 4 MAJOR findings accepted. Please address the inline comments before merging.

Tara-ag:

Review was constrained by truncated tool outputs: I could confirm the PR adds a new SpeechSynthesisController export, types, tests, and docs, but I could not inspect the full controller implementation line-by-line. No gate-verified CRITICAL/MAJOR/MINOR findings were produced, so the PR is not blocked. Before merging, the author should manually verify: (1) SSR safety — no unguarded window/speechSynthesis access at module load or constructor time; (2) cleanup in destroy() clears the error toast timer and the voiceschanged listener; (3) $effect dependencies in the controller are correct and cannot loop; (4) docs/_index.json actually contains the new SpeechSynthesis entry (only the tail of the file was returned).

murdore:
## Re-pushed at `8f18cc9` — all four MAJOR findings closed

The verdict above reviewed `e4b447d`. Every finding it raised was already fixed in a commit that had not been pushed. Mapping each one to the code that closes it:

| Finding | Closed by |
| --- | --- |
| Controllers clobber `synth.onvoiceschanged` | `addEventListener('voiceschanged', ...)` — additive registration, so one controller's subscription never overwrites another's. Teardown removes only its own listener. |
| Numeric options unvalidated | `clampSpeechNumber` / `clampErrorTimeoutMs`. Non-finite falls back rather than clamping garbage into range silently. |
| `selectVoice` hook can crash `speak()` | `resolveVoice()` wraps it in try/catch, reports via `onDiagnostic`, falls back to the built-in default. |
| Docs say `onerror`, code says `onError` | They now agree on `onerror`, which is the repo-wide lowercase convention. |

### The first one was a hole in my own verification, not just in the code

My original independence test used a **fake engine**, so two controllers never shared anything and the test passed while the real bug was live — the real path shares the `window.speechSynthesis` singleton. The test now exercises the shared-singleton path, which is the only place the defect could exist.

### Verification at `8f18cc9`

```
check   exit 0
lint    exit 0
vitest  1063 passed
```

No recording: this is a headless controller with no UI and deliberately no demo route, being the mirror of `SpeechToTextController`. The evidence is the assertions.

docs/SpeechSynthesis.md
```diff
@@ -0,0 +1,66 @@
...
+Headless text-to-speech controller over the browser's `speechSynthesis` API — the symmetric counterpart to `SpeechToTextController`. It owns voice enumeration (including the async `voiceschanged` population), `speak()`/`stop()` with rate/pitch/volume/voice control, toggle-to-stop orchestration, and a self-hiding error toast. It renders nothing: pair it with a "read aloud" button of your own.
+
+## Usage
+
+```svelte
+<script lang="ts">
+  import { SpeechSynthesisController } from '@juspay/svelte-ui-components';
+
+  const tts = new SpeechSynthesisController({ lang: 'en-US' });
+</script>
```
Tara-ag, line 16: 💡 MINOR: The usage example constructs a controller but never tears it down. Without `destroy()`, the error toast timer can leak when the component unmounts.

**Fix:** Update the example to show teardown:
```svelte
<script lang="ts">
  import { SpeechSynthesisController } from '@juspay/svelte-ui-components';

  const tts = new SpeechSynthesisController({ lang: 'en-US' });
  onDestroy(() => tts.destroy());
</script>
```

src/lib/SpeechSynthesis/controller.svelte.ts
```diff
@@ -0,0 +1,395 @@
...
+    if (!this.everInitialized) {
+      this.initialize();
+    }
+    if (!this.supported || this.engine === null) {
+      this.showError(this.messages.unsupported);
+      return;
+    }
+    if (text.length === 0) {
+      return;
+    }
```
Tara-ag, line 251: ⚠️ MAJOR: The error toast timer is only cleared in `destroy()`. A consumer who instantiates this inside a Svelte component (the expected use case) and forgets to call `destroy()` will leak the timeout. Worse, the docs usage example never shows teardown, so most copy-paste consumers will leak it.

**Fix:** Update the docs example to import `onDestroy` from `'svelte'` and call `onDestroy(() => tts.destroy())`. Consider also exposing a small `useSpeechSynthesis(options)` helper that wires teardown automatically.

src/lib/SpeechSynthesis/controller.svelte.ts
```diff
@@ -0,0 +1,395 @@
...
+      utterance.voice = request.voice;
+      utterance.onstart = () => callbacks.onstart();
+      utterance.onend = () => callbacks.onend();
+      utterance.onerror = (event) => callbacks.onerror(event.error);
+      synth.speak(utterance);
+    }
+  };
+}
+
+function defaultSynthesisEngine(): SpeechSynthesisEngineLike | null {
```
Tara-ag, line 139: ⚠️ MAJOR: `createBrowserEngine` is exported and constructs `new SpeechSynthesisUtterance(...)` unconditionally when `speak()` is called. Although the controller's default path guards `window`, this exported helper can be invoked directly in SSR/Node and will throw `ReferenceError: SpeechSynthesisUtterance is not defined`.

**Fix:** Guard `speak` with an explicit check, e.g.:
```ts
speak: (request, callbacks) => {
  if (typeof SpeechSynthesisUtterance === 'undefined') {
    throw new Error('SpeechSynthesisUtterance is not available in this environment');
  }
  const utterance = new SpeechSynthesisUtterance(request.text);
  ...
}
```

src/lib/SpeechSynthesis/types.ts
```diff
@@ -0,0 +1,79 @@
...
+  errorTimeoutMs?: number;
+  /** Per-synthesis-error-code messages, merged over the built-in map. */
+  errorMessages?: Record<string, string>;
+  /** Copy overrides, merged over the built-in defaults. */
+  messages?: Partial<SpeechSynthesisMessages>;
+  /** Resolves the synthesis engine; defaults to the browser's. Injectable for tests. */
+  getSynthesisEngine?: () => SpeechSynthesisEngineLike | null;
+  onSpeakingChange?: (speaking: boolean) => void;
+  // Lowercase to match what scripts/migrate/rename-doc-usages.ts forces docs/SpeechSynthesis.md
+  // to say (its unambiguous-pair rewrite is string-based across all docs, not scoped to
```
Tara-ag: ⚠️ MAJOR: Inconsistent callback casing with sibling `SpeechToTextController`. The existing public API uses `onError` (camelCase) at `SpeechToTextOptions.onError`; this new controller exposes `onerror` (lowercase). Shipping two headless speech controllers with different conventions for the same concept is confusing and looks accidental. The inline comment even says the lowercase spelling was chosen only because a docs migration script rewrites docs globally — that is not a sound basis for a public API shape.

**Fix:** Rename the option to `onError` and update `docs/SpeechSynthesis.md`. If the migration script forces the wrong casing, fix the script or the doc source, not the exported option.

Tara-ag (changes requested):
Reviewed file-by-file. The new controller is well-tested and the implementation is solid, but there are a few public-API and lifecycle issues that should be addressed before merging:

1. **Public API consistency (MAJOR):** `onerror` is lowercase while the sibling `SpeechToTextController` uses `onError`. Align them.
2. **Lifecycle/teardown (MAJOR):** The docs example does not show `destroy()`, which is required to clear the toast timer. Update the example.
3. **SSR safety (MAJOR):** Exported `createBrowserEngine` can throw `ReferenceError` in non-browser environments; guard `SpeechSynthesisUtterance`.
4. **Docs teardown (MINOR):** Same as #2 — add `onDestroy` to the usage snippet.
5. **Web-component build (MINOR):** Confirm `npm run check:wc` still passes with the new controller export.

Requesting changes.

murdore:
## Review threads worked through at `128a8cb`

8 unresolved threads: 2 fixed · 1 refuted · 5 already correct.


### **Fixed**

**`src/lib/SpeechSynthesis/controller.svelte.ts:207`** — initialize() replacing the engine leaves the previous engine speaking with its onvoiceschanged handler still attached

> Confirmed real gap: pre-fix initialize() at controller.svelte.ts:196-216 did `this.engine = engine` with no teardown of a prior different engine, and no existing test exercised initialize() being called twice with two distinct injected engines (grep for 'getSynthesisEngine' in the test file showed every controller using a single fixed `() => engine` closure). Fixed by adding a private teardownEngine() (bumps speechGeneration, resets internalSpeaking/speaking, cancels the old engine, clears its onvoiceschanged) called from initialize() only when `this.engine !== null && this.engine !== engine`. Added regression tests 're-initializing onto a different engine' (stops/detaches the superseded engine, routes further speak() to the new one, ignores a late event from the old one) and 'onto the same engine leaves it running, untouched' -- both pass in the 1066-test vitest run.

**`src/lib/SpeechSynthesis/controller.svelte.ts:187`** — `options.voice ?? resolveVoice(lang)` silently overrides an explicit `voice: null` opt-out

> Confirmed still present: controller.svelte.ts previously had `voice: options.voice ?? this.resolveVoice(lang)` (SpeechUtteranceOptions.voice is typed `SpeechSynthesisVoice | null` in types.ts:39, so `??` treated an explicit null as absent). Changed to `voice: typeof options.voice === 'undefined' ? this.resolveVoice(lang) : options.voice`, matching the `typeof x === 'undefined'` idiom already used elsewhere in this repo (e.g. Chat/controller.svelte.ts:16,73,210) since the bare `undefined` identifier is eslint-banned here. Added regression test 'preserves an explicit voice: null override instead of running default selection (Finding 4 regression)' -- passes.


### **Refuted**

**`docs/SpeechSynthesis.md:48`** — Docs claim the callback is `onerror` but implementation allegedly uses `onError`

> grep -n "onError" src/lib/SpeechSynthesis/types.ts src/lib/SpeechSynthesis/controller.svelte.ts finds only a code comment (types.ts:75) explaining why the option is deliberately kept lowercase; grep -n "onerror" src/lib/SpeechSynthesis/types.ts docs/SpeechSynthesis.md shows the actual option is `onerror` (types.ts:30, :76) and the docs table at docs/SpeechSynthesis.md:57 already documents it as `onerror`, matching. controller.svelte.ts:389 calls `this.options.onerror?.(message, code)`. No code path uses camelCase `onError` for this option -- the claim describes a state that does not exist in this file.


### **Already correct**

**`src/lib/SpeechSynthesis/controller.svelte.ts:51`** — createBrowserEngine assigns window.speechSynthesis.onvoiceschanged directly, clobbering a sibling controller's handler

> controller.svelte.ts:88-128 createBrowserEngine wraps SpeechSynthesisEngineLike.onvoiceschanged as a get/set that internally calls synth.addEventListener('voiceschanged', dispatch) / removeEventListener additively -- it never touches synth.onvoiceschanged. Test 'two engines built from one shared browser-like synth both receive voiceschanged (additive registration, not a clobbered single slot)' at controller.svelte.test.ts:624-651 and 'destroy() removes exactly that engine's own listener, leaving a sibling engine's subscription intact' at :671-690 pass (vitest run: 1066 passed).

**`src/lib/SpeechSynthesis/controller.svelte.ts:93`** — Voice matching uses a fixed two-character lang prefix, so 'fil-PH' could match a 'fi-FI' voice

> controller.svelte.ts:137-156 defines primarySubtag() (splits on the first '-', not a fixed slice(0,2)) and defaultSelectVoice() compares full primary subtags. Regression test 'matches on the full primary language subtag, not a fixed 2-character prefix (Finding 3 regression)' at controller.svelte.test.ts:279-300 passes.

**`src/lib/SpeechSynthesis/controller.svelte.ts:131`** — Duplicate framing of the same createBrowserEngine onvoiceschanged-clobbering concern (older review pass, different line anchor)

> Same code and same passing tests as the :51 thread above (createBrowserEngine's addEventListener-based get/set at controller.svelte.ts:88-128); this is the identical finding restated against a stale line number.

**`src/lib/SpeechSynthesis/controller.svelte.ts:268`** — Caller-supplied selectVoice hook is unguarded and can crash speak()

> controller.svelte.ts:342-350 private resolveVoice() wraps the selectVoice() call in try/catch, reports 'selectVoiceThrew' via this.diagnose(), and falls back to defaultSelectVoice(). Test 'a throwing selectVoice hook does not crash speak(): falls back to the default selection and reports a diagnostic (Finding 2 regression)' at controller.svelte.test.ts:335-353 passes.

**`src/lib/SpeechSynthesis/controller.svelte.ts:265`** — rate/pitch/volume/errorTimeoutMs are passed to the engine/timer unvalidated

> controller.svelte.ts:54-70 defines clampSpeechNumber() (finite-check + clamp to RATE/PITCH/VOLUME range, else fallback) and clampErrorTimeoutMs() (finite-check, floors at 0), both exported and applied in speak() (lines ~242-250) and showError() (line ~387/447). 'clampSpeechNumber / clampErrorTimeoutMs (Finding 5: pure, browser-free unit tests)' describe block passes.


### Verification

```
check   exit 0
lint    exit 0
vitest  Test Files 53 passed (53); Tests 1066 passed (1066); Duration 54.04s
```

murdore:
## The four new threads, verified

### `onerror` vs the sibling's `onError` — the finding is right, and the fix is blocked by one of our own guards

Confirmed, and it is worse than the thread says. The new type is inconsistent **with itself**, not just with its sibling:

```
SpeechToTextOptions.onError          camelCase   (shipped)
SpeechSynthesisOptions.onerror       lowercase
SpeechSynthesisOptions.onSpeakingChange, onDiagnostic   camelCase
```

So I renamed it to `onError` — and the repo rejected it:

```
FAIL  scripts/migrate/rename-doc-usages.test.ts
      > never recommend a spelling the library deprecates
```

`planDocRenames` rewrites deprecated prop spellings across **all** of `docs/`, string-based, unscoped to the component that actually deprecates the name. `onError` is deprecated as a *component prop*, so the guard treats `onError` anywhere in docs — including a headless controller's option bag, which is not a component prop at all — as an instruction to use something 4.0.0 removes.

The lowercase spelling was therefore not an accident or a casing preference. It was a deliberate workaround, and the code comment says exactly that.

**I have reverted the rename**, because the correct fix is to narrow the guard so it applies to component props rather than every string in `docs/`, and that is a change to migration tooling rather than to this PR. Doing it here would either break the guard or smuggle a tooling change into a feature PR.

The inconsistency is real and worth an issue. It is cheapest to fix now while this API is unreleased — but it needs the guard narrowed first.

### `createBrowserEngine` and SSR — real but narrower than stated

It is `export`ed from `controller.svelte.ts`, but **not re-exported from `src/lib/index.ts`**, so it is not part of the package's public surface — only a deep import reaches it. The controller's own path guards `window` before ever constructing an utterance. Worth a guard for defence in depth; not the exposure the thread describes.

### Docs example lacks teardown, and the toast timer leaks without `destroy()`

Both fair, and they are the same point. Taking them together rather than separately.

```
scripts/migrate  75 passed (8 files)  — after the revert
```

murdore:
## Two of the four addressed in code

**SSR guard on `createBrowserEngine`** — added. It reports through the normal error channel rather than throwing out of `speak()`, matching how every other engine failure surfaces:

```ts
if (typeof SpeechSynthesisUtterance !== 'function') {
  callbacks.onerror('synthesis-unavailable');
  return;
}
```

`synthesis-unavailable` is an existing code in `DEFAULT_ERROR_MESSAGES`, so a consumer already has copy for it.

**Docs example now shows teardown** — the prose said *"Call `destroy()` on teardown"* but the example did not, and people copy the example, not the prose. It now imports `onDestroy` and calls it, with a comment saying what leaks without it.

That covers the toast-timer thread too: both threads describe the same leak, and the fix for both is the example.

**The `onerror` casing thread stays open** for the reason given above — the rename is blocked by `rename-doc-usages`, and narrowing that guard is a tooling change that does not belong in this PR.

```
check   exit 0
lint    exit 0
vitest  1066 passed
```

src/lib/SpeechSynthesis/controller.svelte.ts
```diff
@@ -0,0 +1,429 @@
...
+  }
+
+  /** Clears the toast timer and tears the engine connection down. */
+  destroy(): void {
+    this.speechGeneration += 1;
+    if (this.errorTimer !== null) {
+      clearTimeout(this.errorTimer);
+      this.errorTimer = null;
+    }
+    if (this.engine !== null) {
```
Tara-ag, line 329: ⚠️ MAJOR: `refreshVoices` assigns `this.voices = this.engine.getVoices()` directly. If the engine returns the same array reference on subsequent calls (which the real `speechSynthesis` implementation commonly does), Svelte 5's `$state` will see a reference-equal assignment and skip reactivity updates. Consumers binding to `controller.voices` may never see voices populate after `voiceschanged`.

**Fix:** force a new array reference so reactivity always fires:
```ts
this.voices = [...this.engine.getVoices()];
```

src/lib/SpeechSynthesis/controller.svelte.ts
```diff
@@ -0,0 +1,429 @@
...
+        return;
+      }
+      const utterance = new SpeechSynthesisUtterance(request.text);
+      utterance.lang = request.lang;
+      utterance.rate = request.rate;
+      utterance.pitch = request.pitch;
+      utterance.volume = request.volume;
+      utterance.voice = request.voice;
+      utterance.onstart = () => callbacks.onstart();
+      utterance.onend = () => callbacks.onend();
```
Tara-ag, line 132: ⚠️ MAJOR: `typeof SpeechSynthesisUtterance !== 'function'` is not SSR-safe. In Node/jsdom the identifier `SpeechSynthesisUtterance` is undeclared in the global scope; `typeof` on an undeclared global throws a `ReferenceError`. Because this file is exported from `src/lib/index.ts`, it can be loaded server-side.

**Fix:** access the constructor through `globalThis`:
```ts
const UtteranceCtor = (globalThis as any).SpeechSynthesisUtterance;
if (typeof UtteranceCtor !== 'function') {
  callbacks.onerror('synthesis-unavailable');
  return;
}
const utterance = new UtteranceCtor(text);
```

Tara-ag (changes requested):
Two MAJOR issues need fixing before this can land:

1. **SSR safety**: `createBrowserEngine` accesses `SpeechSynthesisUtterance` as a bare global, which throws `ReferenceError` in Node/jsdom because the identifier is undeclared. Use `globalThis.SpeechSynthesisUtterance` instead.
2. **Svelte 5 reactivity**: `refreshVoices` assigns `this.engine.getVoices()` directly. Real browser engines often return the same array reference, so `$state` won't trigger updates. Spread into a new array.

One MINOR note about `src/wc/index.ts` parity could not be posted inline because the file is unchanged in this PR.

murdore:
## Three MAJORs: one refuted, one fixed, one still blocked

### Refuted — `typeof SpeechSynthesisUtterance` is SSR-safe

> In Node/jsdom the identifier is undeclared in the global scope; `typeof` on an undeclared global throws a `ReferenceError`.

It does not. `typeof` is specifically the operator that does **not** throw on an undeclared identifier — that is its whole purpose, and why this is the standard SSR idiom. In Node:

```
$ node -e "console.log(typeof SpeechSynthesisUtterance)"
undefined

$ node -e "try { SomeUndeclared } catch (e) { console.log(e.constructor.name) }"
ReferenceError          ← only a BARE reference throws
```

`typeof SpeechSynthesisUtterance !== 'function'` evaluates to `true` in Node and the guard returns cleanly. Routing through `globalThis` would work too, but it is not fixing a defect. No change made.

### Fixed — `refreshVoices` now copies the array

This one is real, and worth being precise about why. `getSynthesisEngine` is documented as "the seam for tests, demos, and non-browser hosts", so an engine returning a stable array it mutates in place is a **supported** shape. The real `speechSynthesis.getVoices()` returns a fresh array each call (WebIDL sequence), which is why the browser path never surfaced it.

```ts
this.voices = [...this.engine.getVoices()];
```

Honest note on the evidence: I wrote three attempts at a test that would fail before this change and pass after, and none of them discriminated — the first two read `controller.voices.length`, which reflects a mutation through the same object whether or not `$state` notified anyone. What the isolation experiment *did* show is that a raw array mutated behind a `$state` proxy is not reliably observed, which is enough of a hazard that copying is justified on its own terms. I am not claiming a reproduction I did not manage to pin.

### Still blocked — `onerror` casing

Unchanged: the rename is right, and #614 is the prerequisite. `rename-doc-usages` currently rewrites any bare `onError` in `docs/` as a deprecated component prop, including a headless controller's option. Once #614 lands this becomes a one-line change here.

```
check   exit 0
lint    exit 0
vitest  1066 passed
```

murdore:
## Addressing the verdict's third finding, which never appeared as a thread

The current verdict carries a MINOR that says it *"could not be posted inline because the file is unchanged"* — so it exists only in the verdict body and I had not seen it until reading that body directly.

### MINOR — `SpeechSynthesisController` missing from `src/wc/index.ts`

No change, and the finding says why itself:

> headless controllers are not typically rendered as custom elements … the existing SpeechToText controller is also absent from `src/wc/index.ts`, so this is consistent

Confirmed against the tree — neither controller is registered, because neither renders anything. A custom element for a headless controller would be an element with no DOM. Registering it would create the inconsistency rather than remove one.

### The `typeof` MAJOR is still stated, and is still wrong

The verdict repeats:

> The `typeof` check only works for declared-but-undefined globals; an undeclared identifier still throws.

That is not how `typeof` works — being safe on undeclared identifiers is its defining property. Demonstrated in Node:

```
$ node -e "console.log(typeof SpeechSynthesisUtterance)"
undefined

$ node -e "try { SomeUndeclared } catch (e) { console.log(e.constructor.name) }"
ReferenceError          ← only a bare reference throws
```

The guard evaluates correctly under SSR and returns without constructing anything. No change.

### The voices MAJOR is fixed

`refreshVoices` now copies: `this.voices = [...this.engine.getVoices()];` — see the previous comment for why a copy is justified even though the browser path returns a fresh array.

So of the three: **one fixed, one refuted with evidence, one non-actionable by its own reasoning.**

src/lib/SpeechSynthesis/types.ts
murdore: Fixed in `a8b5743` — the option is `onError`, matching `SpeechToTextOptions.onError` exactly (same name, same `(message, code)` signature).

You were right that the migration script was not a sound basis for the API shape, and it is no longer even a constraint: #614 scoped `rename-doc-usages.ts` to components that actually deprecate `onError` as a prop. Verified both directions against this tree — the pre-#614 script reports `docs/SpeechSynthesis.md` / `would rewrite 1 file(s)`, the merged one reports `would rewrite 0 file(s)`. The comment explaining the lowercase is gone with it.

The internal `SpeechUtteranceCallbacks.onerror` stays lowercase deliberately: that one mirrors `SpeechSynthesisUtterance.onerror` from the Web Speech API, not a library option.

Gate on this commit: prettier clean, eslint 0, event-casing 0, svelte-check 0 errors, 1099 unit tests.

src/lib/SpeechSynthesis/controller.svelte.ts
```diff
@@ -0,0 +1,436 @@
...
+
+  /**
+   * Speaks `text`. Calling this while already speaking stops instead of starting a new
+   * utterance -- the same toggle-to-stop behavior as a mic button's start/stop click.
+   */
+  speak(text: string, options: SpeechUtteranceOptions = {}): void {
+    if (this.internalSpeaking) {
+      this.stop();
+      return;
+    }
```
Tara-ag, line 241: ⚠️ MAJOR: `speak()` resolves a voice from `this.voices` immediately after a synchronous `initialize()`. If the platform hasn't fired `voiceschanged` yet, the list may be empty and the utterance falls back to the platform default even though a better match is about to arrive. Document this explicitly in the docs, or expose an `awaitVoices()` / ready signal for hosts that need deterministic voice selection before speaking.

src/lib/SpeechSynthesis/controller.svelte.ts
```diff
@@ -0,0 +1,436 @@
...
+      utterance.pitch = request.pitch;
+      utterance.volume = request.volume;
+      utterance.voice = request.voice;
+      utterance.onstart = () => callbacks.onstart();
+      utterance.onend = () => callbacks.onend();
+      utterance.onerror = (event) => callbacks.onerror(event.error);
+      synth.speak(utterance);
+    }
+  };
+}
```
Tara-ag, line 137: 🔒 CRITICAL: `typeof SpeechSynthesisUtterance` throws a `ReferenceError` in SSR/Node because the bare identifier is evaluated before the guard can run. The controller's own path guards `window`, but `createBrowserEngine` is exported and reachable directly. Guard with `'SpeechSynthesisUtterance' in globalThis` (or `window`) before touching the identifier:
```ts
if (typeof globalThis === 'undefined' || !('SpeechSynthesisUtterance' in globalThis)) {
  callbacks.onerror('synthesis-unavailable');
  return;
}
```

src/lib/SpeechSynthesis/controller.svelte.ts
```diff
@@ -0,0 +1,436 @@
...
+      return synth.speaking;
+    },
+    get onvoiceschanged() {
+      return voicesChangedHandler;
+    },
+    set onvoiceschanged(handler) {
+      voicesChangedHandler = handler;
+      if (handler === null) {
+        if (listening) {
+          synth.removeEventListener('voiceschanged', dispatch);
```
Tara-ag, line 105: 💬 SUGGESTION: Exporting `createBrowserEngine` expands the public API surface and exposes the SSR-unsafe `SpeechSynthesisUtterance` reference. Consider keeping it internal (or at least guarding it robustly as noted above) and exposing it only through a test seam if needed.

docs/SpeechSynthesis.md
```diff
@@ -0,0 +1,70 @@
...
+```svelte
+<script lang="ts">
+  import { onDestroy } from 'svelte';
+  import { SpeechSynthesisController } from '@juspay/svelte-ui-components';
+
+  const tts = new SpeechSynthesisController({ lang: 'en-US' });
+
+  // Without this the error-toast timer outlives the component.
+  onDestroy(() => tts.destroy());
+</script>
```
Tara-ag, line 16: 💡 MINOR: The example uses Svelte 5's `onclick` syntax. Verify this matches the docs site's convention; if examples elsewhere still use `on:click`, align this one for consistency.

src/lib/SpeechSynthesis/controller.svelte.test.ts
```diff
@@ -0,0 +1,780 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import {
+  SpeechSynthesisController,
+  clampErrorTimeoutMs,
+  clampSpeechNumber,
+  createBrowserEngine,
```
Tara-ag, line 6: 💡 MINOR: The test imports from `'./controller.svelte'` rather than the explicit `'./controller.svelte.ts'` extension used by the source file. Align with the repo's `.svelte.ts` runes-controller convention to avoid module-resolution ambiguity.

Tara-ag (changes requested):
Blocking review: one CRITICAL SSR-safety issue in `createBrowserEngine` must be fixed before merge. Additional MAJOR/MINOR/SUGGESTION items are inline.

Tara-ag (changes requested):
Blocking on the CRITICAL SSR-safety issue in `createBrowserEngine`: the bare `SpeechSynthesisUtterance` identifier throws a `ReferenceError` in Node/SSR. Additional MAJOR voice-selection race, MINOR import/docs consistency, and SUGGESTION public-surface items are noted inline.

src/lib/SpeechSynthesis/controller.svelte.ts
murdore: Refuting this one, with the measurement rather than an argument: **`typeof` on an undeclared identifier does not throw.** It is the single expression in JavaScript that can safely reference a name that was never declared — that is what it is for, and it is why the guard is written this way.

Ran the exact line from `controller.svelte.ts:121` in bare Node, no DOM:

```
$ node -e "console.log(typeof NoSuchGlobalAnywhere)"
undefined          # no ReferenceError
```

And the whole guard, as it actually appears in `speak()`:

```
$ node -e "
const callbacks = { onerror: (c) => console.log('onerror called with:', c) };
function speak() {
  if (typeof SpeechSynthesisUtterance !== 'function') { callbacks.onerror('synthesis-unavailable'); return 'returned cleanly'; }
  return 'would have constructed';
}
console.log('result:', speak());"

onerror called with: synthesis-unavailable
result: returned cleanly
```

So in SSR/Node this already does exactly what the suggestion asks for: reports `synthesis-unavailable` through the normal error channel and returns. No `ReferenceError`, no throw out of `speak()`.

The proposed replacement is not wrong — `'SpeechSynthesisUtterance' in globalThis` is also false in Node — it is just equivalent, and strictly weaker: `in globalThis` is true for a global that exists but is not constructible, whereas `typeof x !== 'function'` also rejects that case.

Worth noting the premise "the bare identifier is evaluated before the guard can run" is the part that is mistaken. `typeof` does not evaluate the identifier as a reference; the spec resolves it and returns `"undefined"` for an unresolvable one instead of throwing (ECMA-262, `typeof` operator, step 1: "If val is a Reference Record and IsUnresolvableReference(val) is true, return `"undefined"`").

The guard and the reason for it are already commented in place at `controller.svelte.ts:116-120`. No change.

docs/SpeechSynthesis.md
murdore: Refuting — `onclick` **is** the convention, and `on:click` would be the deviation.

```
$ grep -rl "on:click" docs/*.md | wc -l
0
$ grep -rl "onclick" docs/*.md | wc -l
40
```

It is also machine-enforced rather than a matter of taste. `scripts/check-event-casing.js` runs in `npm run lint`:

> Enforces DESIGN_PRINCIPLES.md's event-casing rule: every event prop is `on` followed by the event name in lowercase — onclick, onrowclick, onoverlayclick — whether the browser fires the event or the component invents it. One rule, no native/synthetic judgement call.

This library is Svelte 5 throughout; `on:click` is Svelte 4 directive syntax and is not used anywhere in the repo. No change.

src/lib/SpeechSynthesis/controller.svelte.test.ts
murdore: Refuting — the extensionless form is the convention in `src/`, and the `.ts` form would not resolve there.

```
$ grep -rn "from '.*\.svelte\.ts'" src/ --include="*.ts"
(no matches)
```

No file under `src/` imports a `.svelte.ts` module with its extension, because `allowImportingTsExtensions` is not enabled in the app's `tsconfig.json`. Turning it on for one import is not a formatting tweak.

I think this comes from `scripts/migrate/`, which genuinely does use `./alias-wc-props.ts`-style imports — but that directory has its own `scripts/migrate/tsconfig.json` with `"allowImportingTsExtensions": true`. Two different compiler scopes, two different correct answers. `src/` is the first one. No change.

src/lib/SpeechSynthesis/controller.svelte.ts
murdore: The behaviour is real and inherent to the Web Speech API — but the "document this explicitly" half is already done, in four places, so let me point at them rather than add a fifth.

`docs/SpeechSynthesis.md:3` — "It owns voice enumeration (including the async `voiceschanged` population)"

`docs/SpeechSynthesis.md:33`, the `voices` row — "Often empty until the platform's `voiceschanged` fires — see `initialize()`."

`docs/SpeechSynthesis.md:68`, its own bullet — "Voices are commonly empty on the very first call and populate asynchronously once the platform fires `voiceschanged`; `initialize()` wires that subscription for the lifetime of the controller, not just once."

`controller.svelte.ts:177`, on the field itself — "Often empty until the platform's voiceschanged fires -- see `initialize()`."

On the substance: `initialize()` sets `engine.onvoiceschanged = () => this.refreshVoices()` **before** calling `refreshVoices()`, so the subscription is live for the lifetime of the controller and a late-arriving list does populate `voices`. What it cannot do is retroactively re-voice an utterance already handed to the platform — and no library can, because `speechSynthesis.speak()` takes the voice at call time.

`awaitVoices()` is the interesting half of your suggestion and I am not taking it in this PR. It is a new public API on a controller that is not merged yet, it needs a decision about what it resolves to on a platform that never fires `voiceschanged` at all (several do not, and a promise that never settles is worse than an empty list), and it wants a timeout policy that is a host concern. Tracking it as a follow-up rather than inventing that contract here.

A host that needs determinism today can wait on the existing signal: `voices` is `$state`, so `$derived`/a store subscription already tells you when the list arrives, and `selectVoice` is called at `speak()` time with whatever is current.

src/lib/SpeechSynthesis/controller.svelte.ts
murdore: Half agreed, and the half I disagree with is load-bearing.

Agreed that exporting it widens the surface. It is exported deliberately as the injection seam: `SpeechSynthesisOptions.getSynthesisEngine` lets a host — and the test suite — supply an engine, and `createBrowserEngine` is the reference implementation a host wraps when it wants the real `speechSynthesis` with one behaviour changed. Keeping it internal would mean either duplicating it in consumer code or exporting a test-only shim, which is a worse public surface, not a smaller one.

On "exposes the SSR-unsafe `SpeechSynthesisUtterance` reference" — that premise is the CRITICAL on this same file, and it does not hold: `typeof` on an undeclared identifier returns `"undefined"` rather than throwing, verified by running the exact guard in bare Node. Details on that thread. The guard already returns `synthesis-unavailable` through the normal error channel for exactly the direct-caller case you are describing, and says so in the comment above it.

So the "guard it robustly" ask is already satisfied; the "keep it internal" ask trades a documented seam for a hidden one. No change.

src/lib/SpeechSynthesis/controller.svelte.ts
murdore: Refuting this one: `typeof` on an undeclared identifier does not throw.

That is the defining property of the `typeof` operator — it is the safe probe for "does this global exist", and it returns the string `"undefined"` rather than raising `ReferenceError`. (The single exception is a TDZ binding — a `let`/`const` in the same scope before initialisation — which is not the case for a missing DOM global.)

```
$ node -e "console.log(typeof SpeechSynthesisUtterance)"
undefined          # no ReferenceError
```

And the guard the suggestion asks for is already the code being flagged, with a comment stating exactly the reachability this finding is about:

```ts
// src/lib/SpeechSynthesis/controller.svelte.ts:116
// `createBrowserEngine` is exported, so a caller can reach it directly and
// run it somewhere `SpeechSynthesisUtterance` does not exist. The
// controller's own path guards `window` before ever getting here; this
// guard is for the direct caller, and reports through the normal error
// channel rather than throwing out of `speak()`.
if (typeof SpeechSynthesisUtterance !== 'function') {
  callbacks.onerror('synthesis-unavailable');
  return;
}
```

So the SSR path is safe, and switching to `'SpeechSynthesisUtterance' in globalThis` would be equivalent, not safer. Resolving as not-a-defect.

src/lib/SpeechSynthesis/controller.svelte.ts
murdore: This is real behaviour, and it is already documented explicitly — which is the first of the two remedies you offered.

On the field itself:

```ts
/** Often empty until the platform's voiceschanged fires -- see `initialize()`. */
voices: readonly SpeechSynthesisVoice[] = $state([]);
```

And in `docs/SpeechSynthesis.md`, twice — the prop table row, and a dedicated note:

> Voices are commonly empty on the very first call and populate asynchronously once the platform fires `voiceschanged`; `initialize()` wires that subscription for the lifetime of the controller, not just once.

On the second remedy — an `awaitVoices()` / ready signal — I'd rather not add one without a caller that needs it. It would be a promise that can never reliably settle: the spec does not guarantee `voiceschanged` fires at all on platforms whose list is synchronously complete, so `awaitVoices()` would either resolve immediately (adding nothing) or hang on exactly the platforms it was meant to help. `voices` is already `$state`, so a caller who needs deterministic selection can react to it populating, which is the same information without the trap.

Happy to reopen if a concrete case wants it. Resolving as documented-as-designed.

src/lib/SpeechSynthesis/controller.svelte.ts
murdore: Deliberate, and it is guarded — but you're right that it widens the surface, so here is the reasoning rather than just a "works as intended".

`createBrowserEngine` is exported as the seam that makes the controller testable against a fake `SpeechSynthesis` without a browser: the tests construct an engine over a stub rather than monkey-patching a global. Keeping it internal would mean either testing through a global patch (which leaks across tests in a shared jsdom) or shipping no coverage of the engine at all.

On the SSR exposure specifically — that is handled at the point of use, not by hiding the symbol. `speak()` checks `typeof SpeechSynthesisUtterance !== 'function'` first and reports through `onerror` rather than throwing, so a direct caller in Node gets a clean `'synthesis-unavailable'`. Merely importing the module touches no DOM global. (See the sibling thread — the `ReferenceError` premise there doesn't hold; `typeof` on a missing global is safe.)

Resolving. If the API surface is the real concern rather than the SSR safety, that's worth its own issue about what the package should export.

docs/SpeechSynthesis.md
murdore: Verified, and it already matches — `onclick` is the repo convention, not the exception:

```
$ grep -rl "onclick=" docs/ | wc -l    → 26
$ grep -rl "on:click" docs/ | wc -l    → 0
```

Not a single doc uses `on:click`. That is expected: this is a Svelte 5 library, `on:` directive syntax is the Svelte 4 spelling, and the 4.0.0 release removed the legacy event-prop spellings entirely. Aligning this example toward `on:click` would make it the only doc in the repo showing syntax the library no longer supports.

No change needed. Resolving.

src/lib/SpeechSynthesis/controller.svelte.test.ts
murdore: Checked the convention across the repo, and the import is already consistent with it — the premise here is inverted.

Sibling runes-controller and component tests all import **without** the extension:

```
src/lib/Checkbox/Checkbox.svelte.test.ts:  import Checkbox from './Checkbox.svelte';
src/lib/Sheet/Sheet.svelte.test.ts:        import Sheet from './Sheet.svelte';
src/lib/Stepper/Stepper.svelte.test.ts:    import Step from './Step.svelte';
src/lib/Table/labels.svelte.test.ts:       import Table from './Table.svelte';
```

`from './controller.svelte'` is the same shape. Adding `.ts` would make this the only file in the repo spelling it that way, and there is no module-resolution ambiguity to avoid — `check` and the build both resolve it today (`npm run check` exits 0).

No change needed. Resolving.

PR #605, opened 2026-09-11
feat(checkbox,toggle,select): participate in native form submission

murdore:
Checkbox and Toggle gain name/value wired directly onto their existing
native <input type="checkbox">, so both now submit through a parent
<form>'s FormData like any native checkbox (absent when unchecked,
"on" by default when checked) instead of every consumer hand-mirroring
a hidden input. Select has no native control to carry name/value at
all, so it renders one hidden input per selected id, matching how a
native <select multiple> submits.

A hidden mirror inside the library was the alternative and was rejected:
it would relocate the workaround rather than remove it, and make it
harder to see. Where a real native control already exists, it carries
the attribute itself.

Checkbox also genuinely suppressed native `change`: the only element a
user can reach is the role="checkbox" span, the real input is
aria-hidden/tabindex=-1/pointer-events:none, and its checked state only
ever moved via a JS property write, which the browser does not treat as
a change. Proven with a failing-first Playwright test against the
pre-fix code, then fixed by dispatching bubbling input/change events
from the native input after the reactive write lands. Toggle's input
was already genuinely clickable, so it never had this bug.

The consuming team that reported this independently measured the same
defect from the other direction: at the checkbox's own centre point
document.elementFromPoint returns an ancestor element rather than the
control, so a synthetic click never reaches it and Playwright fails
actionability. Their E2E specs cannot drive this control at all today.

Checkbox, Toggle and Select each gain a "Native form submission"
section on their own demo page -- a real <form>, a submit button and a
live FormData readout. That is documentation first: a consumer reading
any of the three sees exactly how name/value behave, including that an
unchecked box is absent rather than empty. It also gives this suite
somewhere to drive. An earlier revision put that form in a Playwright-
only fixture route, which would have published to the docs site, since
the Pages workflow uploads the whole build directory.

Purely additive: name/value default to unset, leaving every existing
consumer's DOM byte-identical. Verified against the wc-parity checker
(scripts/wc-parity/prop-parity.ts) that name/value are not reserved
HTMLElement accessors, matching Input.wc.svelte's existing precedent,
and declared them in each .wc.svelte wrapper accordingly.

Verified on this commit: check 0, lint 0, 1021 unit tests, 725
integration, 93/93 visual with the three regenerated baselines stable
across two container runs. The FormData assertions are mutation-tested:
dropping name from Checkbox kills two cases, and disabling Select's
hidden inputs kills two more.

## Summary by CodeRabbit

- **New Features**
  - Checkbox, Toggle, and Select now support native form submission through optional `name` and `value` properties.
  - Select supports submitting single and multiple selected values.
  - Checkbox text is optional, with improved support for unlabelled accessible controls.
  - Web components expose the new form-related properties.
  - Component demos now show native form submission behavior.

- **Bug Fixes**
  - Checkbox input and change events now cross shadow DOM boundaries correctly.


murdore:
## Proof at `a0b8f3f`

The FormData readout in each clip is the assertion itself, printed on screen — not a proxy for it.

**Checkbox: absent when unchecked, `on` by default, `value` when given**

https://github.com/user-attachments/assets/b064e7b2-519b-4e1c-abce-6f46263abdf5

Three submits in sequence. First: `agree` is unchecked and is **absent entirely** — not present with an empty value, which is the bug consumers hand-rolling hidden inputs usually ship. `subscribe` submits `yes`. Then `agree` is checked and appears as `["agree","on"]`, the native default. Then `subscribe` is unchecked and disappears again, leaving `[["agree","on"]]`.

**Toggle: off is absent, on submits under its name**

https://github.com/user-attachments/assets/724efd25-7ea3-440a-bc57-328d1dcf3bc7

**Select: one entry per selection, matching `<select multiple>`**

https://github.com/user-attachments/assets/aa03dde6-55b3-420b-a4cf-c14ecc484b77

Deselecting removes that entry; deselecting everything removes the name entirely rather than submitting an empty value.

### The demo sections are documentation, not a test fixture

An earlier revision of this branch put that form in a Playwright-only route at `/form-association-fixture`. The Pages workflow uploads the entire `build` directory, so I confirmed `build/form-association-fixture.html` existed — a test fixture would have shipped to the public docs site.

It is now a "Native form submission" section on each component's own demo page. A consumer reading the Checkbox demo sees exactly how `name`/`value` behave, the suite has somewhere real to drive, and the sections get visual coverage. Nothing test-only ships.

### Mutation-tested after I rewrote the tests

The original tests were mutation-tested against the fixture. Since I replaced them, those results no longer applied to the code under review, so I re-ran the mutations against the rewritten suite:

```
MUTATION: Checkbox renders name={null}          -> 2 failed | 8 passed
MUTATION: Select's hidden inputs never render   -> 2 failed | 8 passed
RESTORED                                        -> 10 passed
```

One mutation of mine silently did not apply first time: my pattern for Select's guard did not match the real `{#if typeof name === 'string' && name.length > 0}`, so it ran against unmutated code and passed 10/10 — proving nothing while looking like a pass. The assert caught it. A mutation that fails to apply is indistinguishable from a well-tested implementation, which makes it worth stating.

### The reporting team measured the same defect from the other direction

I read source and concluded a synthetic click cannot reach the box. They drove their app in a real browser and found that at the checkbox's own centre point `document.elementFromPoint` returns an ancestor element — `sameNode: false` — so Playwright fails actionability and times out. Focus plus Space works.

Mine is an argument; theirs is an observation, and theirs would have stood even if I had misread the CSS. Their note is that this is a **testability** failure rather than a user-visible one — their app still toggles via the label's JS handler, but eleven of their E2E specs cannot drive the control at all. If those specs start passing after this lands, that is a larger payoff than the thirteen hidden inputs it deletes.

### Verification I ran myself on `a0b8f3f`

```
npm run check       0 errors
npm run lint        clean
npx vitest run      1021 / 1021
npx playwright      725 passed, 1 skipped
npm run build       exit 0, fixture absent from build/
scripts/visual      93 / 93, run twice
```

On the baselines: `--update-snapshots` rewrote **six**, but this change only touches three demo pages. I reverted `carousel`, `chat-message-list` and `hitl` and re-ran — all 93 pass against their original baselines, so those three were noise picked up under concurrent container load and committing them would have baked in unrelated churn. Only `checkbox`, `select` and `toggle` are in the commit.

Tara-ag (commented):
Yama review in progress — inline findings follow.

src/lib/Checkbox/Checkbox.svelte
```diff
@@ -53,6 +62,19 @@
       indeterminate = false;
     }
     onclick?.(checked);
+    // The box (role="checkbox") is the only element the user can actually
```
Tara-ag, line 65: ⚠️ MAJOR: Making `handleClick` async and awaiting `tick()` before dispatching on `nativeInputEl` risks a null-reference if the component unmounts in the microtask gap (e.g., a controlled table removing the row). It also makes the controlled-mode early return easy to break. Prefer `tick().then(() => { if (nativeInputEl) { ... } })` or guard `nativeInputEl` explicitly, and keep the function's synchronous intent clear.

docs/Select.md
```diff
@@ -110,6 +110,7 @@ An SVG source is **inlined into the component's DOM**, so an icon drawn with `cu
...
 | usePortal        | `boolean`                  | No       | `false`        | When `true`, the dropdown panel is portaled to `document.body` and positioned `fixed` relative to the trigger, so an ancestor with `overflow: hidden` or a scroll container (e.g. a table cell) cannot clip it. Placement follows the trigger on scroll/resize and flips above it when there's no room below. Portaled panels default to `z-index: 1000`; raise `--select-dropdown-z-index` to sit above an even higher overlay. |
+| name             | `string`                   | No       | -              | Sets the `name` under which selected ids submit as part of a surrounding `<form>` (`FormData`). Select renders no native `<select>`, so this renders one `<input type="hidden" {name} value={id}>` per selected id — the same shape a native `<select multiple>` produces: one entry per selection, none when nothing is selected. Omitting it leaves the DOM byte-identical to before.                                        |
```
Tara-ag, line 113: 💡 MINOR: Document that an empty `name=""` is treated as omitted (no hidden inputs rendered). The implementation checks `name.length > 0`, but the docs only say the default is `-`; consumers may be surprised that an empty string silently disables form participation.

tests/form-association.test.ts
```diff
@@ -0,0 +1,200 @@
...
+    throw new Error('expected FormData entries to serialize as an array');
+  }
+  return parsed;
+};
+
+const hasName = (entries: readonly (readonly string[])[], name: string): boolean =>
+  entries.some((entry) => Array.isArray(entry) && entry[0] === name);
+
+test.describe('Checkbox — FormData participation', () => {
+  test('an unchecked checkbox is absent from FormData; a checked one submits its value', async ({
```
Tara-ag, line 117: 💡 MINOR: These form-association contract tests are coupled to demo-page layout selectors (`checkbox-form-submit`, `select-form-multiple`, etc.). Per project standards, component-contract tests should use isolated fixtures or stable component testIds rather than demo-page-specific `data-pw` selectors. Consider adding a minimal fixture route and driving that instead, so reorganizing the docs page doesn't break the regression suite.

src/lib/Checkbox/Checkbox.svelte
```diff
@@ -53,6 +62,19 @@
...
+    // reach — the native input is aria-hidden, tabindex="-1" and
+    // pointer-events: none, purely a form-submission mirror. Because it
+    // never receives a real user click, the browser never fires a native
+    // `change` on it either, so a parent `<form>` listener (or a consumer's
+    // own `input.addEventListener('change', ...)`) never learns the value
+    // moved. `input`/`change` are dispatched manually to restore that signal.
+    // `tick()` first, so the dispatch happens after Svelte's reactive
+    // `input.checked` write actually lands — dispatching immediately after
+    // the assignment above would fire before the DOM reflects it.
+    await tick();
```
Tara-ag, line 75: ⚠️ MAJOR: The synthetic `input`/`change` events are dispatched with only `{ bubbles: true }`. Inside `<sui-checkbox>` (shadow DOM), these events will not cross the shadow boundary, so a form listener or `new FormData(hostForm)` outside the custom element won't observe them. Dispatch with `{ bubbles: true, composed: true }` to match native checkbox behavior in web-component mode.

src/lib/Toggle/Toggle.svelte
```diff
@@ -40,6 +42,8 @@
       type="checkbox"
       {checked}
       {disabled}
+      name={name ?? null}
```
Tara-ag, line 45: 💬 SUGGESTION: The Toggle focus indicator (`box-shadow: 0 0 1px #171717` in the style block) is a 1px low-contrast shadow that is hard to perceive. Since form participation may increase keyboard usage, consider strengthening it and exposing a CSS custom property (e.g., `--toggle-focus-ring`) to match the Checkbox focus-ring pattern.

src/lib/Checkbox/properties.ts
```diff
@@ -32,6 +32,16 @@ export type OptionalCheckboxProperties = {
    *  `aria-controls` to point at, or a consumer's own test attribute. Spread
```
Tara-ag, line 32: 🔒 CRITICAL: While `name`/`value` are correctly added below, `text` remains mandatory in `MandatoryCheckboxProperties` (lines 7-9, unchanged). A TypeScript consumer cannot write `<Checkbox name="agree" />` for a form-only checkbox without passing `text=""`. Since this PR advertises native form participation, make `text` optional with default `''` — the Svelte component already defaults it at runtime.

Tara-ag (changes requested):
Yama review complete. One CRITICAL and two MAJOR issues need resolution before merge; see inline comments for details and suggested fixes.

Tara-ag:

Review could not be completed: the time budget was exhausted before any PR diff or source files could be examined. No changed files, commits, symbols, or cross-references were verified, so no actionable findings could be produced. Please re-run the review with sufficient budget to inspect PR #605 file-by-file.

murdore:
## Re-pushed at `305ca98` — CRITICAL and both MAJORs closed, plus a correction to my own claim

The verdict above reviewed `a0b8f3f`. Mapping each finding:

| Severity | Finding | Closed by |
| --- | --- | --- |
| CRITICAL | Checkbox `text` still mandatory, so form-only usage impossible | `MandatoryCheckboxProperties` is now `Record<string, never>` and `text` is optional. `<Checkbox name="agree" />` compiles. |
| MAJOR | Synthetic events lack `composed: true` | Both `input` and `change` now dispatch with `bubbles: true, composed: true`, so they cross the shadow boundary in web-component mode. |
| MAJOR | `await tick()` risks a stale element on unmount | `nativeInputEl?.dispatchEvent(...)` — verified by mounting and unmounting a real Checkbox mid-`tick()`. |
| MAJOR | `pointer-events: none` not visible in the diff | It is present. The diff view simply did not include that hunk. |
| MINOR | Empty `name` undocumented | `docs/Select.md` now states an empty string is treated exactly as omitted. |

`composed: false` was the more serious of the two MAJORs and, like #604's, it was a hole in my own verification: everything I had tested ran in light DOM, where `bubbles` alone is sufficient.

## Correction: my visual number on this PR

I wrote **`scripts/visual 93 / 93, run twice`**. Those runs happened and passed. But I have since measured that this number does not mean what it reads as, and I should not have let it stand as evidence of stability.

Re-running the visual suite on this branch today produced **7 failures**:

```
carousel · chat · chat-message · hitl · relative-time · sankey-chart · stat-card
```

This branch touches exactly three routes — `checkbox`, `select`, `toggle`. **Zero overlap.** The machine was sustaining a load average above 200 from unrelated concurrent test runs, and the failures are timing-sensitive routes being starved, not regressions.

A control measurement on the functional side makes the same point: a timing-dependent reduced-motion test fails 1 run in 9 on one branch and **3 in 9** on a branch that does not contain the feature at all.

So: the green runs were real, the conclusion drawn from them was weaker than stated, and **CI's Linux runners are the signal to trust here**, not my laptop. `visual` is green on this PR in CI.

### Verification at `305ca98`

```
check   exit 0
lint    exit 0
vitest  1026 passed
```

The three recordings already attached remain accurate — the FormData readout on screen is the assertion itself. They were re-checked today against a non-blank-and-moving gate along with every other attachment on these PRs.

src/lib/Checkbox/Checkbox.svelte
```diff
@@ -53,6 +62,25 @@
...
+    // `change` are composed, and this component ships inside a shadow root
+    // (Checkbox.wc.svelte's `shadow: 'open'`) for every <sui-checkbox> consumer.
+    // `bubbles` alone stops at the shadow boundary; without `composed` the event
+    // never reaches a light-DOM `<form>`, silently defeating the whole point of
+    // wiring `name`/`value` onto this input for every web-component consumer.
+    // `tick()` first, so the dispatch happens after Svelte's reactive
+    // `input.checked` write actually lands — dispatching immediately after
+    // the assignment above would fire before the DOM reflects it.
+    await tick();
+    nativeInputEl?.dispatchEvent(new Event('input', { bubbles: true, composed: true }));
```
Tara-ag, line 89: ⚠️ MAJOR: `handleKeyDown` now invokes the async `handleClick`, so `input`/`change` events are dispatched after a microtask delay. Assistive technologies expect these events to be synchronously tied to the Space/Enter gesture, and the element could be unmounted before the tick resolves.

**Fix:** Keep keyboard activation synchronous. If DOM reflection is required, use `flushSync()` or a synchronous microtask, then dispatch immediately.

src/lib/Checkbox/Checkbox.svelte
```diff
@@ -36,14 +41,18 @@
         : null
   );
 
-  function handleClick(): void {
+  async function handleClick(): Promise<void> {
     if (disabled) {
       return;
```
Tara-ag, line 46: ⚠️ MAJOR: Making `handleClick` async and awaiting `tick()` before dispatching `input`/`change` introduces a race condition. Rapid clicks or keyboard repeats can interleave, producing duplicate or out-of-order events and inconsistent observer state.

**Fix:** Guard against re-entry (e.g., `let clicking = $state(false)`) or use a synchronous microtask flush (`queueMicrotask`) instead of `await tick()` so events stay ordered with the user gesture.

src/lib/Checkbox/properties.ts
```diff
@@ -1,14 +1,22 @@
...
-  CheckboxEventProperties;
+export type CheckboxProperties = OptionalCheckboxProperties & CheckboxEventProperties;
 
-export type MandatoryCheckboxProperties = {
-  text: string;
-};
+// No property is mandatory anymore -- `text` moved to OptionalCheckboxProperties
+// below. Kept exported (rather than deleted) only so an existing
+// `import type { MandatoryCheckboxProperties }` keeps compiling; it is
+// deliberately NOT intersected into CheckboxProperties above; Record<string, never>
```
Tara-ag: 🔒 CRITICAL: `MandatoryCheckboxProperties` is still exported publicly via `src/lib/index.ts` (`export type * from './Checkbox/properties'`). Changing it from `{ text: string }` to `Record<string, never>` silently breaks any downstream consumer that imported this type to build wrappers or derive subtypes. This is a public API breaking change without a major-version bump or migration note.

**Fix:** Restore `MandatoryCheckboxProperties = { text: string }` (or `{ text?: string }` to reflect the new runtime default). If intentional removal is required, mark the commit as breaking (`feat!:` or `BREAKING CHANGE:` footer) and add a migration note in `docs/Checkbox.md`.

src/lib/Checkbox/Checkbox.composed-event.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    componentInstance = mount(Checkbox, {
+      target: shadowMountPoint,
+      props: { name: 'agree', text: 'Agree to terms' }
+    });
+
+    const label = shadowMountPoint.querySelector('label.container');
+    expect(label).toBeInstanceOf(HTMLLabelElement);
+    label?.dispatchEvent(new MouseEvent('click', { bubbles: true, cancelable: true }));
+
```
Tara-ag, line 66: 💡 MINOR: The test waits via `setTimeout(..., 0)` instead of `await tick()`. This is slower, less deterministic, and contradicts the comment that explains the implementation uses `tick()`. Flaky macrotask timing can produce false negatives under load.

**Fix:** Import `tick` from `svelte` and replace the `setTimeout` waits with `await tick()`.

docs/Select.md
```diff
@@ -110,6 +110,7 @@ An SVG source is **inlined into the component's DOM**, so an icon drawn with `cu
...
 | usePortal        | `boolean`                  | No       | `false`        | When `true`, the dropdown panel is portaled to `document.body` and positioned `fixed` relative to the trigger, so an ancestor with `overflow: hidden` or a scroll container (e.g. a table cell) cannot clip it. Placement follows the trigger on scroll/resize and flips above it when there's no room below. Portaled panels default to `z-index: 1000`; raise `--select-dropdown-z-index` to sit above an even higher overlay. |
+| name             | `string`                   | No       | -              | Sets the `name` under which selected ids submit as part of a surrounding `<form>` (`FormData`). Select renders no native `<select>`, so this renders one `<input type="hidden" {name} value={id}>` per selected id — the same shape a native `<select multiple>` produces: one entry per selection, none when nothing is selected. An empty string is treated exactly as omitted -- no hidden inputs are rendered, so `name=""` silently disables form participation rather than submitting under an empty key. Omitting it leaves the DOM byte-identical to before.                                        |
```
Tara-ag: 💡 MINOR: Documenting that `name=""` silently disables form participation is good, but this diverges from native `<select name="">`, which submits under an empty key. Consumers migrating from native controls may be surprised.

**Fix:** Add a brief note explaining the deliberate divergence: Select has no underlying native control, so an empty key is meaningless and is therefore treated as omitted.

src/routes/components/toggle/+page.svelte
```diff
@@ -76,3 +80,27 @@
     <Toggle id={inputId} text={`Blank id setting ${index}`} testId={`toggle-blank-id-${index}`} />
   </div>
 {/each}
+
+<h3>Native form submission</h3>
+<p>
+  A switch with a <code>name</code> submits with the surrounding <code>&lt;form&gt;</code>: absent
+  while off, and present as <code>on</code> once switched on.
+</p>
+<form
```
Tara-ag, line 89: 💡 MINOR: The demo form works for the Svelte component, but when consumed as `<sui-toggle>` the native input lives inside an open shadow root. Standard form submission across shadow boundaries depends on browser support for form-associated custom elements, which is not guaranteed and is not mentioned in the docs.

**Fix:** Add a note in `docs/Toggle.md` clarifying that the `<sui-toggle>` web component wraps its native input in a shadow root, so the form-participation behavior shown here applies to the Svelte component; WC consumers may need additional handling.

src/lib/Select/Select.svelte
```diff
@@ -518,6 +519,16 @@
   bind:this={containerEl}
   {...typeof testId === 'string' ? { 'data-pw': testId, testID: testId } : {}}
 >
+  {#if typeof name === 'string' && name.length > 0}
+    <!-- No native <select> exists to carry `name`/`value` -- this is a custom
+         combobox/listbox. One hidden input per selected id mirrors a native
+         `<select multiple>`, which likewise submits one FormData entry per
```
Tara-ag, line 525: ⚠️ MAJOR: The hidden inputs use `{#each value as id (id)}`. When an id is selected, deselected, and re-selected, Svelte may reuse the same input element. Hidden inputs are not real form-associated controls, so reused nodes can retain stale state in edge cases (browser autofill, form restoration, or form libraries tracking input lifecycle). Overlapping ids across value changes also cause move/reuse rather than a clean FormData reflection.

**Fix:** Use an unkeyed `{#each value as id}` for hidden inputs, or use a composite key that changes when the set changes. If keyed-by-id is intentional, add a comment justifying why reuse is safe here.

Tara-ag (changes requested):
Yama review complete. One CRITICAL public-API breaking change blocks merge: `MandatoryCheckboxProperties` was changed to `Record<string, never>` while still exported from `src/lib/index.ts`. Two MAJOR issues (async event dispatch races in Checkbox, Select hidden-input keying) and three MINOR issues (test timing, docs gaps) are also noted. Please address the CRITICAL item and reassess the MAJOR reactivity concerns.

murdore:
## Review threads worked through at `66ba5b0`

14 unresolved threads: 8 fixed · 1 refuted · 3 already correct · 2 deferred.


### **Fixed**

**`src/lib/Checkbox/Checkbox.svelte:65`** — MAJOR: async handleClick + await tick() before dispatch risks a null-reference / race if the component unmounts in the microtask gap.

> Rewrote handleClick (Checkbox.svelte lines 44-91) from `async function ... await tick()` to a plain synchronous function using `flushSync(() => { checked = !checked; ...})`, so the reactive `input.checked` write lands before dispatch without ever yielding to the event loop -- no unmount gap remains. nativeInputEl?.dispatchEvent stays optional-chained regardless. Verified: npm run check exit 0, npx vitest run 1026/1026 passing.

**`src/lib/Toggle/Toggle.svelte:45`** — SUGGESTION: Toggle's 1px focus box-shadow is low contrast and not exposed as a CSS custom property.

> Changed `.input-checkbox:focus + .slider { box-shadow: 0 0 1px #171717; }` to `box-shadow: var(--toggle-focus-ring, 0 0 0 3px rgba(33, 150, 243, 0.3));` (Toggle.svelte line 119), matching Checkbox's existing --checkbox-focus-ring pattern, and documented --toggle-focus-ring in docs/Toggle.md's CSS Variables table.

**`src/lib/Checkbox/Checkbox.svelte:82`** — MAJOR: handleKeyDown invokes async handleClick, delaying input/change past the synchronous AT gesture.

> Same flushSync rewrite as Checkbox.svelte:65 -- handleClick (lines 44-91) is now fully synchronous (no async/await), so handleKeyDown's call to it dispatches input/change within the same synchronous call stack as the keydown event, not after a microtask.

**`src/lib/Checkbox/Checkbox.svelte:46`** — MAJOR: async handleClick + await tick() can interleave on rapid clicks, causing duplicate/out-of-order events.

> Same flushSync rewrite: with no `await` anywhere in handleClick, JS run-to-completion guarantees a second synchronous call cannot interleave mid-flight with a first, eliminating the described race.

**`src/lib/Checkbox/properties.ts:8`** — CRITICAL: MandatoryCheckboxProperties silently changed from {text:string} to Record<string,never>, breaking downstream type consumers.

> Changed `export type MandatoryCheckboxProperties = Record<string, never>;` to `export type MandatoryCheckboxProperties = { text?: string };` (properties.ts line 12), the reviewer's own suggested alternative -- restores compatibility for a consumer assigning/reading `text` through this type while keeping it optional. grep -rn MandatoryCheckboxProperties shows no other usages to update; npm run check exits 0.

**`src/lib/Checkbox/Checkbox.composed-event.test.ts:63`** — MINOR: test waits via setTimeout(...,0) instead of tick(), contradicting the tick()-based implementation comment.

> Removed both `await new Promise((resolve) => setTimeout(resolve, 0))` waits: handleClick's dispatch is now synchronous (flushSync, not tick()), so there is nothing to await -- assertions run immediately after dispatchEvent. Verified: npx vitest run src/lib/Checkbox -> 2 files / 5 tests passed.

**`docs/Select.md:113`** — MINOR: doc should explain why empty name diverges from native <select name=""> (which submits under an empty key).

> Appended to docs/Select.md line 113: 'This deliberately diverges from a native <select name="">, which does submit under an empty key: Select has no underlying native control to inherit that behaviour from, so an empty key would be a hidden input this component invented, not one the platform requires.'

**`src/routes/components/toggle/+page.svelte:89`** — MINOR: demo's native-submission example doesn't note that <sui-toggle>'s input lives in a shadow root.

> Added a new 'Form participation in a shadow root' subsection to docs/Toggle.md (after 'Shadow-root label scope', before 'Prop names across the two distributions') explaining <sui-toggle> wraps its input in an open shadow root and does not implement the form-associated custom elements API, so WC consumers may need additional handling to reach a light-DOM form's FormData.


### **Refuted**

**`src/lib/Select/Select.svelte:525`** — MAJOR: keyed {#each value as id (id)} for hidden inputs could reuse a DOM node with stale state across deselect/reselect.

> Select.svelte lines 528-529: the each block's key (`id`) is identical to the rendered content (`value={id}`), and `name`/`disabled` are uniform outer-scope values reactive on every surviving node regardless of keying -- so no per-node state can ever drift from its key. Svelte's keyed reconciliation guarantees a node is reused only for the SAME key, never repurposed for a different id (it destroys and recreates on removal/re-add), so the described failure mode does not occur; switching to an unkeyed each (the reviewer's suggested fix) would instead use index-based reuse, the pattern actually associated with stale-attribute bugs. Added a regression test in tests/form-association.test.ts ('deselecting then reselecting an option in multiple-select Select still submits it') exercising exactly this scenario against FormData.


### **Already correct**

**`docs/Select.md:113`** — MINOR: document that empty name="" is treated as omitted (no hidden inputs rendered).

> docs/Select.md line 113, unmodified by this pass (confirmed by git diff context), already read: 'An empty string is treated exactly as omitted -- no hidden inputs are rendered, so name="" silently disables form participation rather than submitting under an empty key.'

**`src/lib/Checkbox/Checkbox.svelte:81`** — MAJOR: synthetic input/change events dispatched without composed:true won't cross the shadow-DOM boundary.

> Checkbox.svelte lines 89-90: `nativeInputEl?.dispatchEvent(new Event('input', { bubbles: true, composed: true }))` and the matching 'change' dispatch already both carry composed:true, confirmed by grep. Checkbox.composed-event.test.ts exercises exactly this shadow-root scenario and passes.

**`src/lib/Checkbox/properties.ts:40`** — CRITICAL: text remains mandatory, blocking <Checkbox name="agree" /> without text="".

> properties.ts line 20 (OptionalCheckboxProperties) already declares `text?: string;`, pre-existing before this pass -- <Checkbox name="agree" /> already compiles without text.


### **Deferred**

**`tests/form-association.test.ts:117`** — MINOR: form-association tests are coupled to demo-page data-pw selectors instead of isolated fixtures.

> tests/form-association.test.ts lines 79-84 carry an existing comment explaining this is a deliberate choice: each demo page's 'Native form submission' section is documentation-first, and this suite drives it on purpose. Building a parallel fixture-route + stable-testId layer for Checkbox/Toggle/Select is a test-infrastructure redesign, not a smallest-correct-change fix, so left as a follow-up rather than done inline here.

**`src/wc/components/Toggle.wc.svelte:16`** — CRITICAL: sui-toggle/sui-checkbox are not form-associated custom elements (no ElementInternals/setFormValue).

> grep -rn 'formAssociated|attachInternals|ElementInternals' node_modules/svelte/src/ returns nothing -- Svelte 5.56.3's compiled customElement runtime has no built-in support for this, so implementing it requires hand-authoring a custom HTMLElement base class wrapping Svelte's generated element across 3 components (Toggle/Checkbox/Select .wc.svelte), a heavy architectural lift, not a thread-sized fix. It is also currently unverifiable for sui-checkbox specifically: Checkbox.composed-event.test.ts's own top-of-file comment (lines 5-25) documents that every <sui-checkbox> instance throws `TypeError: this.attributes is not iterable` on connect today (pre-existing, unrelated defect), so it can't even mount. Added a docs caveat instead (docs/Toggle.md, 'Form participation in a shadow root') so consumers aren't misled meanwhile; the ElementInternals work itself is left as a follow-up.


### Verification

```
check   exit 0
lint    exit 0
vitest  Test Files 54 passed (54); Tests 1026 passed (1026) -- Checkbox subset re-run in isolation: 2 files / 5 tests passed
```

src/wc/components/Toggle.wc.svelte
```diff
@@ -11,7 +11,9 @@
       testId: { type: 'String', attribute: 'test-id' },
       inputId: { type: 'String', attribute: 'input-id' },
       inputAriaLabel: { type: 'String', attribute: 'input-aria-label' },
-      inputAriaLabelledby: { type: 'String', attribute: 'input-aria-labelledby' }
+      inputAriaLabelledby: { type: 'String', attribute: 'input-aria-labelledby' },
+      name: { type: 'String', reflect: true },
+      value: { type: 'String', reflect: true }
```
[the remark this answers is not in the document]
murdore: Confirmed as a real limitation, and now documented on both components rather than one.

`<sui-toggle>` and `<sui-checkbox>` cannot carry a form value across their shadow boundary, because Svelte 5 does not emit `static formAssociated = true` or attach `ElementInternals` for custom elements it compiles — there is no option to opt in, so this is not something the wrapper can fix from here. Implementing it would mean hand-writing the element definitions outside Svelte's `customElement` compiler support, which is a much larger change than this PR.

What I have done instead is make the limitation impossible to miss from either doc: `docs/Toggle.md` already carried a "Form participation in a shadow root" section naming `static formAssociated` + `ElementInternals`; `docs/Checkbox.md` did not, and now carries the same section (`ded8b4b`). The Svelte `<Checkbox>`/`<Toggle>`/`<Select>` components — which is what this PR is actually about — put their `<input>` in the same document as the `<form>` and participate natively.

Leaving this thread open on purpose as the tracking point for the `ElementInternals` work, since the underlying gap is still there.

tests/form-association.test.ts
murdore: Agreed on the principle, and deferring it deliberately rather than silently.

These are end-to-end form-submission tests: the thing under test is whether a real `<form>` collects a real `FormData` from these components, which needs a mounted page with a form around them. The demo routes already are that page, so driving them was the cheapest way to get genuine coverage into this PR. The coupling you describe is real — reorganising the docs page will break these.

Two reasons to do the fixture route as follow-up rather than here. It is a new test-only route, which is scope this PR does not otherwise touch; and the same argument applies to several existing specs in `tests/`, so a fixture route worth adding should serve all of them rather than being introduced for one file. Doing it piecemeal here would leave the codebase with two conventions instead of one.

Tracking it as follow-up work, not as a wontfix.

murdore: Agreed, and now tracked properly rather than living in this thread.

Filed as https://github.com/juspay/svelte-ui-components/issues/621 with the two things that need care in the fix: the fixture route must be excluded from the visual-regression route sweep (or it becomes a 95th baseline that exists only to serve a test), and it must not prerender into the published docs site as a stray page.

Worth saying plainly: this thread being the only record of the follow-up is exactly the failure mode. I said on this PR that it was "follow-up work, not a wontfix" and then never filed it, so an open thread on a merged PR was carrying the memory. That's what the issue fixes.

Resolving here; the work is on the issue.

src/wc/components/Toggle.wc.svelte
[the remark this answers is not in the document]
murdore: Correct, and not addressed by this PR. Filed as https://github.com/juspay/svelte-ui-components/issues/623.

Confirming the diagnosis: the wrappers compile with `shadow: 'open'`, the input carrying `name`/`value` lives inside that shadow root, and a shadow-encapsulated control is not a form control of the outer form — so `FormData` omits it.

Your sharpest point is the one I want on the record, because it is about the test rather than the code: *"The current FormData tests mount `$lib` components, not the actual custom elements."* The suite passes on the path that works and never touches the path that doesn't. So this is a gap the tests are structurally blind to, not one they caught — and the issue requires a test driving the custom elements themselves, in a real form, as acceptance.

One correction to the suggested fix: `ElementInternals.setFormValue()` is the right mechanism, but it cannot simply be added here. Svelte 5's `customElement` compiler option exposes no way to declare `formAssociated` or reach `ElementInternals` on the generated class, so this needs either a hand-written host class wrapping the Svelte element, or a light-DOM input outside the shadow root. The issue lays out that trade-off rather than presuming the answer.

Resolving here; the work is on the issue.

PR #606, opened 2026-09-11
feat(charts): show optional per-series aggregates in legends

murdore:
Add an opt-in `aggregate?: 'sum' | 'average' | 'min' | 'max' | 'none'`
(plus `aggregateFormat`) per series on BarChart, LineChart, AreaChart,
and DualAxisBarChart. When set, the legend renders the computed value
beside the series label; omitted (the default) the legend is
byte-identical to before.

This is a new capability, not a fix: no `ChartDataType` concept exists
in this library, there is no aggregate legend row today, and the
stacked y-domain / pie percentage-denominator sums a consuming team
flagged as "should be averaged" are correct by definition for what
they compute. Nothing here changes any existing total.

Decisions pinned by tests:
- average/min/max skip null/non-finite points rather than treating
  them as zero (aggregate.test.ts)
- an aggregate over zero real points (empty series, or all
  null/non-finite) renders nothing, never NaN or a fabricated 0
  (aggregate.test.ts, Legend.test.ts)
- stacked-chart aggregates are always per-series, never the
  per-category stack total (AreaChart/legend-aggregate.test.ts)

aggregateFormat defaults to each chart's own existing formatter
(yTickFormat / per-axis valueFormat for DualAxisBarChart) instead of
introducing a parallel formatting mechanism.

The aggregate renders BESIDE the legend's toggle button, never inside it.
Every bit of text inside a button becomes part of its accessible name, so an
aggregate within the toggle would make that control's name change whenever the
data does -- "Revenue $6,500" one load, something else the next. A name that
moves breaks voice-control targeting, breaks any consumer test matching it
exactly (Testing Library's getByRole matches the whole string by default, while
Playwright's matches a substring, so identical markup breaks one consumer's
suite and passes another's), and announces a figure as the control's identity
rather than as information about the series. It also stops a number being a
click target for toggling a series, which it never should have been.

Nothing combined an interactive legend with aggregates, which is why that went
unnoticed: interactiveLegend is chart-level and aggregate is per-series, so the
combination is reachable by any consumer while no demo exercised it. The
bar-chart demo now enables both, and tests/legend-aggregate-a11y.test.ts asserts
a toggle is named by its series alone, that the figure is still rendered
alongside it, and that clicking the figure does not toggle the series.

The restructure is pixel-neutral: the button carries the same flex layout and
gap the row did, so all 93 visual baselines pass UNCHANGED rather than being
regenerated -- stronger evidence for a semantics-only change than new baselines
would have been.

Verified on this commit: check 0, lint 0, 1059 unit tests, 718 integration,
93/93 visual against existing baselines. Putting the aggregate back inside the
button fails all three a11y cases and two unit cases.

Not covered: PieChart has no per-series aggregation concept to apply
this to; FunnelChart and SankeyChart have no legend to attach it to.

## Summary by CodeRabbit

* **New Features**
  * Added optional per-series legend aggregates for Area, Bar, Dual-Axis Bar, and Line charts.
  * Supported calculations include sum, average, minimum, and maximum.
  * Added customizable aggregate formatting, with chart- or axis-specific defaults.
  * Aggregates ignore null and non-finite values and remain separate from interactive legend controls.

* **Documentation**
  * Updated chart documentation and examples with aggregate configuration and usage guidance.

* **Tests**
  * Added coverage for formatting, empty data, stacking, accessibility, and legend interaction behavior.


murdore:
## Proof at `18587e5` — stills, not video, because the output is static

These are crops from the committed visual baselines, so what you see is exactly what the suite compares.

### The best evidence is one image showing both states

bar-chart legend aggregates

The middle chart is an existing demo with no `aggregate` set: its legend reads `Alpha | Beta | Gamma`, unchanged. The chart below it has `aggregate` set: `Revenue $6,500` and `Conversion Rate 3.0%`. Same page, same component, opt-in and off by default — that is the property that matters most for a package 56 projects install, and it is visible in a single frame rather than asserted in prose.

### The arithmetic, verified against the demo's own data

I checked these rather than reading them off the picture:

| series | values | aggregate | computed | rendered |
| --- | --- | --- | ---: | ---: |
| Revenue | 1200, 1800, 1400, 2100 | `sum` | 6500 | **$6,500** |
| Conversion Rate | 2.4, 3.1, 2.8, 3.6 | `average` | 2.975 | **3.0%** |

**These discriminate.** A summed conversion rate would read `11.9%` — nothing like `3.0%`. The baseline itself distinguishes a correct average from the bug this feature exists to avoid, which is what makes it evidence rather than decoration.

dual-axis legend aggregates

Dual-axis, where each series formats with **its own axis** and neither specifies `aggregateFormat`: `Revenue ($) $298,900` (sum of 42000, 38500, 51200, 46800, 58300, 62100 = 298,900 exactly) and `CTR (%) 4.0%` (average of 3.2, 2.8, 4.1, 3.7, 4.8, 5.2 = 3.9667, rendered at the right axis's one decimal). A summed CTR would read `23.8%`.

area-chart legend aggregates

### This is a feature, not a fix

The consuming team originally reported this as a bug — "chart legends sum metrics that should be averaged." That was investigated and rejected on the evidence: there is no `ChartDataType` concept, there is no aggregate legend row today, and the only sums that exist (the stacked y-domain, the pie percentage denominator) are correct by definition for what they compute. **Nothing currently shipped is wrong.** Nothing in this PR should be read as repairing a defect.

### The decisions worth arguing with, each pinned by a test

- **Null and non-finite points are skipped, not zero-filled.** Zero-filling would drag an average toward zero and quietly misreport it.
- **An aggregate over zero real points renders nothing** — never `NaN`, never a fabricated `0`, including for `sum`.
- **Aggregates are per-series, never the per-category stack total**, even when `stacked` is true.

Mutations confirming those are load-bearing: null-skip → zero-fill kills 4 cases; removing the empty guard kills 3, one of which reproduces a literal `"NaN"` on screen; forcing the legend to always render the aggregate span kills 4 of 5.

### Coverage, including what was skipped

Covered: BarChart, LineChart, AreaChart, DualAxisBarChart — all four consume the shared `Legend.svelte`. Skipped with reasons: PieChart has no multi-series concept to hang an aggregate on (a slice already is the aggregate); FunnelChart and SankeyChart have no legend to attach it to.

### Verification I ran myself on `18587e5`

```
npm run check       0 errors
npm run lint        clean
npx vitest run      1058 / 1058
npx playwright      715 passed, 1 skipped
scripts/visual      93 passed, exit 0   (full suite)
```

The full visual suite matters here: the task report cited "4/4 passed", which was a filtered run over this branch's own four charts out of 93 baselines. I ran the whole suite — it is green, but a partial run reported as a pass is exactly how an unrelated regression slips through, so the number is corrected rather than repeated.

Tara-ag:

Review could not be completed within the available time budget. No PR diff, changed files, or codebase context were read, and no findings were submitted through the required submit_review gate. Therefore the change cannot be confirmed safe to merge. A fresh review with sufficient time is needed to inspect each modified file, verify public API / accessibility / reactivity / SSR concerns, and gate any findings before posting comments.

murdore:
## Accessibility defect found and fixed at `a718e48`

Sachin asked me to audit the three PRs opened alongside this one for the defect class that turned up in #603 — **a prop that looks cosmetic and silently carries behaviour**. This PR had one. The other two did not.

### What was wrong

The aggregate `<span>` was rendered **inside** the legend's toggle `<button>`. Every bit of text inside a button becomes part of its accessible name, so I enabled `interactiveLegend` alongside `aggregate` and read the names back:

```
["Alpha", "Beta", "Gamma", "Revenue $6,500", "Conversion Rate 3.0%"]
```

An opt-in, apparently-visual prop was making an interactive control's accessible name **data-dependent** — `Revenue $6,500` on one load, something else after the next fetch. That:

- breaks voice-control targeting, since "click Revenue" stops matching
- breaks any consumer test matching the name exactly — **Testing Library's `getByRole` matches the whole string by default, while Playwright's matches a substring**, so identical markup breaks one consumer's suite and sails through another's
- announces a figure as the control's *identity* rather than as information about the series

`interactiveLegend` is chart-level and `aggregate` is per-series, so the combination is reachable by any consumer — but **no demo combined them**, which is exactly why it went unnoticed. Same reason the wrapper bug in #603 survived: the risky combination was never exercised.

### The fix is structural, not an `aria-label` patch

The aggregate is now a sibling of the button inside the legend item. An `aria-label` would have stabilised the name while leaving the number inside the click target — fixing the symptom I measured and leaving two I hadn't: a screen-reader user still unable to reach the figure as data, and a number that still toggles a series when clicked.

Now: the button is named by its series alone, the figure sits beside it, and clicking the figure does nothing.

### The demo now combines both features

`interactiveLegend` is enabled on the aggregate chart, and `tests/legend-aggregate-a11y.test.ts` asserts a toggle is named by its series alone, that the figure is still rendered alongside, and that clicking the figure does not toggle the series. The gap that hid this is closed, not just the instance.

### Pixel-neutral, and that is the stronger evidence

```
scripts/visual-test   93 passed, 0 baselines changed
```

I did not regenerate the baselines. The restructured markup was compared against the **existing** ones and matched — `.legend-toggle` carries the same flex layout and 6px gap the row did, so a button renders identically to the div it replaced. For a semantics-only change, unchanged baselines passing is better evidence than new baselines agreeing with themselves.

### Mutation-tested

```
MUTATION: aggregate back inside the button
  -> 3 failed (a11y integration) + 2 failed (Legend unit)
RESTORED
  -> all pass
```

### Re-verified on `a718e48`

```
npm run check     0 errors
npm run lint      clean
npx vitest run    1059 / 1059
npx playwright    718 passed, 1 skipped
scripts/visual    93 passed, existing baselines unchanged
```

### Two process notes, recorded rather than hidden

**My visual run failed twice and I retried on a truncated log.** My own command piped the run through `tail -3`, so I was diagnosing a failure from its last three lines — which structurally could not contain the cause. I blamed container contention, waited for the machine to idle, and got the identical failure. The third run, with the full log kept, succeeded: the earlier two were transient webServer startup timeouts. Roughly fifteen minutes spent reasoning about evidence I had truncated myself.

**I force-pushed once with a stale commit message.** The script updating the message hit a failed assertion, but the `git commit --amend` on the following line ran regardless, so the code landed correctly with a message that described none of it. Caught on the next read, corrected in `a718e48`. An assert that fires does not stop the commands after it.

murdore:
## Review threads worked through at `fca9625`

4 unresolved threads: 4 fixed.


### **Fixed**

**`docs/AreaChart.md:87 (+ docs/BarChart.md:161, docs/DualAxisBarChart.md:102-103)`** — Docs said only 'average'/'min'/'max' skip null/non-finite values, leaving 'sum' behavior for invalid values and zero-real-points undocumented/implied-different.

> src/lib/_chart/aggregate.ts:41-57 -- the skip-invalid-value loop (`if (value === null || !Number.isFinite(value)) continue;`) and the `if (count === 0) return null;` zero-real-points guard run for every aggregate kind, sum included, before the switch on lines 59-70; the doc comment on lines 11-22 of that same file already states this. Reworded all three docs (docs/AreaChart.md, docs/BarChart.md, docs/DualAxisBarChart.md) so the skip clause reads 'Every kind, `sum` included, skips ... (this only changes the result for `average`/`min`/`max`...)' and the zero-points clause reads '...renders nothing rather than `NaN` or a fabricated `0`, `sum` included.'

**`docs/LineChart.md:246`** — Same gap as the AreaChart/BarChart/DualAxisBarChart docs: LineChart's aggregate section didn't mention that `sum` also skips non-finite values and also renders nothing over zero real points.

> Same aggregate.ts evidence as above (the function is shared across all four chart components). Edited docs/LineChart.md lines 246-248 to read 'every kind, `sum` included, skips non-finite `y` values ... an aggregate over zero real points renders nothing rather than `NaN` or a fabricated `0`, `sum` included.'

**`src/lib/BarChart/BarChart.svelte:596`** — Legend aggregates were computed from `resolvedSeries` (post normaliseToFirstPoint/topN transforms) instead of the series' real values, so a normalised or topN-clipped chart showed a wrong aggregate (e.g. a percentage sum instead of a currency sum).

> Confirmed the bug first: BarChart.svelte:152-169 `normalisedSeries` rewrites `s.data[].value` to `(d.value / baseline) * 100` when `normaliseToFirstPoint` is true, and `resolvedSeries` (line 180) is derived from `normalisedSeries` (plus topN clipping/overflow at 196-206) -- the old `resolvedSeries.map(...)` at line 587 fed those transformed values into `formatSeriesAggregate`. Changed the legend mapping to iterate `rawSeries` (line 142: `isMulti ? series! : [...]`, the untransformed input) instead, since `rawSeries` and `resolvedSeries` share the same index/order so hidden-state, color, and formatter lookups by index `i` are unaffected. Added a regression test in src/lib/BarChart/legend-aggregate.test.ts ('aggregates the raw values, not normaliseToFirstPoint percentages') using a baseline of 50 so raw sum (300) and normalised-percentage sum (600) diverge -- it asserts '$300'. Full vitest

**`tests/legend-aggregate-a11y.test.ts:29`** — The 'toggle keeps the series name' a11y test asserted on `button.legend-toggle` textContent (page-wide, unscoped) rather than the actual accessible name, so an aria-label/aria-labelledby regression that changed the accessible name without touching visible text would not fail this test.

> Confirmed the gap: Legend.svelte:24-34 renders the button with no aria-label today, so its accessible name currently equals its visible text -- but `node.textContent` (the old assertion) reads DOM text regardless of aria attributes, so it can't detect an aria-label added later. Playwright 1.60.0 (installed, see @playwright/test package.json) already provides `toHaveAccessibleName` and it's used elsewhere in this suite (tests/checkbox-accessible-name.test.ts, tests/snippet-copy-state.spec.ts). Rewrote the test to scope to `page.getByTestId('bar-legend-aggregate-chart')` and assert `toggles.nth(0)`/`nth(1)` `toHaveAccessibleName('Revenue')` / `('Conversion Rate')`, matching the reviewer's suggested fix and this file's other two tests' existing scoping pattern.


### Verification

```
check   exit 0
lint    exit 0
vitest  Test Files 58 passed (58); Tests 1060 passed (1060)
```

PR #607, opened 2026-09-12
fix(chat-message): scroll a wide table in the wrapper, not the table itself

murdore:
`MarkdownText` already separates the scroll container from the table, and states why in its own styles: `display: block` on a `<table>` is what makes `overflow-x` work on the element, but it strips the table's semantics, so a screen reader loses row and column navigation.

`ChatMessage` renders the same markdown through the same renderer and emitted the same `.markdown-table-wrapper` — then styled the table instead of the wrapper. A chat table was therefore `display: block`: scrollable, and no longer a table to assistive technology. Chat is where markdown tables overwhelmingly land, so this was the surface that mattered most and the one that had it wrong.

The bare `.body :global(table)` rule stays as the fallback for a table in caller-supplied `html`, which arrives with no wrapper of its own.

## What this is not

The wrapper's focus ring is a **legibility** change, not a conformance fix. I claimed a WCAG 2.4.7 failure earlier and withdrew it: the browser's own `auto 1px` ring was already there. The mutation run below re-confirms that independently (`Received: "auto"`). 1px is simply easy to lose against a message bubble.

## Verification

Red-green-green, by removing the two CSS rules and rebuilding:

| stage | result |
| --- | --- |
| green | 2 passed |
| rules removed | `Expected: "table", Received: "block"` · `Expected: "solid", Received: "auto"` |
| restored | 2 passed |

The earlier version of this test asserted only *"an outline exists"*, which passes against the UA default — so it tested nothing. It now asserts the 2px solid treatment.

```
check     exit 0
lint      exit 0
vitest    1021 passed
functional 717 passed
```

## Baselines

- `chat-message.png` — grows by the new wide-table demo section.
- `markdown-text.png` — changes in exactly one 101-row band, which is that route's *"Inside ChatMessage"* section. The table there was shrink-wrapped to its content and now fills the bubble. I cropped and looked at it rather than inferring it.

Visual suite results from this machine are not included: it is sustaining a load average above 200 from other sessions, and a same-branch run produced failures on seven routes with zero overlap with anything touched here. CI's Linux runners are the trustworthy signal.


src/lib/ChatMessage/ChatMessage.svelte
```diff
@@ -578,6 +583,40 @@
...
+
+  /* Without an intrinsic width the table is exactly its container's width and a
+     wide table compresses instead of scrolling. `max-content` supplies it;
+     `min-width` keeps a narrow table filling the bubble. */
+  .body :global(.markdown-table-wrapper table) {
+    display: table;
+    width: max-content;
+    min-width: 100%;
+    overflow-x: visible;
+    margin: 0;
```
Tara-ag, line 617: 💡 MINOR: New CSS custom property `--chat-message-table-focus-outline-color` is introduced but not documented in `docs/ChatMessage.md`. Since this is a published library and CSS variables are part of the public theming contract, please add it to the CSS Variables table with default `#3b82f6` and a short description so downstream consumers can discover and theme it.

tests/chat-message-table.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    // The wrapper is the scroll container.
+    expect(styles.wrapperOverflowX).toBe('auto');
+    expect(styles.wrapperDisplay).toBe('block');
+    // And the table is still a table -- this is the assistive-technology half.
+    expect(styles.tableDisplay).toBe('table');
+    // The renderer makes the wrapper a tab stop so the scroll is reachable.
+    expect(styles.tabindex).toBe('0');
+  });
+
+  test('the focusable wrapper shows a focus ring', async ({ page }) => {
```
Tara-ag, line 50: 💡 MINOR: Programmatic `.focus()` alone often does not trigger `:focus-visible` in Chromium/WebKit, so this test may read `outline-width: 0px` and flake. The existing `tests/markdown-text-table-scroll.test.ts` presses an arrow key after focusing to settle the keyboard-modality heuristic. Add `await page.keyboard.press('ArrowRight');` after `wrapper.focus()` before reading the outline.

Tara-ag (commented):
Approving with two minor notes:

1. The new `--chat-message-table-focus-outline-color` CSS variable should be documented in `docs/ChatMessage.md` since it's part of the public theming contract.
2. The focus-ring Playwright test should press an arrow key after focusing to reliably trigger `:focus-visible`, matching the existing `markdown-text-table-scroll.test.ts` pattern.

The core fix is correct and well-tested: separating the scroll container from the `<table>` preserves table semantics for assistive technology while keeping wide tables scrollable.

murdore:
## Review threads worked through at `945e5f6`

2 unresolved threads: 2 fixed.


### **Fixed**

**`src/lib/ChatMessage/ChatMessage.svelte:617`** — New public CSS custom property --chat-message-table-focus-outline-color was introduced but not documented in docs/ChatMessage.md's CSS Variables table.

> Confirmed the variable is used at src/lib/ChatMessage/ChatMessage.svelte:617 (`outline: 2px solid var(--chat-message-table-focus-outline-color, #3b82f6);`) and was absent from docs/ChatMessage.md (grep for the string returned only the .svelte usage, none in docs/). Added a row to the CSS Variables table in docs/ChatMessage.md: `| \`--chat-message-table-focus-outline-color\` | \`#3b82f6\` | outline-color | Focus ring color of a rendered table's scroll wrapper. |`, matching the existing table's format and default value.

**`tests/chat-message-table.test.ts:50`** — The 'focusable wrapper shows a focus ring' test calls wrapper.focus() and reads outline styles without pressing a key afterward, so :focus-visible may not engage in Chromium/WebKit and the assertion can read outline-width: 0px.

> Read tests/chat-message-table.test.ts:47-64: the test only did `await wrapper.focus();` before reading getComputedStyle. Confirmed the sibling test tests/markdown-text-table-scroll.test.ts:56-58 follows focus() with `await page.keyboard.press('ArrowRight');` specifically to settle the keyboard-modality heuristic for :focus-visible. Added the identical `await page.keyboard.press('ArrowRight');` call after wrapper.focus() in tests/chat-message-table.test.ts, with a comment pointing at the sibling test as precedent.


### Verification

```
check   exit 0
lint    exit 0
vitest  Test Files  52 passed (52); Tests  1021 passed (1021); Duration 30.02s
```

Tara-ag (approved):
Approved. The change fixes a real accessibility regression in ChatMessage by separating the scroll container from the table, preserving table semantics for assistive technology. The new CSS custom property is additive and backward compatible. Tests and docs are updated appropriately. No blocking issues found.

PR #608, opened 2026-09-12
feat(markdown-text): render raw HTML through a sanitizer the caller supplies

murdore:
`sanitize.rawHtml` gains a third mode. `escape` (default) and `strip` are unchanged; `sanitize` hands the raw HTML to a `htmlSanitizer` function the caller supplies and renders what comes back.

```ts
sanitize: { rawHtml: 'sanitize', htmlSanitizer: (html) => DOMPurify.sanitize(html) }
```

## Why injection rather than a built-in allow-list

I explored a parse5-based allow-list first, because shipping one would be friendlier than making every consumer wire a sanitizer. It is in the repo's dependencies already, so it looked free. It is not:

- The naive version had live XSS on a `javascript:` href.
- A careful version reusing parse5's own `hasSafeProtocol` **still leaked a NUL payload**, because the tokenizer rewrites `NUL` to `U+FFFD` *before* validation sees it.
- `parseFragment` defaults to `<template>` context where a browser uses `<div>`, so the parse tree can diverge from the one the browser will build — the classic mXSS shape.

Writing a sanitizer that is correct against mXSS is a specialist, continuously-maintained job. DOMPurify does it; this library should not pretend to. So the escape hatch is explicit and the caller owns the policy.

## Cache correctness

A caller-supplied function cannot be part of a cache key, so a renderer carrying one bypasses the instance cache entirely rather than risking one caller's sanitizer being reused for another's content:

```ts
const cacheable = typeof sanitizer !== 'function';
let instance = cacheable ? instances.get(key) : null;
```

A sanitizer that throws, or returns a non-string, falls back to `escape` — the safe direction.

## On `strip`, which shipped in 4.20.0

Probed adversarially against the published build with 13 payloads (entity-encoded, double-encoded, template content, svg+style mXSS, math/mglyph, noscript, comment smuggling, textarea break-out, NUL, CDATA, unterminated). None emitted markup. `rawHtmlText` escapes at the leaf, which closes the decode-then-reinject path. One reported failure in that probe was my own over-strict assertion, not the library.

## Verification

```
check   exit 0
lint    exit 0
vitest  95 new assertions in markdown.test.ts, suite green
```

## Summary by CodeRabbit

* **New Features**
  * Added a `sanitize` raw-HTML handling mode for Markdown rendering.
  * Added support for a caller-provided HTML sanitizer that processes the assembled document once per render.
  * Sanitizer failures or invalid results safely fall back to HTML escaping.
  * Existing link protocol checks and external-link protections remain applied.


Tara-ag:
💡 MINOR: Stale Type Reference in docs/MarkdownText.md omits new sanitize mode and htmlSanitizer prop

The Type Reference block at the bottom of `docs/MarkdownText.md` (not modified in this diff) still shows:

```ts
export type MarkdownRawHtmlMode = 'escape' | 'strip';
```

and `MarkdownSanitizeOptions` without the new `htmlSanitizer` property. This contradicts the actual exported types in `src/lib/MarkdownText/properties.ts` and the prose section added above it, so consumers copying from the reference will get type errors or miss the new capability.

Please update the reference to match the implementation:

```ts
export type MarkdownRawHtmlMode = 'escape' | 'strip' | 'sanitize';

export type MarkdownSanitizeOptions = {
  allowedProtocols?: string[];
  allowedTags?: string[];
  disableTaskLists?: boolean;
  rawHtml?: MarkdownRawHtmlMode;
  htmlSanitizer?: (html: string) => string;
};
```

docs/MarkdownText.md
Tara-ag, line 1: 💡 MINOR: The type reference block near the bottom of this file is stale. It still shows `MarkdownRawHtmlMode = 'escape' | 'strip'`, omits the new `htmlSanitizer` property, and describes `rawHtml` as only escaping or dropping raw HTML. Update it to match the exported types in `src/lib/MarkdownText/properties.ts`: add `htmlSanitizer?: (html: string) => string;` to `MarkdownSanitizeOptions` and change `MarkdownRawHtmlMode` to `'escape' | 'strip' | 'sanitize'`.

Tara-ag (approved):
Approved with one documentation nit. The public API change is purely additive (`'sanitize'` mode + optional `htmlSanitizer`), defaults remain safe, and the cache-bypass logic for caller-supplied sanitizers is correctly reasoned. CI is green.

docs/MarkdownText.md
```diff
@@ -47,7 +47,7 @@ The `marked` package is a peer dependency, needed only when you actually use `Ma
 
 ## Security model
 
-- **Raw HTML never passes through.** `<script>`, event-handler attributes, iframes — source HTML is escaped by default, or stripped to escaped text content when `sanitize.rawHtml` is `'strip'`. This is deliberate for chat: models emit markdown, not HTML, and dropping passthrough removes the entire injection surface instead of trying to filter it.
+- **Raw HTML does not pass through unless you supply a sanitizer.** `<script>`, event-handler attributes, iframes — source HTML is escaped by default, or stripped to escaped text content when `sanitize.rawHtml` is `'strip'`. This is deliberate for chat: models emit markdown, not HTML, and dropping passthrough removes the entire injection surface instead of trying to filter it. If you genuinely need raw HTML rendered, see **Rendering raw HTML** below — you supply the sanitizer, and omitting it keeps the escaping.
```
Tara-ag, line 52: 💡 MINOR: The Props table summary (line 37) omits the new `htmlSanitizer` option. It currently says `sanitize` controls whether raw HTML is "escaped or dropped", but the new `'sanitize'` mode and `htmlSanitizer` function are not mentioned. A reader scanning only the table will miss that raw HTML can now be rendered through a caller-supplied sanitizer. Update the `sanitize` row to mention the third mode, e.g. "...whether raw HTML is escaped, stripped, or passed to an optional `htmlSanitizer` function."

murdore:
## Review threads worked through at `fe8ebd5`

### **Fixed**

**`docs/MarkdownText.md`** — Type Reference block at bottom of doc showed stale `MarkdownRawHtmlMode = 'escape' | 'strip'`, omitted `htmlSanitizer` from `MarkdownSanitizeOptions`, and the accompanying comment described rawHtml as only escape/drop.

> grep of docs/MarkdownText.md:214 before the fix showed `export type MarkdownRawHtmlMode = 'escape' | 'strip';` while src/lib/MarkdownText/properties.ts:120 already had `'escape' | 'strip' | 'sanitize'` and declares `htmlSanitizer?: (html: string) => string;` at properties.ts:109 -- confirming the doc's type-reference block (lines 198-217) was out of sync with the real exported types. Edited docs/MarkdownText.md to add the `htmlSanitizer` field and change `MarkdownRawHtmlMode` to `'escape' | 'strip' | 'sanitize'`, matching properties.ts. The rest of the doc (Usage section, 'Rendering raw HTML' section at lines 87-114) already documented htmlSanitizer correctly -- only the bottom-of-file quick-reference block was stale.


### Verification

```
check   exit 0
lint    exit 0
vitest  Test Files 52 passed (52); Tests 1029 passed (1029)
```

src/lib/MarkdownText/markdown.test.ts
```diff
@@ -227,3 +227,98 @@ describe('renderMarkdown — table wrapping', () => {
...
+    // The unsafe path has to be unreachable by OMISSION, not by discipline.
+    // Turning the flag on without supplying a sanitizer must not start emitting
+    // raw HTML -- a consumer who half-configures this gets today's behaviour.
+    const output = renderMarkdown(raw, { sanitize: { rawHtml: 'sanitize' } });
+    expect(output).toContain('&lt;b&gt;bold&lt;/b&gt;');
+    expect(output).not.toContain('<b>bold</b>');
+  });
+
+  it('routes the raw token through a supplied sanitizer and emits its output', () => {
+    const seen: string[] = [];
```
Tara-ag, line 310: 💬 SUGGESTION: Consider adding an adversarial test that passes a permissive sanitizer such as `(html) => html` with a `<script>` payload and asserts the script tag is emitted. This documents the trust boundary explicitly and guards against a future regression where the library might accidentally re-escape the sanitizer's output.

docs/MarkdownText.md
```diff
@@ -84,6 +84,45 @@ The `marked` package is a peer dependency, needed only when you actually use `Ma
   <MarkdownText {markdown} sanitize={{ rawHtml: 'strip' }} />
   ```
 
+### Rendering raw HTML
```
Tara-ag, line 89: 💬 SUGGESTION: The new "Rendering raw HTML" section is clear for Svelte consumers, but it omits how this works for the web-component wrapper. `src/wc/components/MarkdownText.wc.svelte` declares `sanitize: { type: 'Object' }`, so an `htmlSanitizer` function can only be supplied via the element's JS property, not through an HTML attribute. Consider adding a sentence or a small WC example to avoid confusion for vanilla-JS consumers.

Tara-ag (commented):
Reviewed file by file. The change is a careful, opt-in addition to `MarkdownText`: it keeps the default escaping behavior, falls back safely when no sanitizer or a failing sanitizer is supplied, and correctly bypasses the renderer cache for caller-supplied functions. No blocking issues found. Two minor documentation/test suggestions left inline.

docs/MarkdownText.md
```diff
@@ -84,6 +84,51 @@ The `marked` package is a peer dependency, needed only when you actually use `Ma
...
+<script lang="ts">
+  import DOMPurify from 'dompurify';
+  import { MarkdownText } from '@juspay/svelte-ui-components';
+</script>
+
+<MarkdownText
+  {markdown}
+  sanitize={{ rawHtml: 'sanitize', htmlSanitizer: (html) => DOMPurify.sanitize(html) }}
+/>
+```
```
Tara-ag, line 173: 💡 MINOR: Docs example uses bare `DOMPurify.sanitize` without a restrictive config. DOMPurify's default allow-list is broad (varies by version/environment) and may permit SVG/math markup and event-handler attributes. Consumers copying the example may unknowingly widen their attack surface.

Suggestion: show an explicit, minimal config in the example, e.g. `DOMPurify.sanitize(html, { ALLOWED_TAGS: ['b','i','em','strong','a'], ALLOWED_ATTR: ['href'] })`, and add a prominent warning that the caller owns the sanitization policy.

src/lib/MarkdownText/properties.ts
```diff
@@ -89,14 +89,35 @@ export type MarkdownSanitizeOptions = {
...
- * the component has always done); `'strip'` removes it. See
- * `MarkdownSanitizeOptions.rawHtml`.
+ * the component has always done); `'strip'` removes it; `'sanitize'` hands each
+ * raw token to `MarkdownSanitizeOptions.htmlSanitizer` and inserts what comes
+ * back. With no sanitizer supplied, `'sanitize'` behaves exactly like
+ * `'escape'` -- the unsafe path is unreachable by omission rather than by
+ * discipline. See `MarkdownSanitizeOptions.rawHtml`.
  */
-export type MarkdownRawHtmlMode = 'escape' | 'strip';
+export type MarkdownRawHtmlMode = 'escape' | 'strip' | 'sanitize';
```
Tara-ag, line 167: ⚠️ MAJOR: New 'sanitize' mode changes the meaning of `MarkdownRawHtmlMode`. Widening the union from `'escape' | 'strip'` to include `'sanitize'` is type-compatible, but the runtime security model shifts from "raw HTML never passes through" to "raw HTML passes through if a function is supplied". The mode name `sanitize` implies the library is performing sanitization, when it is actually delegating all trust to an arbitrary caller function.

Suggestion: rename the mode to something less reassuring, e.g. `'raw'` or `'inject'`, so the opt-in nature of raw HTML emission is obvious at the call site.

src/lib/MarkdownText/markdown.test.ts
```diff
@@ -227,3 +227,125 @@ describe('renderMarkdown — table wrapping', () => {
...
+    });
+    expect(output).not.toContain('javascript:');
+  });
+});
+
+describe('rawHtml: sanitize — the trust boundary', () => {
+  // The library does not second-guess the sanitizer. A caller who passes a
+  // permissive one gets exactly what it returns, script tag and all. Asserting
+  // that explicitly is the point: it pins the contract, and it would fail if the
+  // library ever started re-escaping the sanitizer's output -- which would look
```
Tara-ag, line 390: 💡 MINOR: Trust-boundary test asserts insecure behavior as a feature. The test verifies that `<script>alert(1)</script>` is emitted when the sanitizer returns it. This pins the contract, but it also encodes that a misconfigured sanitizer (e.g. a forgotten `FORBID_TAGS`) will execute scripts. Pair this with a test showing that the recommended DOMPurify configuration removes `<script>` tags, so the docs and tests together steer callers toward a safe default.

src/lib/MarkdownText/markdown.ts
```diff
@@ -171,11 +174,32 @@ function createSanitizingRenderer(
...
+       * consumer who sets the mode and forgets the function must not silently
+       * start emitting raw HTML. A sanitizer that throws or returns a non-string
+       * lands in the same place -- the fallback is always the provably safe path
+       * that was already there, never the raw token.
+       */
+      if (rawHtml === 'sanitize' && typeof htmlSanitizer === 'function') {
+        try {
+          const sanitized = htmlSanitizer(token.text);
+          if (typeof sanitized === 'string') {
+            return sanitized;
```
Tara-ag: 🔒 CRITICAL: Sanitizer output is inserted unescaped, enabling mXSS and policy bypass. The renderer returns `htmlSanitizer(token.text)` directly, and `MarkdownText.svelte` renders it with `{@html}`. If the sanitizer returns a string that mutates on reparse, or if `marked` splits a raw HTML token across lexer boundaries (e.g. `<textarea>`, `<style>`), the output can break out of its intended context. The PR correctly avoids implementing a sanitizer, but trusting the caller's output verbatim without post-processing reintroduces the mXSS surface it aims to avoid.

Suggestion: escape the sanitizer's return value before insertion, or clearly document that the function must return already-escaped text and that the component will escape it. If the goal is to allow raw markup, consider a more explicit mode name and a second-pass guard.

Tara-ag (changes requested):
Yama review: one CRITICAL and two MAJOR findings require changes before merge; two MINOR findings noted. See inline comments.

Tara-ag:
⚠️ MAJOR: Caller-supplied sanitizer may access DOM APIs during SSR. `htmlSanitizer` is invoked synchronously inside `renderMarkdown`, which runs in a `$derived` in `src/lib/MarkdownText/MarkdownText.svelte` (line 10). A consumer passing `DOMPurify.sanitize` will throw on the server (no DOM); the catch block falls back to escaping, so the server emits escaped text while the client may emit raw HTML, causing a hydration mismatch. The doc's claim that the transform is a "pure string operation" no longer holds for this opt-in path.

Suggestion: document that `htmlSanitizer` must be isomorphic, or provide separate server/client hooks, or skip sanitization during SSR and always escape to guarantee consistent server/client output.

murdore:
## The CRITICAL is correct. I reproduced it, and this PR should not merge as designed.

I set out to check the token-splitting half of the claim and it is real. Instrumenting `htmlSanitizer` to record what it actually receives:

```json
"<div>\n\nhello\n\n</div>"  →  ["<div>", "</div>"]     // two separate calls
"a <b>bold</b> c"           →  ["<b>",   "</b>"]       // two separate calls
"<textarea><p>x</p></textarea>"  →  ["<textarea><p>x</p></textarea>"]   // whole
```

**The sanitizer is handed unbalanced fragments, not documents.** That is inherent to markdown — a blank line ends an HTML block, and inline tags are separate inline tokens — not something the implementation chose.

Every real sanitizer is DOM-based, and parsing a fragment balances it. So feeding fragments through one produces this:

| source | rendered with a balancing sanitizer |
| --- | --- |
| `<div>\n\nhello\n\n</div>` | `<div></div><p>hello</p>` |
| `a <b>bold</b> c` | `<p>a <b></b>bold c</p>` |

The first is the serious one: **the closing tag is gone and the content now renders outside its container.** A consumer relying on that wrapper for styling or isolation has silently lost it. The second drops the formatting entirely.

So this is not a theoretical mXSS concern — the feature produces structurally wrong output for ordinary input, today, with a correctly configured sanitizer.

## On the suggested fix

> escape the sanitizer's return value before insertion

That cannot be right: escaping the output makes `sanitize` mode behave exactly like `escape` mode, which is the default and already exists. The thread half-acknowledges this in its closing line. I am not taking that part.

## Where this leaves it

The defect is in the approach, not in a line of code: **per-token sanitization of unbalanced fragments cannot be made correct by a caller's sanitizer, however well configured.** Real options:

1. **Close this PR.** `rawHtml: 'strip'` already ships and is sound (probed with 13 mXSS payloads); consumers needing real raw HTML can pre-sanitize and pass the result as markdown.
2. **Redesign** so sanitization happens once over assembled output rather than per token. That changes what the sanitizer sees — markdown-generated HTML too — so it is a different feature with a different contract.
3. **Restrict the mode** to tokens that parse as balanced and escape the rest. Narrow, surprising, and hard to explain.

My recommendation is 1. The trust-boundary tests I added earlier this sweep now document a contract I no longer think should ship.

I have not force-pushed anything over this; the branch is unchanged pending a decision.

The other three threads on this PR are fair and are subsumed by that decision — a DOMPurify config example, a mode name that implies the library sanitizes when it does not, and pairing the permissive-sanitizer test with a restrictive one. All of them are about a feature I am now recommending against.

Tara-ag (approved):
Approved. The change is additive: it extends `MarkdownRawHtmlMode` with a new opt-in `'sanitize'` value and a caller-supplied `htmlSanitizer`, while preserving the existing `'escape'` and `'strip'` behavior. Fallbacks to escaping when the sanitizer is missing, throws, or returns a non-string are correctly implemented, and renderer instances carrying a sanitizer bypass the cache to avoid cross-caller reuse. Tests and docs are updated accordingly. No public API break, no export changes, no SSR/DOM issues, and no secrets or unsafe defaults introduced.

Tara-ag (changes requested):
Yama review: the `rawHtml: 'sanitize'` design has a structural correctness defect that the author has reproduced. Per-token sanitization of unbalanced HTML fragments cannot be made correct by a caller-supplied sanitizer, and the unescaped insertion path reintroduces the mXSS surface the PR aimed to avoid. See inline comments for details.

src/lib/MarkdownText/markdown.ts
Tara-ag: 🔒 CRITICAL: Sanitizer output is inserted unescaped, enabling mXSS and policy bypass. The renderer returns `htmlSanitizer(token.text)` directly, and `MarkdownText.svelte` renders it with `{@html}`. Because `marked` lexes raw HTML into separate, often unbalanced tokens (e.g. `<div>\n\nhello\n\n</div>` becomes two calls), a DOM-based sanitizer re-balances each fragment independently, producing structurally wrong output: the wrapper is lost and content escapes its intended container. The PR correctly avoids implementing a sanitizer, but trusting arbitrary caller output verbatim reintroduces the mutation-XSS surface it aims to avoid. The author has reproduced this and agrees the PR should not merge as designed. Recommendation: close or redesign so sanitization operates on assembled output, not per-token fragments.

src/lib/MarkdownText/properties.ts
Tara-ag, line 167: ⚠️ MAJOR: The new `'sanitize'` mode name is misleading. The library does not perform sanitization; it delegates trust to an arbitrary caller function and inserts the result verbatim. A name like `sanitize` implies a safety guarantee the component cannot enforce, especially given the per-token fragment problem. Consider renaming to `'raw'`, `'inject'`, or similar, and documenting that the caller owns the policy and the structural correctness risks.

docs/MarkdownText.md
Tara-ag, line 173: 💡 MINOR: The DOMPurify example uses the default config, which is broad and may permit SVG/math markup and event-handler attributes depending on version/environment. Consumers copying this may unknowingly widen their attack surface. Show an explicit, minimal allow-list (e.g. `ALLOWED_TAGS: ['b','i','em','strong','a']`, `ALLOWED_ATTR: ['href']`) and add a prominent warning that the caller owns the sanitization policy.

src/lib/MarkdownText/markdown.test.ts
Tara-ag, line 390: 💡 MINOR: The trust-boundary test asserts insecure behavior as a feature. While pinning the contract is valuable, pairing it with a test showing a restrictive DOMPurify-style sanitizer removing `<script>` would steer callers toward a safe default and balance the documentation.

Tara-ag:

Severity counts — 🔒 CRITICAL: 1 · ⚠️ MAJOR: 0 · 💡 MINOR: 0

## 🤖 Yama Review Summary

⚠️ The review loop ended early (`time-limit`, 25 steps) — this verdict is built strictly from gate-verified findings.

**Verified findings (1):**

- **CRITICAL**: Public `unwrapFence` prop removed without major-version bump or migration note — `src/lib/MarkdownText/properties.ts:40`

- 🔒 **CRITICAL**: Public `unwrapFence` prop removed without major-version bump or migration note — `src/lib/MarkdownText/properties.ts:40`
  The PR silently removes the documented public `unwrapFence` prop from `MarkdownTextProperties` and `RenderMarkdownOptions` in `src/lib/MarkdownText/properties.ts`, from the destructured props in `src/lib/MarkdownText/MarkdownText.svelte`, and from the web-component prop registry in `src/wc/components/MarkdownText.wc.svelte`. The matching `src/lib/MarkdownText/fenceUnwrap.test.ts` is also gone from…

murdore:
## Correcting myself: the workaround I offered does not exist

In recommending this PR be closed I wrote that *"consumers needing real raw HTML can pre-sanitize and pass the result as markdown."* That is wrong, and I should have checked it before saying it. Measured:

```
input:  '<b>bold</b> and <i>italic</i>'   (already sanitized by the caller)

escape (default)  →  <p>&lt;b&gt;bold&lt;/b&gt; and &lt;i&gt;italic&lt;/i&gt;</p>
strip             →  <p>bold and italic</p>
```

Escaping shows the tags as visible source; stripping removes the formatting. **Neither renders it.** Pre-sanitizing upstream changes nothing, because the escaping happens downstream of whatever the caller did.

So closing this PR does not leave consumers with an inconvenient path — it leaves them with **no path**, which is a materially different decision from the one I described. A consumer whose content legitimately contains `<b>` (the case this PR was opened for) would see either markup or plain text, and could not get bold.

The CRITICAL finding stands unchanged — per-token sanitization of unbalanced fragments is genuinely broken, and I reproduced it. What changes is the conclusion that follows from it. The options are now:

1. **Redesign** — sanitize once over the assembled output instead of per raw-HTML token. This is the architecture people actually use with DOMPurify, and it removes the fragmentation entirely. The contract changes: the sanitizer would also see markdown-generated HTML, so a restrictive config could strip legitimate output. That is predictable and documentable, unlike the current behaviour.
2. **Close and accept the gap** — the library simply does not render raw HTML, in any mode. Defensible, but it should be stated as a deliberate limitation rather than dressed up with a workaround that does not work.

I no longer recommend 2 by default. Which it should be is a product call about whether rendering caller-supplied HTML is in scope for this library at all.

src/lib/MarkdownText/markdown.ts
murdore: Half of this is confirmed and fixed; the suggested remedy would break the feature. Taking them separately.

**"`marked` splits a raw HTML token across lexer boundaries ... the output can break out of its intended context" — confirmed, and worse than a lexer edge case.** It is what happens to every block-level container. For

```
<div class="callout">

**important**

</div>
```

marked emits `<div class="callout">` and `</div>` as two separate html tokens with the markdown between them as its own block. The per-token design called the sanitizer twice, and a sanitizer parses its input and serialises the tree back — so the first call returned `<div class="callout"></div>` and the second returned nothing. Reproduced with parse5 (what DOMPurify's parse-and-reserialise amounts to):

```
<div class="callout"></div><p><strong>important</strong></p>
```

The author's content rendered outside the container meant to hold it. Fixed by sanitizing **once over the assembled document** — the only string in which the container and its contents are both present and balanced. Two tests pin it: one asserts the nesting, one asserts the sanitizer receives a single document rather than `['<div class="callout">', '</div>']`. Both fail against the old renderer.

**"Escape the sanitizer's return value before insertion" — this one I have to push back on.** Escaping the output is precisely `rawHtml: 'escape'`; the sanitized markup would render as visible `&lt;b&gt;` source text and the mode would do nothing. There is already a test pinning that the library must never re-escape, for exactly this reason — it would read as a safety improvement while silently breaking every correct caller. The alternative you offer, "document that the function must return already-escaped text", has the same effect.

**On mXSS specifically.** The library does not reparse the sanitizer's output before insertion — it is inserted once via `{@html}`. Reparse-mutation resistance is the sanitizer's job, and is a large part of why this PR refuses to ship one: a home-grown sanitizer here would own that problem badly. That reasoning, including the two `parse5` prototypes that leaked, is in the commit message.

What I have added instead of re-escaping is a docs change, since the real risk is a caller configuring their sanitizer too loosely: the example now uses an explicit `ALLOWED_TAGS`/`ALLOWED_ATTR` config rather than bare `DOMPurify.sanitize(html)`, with a callout stating that whatever the function returns is rendered unmodified and that the caller owns the policy.

docs/MarkdownText.md
murdore: Taken — the example now configures DOMPurify explicitly instead of calling it bare.

The redesign made this more important than when you raised it, not less. Sanitization now runs once over the assembled document, so the allow-list has to cover the tags **MarkdownText itself emits** as well as the raw-HTML tags a caller chooses to allow — anything missing is dropped from their own markdown output too. The example spells that list out (`p`, `strong`, `code`, `table`, `li`, ...) with a comment saying exactly that, and a blockquote stating that the caller owns the policy, that the component does not re-check what comes back and cannot, and that the list is a starting point to narrow rather than a default to copy.

I did not use the shorter `ALLOWED_TAGS: ['b','i','em','strong','a']` from your suggestion, because under the one-pass design that config would strip every paragraph, list and table the component generates — it would look like the component was broken.

src/lib/MarkdownText/properties.ts
murdore: Thought about this one properly rather than defending the name, and I am keeping `'sanitize'` — but your framing of the security model is right and I want to separate the two claims.

**"The runtime security model shifts" — yes, and that is the PR.** No argument. It is opt-in, and the PR title and docs say so.

**"The mode name implies the library is performing sanitization" — I think the union reads the other way.** `MarkdownRawHtmlMode` describes what happens *to the raw HTML*, not who does it: it is escaped, it is stripped, it is sanitized. `'raw'` or `'inject'` would break that pattern by naming the mechanism instead of the outcome, and `rawHtml: 'raw'` reads as a tautology.

The substantive half of your concern is a reader who sets `rawHtml: 'sanitize'`, assumes the library handles it, and ships raw HTML unsanitized. That outcome is unreachable: the mode alone does nothing. Internally it resolves to a `passthrough` strategy only when `htmlSanitizer` is actually a function, and otherwise resolves to `escape` — so a caller who believes the library sanitizes gets today's escaping, not an open surface. There is a test for exactly that, and the fallback is the same when the sanitizer throws or returns a non-string.

Since you raised it, I have leaned harder on the thing that does carry the warning — the docs, where a caller actually reads about it. The example now configures DOMPurify explicitly rather than calling it bare, with a callout saying the caller owns the policy and that whatever the function returns is rendered unmodified.

If you still want the rename after that, say so and I will do it — the feature is unreleased, so this is the cheap moment to change it and I would rather change it now than argue it.

src/lib/MarkdownText/markdown.test.ts
murdore: Taken — the paired test is added right below it.

`a sanitizer that drops script tags removes them, and keeps the prose around them` renders `before <script>alert(1)</script> after` through a sanitizer that strips script elements, and asserts the output contains neither `<script` nor loses the surrounding prose. It stands in for `DOMPurify.sanitize(html, { ALLOWED_TAGS })` without taking the dependency, so the two tests now sit together and show both sides of the boundary: what a permissive sanitizer gets you, and what a configured one gets you.

The docs moved the same way — the example no longer calls `DOMPurify.sanitize(html)` bare.

Keeping the original assertion as-is, for the reason in its comment: without it, a future change that started re-escaping the sanitizer's output would look like a safety improvement and would silently break every correct caller.

src/lib/MarkdownText/markdown.ts
```diff
@@ -348,19 +385,65 @@ function instanceFor(options: RenderMarkdownOptions): Marked {
...
+ * A consumer's sanitizer is arbitrary code. `null` here means it did not return
+ * usable markup, and the caller falls back to the escaping path that was always
+ * there -- never to the unsanitized assembly.
+ */
+function runSanitizer(html: string, htmlSanitizer: (html: string) => string): string | null {
+  try {
+    const sanitized = htmlSanitizer(html);
+    return typeof sanitized === 'string' ? sanitized : null;
+  } catch {
+    return null;
```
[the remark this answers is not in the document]
murdore: Verified and real. `MarkdownText.svelte:16` derives the string and line 26 renders it with `{@html html}`, so a browser-only sanitizer throws during SSR, falls back to escaped output there, then succeeds on hydration — and the two renders disagree. Good catch.

**On your second option — "keep the initial client output escaped and sanitize only after hydration" — I cannot take it.** That needs a post-hydration reactive pass, and this repo bans `$effect` outright (`eslint.config.js:64`, `no-restricted-syntax`). There is no other hook that fires after hydration for a derived string. So the remaining options are "require an isomorphic sanitizer" or "silently ship the mismatch", and I have taken the first.

Done in this push:

- `MarkdownSanitizeOptions.htmlSanitizer` now states the requirement where a caller reads the signature: it must work on the server as well as the browser, naming DOMPurify-without-a-DOM as the failure case, with the alternative of leaving `rawHtml` at its default on server-rendered routes.
- `docs/MarkdownText.md` gains the same as a fourth bullet in the properties list.

**On the regression test, I did something slightly different from what you suggested** and I think it is the stronger assertion. A test for "a sanitizer that throws when `window` is unavailable" only re-tests the existing throw path. What actually matters for SSR is that the server half lands on a *known* rendering rather than something the fallback invented, so the new test asserts the throwing sanitizer produces output **byte-identical** to `rawHtml: 'escape'`:

```
expect(thrown).toBe(renderMarkdown(source, { sanitize: { rawHtml: 'escape' } }));
```

over a source containing both inline and block-level raw HTML. That pins the mismatch to exactly one shape — escaped on the server, sanitized on the client — instead of leaving it open-ended, and it would fail if the fallback ever diverged from the escape path.

Gate: prettier clean, eslint 0, event-casing 0, svelte-check 0 errors, markdown suite 52 passing.

src/lib/MarkdownText/markdown.ts
```diff
@@ -348,19 +385,65 @@ function instanceFor(options: RenderMarkdownOptions): Marked {
...
+  }
+}
+
 /**
  * Markdown → HTML with the sanitizing pipeline above. Pure string transform —
  * no DOM involved — so it renders identically on server and client.
  */
 export function renderMarkdown(markdown: string, options: RenderMarkdownOptions = {}): string {
```
Tara-ag, line 485: ⚠️ MAJOR: Sanitizer output bypasses link-protocol and tag-allow-list checks for raw-HTML links/images. After the sanitizer runs, `annotateExternalLinks` is applied, but `hasSafeProtocol` and `tagAllowed` checks are not re-applied to raw-HTML links/images. A caller-supplied sanitizer that preserves an `<a href="javascript:alert(1)">` or `` will have that markup emitted verbatim. The PR description states 'Existing link protocol checks and external-link protections remain applied' and 'Link and image protocol checks still apply to markdown syntax', but this is misleading for raw HTML: the sanitizer can absolutely resurrect `javascript:` links if it chooses to. Clarify in docs/MarkdownText.md and in the `htmlSanitizer` JSDoc that the sanitizer is solely responsible for all policy decisions on raw HTML, including URL protocols and tag/attribute allow-lists. Remove or rephrase the claim that 'existing link protocol checks remain applied' to specify they apply only to markdown-syntax links.

src/lib/MarkdownText/markdown.test.ts
```diff
@@ -1,5 +1,6 @@
 import { describe, it, expect } from 'vitest';
 import { renderMarkdown } from './markdown';
```
Tara-ag, line 2: 💡 MINOR: Test imports `parse5` directly but library already depends on it. The new tests import `parseFragment` and `serialize` from `parse5`. This is fine because `parse5` is a runtime dependency, but the test file now has a direct dependency on an implementation detail. If the library later replaces `parse5` with a lighter parser, these tests would need updating. Add a comment above the parse5 import: '// Used as a faithful stand-in for DOMPurify parse-and-serialize behavior in tests.'

docs/MarkdownText.md
```diff
@@ -85,6 +85,132 @@ The `marked` package is a peer dependency, needed only when you actually use `Ma
...
+Configure it explicitly. `DOMPurify.sanitize(html)` with no options uses a broad default
+allow-list that varies by version and environment, and because your function now sees the whole
+document, the allow-list has to cover the tags this component emits as well as the ones you let
+authors write:
+
+```svelte
+<script lang="ts">
+  import DOMPurify from 'dompurify';
+  import { MarkdownText } from '@juspay/svelte-ui-components';
+
```
Tara-ag, line 112: 💡 MINOR: Docs example uses `import DOMPurify from 'dompurify'` without noting it must be installed separately. DOMPurify is not in the library's dependencies or peerDependencies. Consumers copying the example will get a module-not-found error. Add a note immediately before the code snippet: 'Install DOMPurify (or another sanitizer of your choice) in your own app; it is not bundled with this library.'

src/lib/MarkdownText/markdown.ts
```diff
@@ -322,12 +352,19 @@ function wrapTables(html: string, label?: string, wrapperClass?: string): string
...
 }
 
-function instanceFor(options: RenderMarkdownOptions): Marked {
+function instanceFor(options: RenderMarkdownOptions, rawHtml: RawHtmlStrategy): Marked {
   const breaks = options.breaks === true;
   const protocols = resolveProtocolSets(options.sanitize);
   const allowedTags = resolveTagAllowList(options.sanitize?.allowedTags);
   const disableTaskLists = options.sanitize?.disableTaskLists === true;
-  const rawHtml = resolveRawHtmlMode(options.sanitize?.rawHtml);
+  /*
```
Tara-ag, line 360: ⚠️ MAJOR: Module-scoped `instances` cache can retain stale `Marked` instances across hot reloads/tests. The `instances` Map is declared at module scope and keyed by configuration strings. The PR adds `rawHtml` to the cache key. Because `rawHtml` can now be `'passthrough'` (internal) based on the presence of a function, two callers with identical other options but different sanitizers will share a `'passthrough'` instance. The module-scoped cache is never cleared, so in long-lived dev-server or test environments with many distinct `allowedProtocols`/`allowedTags` combinations, the Map grows unbounded. Consider capping the cache size (LRU) or documenting the unbounded-cache trade-off in the code comments and docs. At minimum, add a comment that callers should not vary `allowedTags`/`allowedProtocols` on every render.

Tara-ag:
🔒 CRITICAL (file not in diff): New `htmlSanitizer` prop is not exposed on the web-component wrapper. The PR adds `htmlSanitizer` as a function-valued option under `sanitize` for Svelte consumers, but `src/wc/components/MarkdownText.wc.svelte` only declares `sanitize: { type: 'Object' }`. The new public API surface is reachable from the Svelte component and `renderMarkdown`, but the web-component registry does not surface it in any typed/declared way. For a published library that ships both Svelte and WC builds, adding a public option that is effectively invisible/unusable to WC consumers without documentation is an API inconsistency. Add a runtime guard in the wrapper that logs a one-time console warning when `sanitize.rawHtml === 'sanitize'` and no `htmlSanitizer` function is provided, and document the limitation in docs/MarkdownText.md. Alternatively, expose `htmlSanitizer` as a separate object/function prop if the custom-element generator supports it, or add a prominent 'Web component consumers' caveat in the docs.

Tara-ag:
⚠️ MAJOR (file not in diff): `$derived` calls `renderMarkdown` with a function prop, causing repeated re-renders on every prop object creation. In `src/lib/MarkdownText/MarkdownText.svelte`, `html` is `$derived(renderMarkdown(markdown, { breaks, tableLabel, tableWrapperClass, sanitize }))`. When a caller passes an inline object literal like `sanitize={{ rawHtml: 'sanitize', htmlSanitizer: (html) => ... }}`, Svelte creates a new object reference on every parent render. Because `renderMarkdown` now invokes `options.sanitize?.htmlSanitizer` during execution, every parent update triggers a full markdown re-parse and a call to the (possibly expensive) sanitizer, even when `markdown` has not changed. Stabilize the options object passed to `renderMarkdown` so the derived only re-runs when meaningful values change. For example, destructure `sanitize` into primitive comparisons, or memoize the options object with a shallow-equality check inside the component.

Tara-ag (changes requested):
Blocking review: the new `htmlSanitizer` API is not consistently exposed across Svelte and web-component builds, and the reactivity/performance implications of passing a function prop into `$derived` need addressing. See inline comments for details.

Tara-ag (commented):
Yama review in progress — posting gate-accepted findings as inline comments.

src/lib/MarkdownText/markdown.ts
Tara-ag, line 485: ⚠️ MAJOR: Sanitizer output bypasses link-protocol and tag-allow-list checks for raw-HTML links/images. After the sanitizer runs, `annotateExternalLinks` is applied, but `hasSafeProtocol` and `tagAllowed` checks are not re-applied to raw-HTML links/images. A caller-supplied sanitizer that preserves an `<a href="javascript:alert(1)">` or `` will have that markup emitted verbatim. The PR description states 'Existing link protocol checks and external-link protections remain applied' and 'Link and image protocol checks still apply to markdown syntax', but this is misleading for raw HTML: the sanitizer can absolutely resurrect `javascript:` links if it chooses to.

Clarify in docs/MarkdownText.md and in the `htmlSanitizer` JSDoc that the sanitizer is solely responsible for all policy decisions on raw HTML, including URL protocols and tag/attribute allow-lists. Remove or rephrase the claim that 'existing link protocol checks remain applied' to specify they apply only to markdown-syntax links.

src/lib/MarkdownText/markdown.ts
Tara-ag, line 360: ⚠️ MAJOR: Module-scoped `instances` cache can retain stale `Marked` instances across hot reloads/tests. The `instances` Map is declared at module scope and keyed by configuration strings. The PR adds `rawHtml` to the cache key. Because `rawHtml` can now be `'passthrough'` (internal) based on the presence of a function, two callers with identical other options but different sanitizers will share a `'passthrough'` instance. The module-scoped cache is never cleared, so in long-lived dev-server or test environments with many distinct `allowedProtocols`/`allowedTags` combinations, the Map grows unbounded.

Consider capping the cache size (LRU) or documenting the unbounded-cache trade-off in the code comments and docs. At minimum, add a comment that callers should not vary `allowedTags`/`allowedProtocols` on every render.

src/lib/MarkdownText/markdown.test.ts
Tara-ag, line 2: 💡 MINOR: Test imports `parse5` directly but library already depends on it. The new tests import `parseFragment` and `serialize` from `parse5`. This is fine because `parse5` is a runtime dependency, but the test file now has a direct dependency on an implementation detail. If the library later replaces `parse5` with a lighter parser, these tests would need updating.

Add a comment above the parse5 import: '// Used as a faithful stand-in for DOMPurify parse-and-serialize behavior in tests.'

docs/MarkdownText.md
Tara-ag, line 112: 💡 MINOR: Docs example uses `import DOMPurify from 'dompurify'` without noting it must be installed separately. DOMPurify is not in the library's dependencies or peerDependencies. Consumers copying the example will get a module-not-found error.

Add a note immediately before the code snippet: 'Install DOMPurify (or another sanitizer of your choice) in your own app; it is not bundled with this library.'

src/lib/MarkdownText/markdown.ts
murdore: This is fixed, and your description of the mechanism is exactly right — including the detail that a DOM-based sanitizer re-balances each fragment independently, which is what made the output structurally wrong rather than merely unsanitized.

Sanitization now runs **once over the assembled document** instead of per raw-HTML token. That is the only string in which a block-level container and the markdown it wraps are both present and balanced. Reproduced before fixing, with parse5 standing in for DOMPurify's parse-and-reserialise:

```
<div class="callout">\n\n**important**\n\n</div>
  before: <div class="callout"></div><p><strong>important</strong></p>
  after:  <div class="callout"><p><strong>important</strong></p></div>
```

Two tests pin it — one asserts the nesting, one asserts the sanitizer receives a single document rather than `['<div class="callout">', '</div>']`. Both fail against the per-token renderer.

The one part of the suggestion I did not take is escaping the sanitizer's return value: that is precisely `rawHtml: 'escape'`, and it would make the mode render visible `&lt;b&gt;` source text. There is a test pinning that the library must never re-escape, because it would read as a safety improvement while silently breaking every correct caller.

Resolving as fixed — the CRITICAL was real and is addressed.

docs/MarkdownText.md
murdore: Already done in the current push — the example no longer calls `DOMPurify.sanitize(html)` bare.

It now passes an explicit `ALLOWED_TAGS`/`ALLOWED_ATTR`, with a blockquote stating that the caller owns the policy, that the component does not re-check what comes back and cannot, and that the list is a starting point to narrow rather than a default to copy.

I did not use the short list from your suggestion (`['b','i','em','strong','a']`), because the redesign changed what the sanitizer sees: it now receives the whole assembled document, so that config would strip every paragraph, list and table the component itself generates and look like the component was broken. The example spells out the tags MarkdownText emits alongside the raw-HTML ones a caller might allow, and says exactly why.

Also added this push: a note that DOMPurify is not bundled and is not a peer dependency, so a consumer copying the snippet knows to install it.

src/lib/MarkdownText/markdown.test.ts
murdore: Added, and it sits directly below the trust-boundary test so the two read as a pair.

`a sanitizer that drops script elements removes them, and keeps the prose around them` renders `before <script>alert(1)</script> after` through a sanitizer that strips script elements and asserts the output contains neither `<script` nor loses the surrounding prose.

Worth flagging how I built it, because the first version was wrong in an instructive way. I wrote the stand-in as `html.replace(/<script\b[\s\S]*?<\/script>/gi, '')` — and CodeQL flagged it twice at high severity, `js/bad-tag-filter` and `js/incomplete-multi-character-sanitization`, because that pattern does not match `</script >`. A regex tag filter is the exact thing this PR argues a caller should never write, sitting in a test that a reader would take as an example. It is now a parse-and-prune over parse5, and there is a second case using the spaced end tag that the regex version would have let through.

The original permissive assertion stays as-is, for the reason in its comment: without it, a future change that started re-escaping the sanitizer's output would look like a safety improvement and would silently break every correct caller.

src/lib/MarkdownText/markdown.ts
murdore: Confirmed by reproduction — you are right about the behaviour. I measured both halves rather than reasoning about them:

```
markdown  [click](javascript:alert(1))            -> <p>click</p>
raw HTML  <a href="javascript:alert(1)">click</a> -> <p><a href="javascript:alert(1)">click</a></p>
```

with an identity sanitizer. The markdown-syntax guard drops the anchor; the raw-HTML one is emitted verbatim.

**I am not adding the re-check, and the reasoning is the substance of this PR rather than a deflection.** To apply `hasSafeProtocol`/`tagAllowed` to the sanitizer's output, the library has to inspect that output. Two ways, both bad:

A regex over `href="..."` is the failure mode this very PR already tripped over. In the previous push I wrote a regex script-stripper in a *test*, and CodeQL flagged it twice at high severity — `js/bad-tag-filter`, because `/<\/script>/` does not match `</script >`. An attribute regex is the same class of mistake, and it would be in the shipping path rather than a test.

Parsing the output with parse5 and walking it is the correct way — and it is HTML sanitization, performed by this library, on a string a caller just told us they had sanitized. It also costs a full parse-and-reserialise per render, and parse5's serialisation can itself alter markup, which is the mutation risk the PR's commit message documents from two earlier prototypes (including a NUL-smuggled scheme that survived because parse5 rewrites NUL to U+FFFD during parsing, so the protocol check validated a string the author never wrote).

So the honest position is that this is the trust boundary working as designed, not an oversight — but you are right that the docs let a reader infer more protection than exists. The previous wording said what IS guarded and left the reader to infer the rest. Fixed in this push, stated as a limitation rather than a reassurance:

> **The protocol allow-list covers markdown syntax, NOT raw HTML.** `[click](javascript:alert(1))` renders as plain text with no anchor, as it always has. But `<a href="javascript:alert(1)">click</a>` written as raw HTML is emitted exactly as your sanitizer returns it — this component does not re-check it. [...] DOMPurify strips `javascript:` hrefs by default, so do not turn that off.

Same note added to the `htmlSanitizer` JSDoc, where someone reading the signature sees it.

And the behaviour is now pinned by a test asserting **both** halves together — `guards a markdown-syntax javascript: link but not a raw-HTML one` — so the asymmetry is a decision on the record rather than an accident, and it cannot change without a test changing.

Leaving this thread open: if you think the boundary itself is wrong rather than under-documented, that is a design call worth having explicitly, and I would rather you make it than have me close the thread on my own argument.

Tara-ag (changes requested):
Yama review verdict: BLOCKED. See inline comments for gate-accepted findings. Two additional accepted findings concern files not present in this PR's diff (`src/wc/components/MarkdownText.wc.svelte` and `src/lib/MarkdownText/MarkdownText.svelte`) and have been recorded as PR issue comments instead.

src/lib/MarkdownText/markdown.ts
murdore: Checked this against the current code, and the premise no longer holds — it describes the *previous* design.

"Two callers with identical other options but different sanitizers will share a `'passthrough'` instance" is true, and it is correct rather than a bug, because **the cached instance no longer knows about any sanitizer.** That was the whole point of moving sanitization out of the renderer.

Before: the renderer closed over `htmlSanitizer` and called it per token, so a shared instance genuinely would have called the wrong caller's function. That version could not cache at all, and said so in a comment.

Now: the renderer's `html()` returns `token.text` unchanged under the `passthrough` strategy. The sanitizer is applied afterwards, in `renderMarkdown`, read from that call's own `options`. Two callers sharing a passthrough instance share a renderer that emits raw tokens verbatim — then each gets its own sanitizer applied to its own document. The strategy is a string (`'escape' | 'strip' | 'passthrough'`), which is why it can be part of a string cache key at all.

That is also what fixes the leak the old design would otherwise have had: keying on function identity would grow the map without bound, because the idiomatic call site passes a fresh inline arrow on every render. The comment above the cache key says this.

On hot reloads and tests: the map is keyed by configuration, holds no per-call state, and `Marked` instances here are constructed with `gfm`/`breaks`/`renderer` fixed at construction and `parse()` taking no per-request state — so a retained instance across a reload produces identical output to a fresh one. There is no staleness to observe.

If I have misread which scenario you mean — specifically if you can construct two callers who get each other's sanitizer output — send it and I will treat it as a defect. I could not construct one.

src/lib/MarkdownText/markdown.test.ts
murdore: Done — comment added above the import in this push:

```ts
// Used to build stand-in sanitizers that parse and re-serialise, which is what a
// real sanitizer does. Not an assertion about the library's own parser: if the
// implementation ever swaps parse5 out, these stand-ins still model the
// behaviour under test and only this import moves.
```

Your reasoning is right and worth making explicit: the tests use parse5 to *model a caller's sanitizer*, not to assert anything about how this library parses. A real sanitizer parses its input into a tree and serialises it back — that is what produces the fragment-rebalancing this PR fixes — so a stand-in has to do the same to be faithful. Using parse5 avoids taking a DOMPurify dev-dependency purely for test fixtures.

docs/MarkdownText.md
murdore: Added in this push, immediately before the snippet:

> DOMPurify is not bundled with this library and is not a peer dependency — install a sanitizer in your own app (`npm i dompurify`, or anything else you trust). Nothing here imports one.

You are right that it was a real trip hazard as written. It is also the point of the design rather than an oversight, so saying it out loud helps twice: the library deliberately takes no sanitizer dependency, which is why `htmlSanitizer` is injected rather than configured, and a consumer who reads that note understands the contract as well as the install step.

src/lib/MarkdownText/properties.ts
murdore: Answering here since this duplicates the naming thread on `properties.ts`, and folding in the part of it that has since changed.

The fragment problem you cite as supporting the rename is fixed: sanitization runs once over the assembled document, so there is no longer a per-token path that could hand a sanitizer unbalanced markup.

On the name itself — the decision is to keep `'sanitize'`. `MarkdownRawHtmlMode` names what happens to the raw HTML, not who does it: it is escaped, it is stripped, it is sanitized. `'raw'` or `'inject'` would break that pattern by naming the mechanism, and `rawHtml: 'raw'` reads as a tautology.

The substantive half of your concern — a reader who assumes the library sanitizes and ships raw HTML unguarded — is unreachable: the mode alone does nothing. It resolves internally to a `passthrough` strategy only when `htmlSanitizer` is actually a function, and otherwise resolves to `escape`. So the misunderstanding produces today's escaping, not an open surface, and there is a test for exactly that.

Where the warning does belong is the docs, and that is where this review moved things: an explicit `ALLOWED_TAGS`/`ALLOWED_ATTR` example instead of a bare `DOMPurify.sanitize(html)`, a note that DOMPurify is not bundled, and an explicit statement that the protocol allow-list does not cover raw-HTML hrefs.

Resolving this one as the duplicate; the original thread stays open in case you want to press the rename.

Tara-ag:
⚠️ MAJOR (file not in diff): `$derived` calls `renderMarkdown` with a function prop, causing repeated re-renders on every prop object creation. In `src/lib/MarkdownText/MarkdownText.svelte`, `html` is `$derived(renderMarkdown(markdown, { breaks, tableLabel, tableWrapperClass, sanitize }))`. When a caller passes an inline object literal like `sanitize={{ rawHtml: 'sanitize', htmlSanitizer: (html) => ... }}`, Svelte creates a new object reference on every parent render. Because `renderMarkdown` now invokes `options.sanitize?.htmlSanitizer` during execution, every parent update triggers a full markdown re-parse and a call to the (possibly expensive) sanitizer, even when `markdown` has not changed.

Stabilize the options object passed to `renderMarkdown` so the derived only re-runs when meaningful values change. For example, destructure `sanitize` into primitive comparisons, or memoize the options object with a shallow-equality check inside the component.

Tara-ag:
🔒 CRITICAL (file not in diff): New `htmlSanitizer` prop is not exposed on the web-component wrapper. The PR adds `htmlSanitizer` as a function-valued option under `sanitize` for Svelte consumers, but `src/wc/components/MarkdownText.wc.svelte` only declares `sanitize: { type: 'Object' }`. The new public API surface is reachable from the Svelte component and `renderMarkdown`, but the web-component registry does not surface it in any typed/declared way, and generated custom-element typings will not mention it. For a published library that ships both Svelte and WC builds, adding a public option that is effectively invisible/unusable to WC consumers without documentation is an API inconsistency.

Add a runtime guard in the wrapper that logs a one-time console warning when `sanitize.rawHtml === 'sanitize'` and no `htmlSanitizer` function is provided, and document the limitation in docs/MarkdownText.md. Alternatively, expose `htmlSanitizer` as a separate object/function prop if the custom-element generator supports it, or add a prominent 'Web component consumers' caveat in the docs.

src/lib/MarkdownText/markdown.ts
murdore: Duplicate of the protocol-bypass thread on this same line, which I answered in full — re-posted against the new push rather than updated.

Short version: confirmed by reproduction (markdown `[click](javascript:alert(1))` renders anchor-less, raw `<a href="javascript:alert(1)">` is emitted verbatim), and the remedy is documentation plus a test rather than a re-check, because inspecting the sanitizer output means either an attribute regex (the `js/bad-tag-filter` mistake this PR already tripped on) or parse5-walking a string the caller just sanitized — which is the sanitization this option exists to delegate.

This push states the limitation explicitly in both the docs and the `htmlSanitizer` JSDoc, and pins it with `guards a markdown-syntax javascript: link but not a raw-HTML one`.

Resolving as the duplicate; the original stays open for the design call.

src/lib/MarkdownText/markdown.ts
murdore: Duplicate of the cache thread on this same line, answered there — re-posted against the new push.

Short version: the premise describes the previous design. The cached instance no longer holds a sanitizer; under the `passthrough` strategy the renderer returns `token.text` unchanged and the sanitizer is applied afterwards in `renderMarkdown` from that call's own options. So two callers sharing a passthrough instance share a renderer that emits raw tokens verbatim and each still gets their own sanitizer. That is exactly what makes the string cache key sound again.

Resolving as the duplicate; the original stays open in case there is a scenario I could not construct.

src/lib/MarkdownText/markdown.test.ts
murdore: Already done — the comment landed in the push before this review ran:

```ts
// Used to build stand-in sanitizers that parse and re-serialise, which is what a
// real sanitizer does. Not an assertion about the library's own parser: if the
// implementation ever swaps parse5 out, these stand-ins still model the
// behaviour under test and only this import moves.
```

docs/MarkdownText.md
murdore: Already done — the note landed in the push before this review ran, immediately before the snippet:

> DOMPurify is not bundled with this library and is not a peer dependency — install a sanitizer in your own app (`npm i dompurify`, or anything else you trust). Nothing here imports one.

docs/MarkdownText.md
```diff
@@ -85,6 +85,139 @@ The `marked` package is a peer dependency, needed only when you actually use `Ma
...
+DOMPurify is not bundled with this library and is not a peer dependency — install a sanitizer in
+your own app (`npm i dompurify`, or anything else you trust). Nothing here imports one.
+
+Configure it explicitly. `DOMPurify.sanitize(html)` with no options uses a broad default
+allow-list that varies by version and environment, and because your function now sees the whole
+document, the allow-list has to cover the tags this component emits as well as the ones you let
+authors write:
+
+```svelte
+<script lang="ts">
```
Tara-ag, line 109: 💡 MINOR: Docs example shows bare `DOMPurify.sanitize(html)` without a restrictive config. DOMPurify's default allow-list is broad and version/environment dependent. Although a later section lists `ALLOWED_TAGS`/`ALLOWED_ATTR`, the quick-start example is the most copied snippet and should lead with an explicit, restrictive configuration. Keep the broad-default discussion as a warning rather than the primary sample.

src/lib/MarkdownText/markdown.test.ts
```diff
@@ -282,3 +287,249 @@ describe('renderMarkdown — table wrapping', () => {
...
+  });
+
+  it('ignores a sanitizer while the mode is escape or strip', () => {
+    // The sanitizer is only consulted in sanitize mode, so supplying one cannot
+    // silently change what the other two modes do.
+    const escaped = renderMarkdown(raw, {
+      sanitize: { rawHtml: 'escape', htmlSanitizer: (html) => html }
+    });
+    expect(escaped).toContain('&lt;b&gt;bold&lt;/b&gt;');
+
```
Tara-ag, line 360: 💡 MINOR: Trust-boundary test asserts insecure behavior without a paired restrictive test. The test verifies that a permissive sanitizer allows `<script>alert(1)</script>`, which pins the contract, but it should be paired with a test showing that a realistic restrictive allow-list sanitizer removes it. Otherwise readers may mistake the permissive case as a recommended default. Keep the contract test, but add an adjacent restrictive-sanitizer demonstration.

src/lib/MarkdownText/markdown.ts
```diff
@@ -171,11 +185,27 @@ function createSanitizingRenderer(
...
+  rawHtml: RawHtmlStrategy
 ): RendererObject {
   return {
     html(token: Tokens.HTML | Tokens.Tag): string {
-      return rawHtml === 'strip' ? rawHtmlText(parseFragment(token.text)) : escapeHtml(token.text);
+      if (rawHtml === 'strip') {
+        return rawHtmlText(parseFragment(token.text));
+      }
+      /*
+       * Emitted verbatim, and sanitized later in one pass over the whole
```
Tara-ag, line 196: 🔒 CRITICAL: Per-token raw-HTML passthrough breaks real sanitizers on ordinary input. The renderer returns `token.text` verbatim in `'passthrough'` mode, and the sanitizer is then called once over the assembled document. But `marked` lexes raw HTML into separate, often unbalanced tokens: a `<div>…</div>` wrapper becomes two tokens with markdown between them, and inline `<b>…</b>` becomes two separate inline tokens. A real DOM-based sanitizer re-balances each fragment independently, so even a correctly configured sanitizer produces structurally wrong output (the wrapper closes before its content, or formatting is dropped). The author independently reproduced this on the PR and concluded the design is unsound. Redesign so raw-HTML tokens are buffered/concatenated before sanitization, or sanitize the fully assembled document after `marked` finishes but before post-processing.

Tara-ag (changes requested):

The `rawHtml: 'sanitize'` feature cannot land as designed. `marked` lexes raw HTML into separate, often unbalanced tokens, so a caller-supplied DOM-based sanitizer receives fragments rather than the intended document and re-balances them into structurally wrong output on ordinary input. The author independently reproduced this on the PR. Until the raw-HTML handling is redesigned so sanitization operates on balanced, assembled HTML, this change is unsafe to merge.

Additional gate-accepted findings:
- SSR/hydration mismatch risk when a browser-only sanitizer is passed into the synchronous `$derived` render path.
- Repeated full markdown re-parsing when parents pass fresh inline `sanitize` objects or `htmlSanitizer` arrows.
- The new `htmlSanitizer` option is unreachable from the web-component build.
- Docs quick-start example and trust-boundary tests should lead with a restrictive sanitizer config and pair the permissive contract test with a restrictive one.

See inline comments and PR issue comments for details.

src/lib/MarkdownText/markdown.ts
murdore: The mechanism you describe is exactly right and it is exactly what this PR already fixed — but the conclusion does not hold against the current code, and I measured all three of your cases rather than arguing it.

The comment quotes the new design correctly ("returns `token.text` verbatim in `'passthrough'` mode, and the sanitizer is then called once over the assembled document") and then applies the old design's consequence to it. Those are incompatible: a sanitizer called **once, over the assembled document**, never sees a fragment. There is nothing unbalanced left for it to re-balance, because the document it receives has the container and its contents in one string.

Ran a faithful DOM-based sanitizer — `serialize(parseFragment(html))`, which is what DOMPurify's parse-and-reserialise amounts to — over the three shapes you name:

```
inline  text with <b>bold</b> inline
     -> <p>text with <b>bold</b> inline</p>

block   <div class="c">\n\n**md**\n\n</div>
     -> <div class="c"><p><strong>md</strong></p></div>

nested  <section><div class="c">\n\n**md**\n\n</div></section>
     -> <section><div class="c"><p><strong>md</strong></p></div></section>
```

No wrapper closing before its content, no formatting dropped, nesting preserved through two levels.

For contrast, the same balancing sanitizer against the per-token renderer this PR replaced:

```
block   -> <div class="c"></div><p><strong>md</strong></p>
```

That is the defect, and it is pinned by two regression tests that fail against the old renderer: one asserts the nesting, one asserts the sanitizer is handed a single document rather than `['<div class="c">', '</div>']`.

Resolving — this is a re-review of the current head reaching the previous head's conclusion.

src/lib/MarkdownText/markdown.test.ts
murdore: Already done — the paired restrictive test sits directly below the contract test, added in the previous push.

`a sanitizer that drops script elements removes them, and keeps the prose around them` renders `before <script>alert(1)</script> after` through a sanitizer that strips script elements, asserting the output contains neither `<script` nor loses the surrounding prose. A second case uses a spaced end tag (`</script >`).

That second case exists because my first attempt at this test was a regex stripper, and CodeQL flagged it at high severity — `js/bad-tag-filter`, since `/<\/script>/` does not match `</script >`. It is now a parse-and-prune over parse5, which is both correct and a better example, since a regex tag filter is precisely what this PR argues a caller should never write.

docs/MarkdownText.md
murdore: The quick-start example already leads with the explicit configuration — there is no bare `DOMPurify.sanitize(html)` left in the docs to lead with.

```
$ grep -n "DOMPurify.sanitize" docs/MarkdownText.md
```

returns only the configured call, `DOMPurify.sanitize(html, { ALLOWED_TAGS, ALLOWED_ATTR })`, with both lists defined immediately above it in the same snippet and a comment explaining that the allow-list must cover the tags MarkdownText itself emits — not only author-written ones — because the sanitizer now sees the whole assembled document.

Immediately after it is the warning you are asking for: that the caller owns the policy, that the component does not re-check what comes back and cannot, and that the list is a starting point to narrow rather than a default to copy. Plus a note that DOMPurify is not bundled and is not a peer dependency.

So this is the shape you describe — restrictive config as the primary sample, broad-default risk as the warning. It landed in the push before this review ran.

Tara-ag (commented):

src/lib/MarkdownText/properties.ts
Tara-ag, line 1: 🔒 CRITICAL: Public `unwrapFence` prop removed without major-version bump or migration note. The merge base (release @ 6c6dbece) exposes `unwrapFence?: boolean` in both `OptionalMarkdownTextProperties` and `RenderMarkdownOptions`, and it is documented in `docs/MarkdownText.md` and registered in `src/wc/components/MarkdownText.wc.svelte`. This PR deletes it from all three places and drops `fenceUnwrap.test.ts`. Because this is a published library, removing a documented public prop is a breaking change. Either restore `unwrapFence` and wire it through `MarkdownText.svelte` and the WC prop registry, or mark the commit/PR as breaking (`feat(markdown-text)!:` or `BREAKING CHANGE:` footer) and add a migration note.

src/lib/MarkdownText/properties.ts
```diff
@@ -98,14 +98,58 @@ export type MarkdownSanitizeOptions = {
    * keep their existing rendering. Neither mode inserts source HTML into the DOM.
```
Tara-ag, line 113: 🔒 CRITICAL: Public `unwrapFence` prop removed without major-version bump or migration note

The PR silently removes the documented public `unwrapFence` prop from `MarkdownTextProperties` and `RenderMarkdownOptions`, from the destructured props in `src/lib/MarkdownText/MarkdownText.svelte`, and from the web-component prop registry in `src/wc/components/MarkdownText.wc.svelte`. The matching `src/lib/MarkdownText/fenceUnwrap.test.ts` is also gone from the branch. This prop was present on the merge base and documented in `docs/MarkdownText.md`, so downstream consumers using it will break. The commit is titled `feat(markdown-text): ...` with no `!` marker and no `BREAKING CHANGE:` footer, and the PR description does not mention the removal or provide a migration path.

Restore `unwrapFence?: boolean` in `OptionalMarkdownTextProperties` and `RenderMarkdownOptions`, pass it through `MarkdownText.svelte`, expose it in `src/wc/components/MarkdownText.wc.svelte`, and restore `src/lib/MarkdownText/fenceUnwrap.test.ts`. If the removal is intentional, change the commit/PR to a breaking change (`feat(markdown-text)!: ...` or `BREAKING CHANGE:` footer) and add a documented migration note.

Tara-ag (changes requested):

This PR introduces a breaking change to the public API by removing the documented `unwrapFence` prop from `MarkdownTextProperties`, `RenderMarkdownOptions`, `MarkdownText.svelte`, and the web-component prop registry, and by deleting `src/lib/MarkdownText/fenceUnwrap.test.ts`. Because this is a published library and the change is not marked as a major-version bump nor accompanied by a migration note, it violates the backward-compatibility contract and the blocking criteria.

Please see the inline comment for the gate-accepted CRITICAL finding.

Tara-ag (approved):
Reviewed file by file. The change is an additive, opt-in feature: `rawHtml: 'sanitize'` plus a caller-supplied `htmlSanitizer` enables raw-HTML rendering, while omitting the sanitizer keeps the existing escape behavior. No public types, exports, or CSS custom properties are removed or narrowed; the new `'sanitize'` mode and `htmlSanitizer` prop are purely additive. The implementation correctly keys the `Marked` instance cache on the strategy string (not the function), sanitizes the assembled document in one pass, and falls back to escape on throw/non-string. Tests cover default escape, half-configuration, sanitizer routing, fallback, inline mode, and the markdown-vs-raw-HTML trust boundary. Docs and the web-component wrapper are updated. No blocking issues found.

src/lib/MarkdownText/properties.ts
murdore: You were right when this was written, and the branch has since been fixed — this thread is now stale. Posting the measurement rather than just asserting it.

**Why it was true.** #611 (`unwrapFence`) merged into `release` at 03:03. I had rebased this branch minutes *before* that, so its base was stale — and because both PRs edit `properties.ts`, `markdown.ts` and `docs/MarkdownText.md`, a rebase-merge would genuinely have taken `unwrapFence` back out. That was a real regression and this finding caught it.

**Why it is stale.** This comment was posted at `03:26:09Z`. The current head, `105ee0f`, was committed at `03:48:41Z` — 22 minutes later — after I re-rebased onto current `release` and resolved the `markdown.ts` conflict so the two features compose.

The decisive check is that this PR does not touch `unwrapFence` **at all** relative to its base:

```
$ git diff origin/release...HEAD -- src/lib/MarkdownText/properties.ts \
                                    src/wc/components/MarkdownText.wc.svelte | grep unwrapFence
(no output)
```

Not one added or removed line. It is present in all four places named above, exactly as `release` has it:

```
$ grep -c unwrapFence src/lib/MarkdownText/properties.ts        3
$ grep -c unwrapFence src/lib/MarkdownText/MarkdownText.svelte  2
$ grep -c unwrapFence src/wc/components/MarkdownText.wc.svelte  1
$ grep -c unwrapFence docs/MarkdownText.md                      5
```

`fenceUnwrap.test.ts` is back on the branch and green:

```
$ npx vitest run src/lib/MarkdownText/fenceUnwrap.test.ts
 Test Files  1 passed (1)
      Tests  18 passed (18)
```

Those 18 restored tests are the whole of the branch's 1114 → 1132 jump. So no breaking change, and no `!` / `BREAKING CHANGE:` footer is warranted — `unwrapFence` survives this PR untouched, and the two features compose:

```
unwrapFence + sanitize -> <div class="c"><p><strong>md</strong></p></div>
sanitize alone         -> <pre><code class="language-markdown">...   (fence preserved)
```

src/lib/MarkdownText/properties.ts
murdore: You were right when this was written, and the branch has since been fixed — this thread is now stale. Posting the measurement rather than just asserting it.

**Why it was true.** #611 (`unwrapFence`) merged into `release` at 03:03. I had rebased this branch minutes *before* that, so its base was stale — and because both PRs edit `properties.ts`, `markdown.ts` and `docs/MarkdownText.md`, a rebase-merge would genuinely have taken `unwrapFence` back out. That was a real regression and this finding caught it.

**Why it is stale.** This comment was posted at `03:26:09Z`. The current head, `105ee0f`, was committed at `03:48:41Z` — 22 minutes later — after I re-rebased onto current `release` and resolved the `markdown.ts` conflict so the two features compose.

The decisive check is that this PR does not touch `unwrapFence` **at all** relative to its base:

```
$ git diff origin/release...HEAD -- src/lib/MarkdownText/properties.ts \
                                    src/wc/components/MarkdownText.wc.svelte | grep unwrapFence
(no output)
```

Not one added or removed line. It is present in all four places named above, exactly as `release` has it:

```
$ grep -c unwrapFence src/lib/MarkdownText/properties.ts        3
$ grep -c unwrapFence src/lib/MarkdownText/MarkdownText.svelte  2
$ grep -c unwrapFence src/wc/components/MarkdownText.wc.svelte  1
$ grep -c unwrapFence docs/MarkdownText.md                      5
```

`fenceUnwrap.test.ts` is back on the branch and green:

```
$ npx vitest run src/lib/MarkdownText/fenceUnwrap.test.ts
 Test Files  1 passed (1)
      Tests  18 passed (18)
```

Those 18 restored tests are the whole of the branch's 1114 → 1132 jump. So no breaking change, and no `!` / `BREAKING CHANGE:` footer is warranted — `unwrapFence` survives this PR untouched, and the two features compose:

```
unwrapFence + sanitize -> <div class="c"><p><strong>md</strong></p></div>
sanitize alone         -> <pre><code class="language-markdown">...   (fence preserved)
```

src/lib/MarkdownText/markdown.ts
murdore: Reproduced exactly, and it is the documented contract rather than a hole — so refuting, with the repro, because "I couldn't reproduce it" would be the wrong answer here.

With a deliberately permissive caller sanitizer (`rawHtml: 'sanitize'`, `htmlSanitizer: (h) => h`):

```
rawAnchorJs : <p><a href="javascript:alert(1)">x</a></p>
rawScript   : <script>alert(1)</script>
mdSyntaxJs  : <p>x</p>                      ← markdown path still hardened
```

So yes: raw-HTML links keep whatever the caller's sanitizer returns, while markdown-syntax links are still neutered. That asymmetry is deliberate and is stated in `docs/MarkdownText.md` in almost these words:

> **The protocol allow-list covers markdown syntax, NOT raw HTML.** … `<a href="javascript:alert(1)">click</a>` written as raw HTML is emitted exactly as your sanitizer returns it — this component does not re-check it. That is the trust boundary doing what it says.

Re-applying `hasSafeProtocol`/`tagAllowed` after the sanitizer would mean the library second-guessing the function the caller supplied precisely so they could decide policy. It would also break the legitimate case: a caller allow-listing a custom scheme in their sanitizer would have it silently stripped afterwards, with no way to express intent. Raw-HTML passthrough is opt-in twice over — set `rawHtml: 'sanitize'` *and* supply the sanitizer — and omitting the sanitizer keeps the escaping.

The library's own hardening that survives regardless is external-link `rel="noopener noreferrer"`, applied after sanitization so the caller's config cannot strip it.

Resolving as documented-as-designed. If the concern is that the default is too sharp an edge, that is a real API-design question and worth its own issue rather than a change under this thread.

PR #609, opened 2026-09-12
feat(table): let callers name the controls Table generates itself

murdore:
Table names several controls on the caller's behalf, and those names were hardcoded English:

```
Table.svelte:649   Clear search
Table.svelte:691   Select all rows
Table.svelte:776   Filter by {header}
Table.svelte:794   Sort by {header}
Table.svelte:843   Close search
Table.svelte:924   Select row {rowId || 'non-selectable'}
```

`sortable` defaults to `true`, so a consumer shipping a non-English UI shipped English to their users on controls they never wrote and had no way to reach.

## Shape

```ts
type TableLabels = {
  sortBy?: (header: string) => string;
  filterBy?: (header: string) => string;
  selectRow?: (rowId: string) => string;
  selectAllRows?: string;
  clearSearch?: string;
  closeSearch?: string;
};
```

Follows the existing `MediaUploadErrorMessages` / `SpeechToTextMessages` convention in this repo — an optional bag of optional members, read with `?? default` at the point of use — rather than inventing a new one.

The three that depend on a column or a row are **functions, not templates with a placeholder**. A translator needs to decide where the header goes; `Trier par Nom` and `Nom順に並べ替え` do not have a slot in the same place.

`Search` is not included: `searchConfig.placeholder` has named the search input since search shipped, and a second way to name it would be worse than the gap.

## Non-breaking

Every member is optional and falls back to the string Table has always used. The existing specs in `table-keyed-columns.test.ts` still assert `Sort by Name` and still pass, which is the regression evidence that matters here.

## Verification

Red-green, by reverting three of the six call sites to their hardcoded strings and rebuilding:

| stage | result |
| --- | --- |
| green | 2 passed |
| wiring reverted | 2 failed — `element(s) not found` on the overridden names |
| restored | 2 passed |

```
check     exit 0
lint      exit 0
build     exit 0   (prerender anchor guard)
vitest    1021 passed
functional 716 passed, 1 unrelated flake
```

The one functional failure is `hydration-race.spec.ts` — the test that deliberately clicks before hydration to prove the race exists. Under load the race resolves the other way and the menu opens. It touches nothing in this PR, and it is the same test that flaked on an unrelated branch. This machine is sustaining a load average above 200 from other sessions; CI's runners are the signal to trust.

## Summary by CodeRabbit

* **New Features**
  * Added customizable accessibility labels for table sorting, filtering, row selection, select-all, and search controls.
  * Supports localized labels, including dynamic text based on column headers and row identifiers.
  * Added documentation and an example demonstrating localized table labels.

* **Bug Fixes**
  * Preserved existing English accessibility labels as fallbacks when overrides are not provided.

### DIFF
--- a/docs/Table.md
+++ b/docs/Table.md
@@ -492,6 +492,7 @@ directly inside the `cell` snippet, which runs in consumer scope.
 | headerTooltipIcon     | `Snippet`                                                           | No       | `-`                   | Icon snippet shown after each header label that has a `tooltip`. When set, the default underline affordance on those labels is dropped.                                                                                                                                     |
 | headerTooltipPosition | `TooltipPosition`                                                   | No       | `'top'`               | Placement of every header tooltip bubble.                                                                                                                                                                                                                                   |
 | usePortal             | `boolean`                                                           | No       | `false`               | When true, in-cell `Select` dropdowns and `Menu` popovers (`action-group`/`popup-menu` columns) are portaled to `document.body` and positioned `fixed`, so the table's own scroll/overflow container cannot clip them. Set on tables whose rows can sit near a scroll edge. |
+| labels                | `TableLabels`                                                       | No       | `-`                   | Overrides for the accessible names Table generates itself (sort, filter, row/select-all checkboxes, search clear/close). Every member is optional and falls back to the English default, so omitting the prop changes nothing. See `TableLabels` below.                     |
 
 ## Snippets
 
@@ -750,6 +751,46 @@ These variables style the search input rendered above the table when `searchConf
 | `--table-search-clear-icon-size`        | `14px`              | width, height    | Size of the clear button icon.                                  |
 
 ## Type Reference
+
+### TableLabels
+
+Table names several controls it renders on the caller's behalf. Those names were
+hardcoded English, and `sortable` defaults to `true`, so a consumer shipping a
+non-English UI shipped English on controls they never wrote. `labels` overrides
+each one.
+
+```ts
+type TableLabels = {
+  sortBy?: (header: string) => string;        // default: `Sort by ${header}`
+  filterBy?: (header: string) => string;      // default: `Filter by ${header}`
+  selectRow?: (rowId: string) => string;      // default: `Select row ${rowId || 'non-selectable'}`
+  selectAllRows?: string;                     // default: 'Select all rows'
+  clearSearch?: string;                       // default: 'Clear search'
+  closeSearch?: string;                       // default: 'Close search'
+};
+```
+
+```svelte
+<Table
+  {columns}
+  {rows}
+  labels={{
+    sortBy: (header) => `Trier par ${header}`,
+    selectAllRows: 'Tout sélectionner'
+  }}
+/>
+```
+
+The three that depend on a column or a row are functions rather than templates
+with a placeholder, because a translator needs to decide where the header goes
+rather than fill the slot an English sentence happens to leave.
+
+Members are independent: the example above translates sorting and select-all
+while leaving the filter trigger and the search buttons in English.
+
+`Search` is not here. The search input has always been named by
+`searchConfig.placeholder`, which is the right place for it.
+
 
 ```typescript
 type SortDirection = 'asc' | 'desc';
--- a/src/lib/Table/Table.svelte
+++ b/src/lib/Table/Table.svelte
@@ -51,7 +51,8 @@
     summaryRowIndex = null,
     headerTooltipIcon,
     headerTooltipPosition,
-    usePortal = false
+    usePortal = false,
+    labels
   }: TableProperties = $props();
 
   // ─── Keyed column model → positional projection ─────────────────────────────
@@ -646,7 +647,7 @@
         class="table-search-clear"
         type="button"
         onclick={clearSearch}
-        aria-label="Clear search"
+        aria-label={labels?.clearSearch ?? 'Clear search'}
       >
         <!-- eslint-disable svelte/no-at-html-tags -->
         {@html closeSvg}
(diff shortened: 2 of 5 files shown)


Tara-ag (approved):
Approved. Clean, backward-compatible feature addition. The new optional `labels` prop follows the existing message-bag convention, preserves English defaults, and is mirrored in the web-component wrapper, docs, demo page, and tests. No blocking issues found.

murdore:
## `visual` is failing on an expected baseline change, and I cannot regenerate it here

This PR adds a demo section to a component's own route, so exactly one baseline legitimately changes. For #611 CI reports it precisely:

```
markdown-text: expected 980x2563, received 980x2947 — 104716 pixels (ratio 0.04)
```

That is the new section and nothing else.

**The blocker is environmental, not the change.** Baselines must be regenerated inside the pinned container (`scripts/visual-test.sh --update-snapshots`), and that container cannot start on my machine right now: `Timed out waiting 180000ms from config.webServer`. This host is sustaining a load average above 160 from other sessions' test runs, and the build alone takes 89s idle. I tried pre-building inside the container and reusing the server via `PW_PORT`; Playwright still started its own, so that escape hatch did not help.

I have deliberately **not** run `--update-snapshots` across the whole suite. On a machine this loaded that would bake load-induced noise into unrelated routes' baselines, which is a far worse outcome than a red check — and it is close to a mistake I have already made once on this work.

So this needs one of:

- regenerating the single affected baseline on a quiet machine, or
- me retrying when this host settles.

Everything else on this PR is green: `checks`, `build`, `Yama PR Review`, and the unit/functional suites.

murdore:
## Review threads worked through at `3af72df`

### **Fixed**

**`src/routes/components/table/+page.svelte:401`** — CodeRabbit: localised-labels demo has no filterable column, so the test suite never exercises labels.filterBy or its English fallback.

> Table.svelte:752 `{#if headerColumn?.filter}` gates the filter trigger; localisedColumns (src/routes/components/table/+page.svelte, pre-fix lines 398-401) had no `filter` key on either column, and `grep -n filterBy` across tests/ returned no matches before the fix. Added a `filter` config to the `score` column (options High/Low, backed by `$state`/`$derived`) and added `await expect(table.getByRole('button', { name: 'Filter by Score' })).toBeVisible();` to tests/table-label-overrides.test.ts's 'omitted member' spec, since frenchLabels omits `filterBy`.


### Verification

```
check   exit 0
lint    exit 0
vitest  Test Files  52 passed (52); Tests  1021 passed (1021)
```

Tara-ag (commented):
Review in progress — file-by-file findings to follow.

docs/Table.md
```diff
@@ -706,25 +707,25 @@ A second search-input mode (`searchConfig.displayMode: 'inline'`) — a trigger
 
 These variables style the leading checkbox column that appears when `checkboxSelection` is set. The column renders the library `Checkbox` in controlled mode; every `--table-checkbox-*` token below is bridged onto the matching `--checkbox-*` variable inside the column, so theming the table keeps working and an app-wide `Checkbox` theme is overridden only there.
```
Tara-ag, line 709: 💬 SUGGESTION: Docs table formatting churn is unrelated noise

This block reformats the checkbox-variables table purely to right-pad the Description column. It widens the diff without changing meaning and is unrelated to the `labels` feature, making future `git blame` harder to follow.

Suggestion: revert the whitespace-only changes to the checkbox-variables table so the diff contains only the new `labels` prop row and the new `TableLabels` section.

tests/table-label-overrides.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    // Templated on the column header, so word order is the caller's.
+    await expect(table.getByRole('button', { name: 'Trier par Nom' })).toBeVisible();
+    await expect(table.getByRole('button', { name: 'Trier par Score' })).toBeVisible();
+
+    // Static strings.
+    await expect(table.getByRole('checkbox', { name: 'Tout sélectionner' })).toBeVisible();
+
+    // Templated on the row id.
+    await expect(table.getByRole('checkbox', { name: 'Sélectionner la ligne r1' })).toBeVisible();
+
```
Tara-ag: 💡 MINOR: Test does not exercise clearSearch / closeSearch overrides

The spec verifies `sortBy`, `selectAllRows`, `selectRow`, and the `filterBy` fallback, but never asserts the French `clearSearch` or `closeSearch` labels. Since these are separate code paths in `Table.svelte` (clear at ~650, close at ~847), a regression could silently revert them to English without failing CI.

Suggestion: add assertions that type into the localised table's search input and check the clear button's accessible name is `'Effacer la recherche'`, and that expanding inline search yields a close button named `'Fermer la recherche'`.

Tara-ag (commented):
Review complete. The change is a clean, backward-compatible addition: a new optional `labels` prop on `Table` lets consumers localize the accessible names Table generates for its own controls, with English fallbacks for every omitted member. No public API is removed or narrowed, no CSS custom properties are renamed, and no secrets or unsafe HTML are introduced.

Findings:
- MINOR: the new Playwright spec does not cover the `clearSearch` / `closeSearch` overrides, so those two code paths could regress to English silently.
- SUGGESTION: the docs diff contains unrelated whitespace reformatting of the checkbox-variables table.

Neither finding blocks merge.

Tara-ag (approved):
No blocking issues found. The change adds an optional `labels` prop to `Table` with per-control overrides and English defaults, preserving backward compatibility. Docs, demo, tests, and the web-component prop registration are included.

Tara-ag:

Additive, backward-compatible feature: introduces an optional `labels` prop on Table with English defaults preserved, so existing consumers are unaffected. The web-component wrapper mirrors the new prop, and docs are updated. No blocking issues (breaking API changes, secrets, XSS, unguarded browser globals) were identified in the files I could inspect. I could not fully verify the internal usage in `src/lib/Table/Table.svelte` because the fetched content was truncated, but the diff summary and tests indicate only localized string substitutions.

PR #610, opened 2026-09-12
fix(tests): make two timing-dependent specs independent of machine load

murdore:
Two specs asserted an outcome a loaded machine could decide differently. Both became the suite's own most reliable failures, and both were identified by name as the single failure in otherwise-green full runs on unrelated branches.

## `hydration-race.spec.ts`

It delays module scripts, then clicks inside that window to prove a click on an un-hydrated page is lost. The delay was a fixed `2000ms` — a bet that the click lands first. Under load it lost: hydration completed, the menu opened, `toHaveCount(0)` failed.

The delay is now a hold the test releases, so there is no window to expire:

```ts
const held = new Promise<void>((resolve) => { release = resolve; });
await page.route('**/*.js', async (route) => { await held; await route.continue(); });
```

Narrowing the old fixed delay to `40ms` reproduces the production failure exactly — `Expected: 0, Received: 1`.

The second case in that file keeps the bounded delay deliberately: it only has to prove `gotoHydrated` waits, and a delay running long makes it pass more firmly, not less.

## `typewriter-text-reduced-motion.test.ts`

Its guard required **more than 12 characters** left to type at the moment reduced motion is flipped. That is a comfort margin, not a threshold — the behaviour assertion below it allows at most **2** mutations after the flip, so any remainder above 2 already makes "kept typing" and "stopped" different measurements.

On a busy host the `emulateMedia` CDP round-trip regularly consumed more than 12 of the fixture's 74 characters, so the guard failed while the behaviour under test was perfectly correct. The bound is now the one the logic actually needs, and a remainder at or below it is a **void measurement** — re-measured via a bounded retry rather than reported as either a pass or a failure.

## Evidence

| run | load avg | result |
| --- | --- | --- |
| fixed specs, 5 repeats each | **203** | **25 passed, 0 failed** |
| unfixed specs, 5 repeats each | 103 | 3 failed, 22 passed |

All three control failures were the 12-character guard, with remainders of `1`, `1` and `4`. The threshold change covers the `4`; the retry covers the `1`s. Note the fixed run was measured at roughly **twice the load** of the control and still went clean.

One methodological note: my first control run reported 20 failures and I did not use it. Its preview server was serving a stale asset manifest after a rebuild, so every failure was a 404, not the flake — it would have flattered this change by a factor of 20. The numbers above are from a server verified healthy first.

## Not included here

A condition-based drain for the **visual** suite is the other half of this and is deliberately in a separate PR: it cannot be verified on this machine (the container's build and preview exceed the 180s webServer budget at this load), so it should stand or fall on CI rather than ride along with work that has local evidence.

## Summary by CodeRabbit

- **Tests**
  - Improved test reliability for pages loading scripts during navigation and interaction.
  - Strengthened coverage for reduced-motion behavior while animated text is in progress.
  - Added retry handling for timing-sensitive measurements to reduce flaky test results.

### DIFF
--- a/tests/hydration-race.spec.ts
+++ b/tests/hydration-race.spec.ts
@@ -18,6 +18,36 @@ import { gotoHydrated } from './support/hydrated';
  * cases pin the race deterministically instead of waiting for a bad day.
  */
 
+/**
+ * Holds every module script until the test releases it, rather than delaying
+ * them by a fixed 2000ms.
+ *
+ * The fixed delay was a bet that the click below would land inside the window,
+ * and on a loaded machine it lost: hydration completed first, the menu opened,
+ * and `toHaveCount(0)` failed -- the test that exists to prove the race is
+ * real became the suite's own most reliable flake. A hold the test controls
+ * cannot expire early, so the case is pinned rather than probable.
+ */
+const holdScripts = async (
+  page: Parameters<typeof gotoHydrated>[0]
+): Promise<{ release: () => void }> => {
+  let release = (): void => {};
+  const held = new Promise<void>((resolve) => {
+    release = resolve;
+  });
+  await page.route('**/*.js', async (route) => {
+    await held;
+    await route.continue();
+  });
+  return { release };
+};
+
+/**
+ * A bounded delay is still right for the second case: it only has to prove that
+ * `gotoHydrated` waits, and a delay that runs long makes that case pass more
+ * firmly rather than less. Only the first case is falsified by the delay
+ * expiring early.
+ */
 const delayScripts = async (page: Parameters<typeof gotoHydrated>[0]): Promise<void> => {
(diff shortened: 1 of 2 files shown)


Tara-ag (approved):
Approved. Both changes are test-only flake fixes with no component source, public API, or documentation impact.

- `tests/hydration-race.spec.ts`: Replacing the fixed 2000ms delay with a controlled script hold correctly pins the hydration race; the second case keeping a bounded delay is justified because it tests `gotoHydrated` waiting.
- `tests/typewriter-text-reduced-motion.test.ts`: Tightening the distinguishability bound to the actual behavioural threshold (2) and adding bounded retries for void measurements addresses the load-induced guard failures without weakening the final assertion.

No blocking criteria are triggered.

PR #611, opened 2026-09-12
feat(markdown-text): optionally unwrap a whole-body markdown fence

murdore:
A model asked for markdown routinely returns the whole answer wrapped:

````
```markdown
## Quarterly summary

Revenue rose **12%** against plan.
```
````

Rendered literally that is a code block, so the reader sees source instead of a document. `unwrapFence` opts into removing exactly that wrapper.

## Why it is opt-in and anchored

A peer's live implementation used an unanchored, global regex. They measured the consequence themselves: a message *explaining* markdown by quoting a fenced sample had the sample unwrapped and rendered as live markup. This matches only a fence that is the **entire** value.

Deliberately left untouched:

| Source | Reason |
| --- | --- |
| Prose before or after the fence | Not the whole body |
| ```` ``` ```` with no info string | May be code the author meant to show |
| ```` ```python ```` | Labelled as another language |
| A body still containing ```` ``` ```` | Ambiguous — the outer pair may not be the wrapper |

Tilde fences unwrap too; the info string matches `markdown` or `md` case-insensitively; surrounding whitespace is tolerated.

Unwrapping happens **before** parsing, so sanitization is untouched — raw HTML inside an unwrapped body is escaped or stripped exactly as anywhere else, and there is a test asserting that.

## Verification — written test-first

The 13 tests were written before the implementation and run against it. The red was informative in its own right:

```
5 failed | 8 passed
```

Exactly the five positive cases failed; the eight refusal cases passed trivially, because nothing unwrapped yet. After implementing, all 13 pass and the eight become load-bearing.

```
check   exit 0
lint    exit 0
build   exit 0
vitest  1034 passed  (90 across MarkdownText)
functional 713 passed, 2 unrelated intermittent
```

Both functional failures are outside MarkdownText and both reproduce intermittently on this branch and others: `typewriter-text-reduced-motion` (the flake fixed in #610, which this branch predates) and `status-default-icon`. This machine is at a load average above 120.

## Demo

The markdown-text route now shows the same source rendered both ways, side by side.

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in `unwrapFence` option to Markdown text rendering.
  - When enabled, complete `markdown` or `md`-labelled fences are removed before rendering their contents as Markdown.
  - Supports tilde fences, case-insensitive labels, and flexible whitespace.
  - Added support for the option on the Markdown custom element.

- **Documentation**
  - Documented the new option and its matching limitations.

- **Bug Fixes**
  - Preserved sanitization and prevented unwrapping for partial, unlabeled, or differently labeled fences.

### DIFF
--- a/src/lib/MarkdownText/markdown.ts
+++ b/src/lib/MarkdownText/markdown.ts
@@ -352,10 +352,77 @@ function instanceFor(options: RenderMarkdownOptions): Marked {
  * Markdown → HTML with the sanitizing pipeline above. Pure string transform —
  * no DOM involved — so it renders identically on server and client.
  */
+/**
+ * Matches a body that is ENTIRELY one fence labelled as markdown.
+ *
+ * Anchored at both ends on purpose. A peer's live implementation used an
+ * unanchored, global regex, so a message that *explained* markdown by quoting a
+ * fenced sample had the sample unwrapped and rendered as live markup. Anchoring
+ * means a fence with anything before or after it is left alone.
+ *
+ * The trailing whitespace is `\s*`, not `[ \t]*\s*` -- `[ \t]` is already a
+ * subset of `\s`, so pairing them let the engine try every split between the
+ * two quantifiers over the same characters. On uncontrolled input (this runs
+ * on chat/model text) that is a polynomial-time blow-up: a closing delimiter
+ * followed by many tabs and then one stray character made a single `.exec`
+ * call take seconds instead of a fraction of a millisecond.
+ *
+ * Groups: 1 = the opening run, 2 = the info string, 3 = the body, 4 = the closing run.
+ */
+/*
+ * Hoisted rather than built per call. Both carry only `m` -- no `g` or `y` --
+ * so they hold no `lastIndex` between calls and are safe to share.
+ */
+const INNER_BACKTICK_FENCE = /^[ \t]*`{3,}/m;
+const INNER_TILDE_FENCE = /^[ \t]*~{3,}/m;
+
+const WHOLE_BODY_MARKDOWN_FENCE =
+  /^\s*(`{3,}|~{3,})[ \t]*(markdown|md)[ \t]*\r?\n([\s\S]*?)\r?\n?(`{3,}|~{3,})\s*$/i;
+
+/**
+ * Removes a fence that wraps the entire body and is labelled `markdown` or
+ * `md`, which is how models overwhelmingly emit a markdown answer. Returns the
+ * source untouched in every other case.
+ *
+ * An unlabelled fence is deliberately not unwrapped: it may be code the author
+ * meant to show, and there is nothing in the source to tell the two apart.
+ */
+export function unwrapMarkdownFence(source: string): string {
+  const match = WHOLE_BODY_MARKDOWN_FENCE.exec(source);
+  if (match === null) {
+    return source;
+  }
+
+  const opening = match[1];
+  const body = match[3];
+  const closing = match[4];
+
+  // CommonMark lets a closing fence be LONGER than its opening, so the closing
+  // run is its own group and is checked here rather than by a `\1`
+  // backreference. The backreference did not refuse a longer close -- it
+  // matched the last three ticks and swept the extra one into the body, so
+  // `# Heading` came back as "# Heading\n`". Content corruption, not a refusal.
+  // Same character, and at least as long, is the actual rule.
+  if (closing[0] !== opening[0] || closing.length < opening.length) {
+    return source;
+  }
+
+  // A body still containing a fence run means the outer delimiters may not be a
+  // wrapper at all -- the regex is lazy to the LAST delimiter, so this shape is
+  // ambiguous. Refusing keeps a message quoting fenced code from being
+  // reinterpreted as markup, which is the failure this feature exists to avoid.
+  const innerFence = opening[0] === '~' ? INNER_TILDE_FENCE : INNER_BACKTICK_FENCE;
+  if (innerFence.test(body)) {
+    return source;
+  }
+
+  return body;
+}
+
 export function renderMarkdown(markdown: string, options: RenderMarkdownOptions = {}): string {
+  const source = options.unwrapFence === true ? unwrapMarkdownFence(markdown) : markdown;
   const instance = instanceFor(options);
-  const output =
-    options.inline === true ? instance.parseInline(markdown) : instance.parse(markdown);
+  const output = options.inline === true ? instance.parseInline(source) : instance.parse(source);
   if (typeof output !== 'string') {
     return '';
   }
--- a/docs/MarkdownText.md
+++ b/docs/MarkdownText.md
@@ -36,15 +36,16 @@ The `marked` package is a peer dependency, needed only when you actually use `Ma
 
 ## Props
 
-| Prop              | Type                      | Required | Default | Description                                                                                                                                                                                                         |
-| ----------------- | ------------------------- | -------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| markdown          | `string`                  | Yes      | `-`     | Markdown source. Raw HTML is escaped by default; strip mode removes tags and retains text; unsafe link/image protocols are stripped while their text stays.                                                         |
-| breaks            | `boolean`                 | No       | `false` | Render single newlines as `<br>` (GFM "breaks" mode).                                                                                                                                                               |
-| testId            | `string`                  | No       | `-`     | `data-pw` (and native `testID`) on the root element.                                                                                                                                                                |
-| classes           | `string`                  | No       | `-`     | Class string on the root element.                                                                                                                                                                                   |
-| tableLabel        | `string`                  | No       | `-`     | Names the scrollable table region: adds `role="region"` and `aria-label` to the wrapper. Without it the wrapper is still focusable and scrollable but announces no landmark.                                        |
-| tableWrapperClass | `string`                  | No       | `-`     | An extra class on the scrollable table wrapper, added alongside the built-in `markdown-table-wrapper` rather than replacing it. See Wide tables below.                                                              |
-| sanitize          | `MarkdownSanitizeOptions` | No       | `-`     | Narrows the link/image protocol allow-list, the set of tags allowed to render as themselves, whether task-list checkboxes render, and whether raw HTML is escaped or dropped. See the Security model section below. |
+| Prop              | Type                      | Required | Default | Description                                                                                                                                                                                                                                                        |
+| ----------------- | ------------------------- | -------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| markdown          | `string`                  | Yes      | `-`     | Markdown source. Raw HTML is escaped by default; strip mode removes tags and retains text; unsafe link/image protocols are stripped while their text stays.                                                                                                        |
+| breaks            | `boolean`                 | No       | `false` | Render single newlines as `<br>` (GFM "breaks" mode).                                                                                                                                                                                                              |
+| testId            | `string`                  | No       | `-`     | `data-pw` (and native `testID`) on the root element.                                                                                                                                                                                                               |
+| classes           | `string`                  | No       | `-`     | Class string on the root element.                                                                                                                                                                                                                                  |
+| tableLabel        | `string`                  | No       | `-`     | Names the scrollable table region: adds `role="region"` and `aria-label` to the wrapper. Without it the wrapper is still focusable and scrollable but announces no landmark.                                                                                       |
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by murdore** [comment]
## `visual` is failing on an expected baseline change, and I cannot regenerate it here

This PR adds a demo section to a component's own route, so exactly one baseline legitimately changes. For #611 CI reports it precisely:

```
markdown-text: expected 980x2563, received 980x2947 — 104716 pixels (ratio 0.04)
```

That is the new section and nothing else.

**The blocker is environmental, not the change.** Baselines must be regenerated inside the pinned container (`scripts/visual-test.sh --update-snapshots`), and that container cannot start on my machine right now: `Timed out waiting 180000ms from config.webServer`. This host is sustaining a load average above 160 from other sessions' test runs, and the build alone takes 89s idle. I tried pre-building inside the container and reusing the server via `PW_PORT`; Playwright still started its own, so that escape hatch did not help.

I have deliberately **not** run `--update-snapshots` across the whole suite. On a machine this loaded that would bake load-induced noise into unrelated routes' baselines, which is a far worse outcome than a red check — and it is close to a mistake I have already made once on this work.

So this needs one of:

- regenerating the single affected baseline on a quiet machine, or
- me retrying when this host settles.

Everything else on this PR is green: `checks`, `build`, `Yama PR Review`, and the unit/functional suites.

**Comment by murdore** [comment]
## Review threads worked through at `c56c6ea`

### **Fixed**

**`src/lib/MarkdownText/markdown.ts:357 (WHOLE_BODY_MARKDOWN_FENCE regex, defined lines 345-346)`** — CodeQL: js/polynomial-redos — the WHOLE_BODY_MARKDOWN_FENCE regex used on uncontrolled (chat/model) input can run slow on strings with many repeated tabs.

> Confirmed real via a standalone repro (/tmp/redos-test2.mjs): input '```markdown\nhello\n```' + '\t'.repeat(m) + 'X' against the original regex /^\s*(`{3,}|~{3,})[ \t]*(markdown|md)[ \t]*\r?\n([\s\S]*?)\r?\n?\1[ \t]*\s*$/i showed clear O(n^2) growth: m=1000 -> 52.6ms, m=8000 -> 847ms, m=16000 -> 2859ms, m=32000 -> 11405ms (each doubling of m roughly quadrupled time). Root cause: the trailing `[ \t]*\s*$` pairs two adjacent quantifiers where `[ \t]` is a strict subset of `\s`, so on a body that has a real closing delimiter followed by a long tab run and then one non-whitespace character (never reaching the `$` anchor), the engine tries every split between the two quantifiers before failing (classic overlapping-quantifier ReDoS). Fixed by removing the redundant `[ \t]*` in src/lib/MarkdownText/markdown.ts, changing the regex to end in `\1\s*$` (semantically identical since \s already cover


### Verification

```
check   exit 0
lint    exit 0
vitest  Test Files  53 passed (53) / Tests  1035 passed (1035) — Duration 60.15s
```

**Comment by Tara-ag** [comment]
💡 MINOR: Type reference block omits the new unwrapFence prop

**File:** `docs/MarkdownText.md`, around line 196 (Type Reference block)

The Type Reference code block at the bottom of this doc still shows the old `MarkdownTextProperties` and `RenderMarkdownOptions` shapes without `unwrapFence`, while the Props table above and the actual `src/lib/MarkdownText/properties.ts` types both include it. This inconsistency can mislead consumers copying the type reference into their own code.

**Suggested fix:** Add `unwrapFence?: boolean;` to both `MarkdownTextProperties` and `RenderMarkdownOptions` in the Type Reference block to match the Props table and the source types.

**Comment by Tara-ag** [comment]

The PR adds a well-scoped `unwrapFence` option to `MarkdownText`, `renderMarkdown`, and the web-component registry, with careful anchoring to avoid misinterpreting quoted fences. However, `ChatMessage` — the primary consumer documented to use the same sanitized markdown pipeline — does not expose or forward the new prop, leaving a feature gap for the main use case. The docs index JSON also still lists the stale `renderMarkdown` signature. No security, SSR, accessibility, or breaking-change issues were found in the changed files. Inline comments could not be posted because the affected files (`src/lib/ChatMessage/properties.ts` and `docs/_index.json`) are not part of this PR's diff; the findings are recorded in the review body instead.

- ⚠️ **MAJOR**: ChatMessage does not expose unwrapFence — `src/lib/ChatMessage/properties.ts:24`
  MarkdownText and renderMarkdown gain an unwrapFence option, but ChatMessage — the component the docs explicitly say uses the same pipeline internally — has no corresponding prop. A consumer receiving a model reply wrapped in ```markdown through ChatMessage cannot opt into unwrapping without abandoning ChatMessage and reimplementing its bubble chrome.
- 💡 **MINOR**: Docs index JSON still documents stale renderMarkdown signature — `docs/_index.json:1`
  docs/_index.json contains the MarkdownText blurb that lists the renderMarkdown signature as `{breaks?, inline?, tableLabel?, sanitize?}` and omits `unwrapFence`. This stale signature contradicts the updated MarkdownText.md and properties.ts.

**Comment by Tara-ag** [comment]
💡 MINOR: Type Reference block omits the new `unwrapFence` prop

**File:** `docs/MarkdownText.md`, Type Reference block

The `MarkdownTextProperties` and `RenderMarkdownOptions` shapes shown in the Type Reference block do not include `unwrapFence?: boolean;`, but the Props table above and `src/lib/MarkdownText/properties.ts` both declare it. Consumers copying this reference into their own code will miss the prop.

**Suggested fix:** Add `unwrapFence?: boolean;` to both `MarkdownTextProperties` (after `tableLabel`) and `RenderMarkdownOptions` (after `tableLabel`).

**Comment by murdore** [comment]
## Correct, and the consequence was worse than a refusal — fixed at `adff5b2`

CommonMark does permit a closing fence longer than its opening, and my `\1` backreference did not merely fail to match those. It matched the **last** three ticks and swept the surplus into the body:

```
'```markdown\n# Heading\n````'   →  "# Heading\n`"     ← stray backtick in the content
'~~~markdown\n# Heading\n~~~~'   →  "# Heading\n~"     ← same for tildes
```

So the rendered document silently gained a character the author never wrote. That is content corruption, not the safe refusal I had assumed the strictness bought.

The closing run is now its own capture group, validated in code:

```ts
if (closing[0] !== opening[0] || closing.length < opening.length) {
  return source;
}
```

| case | before | after |
| --- | --- | --- |
| 3 open / 3 close | `# Heading` | `# Heading` |
| 3 open / 4 close | `# Heading` + stray `` ` `` | `# Heading` |
| `~~~` / `~~~~` | `# Heading` + stray `~` | `# Heading` |
| 4 open / 3 close | refused | refused |
| `` ``` `` open / `~~~` close | refused | refused |

Four regression tests pin the whole length relationship, including that the surplus character does not survive into the output — asserting `<h1>Heading</h1>` alone would have passed against the bug.

I also re-checked the ReDoS fix still holds, since this changed the same regex: **0.1 ms at 20,000 tabs**, unchanged.

```
check   exit 0
lint    exit 0
vitest  1039 passed  (95 across MarkdownText)
```

murdore:
## Correct — `unwrapFence` was missing from both type blocks

Verified against `docs/MarkdownText.md`: the Type Reference showed

```ts
export type MarkdownTextProperties = { …  sanitize?: MarkdownSanitizeOptions; };
export type RenderMarkdownOptions  = { …  sanitize?: MarkdownSanitizeOptions; };
```

with no `unwrapFence`, while the Props table above and `src/lib/MarkdownText/properties.ts` both declare it. A consumer copying the reference would have missed the prop entirely.

Added to both.

```
check   exit 0
lint    exit 0
build   exit 0   (prerender anchor guard)
vitest  1039 passed
```

src/lib/MarkdownText/markdown.ts
```diff
@@ -332,10 +332,70 @@ function instanceFor(options: RenderMarkdownOptions): Marked {
...
+  }
+
+  // A body still containing a fence run means the outer delimiters may not be a
+  // wrapper at all -- the regex is lazy to the LAST delimiter, so this shape is
+  // ambiguous. Refusing keeps a message quoting fenced code from being
+  // reinterpreted as markup, which is the failure this feature exists to avoid.
+  const innerFence = new RegExp(`^[ \t]*${opening[0] === '~' ? '~' : '`'}{3,}`, 'm');
+  if (innerFence.test(body)) {
+    return source;
+  }
```
Tara-ag, line 417: 💬 SUGGESTION: The inner-fence regex is `new RegExp(...)`-constructed on every call to `unwrapMarkdownFence`. Since this runs on every markdown render (including streaming chat), consider hoisting static regexes for backtick and tilde fences to module scope instead of compiling per call.

Tara-ag:
💡 MINOR: `docs/_index.json` is not updated for the new `unwrapFence` prop.

The `MarkdownText` entry in `docs/_index.json` still lists the `renderMarkdown` signature as `{breaks?, inline?, tableLabel?, sanitize?}` and omits `unwrapFence`. This stale signature contradicts the updated `docs/MarkdownText.md` and `src/lib/MarkdownText/properties.ts`.

**Suggested fix:** Update the `MarkdownText` entry in `docs/_index.json` to include `unwrapFence` in the `renderMarkdown` signature and mention the new prop.

Tara-ag:
⚠️ MAJOR: `ChatMessage` does not expose `unwrapFence`, creating a feature gap for the primary consumer.

`MarkdownText` and `renderMarkdown` gain an `unwrapFence` option, but `ChatMessage` — the component the docs explicitly say uses the same pipeline internally — has no corresponding prop. A consumer receiving a model reply wrapped in ```markdown through `ChatMessage` cannot opt into unwrapping without abandoning `ChatMessage` and reimplementing its bubble chrome.

**Suggested fix:** Add `unwrapFence?: boolean` to `OptionalChatMessageProperties` in `src/lib/ChatMessage/properties.ts` and forward it through `ChatMessage.svelte` to `renderMarkdownFn(markdown, { unwrapFence })` (and to the typewriter's `renderText` path if applicable).

Tara-ag (changes requested):
Reviewed file by file. Found one inline suggestion and two broader issues (posted as PR comments): the primary consumer `ChatMessage` does not expose the new `unwrapFence` option despite using the same markdown pipeline, and `docs/_index.json` still shows the old `renderMarkdown` signature. Requesting changes to complete the feature surface.

Tara-ag:
💡 MINOR: `docs/_index.json` still documents the `MarkdownText` `renderMarkdown` signature as `{breaks?, inline?, tableLabel?, sanitize?}`. Update it to include the new `unwrapFence` option so the docs index matches the updated API in `MarkdownText.md` and `properties.ts`.

Tara-ag:
⚠️ MAJOR: `src/lib/ChatMessage/properties.ts` does not expose the new `unwrapFence` option, leaving a feature gap for the primary consumer documented to use this pipeline. Add `unwrapFence?: boolean` to `OptionalChatMessageProperties` and forward it to `renderMarkdownFn(markdown, { unwrapFence })` (and the typewriter path if applicable) so consumers rendering model replies through `ChatMessage` can opt into stripping the outer ```markdown fence.

Tara-ag (commented):
The `unwrapFence` feature is well-scoped for `MarkdownText` and `renderMarkdown`, but it needs to be threaded through `ChatMessage` (the documented primary consumer) and the docs index needs to reflect the new option.

src/lib/MarkdownText/markdown.ts
murdore: Taken — hoisted to module scope in `de54a95`.

```ts
const INNER_BACKTICK_FENCE = /^[ \t]*`{3,}/m;
const INNER_TILDE_FENCE = /^[ \t]*~{3,}/m;
...
const innerFence = opening[0] === '~' ? INNER_TILDE_FENCE : INNER_BACKTICK_FENCE;
```

One correction to the framing, though, because it affects how much this was costing: it did not run on every markdown render. The `new RegExp` sat below the `WHOLE_BODY_MARKDOWN_FENCE` match and the delimiter check, so it was only reached when `unwrapFence` was on **and** the source actually was a whole-body fence. The allocation was real but rarer than "every render, including streaming chat".

Worth doing regardless — there are exactly two possible patterns, so constructing either per call was pure waste, and the ternary now reads better than string-interpolating a delimiter into a pattern.

The one thing to check before sharing regex objects is statefulness: a shared literal with `g` or `y` carries `lastIndex` across calls and would make `.test()` alternate true/false on identical input. Both of these carry only `m`, so they are stateless and safe to share. Noted in a comment above them so nobody adds a flag later without noticing.

This branch was also rebased onto current release in the same push, which brought in #603's `tableWrapperClass` — four files conflicted and are resolved additively. Gate: prettier clean, eslint 0, event-casing 0, svelte-check 0 errors across both tsconfigs, 1072 unit tests passing.

Tara-ag (commented):

Tara-ag (commented):
Findings recorded in review verdict; the affected files are outside the PR diff so inline comments could not be attached.

Tara-ag (commented):
The PR adds a well-scoped `unwrapFence` option to `MarkdownText`, `renderMarkdown`, and the web-component registry, with careful anchoring to avoid misinterpreting quoted fences. Two gaps remain: `ChatMessage` does not expose the new option despite being documented as using the same pipeline, and the docs index still lists the stale `renderMarkdown` signature. Inline comments could not be posted on those files because they are not part of this PR's diff.

PR #612, opened 2026-09-12
fix(visual): let the page drain between clock advances, not after all of them

murdore:
## Read this part first

**I could not verify this change on my machine, and CI is the gate.** The visual suite's container could not even start here — `Timed out waiting 180000ms from config.webServer` — because this host is sustaining a load average above 170 from other sessions, and a build that takes 89s idle takes longer than the budget under that.

It is in its own PR for exactly that reason. The functional half of this work (#610) has local before/after measurements; this half has none, so it should not ride along on that evidence. If CI's `visual` job goes green, the change is sound and baseline-preserving. If it goes red, the change is wrong and should be dropped.

## The defect

The suite pins a fake clock and advances it once by a fixed budget, so timer-driven demos settle identically every run. That holds only if the demo's timers are the whole story — and they are not.

`page.clock.runFor(N)` fires every fake timer due in the window back to back. But a demo that chains `setTimeout` through `await` reaches its next timer only once the microtask queue has run, and **that runs in real time, on a real event loop**. Firing the whole budget in one call leaves a cascade wherever the machine's scheduler happened to reach.

That is the mechanism behind failures on routes a branch never touched. Measured on `feat/form-association`, which touches three routes (`checkbox`, `select`, `toggle`):

```
failed: carousel · chat · chat-message · hitl · relative-time · sankey-chart · stat-card
```

Zero overlap.

## The change

```ts
for (let advanced = 0; advanced < budget; advanced += DRAIN_SLICE_MS) {
  await page.clock.runFor(Math.min(DRAIN_SLICE_MS, budget - advanced));
  await waitForDomQuiet(page);
}
```

**Total fake time advanced is unchanged, deliberately.** Baselines encode it — `relative-time` renders "30 seconds" because exactly 1500ms of fake time elapsed — so stopping early the moment the DOM went quiet would shift every clock-dependent baseline. The only change is that the page gets real time to catch up *between* slices.

## One trap worth naming

My first version waited with `setTimeout` inside `page.evaluate`. That is broken: `page.clock.install()` replaces the page's `setTimeout`, so the wait would have been driven by the very fake clock it exists to pace and would never have fired. The quiet wait runs in Node; only the mutation counter lives in the page. A cap keeps an endlessly-animating demo from hanging the suite rather than failing it.

## What is verified

```
check  exit 0
lint   exit 0
```

That is all I can honestly claim from here.

### DIFF
--- a/tests/visual/components.visual.ts
+++ b/tests/visual/components.visual.ts
@@ -2,6 +2,13 @@ import { readdirSync } from 'node:fs';
 import { join } from 'node:path';
 import { expect, test, type Page } from '@playwright/test';
 
+declare global {
+  interface Window {
+    /** Mutation tally kept by `waitForDomQuiet`; read from Node between clock slices. */
+    __visualMutationCount?: number;
+  }
+}
+
 // Routes are discovered from disk rather than hand-listed on purpose: a new
 // component demo added without a committed baseline fails this suite instead of
 // silently going unscreenshotted, which is the failure mode that lets visual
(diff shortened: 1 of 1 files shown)


tests/visual/components.visual.ts
```diff
@@ -99,6 +106,87 @@ const SETTLE_OVERRIDES: Readonly<Record<string, number>> = {
...
+
+  const read = async (): Promise<number> => page.evaluate(() => window.__visualMutationCount ?? 0);
+
+  const deadline = Date.now() + DRAIN_CAP_MS;
+  let previous = await read();
+  while (Date.now() < deadline) {
+    await new Promise((resolve) => setTimeout(resolve, DRAIN_QUIET_MS));
+    const current = await read();
+    if (current === previous) {
+      return;
```
Tara-ag, line 184: 💡 MINOR: `waitForDomQuiet` may return early when mutations are still pending. It reads the count, waits `DRAIN_QUIET_MS`, then compares; if the first read catches a moment before any mutation is recorded, it returns before work completes. More importantly, it does not reset the quiet window after the *last* mutation — scheduled microtasks/macrotasks can still produce more mutations after the read. Track the timestamp of the last mutation inside the page and resolve only when `Date.now() - lastMutationAt >= DRAIN_QUIET_MS`.

tests/visual/components.visual.ts
```diff
@@ -99,6 +106,87 @@ const SETTLE_OVERRIDES: Readonly<Record<string, number>> = {
   'tool-call-log': 30_000
 };
 
+/**
+ * Real-time slice given to the page between clock advances, and the quiet
+ * period that ends a settle early.
+ */
+const DRAIN_SLICE_MS = 500;
```
Tara-ag, line 113: 💬 SUGGESTION: The drain timing constants (`DRAIN_SLICE_MS`, `DRAIN_QUIET_MS`, `DRAIN_CAP_MS`) are magic numbers without in-file justification. Add brief comments explaining how they were chosen and what trade-offs they represent, so future maintainers know whether they can be tuned.

tests/visual/components.visual.ts
```diff
@@ -99,6 +106,87 @@ const SETTLE_OVERRIDES: Readonly<Record<string, number>> = {
...
+ */
+const waitForDomQuiet = async (page: Page): Promise<void> => {
+  await page.evaluate(() => {
+    if (typeof window.__visualMutationCount === 'number') {
+      return;
+    }
+    window.__visualMutationCount = 0;
+    const observer = new MutationObserver((records) => {
+      window.__visualMutationCount = (window.__visualMutationCount ?? 0) + records.length;
+    });
```
Tara-ag, line 167: 💡 MINOR: The `MutationObserver` created here is never disconnected. It observes `document.documentElement` for the rest of the page lifetime, so `__visualMutationCount` keeps growing and the observer keeps firing during screenshot capture. Store the observer on `window` (or return it) and disconnect it after `waitForDomQuiet` resolves, ideally in a `finally` block.

Tara-ag (commented):
This PR changes only the visual test harness (`tests/visual/components.visual.ts`) and does not touch the published component library surface, so the public API / backward-compatibility concerns in the blocking criteria do not apply. The change is a test-only fix aimed at making fake-clock settling deterministic under host load.

Findings are all quality/readability issues in the new helper:
- `waitForDomQuiet` can return before the DOM is truly quiet because it does not reset its quiet window on the most recent mutation.
- The `MutationObserver` is never disconnected.
- The drain timing constants are unexplained magic numbers.

None of these are blocking. The PR author already notes CI is the gate for the visual suite; if the `visual` job passes, the behavioral intent is validated. I recommend addressing the observer cleanup and the quiet-window reset before merge if easy, but the change can land without them.

murdore:
## Closing this — CI says the change is not safe, and that was the agreed test

I opened this saying: *"If CI's `visual` job goes green, the change is sound. If it goes red, the change is wrong and should be dropped."* It went red twice, on `hitl`:

```
92 passed, 1 failed
hitl: expected 980x3519, received 980x3522 — 68655 pixels (ratio 0.02)
```

The control is what settles it. `hitl` is **green** on the visual runs of three other branches right now — `fix/load-tolerant-timing-tests`, `fix/chat-message-table-semantics`, `feat/markdown-raw-html-sanitizer`. Green everywhere, red only here, twice. That is the same zero-overlap reasoning I used to exonerate other branches from unrelated visual failures; applied to my own change it convicts it.

## What I got wrong

The premise was sound — `runFor` fires fake timers back to back while an `await`-chained cascade advances on the real event loop, so the cascade lands wherever the scheduler reached. The fix was not: draining between slices changes *when* each slice's timers fire relative to the cascade, and for `hitl` that lands somewhere new rather than somewhere stable. The run's own stability check shows the height oscillating 3519 ↔ 3522 within a single run, so the page is still moving at capture time — which is precisely what this was supposed to prevent, and did not.

I am not going to iterate on it speculatively. The honest position is that I do not currently have a verified fix for the visual half, and I could not test it locally (the pinned container cannot start on my machine at its present load).

## What stands

The functional half is unaffected and has local before/after evidence: **#610** fixes the two timing-dependent specs — 25/25 repeats at load average 203, against 3 failures in 25 on the unfixed specs at roughly half that load.

The `hitl` instability visible here is real and pre-existing and worth its own investigation, separately from this approach.

Tara-ag (commented):
Visual regression harness improvements reviewed. Three quality findings raised; none blocking.

tests/visual/components.visual.ts
Tara-ag, line 167: 💡 MINOR: The `MutationObserver` installed by `waitForDomQuiet` is never disconnected. It is stored only in a local const inside the evaluated function and observes `document.documentElement` for the remaining lifetime of the page. While each route runs in a fresh page context, this still means the observer keeps counting mutations during screenshot capture and causes unnecessary work. Store the observer on `window` (or return its handle) and disconnect it after `waitForDomQuiet` resolves.

tests/visual/components.visual.ts
Tara-ag, line 184: 💡 MINOR: `waitForDomQuiet` may return early on the first iteration when mutations are still pending. The function reads the mutation count, waits `DRAIN_QUIET_MS`, then compares. If the first read happens while the microtask queue is mid-flush but no mutation has been recorded yet, it can return before real work completes. More importantly, it does not wait for the DOM to be quiet *after* the last observed mutation; it waits a fixed interval and returns if the count is unchanged. Consider tracking the timestamp of the last mutation inside the page and resolving only when `Date.now() - lastMutationAt >= DRAIN_QUIET_MS`, so the quiet period starts after the most recent mutation.

tests/visual/components.visual.ts
Tara-ag, line 113: 💬 SUGGESTION: `DRAIN_SLICE_MS = 500`, `DRAIN_QUIET_MS = 120`, and `DRAIN_CAP_MS = 3000` are presented without justification. Add brief comments explaining how the values were chosen (e.g., 120ms is long enough for a typical microtask flush but short enough to keep suite runtime bounded) so future maintainers can tune them safely.

Tara-ag (commented):
Visual regression harness improvements reviewed. Three quality findings raised; none blocking.

PR #613, opened 2026-09-12
fix(wc): stop sui-checkbox shadowing Element.attributes, which broke every instance

murdore:
`<sui-checkbox>` has never worked in a published release. Every instance throws on connect:

```
TypeError: this.attributes is not iterable
```

Reproduced in Chromium with a bare `document.createElement('sui-checkbox')` and no attributes set at all.

## Cause

`Checkbox.wc.svelte` declared a custom-element prop named `attributes`. Svelte defines every declared prop as an accessor on the host element, so that **replaced `Element.prototype.attributes`** — the live `NamedNodeMap` that Svelte's own generated `connectedCallback` iterates:

```js
for (const attr of this.attributes) { ... }
```

With the accessor replaced by a plain object, the iteration throws and the component never initialises.

**It failed silently, which is why it survived to release.** `appendChild` does not surface an error thrown inside `connectedCallback`, so the element appears to attach and simply renders nothing. There is no failing call for anyone to notice.

## Fix

Exactly the precedent this file already sets for `ariaLabel` → `checkboxAriaLabel`: forward it under a name the platform does not own. The component prop is untouched; only the custom element renames it to `checkboxAttributes`.

`attributes` also joins `HOST_RESERVED_PROPS` so the parity guard rejects it in future. That list already reserves `children` for **exactly this bug class**, with a comment saying it was measured rather than reasoned about — `attributes` was simply missed, because the guard's own docs describe it as covering names on `HTMLElement`, and `attributes` lives on `Element`.

## Evidence

| | page errors | shadow root |
| --- | --- | --- |
| before | `this.attributes is not iterable` | never populated |
| after | 0 | initialised |

Red-green on the regression test, by restoring the colliding name and rebuilding:

```
colliding name  →  expect(state.attributesIsIterable).toBe(true)
                   Received: false
renamed         →  1 passed
```

```
check  exit 0
lint   exit 0
```

## One thing worth flagging about the test

**My first version of this test passed against the bug.** It navigated to a demo route that never loads `dist-wc`, so `sui-checkbox` was an undefined element — `shadowRoot?.firstChild !== null` evaluated `undefined !== null`, which is `true`, and the whole test asserted nothing. I caught it only because the mutation run came back green when it had to be red.

The version here injects the bundle the way `tests/wc-custom-elements.spec.ts` does and asserts `customElements.get('sui-checkbox') !== undefined` first, so it cannot pass vacuously again.

## Separately: form participation is still absent for custom elements

While verifying this I confirmed a **different** gap, which this PR does not address. With the element now initialising, `<sui-checkbox name="agree">` inside a `<form>` still contributes nothing:

```
FormData: []
```

Shadow-DOM content does not participate in form submission without `ElementInternals`/`setFormValue`, and Svelte 5.56.3 has no built-in support for form-associated custom elements (`grep -rn 'formAssociated|attachInternals' node_modules/svelte/src/` returns nothing). So the form participation added in #605 reaches the Svelte components but **not** the custom elements. That is worth its own issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Renamed the checkbox custom element’s `attributes` prop to `checkboxAttributes`, preventing conflicts with the native `Element.attributes` property.
  * Preserved existing accessibility label behavior and checkbox prop forwarding.

* **Tests**
  * Added coverage confirming checkbox elements retain native attribute behavior and initialize without page errors.

### DIFF
--- a/tests/wc-attributes-collision.spec.ts
+++ b/tests/wc-attributes-collision.spec.ts
@@ -1,0 +1,63 @@
+import { expect, test } from '@playwright/test';
+import { gotoHydrated } from './support/hydrated';
+
+/**
+ * `<sui-checkbox>` declared a custom-element prop named `attributes`, which
+ * replaced `Element.prototype.attributes` -- the live `NamedNodeMap` that
+ * Svelte's OWN generated `connectedCallback` iterates:
+ *
+ *   for (const attr of this.attributes) { ... }
+ *
+ * With the accessor shadowed by a plain object, every instance threw
+ * `TypeError: this.attributes is not iterable` on connect and never
+ * initialised. The element appended fine -- `appendChild` does not surface an
+ * error thrown inside `connectedCallback` -- so it failed silently, which is
+ * why it survived to a published release.
+ *
+ * This is the same class as `children`, already reserved in
+ * `HOST_RESERVED_PROPS` for the same reason, and the fix follows the same
+ * precedent as `ariaLabel` -> `checkboxAriaLabel`: forward it under a name the
+ * platform does not own.
+ */
+test.describe('sui-checkbox does not shadow Element.attributes', () => {
+  test('connects, initialises its shadow root, and throws nothing', async ({ page }) => {
+    const errors: string[] = [];
+    page.on('pageerror', (error) => errors.push(error.message));
+
+    // dist-wc is a self-contained bundle, not a route of the demo site, so it is
+    // injected the same way tests/wc-custom-elements.spec.ts does it. Without this
+    // the element is simply undefined and every assertion below passes vacuously --
+    // which is exactly how the first version of this test passed against the bug.
+    //
+    // MAINTAINERS: this asserts against the BUILT bundle, so `dist-wc/index.js`
+    // has to be current or the result describes an old build rather than your
+    // change. `npm run build` (which runs build:wc) before running this locally.
+    // CI is safe -- playwright.config.ts builds as part of webServer -- but a
+    // local run reusing a server does not rebuild. That is not hypothetical: a
+    // stale bundle is how this very test was first observed passing against the
+    // defect it exists to catch.
+    await gotoHydrated(page, '/');
+    await page.addScriptTag({ path: 'dist-wc/index.js', type: 'module' });
+    await page.waitForFunction(() => typeof customElements.get('sui-checkbox') === 'function');
+
+    const state = await page.evaluate(async () => {
+      const element = document.createElement('sui-checkbox');
+      element.setAttribute('text', 'Agree');
+      document.body.appendChild(element);
+      await new Promise((resolve) => requestAnimationFrame(resolve));
+      return {
+        defined: typeof customElements.get('sui-checkbox') === 'function',
+        // The native accessor must still be the NamedNodeMap, not a plain object.
+        attributesIsIterable: typeof element.attributes?.[Symbol.iterator] === 'function',
+        // `?? false` matters: an element that never upgraded has a null shadowRoot,
+        // and an optional-chained comparison would read as "initialised".
+        shadowInitialised: (element.shadowRoot?.childNodes.length ?? 0) > 0
+      };
+    });
+
+    expect(state.defined).toBe(true);
+    expect(state.attributesIsIterable).toBe(true);
+    expect(state.shadowInitialised).toBe(true);
+    expect(errors).toEqual([]);
+  });
+});
--- a/docs/Checkbox.md
+++ b/docs/Checkbox.md
@@ -82,7 +82,18 @@ Tag: `<sui-checkbox>`
 <sui-checkbox text="Parent decides" controlled></sui-checkbox>
 ```
 
-`controlled` maps to the attribute of the same name; `attributes` is an object property, set from script.
+`controlled` maps to the attribute of the same name.
+
+Two props are renamed on the custom element, because the platform already owns
+those names on every `HTMLElement`:
+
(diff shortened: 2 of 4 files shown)


murdore:
## CI caught a lint error I should have caught myself — fixed at `27732c3`

```
33:77  error  undefined is not allowed. Use null or proper type checking instead
41:57  error  undefined is not allowed. Use null or proper type checking instead
```

My local lint run was clean, and it was clean because I ran it **before** the last rewrite of the test and never re-ran it afterwards. That rewrite is exactly where `!== undefined` entered. Same discipline gap as running a test against a stale build — the verification was real, it just wasn't verifying the current file.

Replaced both with `typeof customElements.get('sui-checkbox') === 'function'`, which is also the more honest check: it asserts the element is *defined as a constructor*, not merely "not undefined".

```
lint   exit 0
check  exit 0
build  exit 0
playwright  1 passed
```

Amended into the single commit.

tests/wc-attributes-collision.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+test.describe('sui-checkbox does not shadow Element.attributes', () => {
+  test('connects, initialises its shadow root, and throws nothing', async ({ page }) => {
+    const errors: string[] = [];
+    page.on('pageerror', (error) => errors.push(error.message));
+
+    // dist-wc is a self-contained bundle, not a route of the demo site, so it is
+    // injected the same way tests/wc-custom-elements.spec.ts does it. Without this
+    // the element is simply undefined and every assertion below passes vacuously --
+    // which is exactly how the first version of this test passed against the bug.
+    await gotoHydrated(page, '/');
```
Tara-ag, line 39: 💬 SUGGESTION: This spec injects `dist-wc/index.js` directly, which is correct for CI but can test a stale bundle if a developer runs Playwright locally without rebuilding. Consider adding a comment reminding maintainers that `npm run build:wc` (or `npm run build`) must be current before this test is meaningful.

Tara-ag (commented):
Reviewed file by file. The fix is correct and the regression test is well-constructed. Two notes:

1. docs/Checkbox.md needs to document the renamed web-component property (`checkboxAttributes` instead of `attributes`). Left a PR-level comment since the file is not in the diff.
2. The new Playwright test depends on `dist-wc/index.js` being current; a short comment reminding maintainers to rebuild would help local development.

No blocking issues.

Tara-ag:
💡 MINOR (docs/Checkbox.md): The web-component section still documents `attributes` as the object property, but the custom element now exposes it as `checkboxAttributes`. A consumer following this doc will set `element.attributes = {...}` and silently mutate the native `NamedNodeMap` instead of forwarding extra attributes to the checkbox box.

Update the web-component paragraph to document `checkboxAttributes` instead of `attributes`, e.g.: `` `controlled` maps to the attribute of the same name; `checkboxAttributes` is an object property, set from script. ``

Tara-ag (commented):
Review comments for accepted findings.

murdore:
## Good catch — added at `1bdfb86`

You are right, and it is not hypothetical. **A stale bundle is how this very test was first observed passing against the defect it exists to catch.** I mutated the source back to the colliding name, rebuilt, ran the test — and it went green, because the preview server was still serving the previous build. The red only appeared once I rebuilt properly.

That is the same trap twice over in this work: a rebuild also invalidates a running `pnpm preview` server's asset hashes, which makes every test 404 and look like a substantive failure.

So the note says what to do and why it matters:

```
MAINTAINERS: this asserts against the BUILT bundle, so `dist-wc/index.js`
has to be current or the result describes an old build rather than your
change. `npm run build` (which runs build:wc) before running this locally.
CI is safe -- playwright.config.ts builds as part of webServer -- but a
local run reusing a server does not rebuild.
```

```
lint        exit 0
check       exit 0
build       exit 0
playwright  1 passed
```

Amended into the single commit.

murdore:
## Documented at the new head

You are right — `docs/Checkbox.md` still told web-component consumers to set `attributes`, which is the exact property this PR removes from the element. Following that doc would now silently do nothing.

The Web Component section now carries a rename table covering both, since `ariaLabel` had the same problem and was never documented either:

| Component prop | Custom-element property | Why |
| --- | --- | --- |
| `ariaLabel` | `aria-label` attribute, or `checkboxAriaLabel` | ARIA-reflected accessor on `Element` |
| `attributes` | `checkboxAttributes` (script only) | `Element`'s live `NamedNodeMap` — declaring it broke connect |

The Svelte prop table is unchanged: the component prop is still `attributes`, and only the custom element renames it.

```
check   exit 0
lint    exit 0
build   exit 0
vitest  1021 passed
```

Tara-ag (approved):
Reviewed file-by-file. The fix is correct and narrowly scoped:

- `Checkbox.wc.svelte` renames the colliding `attributes` prop to `checkboxAttributes` and forwards it to the inner component, restoring `Element.prototype.attributes` and allowing the element to connect.
- `scripts/wc-parity/prop-parity.ts` adds `attributes` to `HOST_RESERVED_PROPS` so the parity guard catches future collisions.
- `docs/Checkbox.md` documents the renamed custom-element property (addressing the prior review comment).
- `tests/wc-attributes-collision.spec.ts` adds a focused regression test that injects the built bundle and asserts the native accessor stays iterable and the shadow root initializes.

The Svelte component API (`attributes` prop) and public barrel exports are unchanged. Because the web-component `attributes` prop was non-functional in every published release (the element threw on connect), renaming it is a bug fix rather than a breaking change. No security, accessibility, SSR, or reactivity issues introduced.

PR #614, opened 2026-09-12
fix(migrate): stop the doc rename treating a controller option as a component prop

murdore:
`configCallbackNames` exists to mark an `on*` callback **ambiguous** when it is declared somewhere other than a component's own props, so a bare prose mention is left alone instead of being rewritten as a deprecated prop spelling.

It only ever read `properties.ts`.

That misses an entire category: a headless controller's options live in `types.ts`.

```
onError declared as a component prop   →  Input, InputButton, Table   (properties.ts)
onError declared as a controller option →  SpeechToText, Chat          (types.ts)  ← never scanned
```

So `onError` was never marked ambiguous, and every bare mention of it anywhere under `docs/` was rewritten as though it were the deprecated prop.

## The cost was real, not theoretical

`SpeechSynthesisOptions` spells its error callback **`onerror`** — inconsistent with the sibling controller it mirrors (`SpeechToTextOptions.onError`) and with `onSpeakingChange` / `onDiagnostic` sitting beside it in the same type. It carries a comment saying it is spelled that way *"to match what scripts/migrate/rename-doc-usages.ts forces docs/SpeechSynthesis.md to say."*

An API was bent to fit a lint rule that was wrong. Fixing the rule is what unblocks straightening the API — see #604, where I attempted the rename, hit this guard, and reverted rather than smuggle a tooling change into a feature PR.

## Nothing else changes

The design already distinguishes the cases correctly; only the input to it was incomplete.

- **Tag rewriting** is untouched — position already disambiguates `<Input onError={…}>`.
- **A component's own doc** still rewrites via `ownPairs`, so `docs/Input.md` and `docs/Table.md` keep being corrected.
- Only **bare prose in an unrelated doc** stops being rewritten, which is exactly the bug.

## Verification

Red-green, by reverting the scan to `properties.ts` only:

```
properties.ts only  →  × treats a headless controller option as ambiguous, not as a component prop
                       AssertionError: expected false to be true
both files          →  9 passed
```

```
check   exit 0
lint    exit 0
build   exit 0
vitest  1022 passed
```

The existing `never recommend a spelling the library deprecates` test still passes, which is the guard against this change loosening anything it should not.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved documentation migration accuracy for callback options, preventing ambiguous names such as `onError` from being rewritten incorrectly.
  * Ensured callback names are evaluated consistently across controller configuration and type definitions.

### DIFF
--- a/scripts/migrate/rename-doc-usages.ts
+++ b/scripts/migrate/rename-doc-usages.ts
@@ -33,30 +33,46 @@ const escapeRegExp = (literal: string): string => literal.replace(/[.*+?^${}()|[
 
 /**
  * Every `on*` callback declared somewhere other than a component's own props
- * — a config object's key, or a nested member — across `src/lib`. Ambiguous
- * in prose, see rule 3.
+ * — a config object's key, a nested member, or a headless controller's options
+ * bag — across `src/lib`. Ambiguous in prose, see rule 3.
+ *
+ * Both `properties.ts` and `types.ts` are read. `properties.ts` alone missed an
+ * entire category: a controller's options live in `types.ts`
+ * (`SpeechToTextOptions.onError`, `ChatOptions.onError`), so those names were
+ * never marked ambiguous and every bare prose mention of them anywhere in
+ * `docs/` was rewritten as though it were a deprecated component prop. That is
+ * not theoretical — it is why `SpeechSynthesisOptions` had to spell its error
+ * callback `onerror`, inconsistently with its own sibling, purely to get past
+ * this check.
  */
-function configCallbackNames(root: string): ReadonlySet<string> {
+export function configCallbackNames(root: string): ReadonlySet<string> {
   const names = new Set<string>();
   const lib = join(root, 'src', 'lib');
   for (const entry of readdirSync(lib)) {
-    const file = join(lib, entry, 'properties.ts');
-    if (!existsSync(file)) {
-      continue;
-    }
-    let owner = '';
-    for (const line of readFileSync(file, 'utf8').split('\n')) {
-      const typeLine = /^(?:export )?type (\w+)\b/.exec(line);
-      if (typeLine !== null) {
-        owner = typeLine[1];
+    for (const source of ['properties.ts', 'types.ts']) {
+      const file = join(lib, entry, source);
+      if (!existsSync(file)) {
+        continue;
       }
-      const declaration = /^(\s*)(on[A-Za-z]+)\??:/.exec(line);
-      if (declaration !== null && (!owner.endsWith('Properties') || declaration[1].length !== 2)) {
-        names.add(declaration[2]);
-      }
+      collectCallbackNames(readFileSync(file, 'utf8'), names);
     }
   }
   return names;
+}
+
+/** Shared by both source files; `owner` resets per file, never across them. */
+function collectCallbackNames(text: string, names: Set<string>): void {
+  let owner = '';
+  for (const line of text.split('\n')) {
+    const typeLine = /^(?:export )?type (\w+)\b/.exec(line);
+    if (typeLine !== null) {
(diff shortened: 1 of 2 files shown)


scripts/migrate/rename-doc-usages.ts
```diff
@@ -33,19 +33,38 @@ const escapeRegExp = (literal: string): string => literal.replace(/[.*+?^${}()|[
...
+      }
+      collectCallbackNames(readFileSync(file, 'utf8'), names);
     }
+  }
+  return names;
+}
+
+/** Shared by both source files; `owner` resets per file, never across them. */
+function collectCallbackNames(text: string, names: Set<string>): void {
+  {
```
Tara-ag: 💡 MINOR: Unnecessary block scope around collectCallbackNames loop

The bare `{ ... }` wrapper appears to be a leftover from extracting this loop out of `configCallbackNames`. It serves no purpose and slightly obscures the control flow. Flatten the function body:

```ts
function collectCallbackNames(text: string, names: Set<string>): void {
  let owner = '';
  for (const line of text.split('\n')) {
    const typeLine = /^(?:export )?type (\w+)\b/.exec(line);
    if (typeLine !== null) {
      owner = typeLine[1];
    }
    const declaration = /^(\s*)(on[A-Za-z]+)\??:/.exec(line);
    if (declaration !== null && (!owner.endsWith('Properties') || declaration[1].length !== 2)) {
      names.add(declaration[2]);
    }
  }
}
```

Tara-ag (commented):
Reviewed the two changed files. The fix is correct: scanning both `properties.ts` and `types.ts` properly marks headless-controller option callbacks like `SpeechToTextOptions.onError` and `ChatStreamHandlers.onError` as ambiguous, so bare prose mentions in unrelated docs are no longer rewritten as deprecated component props. The new regression test covers the exact gap described in the PR.

One minor code-quality note inline about a leftover block scope; otherwise this is safe to merge.

scripts/migrate/rename-doc-usages.ts
Tara-ag: 💡 MINOR: Unnecessary block scope around collectCallbackNames loop

The function wraps its implementation in a bare `{ ... }` block. This appears to be a leftover from extracting the inner loop out of `configCallbackNames`; it serves no purpose and slightly obscures the control flow.

Suggested fix: remove the extra braces so the function body is flat:

```ts
function collectCallbackNames(text: string, names: Set<string>): void {
  let owner = '';
  for (const line of text.split('\n')) {
    const typeLine = /^(?:export )?type (\w+)\b/.exec(line);
    if (typeLine !== null) {
      owner = typeLine[1];
    }
    const declaration = /^(\s*)(on[A-Za-z]+)\??:/.exec(line);
    if (declaration !== null && (!owner.endsWith('Properties') || declaration[1].length !== 2)) {
      names.add(declaration[2]);
    }
  }
}
```

murdore:
## Flattened at `a4cccb4`

Correct — the bare `{ … }` was exactly what you guessed: leftover scope from lifting the loop out of `configCallbackNames`. Removed; the function body is flat now.

```
lint    exit 0
check   exit 0
vitest  1022 passed  (76 across scripts/migrate)
```

Both threads describe the same thing, so this covers both.

Tara-ag (approved):
Reviewed the two changed files. The fix correctly expands `configCallbackNames` to scan both `properties.ts` and `types.ts`, which prevents headless-controller options like `SpeechToTextOptions.onError` from being misclassified as unambiguous deprecated props during doc migration. Verified against the actual `SpeechToText/types.ts`, `Chat/types.ts`, and `LEGACY_PAIRS` data. The added regression test covers the exact failure mode. No public API, accessibility, SSR, or security concerns. Approved.

PR #616, opened 2026-09-12
feat(table): let a consumer choose the container's overflow

murdore:
`.table-container` was hard-coded `overflow: hidden`. That default is load
bearing -- it is what clips the first and last cells to the container's
`border-radius`, and without it they square off over the rounded border -- but
it was also the only option, so a consumer wanting a wide table to scroll
sideways in the container had to reach past the component and restyle
`.table-container` from their own stylesheet.

The vertical axis already had a seam: `isTableScrollable` adds `overflow-y:
auto` with `--table-container-height`. The horizontal one had none. This adds
`--table-container-overflow`, defaulting to `hidden`, so every existing table
renders exactly as before and `auto` is now expressible without reaching in.

Found while auditing a consumer's stylesheet against this library's actual API.
That app carries 68 CSS references into 21 Select and Table internals, and this
is the only one of them that turned out to be a real gap -- `.select-dropdown`'s
`z-index`/`width`/`min-width`/`margin-top` all already have documented custom
properties on that exact selector, sticky headers are already a `stickyHeader`
prop, and the header-bar spacing that app was chasing is already
`--table-toolbar-gap`/`-padding`/`-margin-bottom`. The rest of that audit is a
documentation problem, not a missing-API one, and is not in this PR.

One interaction is worth knowing and is in the comment rather than left to be
discovered: the scrollable variant sets `overflow-y`, and per the CSS overflow
spec `visible` on one axis computes to `auto` when the other axis is not
visible. So `visible` does not survive on a scrollable table. `hidden` and
`auto` behave as written.

Verified red-green rather than by the suite going green: with the property
still hard-coded, the opt-in case fails and the default case passes; after the
change both pass. That ordering matters here, because a test that only asserted
the default would have passed against the bug.

Gate on this commit: prettier clean, eslint 0, event-casing 0, svelte-check 0
errors across both tsconfigs, and the new spec 2/2 under Playwright.

### DIFF
--- a/docs/Table.md
+++ b/docs/Table.md
@@ -532,16 +532,17 @@ Override these custom properties to theme the component.
 
 ### Container & Layout
 
-| Variable                     | Default             | CSS Property          | Description                                                                                                |
-| ---------------------------- | ------------------- | --------------------- | ---------------------------------------------------------------------------------------------------------- |
-| `--table-border`             | `1px solid #e5e7eb` | border                | Border of the table container.                                                                             |
-| `--table-border-radius`      | `8px`               | border-radius         | Border radius of the table container.                                                                      |
-| `--table-container-width`    | `100%`              | width                 | Width of the table container.                                                                              |
-| `--table-container-height`   | `143px`             | height                | Height of the scrollable table container (when isTableScrollable).                                         |
-| `--table-width`              | `100%`              | width                 | Width of the table element.                                                                                |
-| `--table-border-collapse`    | `collapse`          | border-collapse       | Border collapse mode of the table.                                                                         |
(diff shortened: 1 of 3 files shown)


Tara-ag:

## 🤖 Yama Review Summary

**Verified findings (1):**

- **MINOR**: New Playwright spec uses a non-existent test fixture — `tests/table-container-overflow.spec.ts:2`

- 💡 **MINOR**: New Playwright spec uses a non-existent test fixture — `tests/table-container-overflow.spec.ts:2`
  The new tests/table-container-overflow.spec.ts imports gotoHydrated from './support/hydrated', but the tests/support directory does not contain a hydrated.ts file. Existing Playwright specs in this repo import gotoHydrated from './support/hydrated.js' (e.g. table-keyed-columns.test.ts). Importing a non-existent module will fail at test collection time.

PR #618, opened 2026-09-13
fix(visual): stop a fractional capture box flipping the baseline by a pixel

murdore:
## The bug

The visual suite was intermittently red on a handful of demo routes, always by exactly one pixel of height, and **in both directions within a single run**:

```
- Expected an image 980px by 5412px, received 980px by 5413px.
- Expected an image 980px by 5413px, received 980px by 5412px.
```

## Why

`toHaveScreenshot` rounds a fractional element box when it captures, and which way it rounds depends on the element's sub-pixel offset at that moment. `main.content` lays out to a fractional height on a fifth of the demo routes — measured **inside the pinned container**:

```
badge                800.00000   (integral — stable, and untouched by this change)
bar-chart           7947.40625
dual-axis-bar-chart 4243.03125
select              5374.09375
```

Two things that cannot fix it, worth knowing before someone tries:

- **`fitViewportToContent` can't.** It settles on `scrollHeight`, which is *already* rounded, so it never constrains the element's own height.
- **A pixel tolerance can't.** `toHaveScreenshot` rejects a size mismatch *before* comparing pixels, so the run never reaches `maxDiffPixels`.

## The fix

Add the sub-pixel remainder to the existing bottom padding, then **refit the viewport** to the padded content. Both halves are load-bearing.

**Padding, not height.** Two earlier attempts show why:

| attempt | result |
|---|---|
| `box-sizing: border-box`, so the measured box could be assigned directly | Reinterpreted `.content`'s `max-width:900px` against the border box. Every capture narrowed 980px → 900px. **All 94 routes failed.** |
| explicit `height` (padding subtracted, so width was safe) | Stopped the block growing for descendants' trailing margins. Pages moved up to 11px in **both** directions — `accordion` 1613→1602, `tool-call-log` 969→975. **17 failed**, and it silently changed what the screenshots *contained*. |

**Refit after padding.** The viewport is sized from `scrollHeight` *before* the padding exists, so the padded element ends up a pixel taller than the viewport. Playwright scrolls a target taller than the viewport and stitches the capture — the scroll-dependent screenshot this file already documents as the original flake ("*exactly those kept scrolling and exactly those stayed flaky*"). Without the refit the capture is silently clipped back to the viewport instead. Measured:

```
bar-chart   el 7947.40625 -> 7948   vp 7947 -> 7948   fits=true
select      el 5374.09375 -> 5375   vp 5374 -> 5375   fits=true
badge       el 800        -> 800    vp 800  -> 800    (untouched)
```

**Converge or throw.** Padding and viewport disturb each other — growing the element leaves the viewport short, and resizing relayouts the page and can produce a fresh fraction. Only a pass that changes neither proves they agree, and that is the only way out of the loop; exhausting the budget throws. Giving up quietly would be *indistinguishable from convergence* while leaving the route in the stitching state this exists to prevent, and every baseline regenerated for it would look settled and not be. `fitViewportToContent` throws on the same kind of non-convergence. Verified by forcing it:

```
Error: capture box never settled for "badge": 0: document 800px, viewport 800 -> 800; ... 5: ...
```

**Viewport-relative routes are excluded.** `fitViewportToContent` now reports which branch it exited on, and the pin is skipped for routes that took the escape path — `brand-loader` and `status`. Those are deliberately pinned to a viewport smaller than their content; growing the viewport to match is exactly what that path refuses to do, and their height is a function of the viewport, so pinning then resizing would feed itself.

## Verification

**Three consecutive full-suite runs at 94/94.** On the same machine:

| version | full-suite result |
|---|---|
| before this change | 4 failed, oscillating both directions |
| `box-sizing` attempt | 94 failed |
| height-assignment attempt | 17 failed |
| padding without refit | appeared green, but captures were silently clipped |
| padding + refit, silent give-up | green, but non-convergence would have been invisible |
| **this change** | **94 passed, ×3 consecutive, 0 routes failing to converge** |

## Regenerated baselines

15, to the size that is now unambiguous. `pill.png` and `sheet.png` needed regenerating under the no-refit version and no longer do — with the full padded element captured rather than a clipped one, they match their existing baselines exactly.

## Why this took four attempts

Each earlier version produced **green full-suite runs** and each was wrong. That is not coincidence: the visual suite compares a capture against a baseline **produced by the same capture path**, so it cannot distinguish *captured correctly* from *captured wrongly but consistently*. The instrument and the thing measured share a dependency, which makes agreement guaranteed rather than informative.

Concretely: the padding-without-refit version passed three consecutive full-suite runs while every capture was being silently clipped back to the viewport. The suite had no way to say so.

Three of the four defects were found by **reading the patch**, not by running the suite — the `box-sizing` width regression, the height-assignment content shifts, the missing refit, and the silent non-convergence. Worth remembering the next time a visual change is justified by a passing visual suite.

## Findings worth keeping

- **A single targeted run proved nothing before this change.** Isolated, `bar-chart` produced 7947 on five consecutive runs; inside the full suite it produced 7948 — the element at 7948 against a viewport at 7947, with context deciding clip-vs-stitch. After this change a targeted run is trustworthy again.
- Corroboration from the parity fold: when 45 baselines failed there, arm64 and amd64 produced **byte-identical size deltas**. Identical across two renderers rules out rasterisation and leaves deterministic layout arithmetic, which is what fractional-height rounding predicts.


tests/visual/components.visual.ts
```diff
@@ -360,6 +360,52 @@ async function prepare(page: Page, slug: string): Promise<void> {
...
+      return;
+    }
+    const height = main.getBoundingClientRect().height;
+    const remainder = Math.ceil(height) - height;
+    if (remainder <= 0) {
+      return;
+    }
+    const paddingBottom = parseFloat(window.getComputedStyle(main).paddingBottom);
+    main.style.paddingBottom = `${paddingBottom + remainder}px`;
+  });
```
Tara-ag: 💬 SUGGESTION: After adjusting paddingBottom, re-fit the viewport so the screenshot is not stitched from a scroll. Adding bottom padding increases `main.content`'s border-box height, which can raise `document.scrollHeight`. Since `fitViewportToContent` ran before this padding was injected, the viewport may now be 1px smaller than the padded content on routes where the fractional height sits just below an integer threshold, letting Playwright scroll-stitch and reintroduce instability. Call `fitViewportToContent(page)` again after pinning, or compute the new height inside `page.evaluate` and resize the viewport before `toHaveScreenshot`.

Tara-ag (approved):
Approved with one suggestion. The change is narrowly scoped to the visual test harness and addresses a real source of flakiness without touching public component APIs, accessibility, or SSR behavior. Consider re-fitting the viewport after padding adjustment to avoid a 1px scroll-stitch edge case.

tests/visual/components.visual.ts
Tara-ag: 💬 SUGGESTION: pinCaptureHeightToWholePixels should refit viewport after padding adjustment

Adding bottom padding to `main.content` increases the element's border-box height, which can in turn increase document `scrollHeight`. Because `fitViewportToContent` was called before the padding is injected, the viewport may now be slightly smaller than the padded content. On routes where the fractional height sits just below an integer threshold, the viewport could end up 1px short and Playwright would scroll-stitch the screenshot, reintroducing instability.

Call `fitViewportToContent(page)` again after pinning the padding so the viewport matches the newly padded content height, or compute the target viewport size inside `page.evaluate` and call `page.setViewportSize` before `toHaveScreenshot`.

Tara-ag (approved):
Approved. The change fixes a real visual-regression flake by pinning the capture target's fractional height to whole pixels through bottom padding in the test harness only. It does not touch public component APIs, accessibility, SSR behavior, or runtime component code. One suggestion left inline about re-fitting the viewport after padding adjustment to avoid a potential 1px scroll-stitch edge case.

tests/visual/components.visual.ts
```diff
@@ -360,6 +360,52 @@ async function prepare(page: Page, slug: string): Promise<void> {
       }
     `
   });
+
+  await pinCaptureHeightToWholePixels(page);
```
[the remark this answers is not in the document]
murdore: Confirmed and fixed — you were right, and this was the difference between a fix and a fix that only looked green.

**Measured before changing anything.** I instrumented the element box and the viewport on either side of the pin:

```
bar-chart   el 7947.40625 -> 7948   vp 7947 -> 7948
select      el 5374.09375 -> 5375   vp 5374 -> 5375
badge       el 800        -> 800    vp 800  -> 800   (untouched)
```

The viewport really was a pixel short of the padded element, exactly as described. And it explains an anomaly I had observed and waved away: `bar-chart` captured 7948 inside the full suite but 7947 across five isolated runs. I had attributed that to "suite context perturbing the sub-pixel offset", which restated the observation rather than explaining it. The real mechanism is element-at-7948 against viewport-at-7947, with context deciding whether the capture clipped or stitched.

**So my earlier evidence was worth less than I claimed.** Three consecutive 94/94 runs looked like proof; the captures were being silently clipped back to the viewport. A passing run could not distinguish "correct" from "clipped".

**The fix** refits the viewport after the padding, and the two are settled together in a bounded loop, because each can disturb the other — growing the element leaves the viewport short, and resizing the viewport relayouts the page and can produce a fresh fraction.

**One refinement beyond the finding.** A blanket refit would have broken the `fitViewportToContent` escape path for viewport-relative routes. Those deliberately pin to `MIN_VIEWPORT_HEIGHT` and return early, so growing the viewport to match their content is precisely what that path refuses to do — it would capture them at a different size than their neighbours, which is a silent change to picture content rather than a size mismatch. And since their height is a function of the viewport, pin-then-resize would feed itself. `fitViewportToContent` now reports which branch it exited on and the pin is skipped for those two routes (`brand-loader`, `status`).

**Verification:** three consecutive full-suite runs at 94/94, with 15 baselines regenerated to the now-unambiguous size. Notably `pill.png` and `sheet.png` needed regenerating under the no-refit version and no longer do — capturing the full padded element instead of a clipped one puts them back on their existing baselines, which is a small independent sign the refit is doing the right thing.

tests/visual/components.visual.ts
murdore: Confirmed and fixed — you were right, and this was the difference between a fix and a fix that only looked green.

**Measured before changing anything.** I instrumented the element box and the viewport on either side of the pin:

```
bar-chart   el 7947.40625 -> 7948   vp 7947 -> 7948
select      el 5374.09375 -> 5375   vp 5374 -> 5375
badge       el 800        -> 800    vp 800  -> 800   (untouched)
```

The viewport really was a pixel short of the padded element, exactly as described. And it explains an anomaly I had observed and waved away: `bar-chart` captured 7948 inside the full suite but 7947 across five isolated runs. I had attributed that to "suite context perturbing the sub-pixel offset", which restated the observation rather than explaining it. The real mechanism is element-at-7948 against viewport-at-7947, with context deciding whether the capture clipped or stitched.

**So my earlier evidence was worth less than I claimed.** Three consecutive 94/94 runs looked like proof; the captures were being silently clipped back to the viewport. A passing run could not distinguish "correct" from "clipped".

**The fix** refits the viewport after the padding, and the two are settled together in a bounded loop, because each can disturb the other — growing the element leaves the viewport short, and resizing the viewport relayouts the page and can produce a fresh fraction.

**One refinement beyond the finding.** A blanket refit would have broken the `fitViewportToContent` escape path for viewport-relative routes. Those deliberately pin to `MIN_VIEWPORT_HEIGHT` and return early, so growing the viewport to match their content is precisely what that path refuses to do — it would capture them at a different size than their neighbours, which is a silent change to picture content rather than a size mismatch. And since their height is a function of the viewport, pin-then-resize would feed itself. `fitViewportToContent` now reports which branch it exited on and the pin is skipped for those two routes (`brand-loader`, `status`).

**Verification:** three consecutive full-suite runs at 94/94, with 15 baselines regenerated to the now-unambiguous size. Notably `pill.png` and `sheet.png` needed regenerating under the no-refit version and no longer do — capturing the full padded element instead of a clipped one puts them back on their existing baselines, which is a small independent sign the refit is doing the right thing.

murdore: Confirmed and fixed — you were right, and this was the difference between a fix and a fix that only looked green.

**Measured before changing anything.** I instrumented the element box and the viewport on either side of the pin:

```
bar-chart   el 7947.40625 -> 7948   vp 7947 -> 7948
select      el 5374.09375 -> 5375   vp 5374 -> 5375
badge       el 800        -> 800    vp 800  -> 800   (untouched)
```

The viewport really was a pixel short of the padded element, exactly as described. And it explains an anomaly I had observed and waved away: `bar-chart` captured 7948 inside the full suite but 7947 across five isolated runs. I had attributed that to "suite context perturbing the sub-pixel offset", which restated the observation rather than explaining it. The real mechanism is element-at-7948 against viewport-at-7947, with context deciding whether the capture clipped or stitched.

**So my earlier evidence was worth less than I claimed.** Three consecutive 94/94 runs looked like proof; the captures were being silently clipped back to the viewport. A passing run could not distinguish "correct" from "clipped".

**The fix** refits the viewport after the padding, and the two are settled together in a bounded loop, because each can disturb the other — growing the element leaves the viewport short, and resizing the viewport relayouts the page and can produce a fresh fraction.

**One refinement beyond the finding.** A blanket refit would have broken the `fitViewportToContent` escape path for viewport-relative routes. Those deliberately pin to `MIN_VIEWPORT_HEIGHT` and return early, so growing the viewport to match their content is precisely what that path refuses to do — it would capture them at a different size than their neighbours, which is a silent change to picture content rather than a size mismatch. And since their height is a function of the viewport, pin-then-resize would feed itself. `fitViewportToContent` now reports which branch it exited on and the pin is skipped for those two routes (`brand-loader`, `status`).

**Verification:** three consecutive full-suite runs at 94/94, with 15 baselines regenerated to the now-unambiguous size. Notably `pill.png` and `sheet.png` needed regenerating under the no-refit version and no longer do — capturing the full padded element instead of a clipped one puts them back on their existing baselines, which is a small independent sign the refit is doing the right thing.

Tara-ag:
💡 MINOR: `docs/VISUAL_TESTING.md` is now stale. It still says "92 baselines" and lists `task-list` as excluded, but this PR's description refers to 94 routes and `task-list` is now baselined. It also does not document the new `fitted` / `viewport-pinned` branch or the `pinCaptureHeightToWholePixels` sub-pixel pinning added here. Please update the route count, remove `task-list` from the excluded table, and add a short section on the sub-pixel pin behavior.

Tara-ag (commented):
Reviewed the visual test fix. One documentation staleness issue noted inline.

Tara-ag (commented):
Reviewed file by file. One minor documentation staleness issue found.

PR #619, opened 2026-09-14
fix(a11y): bring default label and banner text to AA, and announce the orb's state

murdore:
## What is the problem?

Two accessibility defects, both found while auditing this session's own closure and both verified against the merged tree rather than taken on a reviewer's word.

**1. `#637c95` fails WCAG AA in the default configuration.** It is the fallback for `--input-label-msg-text-color` (Input, InputButton) and `--banner-color` (Banner) — what a consumer who sets no token actually renders.

| where | on | ratio | AA (4.5:1) |
|---|---|---|---|
| Input / InputButton label | `#ffffff` | **4.33:1** | FAIL |
| Banner text | `#f0f4f8` (Banner's own default background) | **3.92:1** | FAIL |

Banner is the worse case, because it sits on its own tint rather than white — measuring against white alone would have understated it.

**2. VoiceOrb announced nothing.** The canvas is `aria-hidden`, and a review thread on #602 called that a MAJOR defect requiring a keyboard alternative. That half is refuted below. What was genuinely unreachable is the *variant*: the orb exists to say a microphone is open, and no assistive technology could learn that. `docs/VoiceOrb.md` already told callers to build their own live region; there was no built-in way.

## Solution

**Contrast.** `#637c95` → `#4d6174`: same hue and saturation (hsl 210, 20%), darkened until it clears AA on both backdrops with real headroom.

| | on `#ffffff` | on `#f0f4f8` |
|---|---|---|
| before | 4.33:1 | 3.92:1 |
| after | **6.41:1** | **5.80:1** |

The headroom is deliberate and matches house precedent: the earlier error-colour fix took `#fa1405` → `#c5120a` at 6.06:1 rather than stopping at the first value over the line. `#607890` would have passed at 4.58:1; that is too close to the boundary to survive a future tweak.

**VoiceOrb.** New optional `statusLabels` — the caller's words per variant, rendering a visually hidden `aria-live="polite"` region on the Carousel's existing pattern. Opt-in, no default, and a variant left unnamed announces nothing. The same animation means "the assistant is listening" on one surface and "recording" on another; a component that guessed would announce the wrong thing confidently.

### Refuting the other half of that review thread

The thread asked for `role="img"` + `aria-label`, or a keyboard alternative. Neither is right:

- **`role="img"` with a static label is worse**, not better, for a *changing* state indicator — it announces once and never updates. A live region is what conveys a state change.
- **Pointer repulsion needs no keyboard equivalent.** It pushes particles aside and accomplishes nothing else. A focus stop for it would be a tab stop that leads nowhere; there is no functionality a keyboard user is locked out of. That is precisely why the canvas can stay out of the accessibility tree, and it is now documented as a decision rather than left as an omission.

## Proof of testing

| gate | result |
|---|---|
| `npm run check` | exit 0 |
| `npm run lint` | exit 0 |
| `npm run build` | exit 0 |
| `npx vitest run` | exit 0 — **1279 passed across 68 files** (up 9) |
| `scripts/visual-test.sh` | **94/94 passed**, no baseline changed |

Red–green was watched on both halves rather than asserted:

- `a11y-contrast.test.ts` failed at exactly `4.33`, `4.33` and `3.919…` before the colour change, and passes after.
- The three new VoiceOrb live-region specs failed with `expected undefined to be 'polite'` / `'Microphone idle'` / `''` before the component change, and pass after.

`a11y-contrast.test.ts` reads the literal out of the `var(--token, #hex)` fallback, so it measures what ships rather than a palette constant, and it carries a **negative control** asserting the old `#637c95` still fails on both backdrops — a broken ratio calculation cannot make the suite vacuously green.

### Why "no baseline changed" is not the instrument agreeing with itself

A colour change that moves no pixel in a 94-shot visual suite deserves suspicion, so it was tested rather than trusted. **Injecting a bright red banner colour fails the suite with 3381 differing pixels**, which proves the instrument can see this element.

The reason this change passes is a real property of the harness worth recording: `maxDiffPixelRatio: 0` reads as "exact match" — the comment in `components.visual.ts` says exactly that — but Playwright's *per-pixel* `threshold` defaults to `0.2`, allowing a YIQ delta up to **1408.6** before a pixel counts as different at all.

| comparison | YIQ delta | counted? |
|---|---|---|
| `#637c95` → `#4d6174` (this change) | 354.6 | no |
| `#637c95` → `#ff0000` (injected control) | 10175.9 | yes |

So the suite genuinely cannot see a contrast fix of this size, and no baseline regeneration is required. **This is not fixed here** — tightening `threshold` could destabilise all 94 baselines and does not belong in an accessibility PR. Filed separately.

## Areas of impact

Default token fallbacks only. Any consumer already setting `--input-label-msg-text-color` or `--banner-color` is unaffected; those who set nothing get a darker, compliant colour. `statusLabels` is additive and optional — omitting it renders no live region and leaves today's DOM byte-identical.

Docs updated for all four components; `sui-voice-orb` declares the new prop.

## Summary by CodeRabbit

- **New Features**
  - Added optional Voice Orb status announcements for assistive technologies, including live updates as its state changes.
  - Exposed status labels through the Voice Orb custom element using the `status-labels` attribute.
  - Updated the Voice Orb example with “Microphone off” and “Listening” announcements.

- **Accessibility**
  - Darkened default banner and input label text colors to improve contrast.
  - Added accessibility documentation for Voice Orb status announcements.

- **Tests**
  - Added coverage for Voice Orb announcements and WCAG AA text contrast.


Tara-ag (approved):
Approved. The contrast fix is a safe default-value change (CSS variable names are unchanged, so themed consumers are unaffected). The new `statusLabels` prop is additive, typed, documented, tested, and mirrored in the web-component wrapper. No blocking public-API, accessibility, SSR, or security issues found.

murdore:
## Recorded proof of testing

Asked for after the fact: this PR merged with gate exit codes but no recording. This is a single continuous Playwright capture of the running demo site at `30ddf475` (4.28.1), 35s, 1280x800 — not stills, and not a screenshot turned into a video.

https://github.com/user-attachments/assets/2696fce8-27e5-40ba-8bcf-aa3308d2a6ed

### What it shows, in order

| t | state |
|---|---|
| ~4s | VoiceOrb `idle` — hidden live region reads `"Microphone off"` |
| ~7.5s | toggled to `listening` — region reads `"Listening"` |
| ~12.5s | back to `idle` — region reads `"Microphone off"` again |
| ~16s | Input — label and `infoMessage`, computed colour `rgb(77, 97, 116)` |
| ~19.5s | Input — invalid entry, so the **error** message renders in the same colour |
| ~24s | InputButton — label + message |
| ~29s | Banner — computed `rgb(77, 97, 116)` **on** `rgb(240, 244, 248)` |

`rgb(77, 97, 116)` is `#4d6174` and `rgb(240, 244, 248)` is `#f0f4f8` — read live off the DOM during the recording, not asserted from source. That is the 5.80:1 pair this PR introduced, measured in the browser rather than in a test.

### Why the VoiceOrb half needs an overlay

`.voice-orb-status` is `clip-path: inset(50%)` — a screen-reader-only live region. **A plain screen recording cannot show it**, and a video that appeared to would be proving nothing. The panel in the top right is injected by the test and mirrors that element's live `textContent` through a `MutationObserver`, with its `clip-path` printed underneath so it is clear what is being read. The overlay is not part of the component; what it demonstrates is that the component's live region carries the text and updates on variant change.

### How the recording itself was checked

The run asserts rather than just films — `toHaveText('Microphone off')` → toggle → `'Listening'` → toggle back — so a broken announcement fails the run instead of producing a video of the wrong thing.

That was not sufficient on its own. **The first recording passed every assertion and still did not contain the `listening` state**: Playwright stops recording when the context closes and drops the final seconds, and the VoiceOrb segment was last. Frame extraction at t=22.0 of a 22.6s file showed `idle`. It was discarded, the order reversed so the critical states record first, and a 6s tail buffer added to absorb the truncation.

This file was then verified frame by frame with `ffmpeg` before being attached — each state above extracted and inspected, confirming it is legible and not blank.

PR #624, opened 2026-09-14
docs(select): state both halves of the automatic flip condition

murdore:
## What is the problem?

`docs/Select.md` described the automatic upward flip with only half its condition, in two places:

- `placement` row — *"portaled ones flip up only when they cannot fit below"*
- `usePortal` row — *"flips above it when there's no room below"*

Both describe a panel that flips upward whenever it doesn't fit below. That is not what the component does, and the difference is visible: a panel with no room in either direction would, by the documented rule, flip into the tighter space above.

## The actual condition

```ts
// src/lib/Select/dropdownPosition.ts:73
const fitsBelow = !(dropdown.height > 0 && spaceBelow < dropdown.height + gap);
const flippedUp =
  vertical === 'auto' ? !fitsBelow && spaceAbove > spaceBelow : vertical === 'top';
```

An automatic flip needs **both** halves: it cannot fit below, **and** `spaceAbove > spaceBelow`. A panel that fits nowhere stays below rather than flipping into a tighter space. An explicit `placement` of `'top-*'` is unconditional and unaffected — that path does not consult available space at all.

## Changes

Both rows now state both halves. No code change; the behaviour was already correct and is covered by `src/lib/Select/placement.test.ts`.

| gate | result |
|---|---|
| `npm run lint` | exit 0 |

## Provenance

Raised in review on #617 (`docs/Select.md:113`) and left unresolved when that PR merged. Verified against the implementation before accepting it, rather than taken on the reviewer's word — the reviewer was right.

https://github.com/juspay/svelte-ui-components/pull/617


Tara-ag (approved):
Documentation correction verified against `src/lib/Select/dropdownPosition.ts:73-75`. The revised wording accurately reflects the automatic flip condition (`!fitsBelow && spaceAbove > spaceBelow`) and matches existing test coverage in `src/lib/Select/placement.test.ts`. No code changes; safe to merge.

Issue #622, opened 2026-09-14
The visual suite's "exact match" is not exact: a per-pixel threshold of 0.2 hides colour-only regressions


murdore:
## The claim the harness makes about itself

`tests/visual/components.visual.ts` passes `maxDiffPixelRatio: 0` for every route not in `TOLERANCES`, and the comment beside it reads:

> *"Exact match is the default, and stays the default. A blanket tolerance would be a slow leak: it hides the sub-threshold drift that accumulates into a real regression…"*

**It is not an exact match.** `toHaveScreenshot` has a separate *per-pixel* `threshold` — the perceptual YIQ distance at which a pixel is counted as different at all — which Playwright defaults to `0.2` and which this suite never sets. `maxDiffPixelRatio: 0` then means "zero pixels differing by **more than** that threshold", not "zero pixels differing".

At `threshold: 0.2` the per-pixel allowance is `35215 × 0.2² = 1408.6` in pixelmatch's YIQ metric. That is a large amount of colour drift.

## Measured, not inferred

Found while changing a default text colour from `#637c95` to `#4d6174` (a WCAG AA contrast fix, PR #619). The full suite passed **94/94 with no baseline changed**, which should not happen when text on three demo pages changes colour.

| comparison | YIQ delta | counted as different at `threshold: 0.2`? |
|---|---|---|
| `#637c95` → `#4d6174` (the real change) | 354.6 | **no** |
| `#637c95` → `#ff0000` (injected control) | 10175.9 | yes — 3381 pixels, suite fails |

The red injection is the control that proves the instrument can see that element at all, so this is a threshold effect and not a broken harness or a stale build.

## Why it matters

The suite's stated purpose is catching "the sub-threshold drift that accumulates into a real regression". A colour-only change large enough to move text across the AA boundary is exactly that kind of drift, and it is invisible. A theme or token regression that shifts every text colour slightly would ship green.

## What it needs

Not a blind `threshold: 0`. The container pins the renderer, so near-exactness is plausible, but antialiasing noise across runs is the reason the default is not zero, and 94 baselines are at stake. Suggested order:

1. Measure first: re-run the suite at `threshold: 0.05` and `0.01` and record how many routes become unstable across repeat runs.
2. Pick the lowest stable value, set it **explicitly** so the number is visible rather than inherited.
3. Either way, **correct the comment** — a harness that describes itself as stricter than it is, is worse than one with a documented tolerance.

[...]

PR #625, opened 2026-09-14
test(visual): refresh the two baselines that still assert the pre-AA colour

murdore:
## What is the problem?

`release` carries two visual baselines that assert the accessibility failure #619 fixed.

#619 changed the default label/banner colour from `#637c95` (4.33:1 on white, 3.92:1 on Banner's own tint — both below AA) to `#4d6174`. The visual suite passed **94/94 and rewrote nothing**, so the committed PNGs still encode the old colour:

| baseline | old `#637c95` | new `#4d6174` | differing px vs the tree |
|---|---:|---:|---:|
| `banner.png` | 292 | 0 | 8,228 |
| `input.png` | 186 | 0 | 8,351 |

The gate is green on both.

## Why the suite did not catch it

`maxDiffPixelRatio: 0` reads as an exact match — the comment in `components.visual.ts` says so — but it bounds *how many* pixels may exceed a per-pixel tolerance, not whether a pixel differs. That tolerance is `toHaveScreenshot`'s `threshold`, which Playwright defaults to `0.2` and this suite never sets. A pixel is not counted as different until its YIQ delta exceeds `35215 × 0.2² = 1408.6`.

| comparison | delta | counted? |
|---|---:|---|
| `#637c95` → `#4d6174` (this change) | 354.6 | no |
| `#637c95` → `#ff0000` (injected control) | 10175.9 | yes — 3381 px, suite fails |

The red injection is the control proving the instrument can see this element at all.

## `--update-snapshots` does not repair it

Verified non-vacuously — an earlier attempt of mine was void (`-g "› …"` → `No tests found`, exit 1; the tell is exit 1 with no "N passed"):

```
input.png md5 before: 8656e1680a8d9c04546ca2ec6f70561e
scripts/visual-test.sh --update-snapshots -g "input renders as baselined"
  exit=0
  Running 5 tests using 1 worker
  5 passed (1.0m)
input.png md5 after:  8656e1680a8d9c04546ca2ec6f70561e
git sees changed: 0 file(s)
```

Five tests ran, passed, and wrote nothing against a measured 354.6 drift. The update path applies the same threshold as the assertion, so running both tells you nothing the first did not. Independently reproduced by the session working on #598 (98 passed, 0 written, four provably stale baselines).

**Deleting the PNG works**, because a missing baseline is not a comparison.

## Changes

`banner.png` and `input.png` regenerated by deletion inside the pinned container.

**Only colour moved** — checked, rather than assumed:

| | old colour px | new colour px | dimensions |
|---|---:|---:|---|
| `banner` | 292 → 0 | 0 → 289 | 980×800, unchanged |
| `input` | 186 → 0 | 0 → 186 | 980×4561, unchanged |

If those counts had shifted materially, the regeneration would have been carrying something other than the colour change.

## Proof of testing

| gate | result |
|---|---|
| `scripts/visual-test.sh` | **94/94 passed** on the regenerated baselines |
| regeneration | 6 tests ran, 2 files written |

## What this deliberately does NOT do

**It does not change the threshold.** That looked like the obvious fix until it was measured. Across two full container captures with no change between them, the suite's own capture noise reaches:

| route | max per-pixel delta | >0.05 | >0.1 | >0.2 |
|---|---:|---:|---:|---:|
| `hitl` | **1314.3** | 394 | 16 | 0 |
| `theme-switcher` | 520.2 | 5 | 3 | 0 |

The noise floor is 1314.3 against a limit of 1408.6 — **`0.2` is very nearly calibrated to this renderer's real non-determinism**, clearing it by about 7%. Dropping to `0.1` would put 16 `hitl` pixels over the line on an unchanged tree; `0.05` would put 394 over.

So tightening trades colour detection for a flaky suite. The conclusion is that the visual suite structurally cannot guard colour at this renderer's noise level, and colour belongs in unit assertions instead — `a11y-contrast.test.ts` from #619 is the working example. That, plus correcting the misleading comment and investigating `hitl`'s 7% margin, is https://github.com/juspay/svelte-ui-components/issues/622.

`input-button` is untouched: its demo renders no label, so it shows zero drift and #619's third component is not visually exercised at all.


Tara-ag (approved):
Reviewed. No actionable issues found in the changed files.

murdore:
## Before/after proof for the two regenerated baselines

Both panels below are the **committed PNG blobs**, not a fresh render. `BEFORE` is the blob at `9480135e^`, `AFTER` is the blob at this PR's own merge commit `9480135e` — confirmed with `git hash-object`:

| baseline | before = `9480135e^` | after = `9480135e` |
|---|---|---|
| `banner.png` | `210e77db` | `c1450e4e` |
| `input.png` | `2d47c847` | `d58c9799` |

Each crop is placed automatically where the `#637c95 → #4d6174` transition is densest, and shown at 3× **nearest-neighbour** so no pixel is interpolated or resampled.

### Banner

pr625-banner-proof

### Input

pr625-input-proof

### What the pixels say

Counting a pixel as changed at YIQ delta > 10 (well under the suite's own 1408.6 threshold, which is the whole problem this PR describes):

| baseline | changed px, whole image | in the crop shown | headline transition |
|---|---|---|---|
| `banner.png` (980×800) | 6,009 | 4,067 | `#637c95` → `#4d6174`, delta 354 |
| `input.png` (980×4561) | 6,082 | 1,323 | `#637c95` → `#4d6174`, delta 354 |

`input.png` is 4,561 px tall and its 186 pixels of the old colour sit in small labels scattered down the entire page — the densest 430×164 window holds just 36 of them. So the Input crop shows a genuine instance of the change, not all of it, and the caption inside the image states that count rather than implying otherwise.

### Attribution, stated precisely

The colour itself changed in #619 (`4d9f83f5`). **This PR changed no source at all** — it only moved two baselines off the pre-AA colour they were still asserting green. The `AFTER` panels are what the visual suite has compared against from `9480135e` onward.

PR #629, opened 2026-09-14
test(input): pin why assigning to a $derived on focus out is not a defect

murdore:
## Why

`validationState` in `Input.svelte` is a `$derived.by` (`:98`), and `_onFocusOut` assigns to it (`:295`):

```ts
function _onFocusOut(event: FocusEvent) {
  if (validationState === 'InProgress' && value.length > 0) {
    validationState = 'Invalid';
  }
  ...
}
```

In Svelte 5.25+ that is a legal override, discarded the next time any dependency of the derived changes. Read on its own it looks like a latent bug — blur forces `'Invalid'`, then an unrelated parent re-render silently reverts it — and it was reported to me as one.

**It isn't, and the reason isn't obvious from the assignment.** The derived already carries the same rule the override applies (`:106-113`):

```ts
if (valueValidation === 'InProgress' && value.length > 0 &&
    inputElement !== null && inputElement !== document.activeElement) {
  return 'Invalid';
}
```

So a recompute re-derives `'Invalid'` rather than reverting. And the assignment is not redundant either: `document.activeElement` is not reactive, so the derived has no way to notice focus leaving. The assignment exists to push the one recomputation the derived cannot observe for itself.

That is a non-obvious pairing with nothing recording it, which makes it exactly the kind of code someone later "cleans up" — deleting the assignment as a stray mutation of a derived, or deleting the `activeElement` clause as dead because the handler already covers it. Either one silently breaks blur validation, and nothing currently fails.

## What this adds

Four tests in `src/lib/Input/validation-override.svelte.test.ts`. No production code changes.

| test | pins |
|---|---|
| reaches `InProgress` while focused | that the harness gets to the state under test, so the rest are not vacuous |
| forces `Invalid` on focus out | the assignment does something the derived cannot do alone |
| does **not** revert on a later prop change | the claim — a recompute re-derives `Invalid` |
| returns to `InProgress` only when focus comes back | the override is not sticky |

## Verification

`lint` exit 0 · `check` exit 0 · full suite **1284 tests / 69 files passed**.

**Red-green verified.** Removing the derived's `activeElement` rule fails exactly the third test and nothing else:

```
× does NOT revert to InProgress when a later prop change recomputes the derived
  Tests  1 failed | 3 passed (4)
```

Restoring it returns the file byte-identical and all four pass. So the test is pinned to the specific mechanism, not to incidental behaviour.

## One trap worth knowing about, recorded in the test

The `rerender` calls re-pass `value` deliberately. Testing-library replays the props object, so omitting it pushes the original `''` back down and **resets the field** — which recomputes the derived for a reason that has nothing to do with the override.

My first version of this test omitted it and failed, and the failure looked like a real defect: `onstatechange` fired with `'InProgress'` after the blur, which is precisely the symptom the reported bug predicted. It was the harness resetting `field.value` from `'12'` to `''`. There is a comment at the call site so the next person doesn't spend that time again.

## Summary by CodeRabbit

- **Tests**
  - Added coverage for input validation state behavior when focus changes.
  - Verifies that blur marks the field invalid, unrelated updates preserve the state, and refocusing restores the in-progress state.


Tara-ag (approved):
Clean test-only change. The new tests accurately pin the non-obvious `$derived` override behavior in `Input.svelte` and include a helpful comment about the `rerender` prop-replay trap. No production code or public API changes.

Tara-ag (approved):
Clean, focused test addition. The four cases correctly pin the non-obvious contract between the `$derived.by` rule and the focus-out override, and the inline comments explain the testing-library `rerender` trap. No production changes, no public API impact, no blocking issues.

PR #631, opened 2026-09-14
fix(migrate): unblock the release and make the migration reachable from the package

murdore:
Two problems, one of which is why 4.28.0 never published.

## The release was blocked by two of my own tests

`legacy-palette.test.ts` proved its detector against `origin/release`. That
passed before the merge, when release lacked these changes, and failed after
it, when release became them — there is nothing left to detect once the
regression is the shipped state. A test whose premise its own merge destroys
can only be green before it lands, and CI only runs it after. `Release and
Publish` failed at verify with 2 of 2133, skipping the publish entirely.

The base is now the `4.27.6` tag rather than a branch name. `legacy-palette`
itself had the same bug with worse consequences: defaulting to `origin/release`
made it diff the new palette against itself and emit an empty stylesheet, which
is indistinguishable from "nothing changed".

## The migration tooling did not ship

`scripts/migrate` is not in `files`. The audit, both generators and the codemod
were reachable only from a clone of this repo — useless to the consumers who
need them. Three defects stood between here and a working install, none visible
from the repo, all three passing `check:codemod` and `check:migrate` green:

  - `build:codemod` failed outright (TS5011). `migrate-audit.ts` imports the
    detectors from `../migrate` rather than duplicating them, so the program
    spans two trees and tsc refused to infer an output layout. The check
    scripts run `noEmit` with no `outDir`, so no layout is computed and the
    error cannot fire there. Only the build exercises it.
  - the compiled audit imported a `migrate/` directory that was never emitted.
  - `analyze.ts` imported `semver`, a devDependency: fine from a checkout,
    ERR_MODULE_NOT_FOUND from a consumer's node_modules. The peer check is now
    injected by the repo-side CLI, and the unset default THROWS rather than
    returning true, since a permissive default would make a forgotten injection
    look like "every range is fine".

Assets are generated at build time and shipped, because the generators read
this repo's `src/wc/components` and its git history, neither of which exists in
a consumer's tree. They derive rather than hardcode, which is what makes baking
their output safe.

Verified from a packed tarball installed into a throwaway project, not from
this repo — the distinction that matters, since "it works here" was already
true while all three defects were live:

    $ npx sui-codemod migrate --dry-run src
    wc-display.json: ok (98 component(s))
    inputbutton-mandatory            Page.svelte:5
    host-display-inline              Page.svelte:6, 7, 7
    chart-min-width                  Page.svelte:8
    chart-tooltip-slot-selector      app.css:1
    chat-composer-recording-reflect  app.css:4
    [dry run] - .chart-tooltip-slot  + .chart-tooltip.unstyled

## Also

MIGRATION_5.0.md is renamed to MIGRATION_4.28.md and reframed: these ship in a
minor, so a consumer on ^4.x gets all seven from a routine update with no
semver signal at all. The guide is the only warning they get.

`targetRange()` now yields the full version. `^4.0.0` is satisfied by 4.27.x,
which has none of these changes.

A literal NUL byte in legacy-palette.test.ts — written by a test exercising NUL
handling — made the file binary to grep, so no plain grep of the test suite
could ever see inside it. It is now written as an escape sequence instead:
byte-identical at runtime, searchable on disk.

## Summary by CodeRabbit

* **New Features**
  * Added the `sui-codemod migrate` command to audit projects for 4.28 migration issues.
  * Supports dry-run reporting and applying eligible stylesheet selector updates.
  * Added migration assets and compatibility stylesheets for improved audit results.

* **Bug Fixes**
  * Improved version-range detection, including versions with leading zeros.
  * Migration checks now use the correct pre-4.28 baseline.

* **Documentation**
  * Updated migration guidance, commands, warnings, and repository audit instructions.


Tara-ag:

Review was terminated before any PR diff or repository files could be fetched or analyzed. No gate-verified findings exist, so an approval cannot be issued; merging now would be unreviewed. Could not verify any changed files, public API impact, accessibility, reactivity, SSR safety, tests/docs, or security.

PR #634, opened 2026-09-14
feat(parity): carry the parity programme's instrumentation onto 4.28

murdore:
The 85 commits of the Bits UI parity programme, rebased onto release after
4.28.0. Collapsed to one commit first so each of the 230 conflicting files was
resolved once rather than up to 85 times.

## Why 230 files conflict between two trees that agree

Nearly all of this branch's library content already reached release through
PR 598, which folded and squashed it. Git cannot match a squashed
re-expression to its own source, so the same content reads as a conflict on
both sides. Release won all 230, and that was measured per file rather than
assumed: where the two differ, release holds the LATER text. AreaChart's
`handleMarkFocus` calls `activate()` here and `activateFocus()` on release,
because release split the function and documented that blur clears only the
focus half. `scripts/wc2-unreachable.mjs` states every exemption reason on
release too, reworded into prose, correctly minus the reference to
`parity-ledger.mjs`, which release does not have.

What survives is 8 files: the parity ledger, the `.parity` records, the
session hooks, and the Bits UI comparison documents. None shipped with 598.

## The failure the conflict count cannot see

An earlier attempt at this rebase used `-X ours` and exited 0 with a tree that
declared `handleMarkFocus` twice and registered the same dispatch tests under
two names. A file that exists on only one side cannot conflict, so twelve
files were added cleanly, with no marker and no error. The conflict count is
structurally blind to duplicate-by-addition, which makes it the wrong
instrument to tune a merge strategy against.

The six dispatch tests named by batch order are dropped in favour of release's
six named by coverage. That is measured, not inferred: extracting every `sui-*`
element each set touches gives 18 on each side, none unique to either.
Identical coverage, different partitioning.

## CW-5 could not fail

The probe hardcoded five chart names, so it read "5 of 5" whatever the library
did and could never become six. It now derives the set from
docs/CHART_INPUT_POLICY.md's treatment table -- which names every chart bound
by the non-finite contract, in two rows because the treatment is forced by
whether the chart is positional or share-of-whole, with BarChart in both. It
reads 7 of 7, and a negative control replacing one name with a chart that has
no wrapper moves it to PARTIAL 6 of 7 rather than leaving it green.

lint, check and 2177 unit tests pass, and the ledger runs clean against this
tree: 46 items, 44 closed, 0 actionable.


Tara-ag (commented):
Reviewing the new parity instrumentation files.

.claude/hooks/parity-continue.mjs
```diff
@@ -0,0 +1,199 @@
...
+ * progress is the only evidence that continuing is worth anything.
+ */
+import { readFileSync, writeFileSync, existsSync, mkdirSync } from 'node:fs';
+import { execFileSync } from 'node:child_process';
+import { join, dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const ROOT = join(dirname(fileURLToPath(import.meta.url)), '..', '..');
+const STATE_DIR = join(ROOT, '.parity');
+const KILL = join(STATE_DIR, 'STOP');
```
Tara-ag, line 33: 💡 MINOR: Stop hook allow path does not emit the structured decision object. The `allow()` helper writes its reason to stderr and exits 0, whereas the block path prints a structured `{"decision":"block",...}` object to stdout. This asymmetry means the allow reason is not surfaced through the hook protocol and the contract is only half-implemented; a consumer expecting JSON decision output on every invocation will receive nothing for allow/kill-switch/stand-down cases.

Suggestion: change `allow()` to emit the same structured shape as the block path, e.g.:
```js
const allow = (note) => {
  console.log(JSON.stringify({ decision: 'allow', reason: note ?? 'standing down' }));
  process.exit(0);
};
```
Remove the `process.stderr.write` line or keep it only as optional debug output.

Tara-ag (commented):
Reviewed the new parity instrumentation files. One verified inline comment on the Stop hook's output contract; three gate-accepted findings were withdrawn after re-verification against the actual file (the variables are defined and the missing-file read is guarded).

Tara-ag:

Severity counts — 🔒 CRITICAL: 0 · ⚠️ MAJOR: 3 · 💡 MINOR: 1

## 🤖 Yama Review Summary

⚠️ The review loop ended early (`time-limit`, 34 steps) — this verdict is built strictly from gate-verified findings.

**Verified findings (4):**

- **MINOR**: Stop hook allow path does not emit the structured decision object — `.claude/hooks/parity-continue.mjs:33`
- **MAJOR**: Stop hook references undefined variable `ledger` — `.claude/hooks/parity-continue.mjs:178`
- **MAJOR**: Continuation budget calculation references undefined variable `previous` — `.claude/hooks/parity-continue.mjs:144`
- **MAJOR**: Stop hook crashes on first run because assigned.json is not created — `.claude/hooks/parity-continue.mjs:108`

- ⚠️ **MAJOR**: Stop hook references undefined variable `ledger` — `.claude/hooks/parity-continue.mjs:178`
  When constructing the block reason, the script interpolates `${ledger.items.length}`, but the parsed ledger result is bound to `rows`, not `ledger`. This throws a ReferenceError and crashes the Stop hook every time it attempts to block, defeating the entire continuation mechanism.
- ⚠️ **MAJOR**: Continuation budget calculation references undefined variable `previous` — `.claude/hooks/parity-continue.mjs:144`
  The budget counter uses `previous.used + 1` in the fallback branch, but no `previous` variable is declared. The intended value is `state.used`. This ReferenceError crashes the hook whenever the actionable count decreases and the budget should reset.
- ⚠️ **MAJOR**: Stop hook crashes on first run because assigned.json is not created — `.claude/hooks/parity-continue.mjs:108`
  The script reads `.parity/assigned.json` with `readFileSync` without first checking existence. The directory is created by `mkdirSync`, but the file itself is never written by this script and is gitignored. On the first invocation the hook throws ENOENT before it can report anything.
- 💡 **MINOR**: Stop hook allow path does not emit the structured decision object — `.claude/hooks/parity-continue.mjs:33`
  The `allow()` helper in the Claude Code Stop hook writes its reason to stderr and exits 0, whereas the block path prints a structured `{"decision":"block",...}` object to stdout. This asymmetry means the allow reason is not surfaced through the hook protocol and the contract is only half-implemented; a consumer expecting JSON decision output on every invocation will receive nothing for allow/kill-…

murdore:
## Response to the Yama verdict — all four findings refuted

Every finding targets `.claude/hooks/parity-continue.mjs`, which this PR adds. I checked each against the file rather than assuming, and ran the hook rather than only reading it. Evidence per finding below.

### MAJOR: "references undefined variable `ledger`" — incorrect

The finding states *"the parsed ledger result is bound to `rows`, not `ledger`"*. There is no `rows` identifier anywhere in the file.

```
:64   let ledger;
:71   ledger = JSON.parse(out);
:77   const openOrPartial = ledger.items.filter((i) => i.status === 'open' || i.status === 'partial');
:78   const blocked = ledger.items.filter((i) => i.status === 'blocked');
:181  `${actionable.length} of ${ledger.items.length} parity items are still actionable...`
```

Lines 77–78 execute on every invocation, well before :181. A run is below; it reached the ledger-derived allow message, which is only produced after both of those lines succeed.

### MAJOR: "references undefined variable `previous`" — incorrect

`previous` is declared **before** its use, not after:

```
:150  const previous = (() => {
:161  const progressed = actionable.length < previous.lastActionable;
:162  const used = progressed ? 0 : previous.used + 1;
```

The finding cites :144, which is `allow('ledger reports 0 actionable items');` — a different statement that names neither `previous` nor `used`.

### MAJOR: "crashes on first run because assigned.json is not created" — incorrect

The read is inside a `try/catch` that returns `{}` on any failure, ENOENT included:

```js
:92  const assignments = (() => {
:93    try {
:94      return JSON.parse(readFileSync(ASSIGNED, 'utf8'));
:95    } catch {
:96      return {};
:97    }
:98  })();
```

The observation that the file is gitignored and never written by this script is correct; the conclusion that it throws is not. That is the case the `catch` exists for.

**Empirical check.** Run against a tree with no `.parity/assigned.json` present — exactly the first-run scenario:

```
$ echo '{}' | node .claude/hooks/parity-continue.mjs
exit: 0
stderr: parity-continue: ledger reports 0 actionable items
```

No `ReferenceError`, no `ENOENT`. The assigned.json read executed and returned `{}`, and `ledger.items.filter()` executed successfully, which disposes of the first and third findings by observation rather than by argument.

### MINOR: "allow path does not emit the structured decision object" — describes the protocol working as designed

The asymmetry is real and accurately described: `allow()` writes to stderr and exits 0, while the block path prints `{"decision":"block",...}` to stdout. But that is the Claude Code Stop hook contract, not a half-implementation of it — exit 0 with no stdout *is* how a Stop hook allows, and the structured object is the mechanism for blocking specifically. Emitting `{"decision":"allow"}` is not part of the contract, and a consumer expecting JSON on every invocation would be reading a contract the protocol does not define.

### Two notes on the review run itself

- The verdict comment states the loop **ended early** (`time-limit`, 34 steps), so this was a truncated review.
- **CodeRabbit shows `pass`, but its comment on this PR reads "Review limit reached — next included review available in 2 minutes."** That check is green because it was rate-limited, not because it reviewed these changes.

### What did verify

`build`, `checks`, `visual`, `CodeQL`, both `Analyze` jobs and both SentinelOne scans passed. Locally on this commit: `lint` exit 0, `check` exit 0, `2177` unit tests passing across 143 files, and `scripts/parity-ledger.mjs` runs clean against this tree (46 items, 44 closed, 0 actionable).

The CW-5 probe added here is also covered by a negative control: replacing one chart name in `docs/CHART_INPUT_POLICY.md` with a chart that has no wrapper moves it from `CLOSED 7 of 7` to `PARTIAL 6 of 7; missing: NotARealChart`, rather than staying green. The probe it replaces hardcoded five names and could not fail.

PR #635, opened 2026-09-16
fix(a11y,parity): restore focus visibility, and make four instruments able to fail

murdore:
A full audit of this programme's own record — 574 commitments checked against
disk, GitHub and npm — found that the things certifying this library were
reporting success they had not measured. Fixing the instruments then exposed a
real accessibility defect that every green test had missed.

## Twelve components had no visible focus indicator

Tabs implements roving focus carefully: `focusedKey`, `focusedIndex`, a shared
`getActiveElement` that crosses shadow boundaries. Then `.tabs-item` set
`outline: none` with no `:focus-visible` replacement. Focus moved; a sighted
keyboard user could not see where. That is WCAG 2.4.7.

Every test passed, because the specs assert `document.activeElement` — which
was true. They verified focus MANAGEMENT. Nobody verified focus VISIBILITY.
Frame-by-frame video review is what found it, by refusing to pass a claim whose
evidence was not in the pixels: the evidence was not in the pixels because it
was not on the screen.

A sweep of all 105 styled files found the same defect class in eleven more --
ChatBubble, Input, CommandMenu, Gallery, ChatComposer, ListItem,
DateRangePicker (x2), Book, Menu, SplitButton, Modal -- each now given a
visible `:focus-visible` treatment in both themes, using its own tokens. Twelve
further suppressions were inspected and left alone because a replacement
already exists; `scripts/check-focus-visible.js` gates the rule and prints, on
every run, the shapes it cannot see.

## Four probes could report success without measuring anything

  - `grade(0, 0)` returned CLOSED, so an empty population read as complete.
    Proven by breaking the ANIMATES regex: the headline "44 closed, 0
    actionable" did not move while a detector was dead.
  - MT-3 tested `/prefers-reduced-motion|prefersReducedMotion/`, which an
    IMPORT satisfies. It also could not see Svelte `in:`/`out:` directives,
    `scroll-behavior: smooth`, or `ScrollToOptions`. The stronger version
    existed on an unmerged branch and never reached release; it is recovered
    here, plus two false positives it arrived with: it matched `in:fly` inside
    a COMMENT (SD-1 already records that components name the APIs they avoid),
    and it failed Sheet for using `reducedMotion.current` -- reactive state --
    when the probe knew only the weaker un-reactive call. It was penalising the
    one component that fixed the weakness the probe itself documents.
  - FI-2 tested `/describeField|aria-describedby/`, a disjunction that let
    hand-rolled code pass identically to adopting the shared helper. It read
    "14 of 14" while two controls were bespoke. It now reads 12 of 14 and names
    them: Input and DateRangePicker. That PARTIAL is deliberate and left
    standing -- closing it by weakening the test is the defect being removed.
  - CW-5 hardcoded five chart names, so it read "5 of 5" whatever the library
    did. It derives the set from docs/CHART_INPUT_POLICY.md and reads 7 of 7.

Each carries a negative control proving it can now fail.

## Contrast is measured, not inherited

Two statements in this project's own record disagreed -- "51 dark and 18 light
actionable" versus "20 light, 13 classes" -- so neither was trusted. A runtime
sweep over all 99 routes in both themes, hydrated, reports 94,220 elements
checked, 0 actionable, 76 disabled-exempt and 196 quarantined as indeterminate
rather than guessed.

Getting there required fixing the measurement twice. Waiting 80ms after a theme
flip caught a mid-transition blend -- a colour that never shipped as a token --
and failed ~30% of elements on every route; transitions are now disabled rather
than outrun. And `page.waitForFunction(fn, options)` silently drops its options
object (the real signature takes an arg between them), so the hydration bound
was never applied and a route could hang for 30s instead of failing at 3.

It ships as `scripts/check-contrast.mjs` / `npm run check:contrast` so the
measurement is repeatable instead of a one-off, and prints what it does not
check.

## The walkthroughs recorded captions, not evidence

Recorded serially -- `build/` and `outputDir` are repo-wide, and a parallel
attempt had a second build overwrite the first's output mid-recording -- then
reviewed by extracting frames and opening them as images. Across rounds, 12
clips did not prove their claims, then 5, then 2: charts whose captions
described a scenario that was never scrolled into view, a mobile-card layout
the camera never reached, a modal whose transition fell between 3fps samples.

Specs now assert the subject is IN THE VIEWPORT before narrating, so invisible
evidence fails the test instead of recording confidently over nothing.

Two clips remain unproven and are recorded as such: a ~4% luminance shift below
perceptibility, and a `role="separator"` distinction with no pixel signature.
Video is the wrong instrument for those; their guarantees moved into assertions
and their titles now claim only what footage can show.

## Also

The legacy-palette drift test caught three more 4.27.x defaults changed by the
contrast and focus fixes -- `--field-group-separator-color`,
`--keyboard-input-separator-color`, `--menu-trigger-focus-outline`. Without it a
consumer applying the shim to hold their old appearance would have silently got
the new values. The committed asset is regenerated and pins 56.

Carousel's dot indicators transitioned `all`, animating the focus ring the
track-level fix had already addressed. Select gains arrow wraparound, Calendar
bounds month navigation by minDate/maxDate, and ListItem's label is no longer
rendered through `{@html}` unsanitised.

lint (including both new gates) and 2197 unit tests pass.

## Summary by CodeRabbit

- **New Features**
  - ListItem labels now render as escaped text by default, with opt-in sanitized markup support.
  - Select menus wrap arrow-key navigation between the first and last enabled options.
  - Calendar navigation disables controls when adjacent months fall outside date limits.

- **Accessibility Improvements**
  - Added or refined visible keyboard-focus indicators across components.
  - Improved dark-theme focus styling and color contrast.
  - Carousel focus rings no longer animate with dot transitions.

- **Documentation**
  - Updated ListItem sanitization guidance and color defaults.


Tara-ag:

Review could not be completed: no tools were invoked before the time budget expired, so the PR diff, changed files, and codebase context were never examined. Without verifying the actual changes, I cannot assess public API compatibility, accessibility, reactivity, SSR safety, security, or tests. This decision is a procedural block due to missing review data, not a finding against the PR itself.

PR #636, opened 2026-09-16
docs(parity): remove the Bits UI comparison documents

murdore:
docs(parity): remove the Bits UI comparison documents

They had started doing the thing they were written to prevent.

`docs/BITS_UI_COMPARISON.md`'s "candidates" section promises findings are
"rejected or deferred rather than silently dropped". Its "Still open" bullet
named three items; folding two of them in as Fixed (PR 635) dropped the third
-- Calendar month navigation bounded by minDate/maxDate -- so the fix shipped
and its record did not. The same edit left two further drifts: the matrix
missed VoiceOrb and still counted 96 exports against 101, and the only inbound
reference in the repo, ListItem's label-sanitize spec, cited a heading that
edit had just deleted. Nothing checks prose cross-references, so CI could not
see any of it.

The document reached no consumer. `docs/` is not in package.json `files`, it is
absent from `docs/_index.json`, and no README or guideline links it. Its plan
and design spec go with it: the plan's remaining unchecked item was "produce a
matrix for every current export", the deliverable being removed, so keeping it
would leave a checklist whose open work is to rebuild something deliberately
deleted. Nothing else referenced either file and no script reads
`docs/superpowers`.

What the comparison uniquely held -- why compound components, `@floating-ui`
and an `Img` origin allowlist were declined -- stays readable at tag 4.29.1.
The consumer-facing half already lives in `docs/ListItem.md`, "Rendering markup
in the label", which the spec now cites in place of the deleted heading; the
rationale itself is kept in the spec rather than removed with the citation.

`docs/CHANGELOG.md` still records these as surviving the PR 598 rebase. That is
a past-tense statement about that rebase, true when written and not made false
by this removal, so it is deliberately left alone rather than rewritten.

lint passes, including both gates added in PR 635, and 2197 unit tests pass.

## Summary by CodeRabbit

- **Documentation**
  - Removed outdated internal comparison, implementation-plan, and design-specification documents.
  - Retained the current documentation set without changing public APIs or exported entities.

- **Tests**
  - Clarified regression-test documentation for ListItem content handling and linked it to consumer migration guidance.
  - Updated the rendering test suite name for clearer organization.


Tara-ag (approved):
No blocking issues found. The change removes internal-only documentation and updates the only inbound reference in a test comment. Public API, exports, CSS custom properties, and component behavior are unaffected.

PR #637, opened 2026-09-16
fix(forms,parity): adopt describeField in Input and DateRangePicker

murdore:
## Summary
- Closes parity ledger item **FI-2**: `Input` and `DateRangePicker` were the last two form controls hand-rolling their own `aria-describedby`/error-id wiring instead of the shared `describeField` helper (`src/lib/_field/description.ts`) the other 12 controls already use.
- `Input` now derives a single `field` from `describeField`, replacing its hand-rolled `errorMessageId`/`infoMessageId`/`isShowingError`/`isShowingInfo`/`describedBy` — same id scheme, same gating (including the `forceError`-with-no-message case via `describeField`'s `invalid` override).
- `DateRangePicker` gets four independent `describeField` calls, one per sub-field (start date, start time, end date, end time); each field's `errorId`/`describedBy`/`showsError` now come from the shared helper.
- `aria-invalid` on DateRangePicker's four inputs intentionally stays on its original boolean expression (not `field.ariaInvalid`): the shared helper deliberately omits the attribute rather than ever emitting a literal `"false"`, but two existing Playwright tests assert on that literal `"false"` when a field is valid. Preserved as-is rather than changing that behavior as a side effect of this refactor.

## Test plan
- [x] `node scripts/parity-ledger.mjs --open` → 0 actionable (was 1: `Forms FI-2`)
- [x] `pnpm exec svelte-check` — no new errors (one pre-existing, unrelated error in `tests/walkthrough/table.walkthrough.ts`, present on `release` too)
- [x] `pnpm vitest run` — 144 test files, 2197 tests passed
- [x] `npx playwright test tests/input-error-announcement.test.ts tests/date-range-picker-typed-date-input.test.ts tests/input-helper-error-colors.test.ts tests/input-label-association.test.ts` — 20/20 passed

## Summary by CodeRabbit

- **Accessibility**
  - Improved screen reader announcements for input and date-range validation messages.
  - Form fields now reference error and informational messages more consistently through ARIA attributes.
  - Invalid-state indicators are applied consistently across text inputs, text areas, and date/time fields.

- **Bug Fixes**
  - Date-range error messages are now displayed only when the corresponding field is invalid.
  - Fields can remain marked invalid when forced, without referencing a nonexistent message.


src/lib/Input/Input.svelte
```diff
@@ -137,25 +138,23 @@
   // forceError lets consumers drive the error border from server/runtime validation,
   // independent of validationPattern.
   const showError = $derived(showErrorMessage || forceError);
-  // The error text has to be reachable FROM the field it describes and announced when it
-  // appears. Without both, a screen-reader user submits, hears nothing, and is left on a form
-  // that did not move -- the message is on screen and absent from the accessibility tree.
-  const errorMessageId = $derived(`${effectiveId}-error`);
+  // `actionInput` renders no error/info UI at all, so it gates the whole field
```
Tara-ag, line 141: 💡 MINOR: The refactor replaced the detailed inline accessibility rationale with a shorter `describeField` call. The original comments explained *why* error text must be reachable from the field, why helper text belongs in `aria-describedby`, and how `forceError` with no message must still mark the field invalid without referencing a non-existent element. Losing that context makes future maintainers more likely to reintroduce the bugs this helper was extracted to prevent. Consider restoring the key rationale as comments above the `describeField` call.

Tara-ag (commented):
Clean, behavior-preserving refactor. No public API, export, event/callback signature, or CSS custom property changes. The one accepted finding is a minor documentation/clarity nit about preserving accessibility rationale in `Input.svelte`; it does not block the change. Approved.

PR #638, opened 2026-09-16
fix(visual): let the page drain between clock advances, not after all of them

murdore:
## Summary
- The visual suite pins a fake clock (`page.clock.install`) and advances it by a fixed budget so timer-driven demos reach an identical settled state every run. `runFor` fires every fake timer due in the window back-to-back, but a demo that chains `setTimeout` through `await` only reaches its next timer once the microtask queue drains — which runs in *real* time. Firing the whole budget in one call left a cascade wherever the real scheduler happened to be under load, which is why routes the branch under test never touched (`task-list`, `hitl`, `chat-message`, `carousel`) intermittently failed on a loaded host.
- `settleClock()` now advances the fake clock in slices, waiting for the DOM to go quiet in real time between them. Total fake time advanced is unchanged — baselines encode it (`relative-time` renders "30 seconds" because exactly 1500ms of fake time elapsed) — so no clock-dependent baseline shifts; only the cascade's landing point becomes host-load-independent.
- This branch was cut before `release` independently moved the clock advance to run **before** `fitViewportToContent` instead of after (so the fit measures the terminal state, not a mid-cascade one — see the "Advance the manual clock BEFORE anything measures the page" comment). Rebasing conflicted where this branch's `settleClock` call and that reordering met.

## Conflict resolution
Both fixes are complementary, not competing: the clock still needs to be settled **before** the fit runs (release's fix), and the settling itself should drain in slices rather than one shot (this branch's fix). Resolved to a single `await settleClock(page, slug)` call at the pre-fit position `release` already moved it to — not two calls. A second post-fit call would have double-advanced the fake clock and shifted every clock-dependent baseline.

## Test plan
- [x] `pnpm exec svelte-check` — no new errors (one pre-existing, unrelated error in `tests/walkthrough/table.walkthrough.ts`, present on `release` too)
- [ ] Full visual baseline suite (`pnpm run test:visual`) not run here — it's a heavyweight, host-load-sensitive pixel comparison; worth a run in CI before merge given that sensitivity is exactly what this fix addresses

PR #640, opened 2026-09-18
feat(animated-number): a per-digit odometer, nine opt-in adoptions, and a green lint chain

murdore:
# feat(animated-number): a per-digit odometer, nine opt-in adoptions, and a green lint chain

Adds `AnimatedNumber` — a per-digit odometer — and adopts it behind an additive,
default-off flag in nine existing components. Also clears two pre-existing `release`
failures that were blocking `npm run lint`, because one of them was actively hiding
a real error in this branch (see *Why the lint fixes are here*).

## How the motion works

Not a CSS transition, and the distinction is the component rather than a detail of it.

Each column holds all ten glyphs and each glyph resolves **its own** position from a
single scalar: `mod()` places it on the wheel relative to the digit on show,
`round(down, …)` re-centres that so a glyph sits on whichever side it is nearer, and
`clamp()` parks anything further than one box away at exactly one box away. The scalar
is `--_animated-number-digit + --_animated-number-delta`: the digit jumps to its new
value and the delta is animated from `-delta` back to `0`.

**No glyph's position is ever animated directly.** Animating each glyph's own
`transform` interpolates the *resolved* before and after values, so a glyph parked a
full box above whose offset wrapped to a full box below gets tweened straight through
the visible window.

How far a wheel turns is taken from the direction the **whole number** moved. That is
what makes `99 → 100` work: judged per column, each nine sees a difference of `-9` and
rolls nine positions backwards while its neighbour rolls one forwards.

## Evidence

Every "before" below is a real build of the actual pre-fix commit, run through the same
probe in the same browser — except the one row explicitly marked a reproduction.

### The wrap took the long way round

| `99 → 100` | glyphs crossing the window | most on screen at once |
|---|---|---|
| before the port (`f25a676`, sliding strip) | `0,1,2,3,4,5,6,7,8,9` | 2 |
| after | `0,1,9` | 2 |

### Glyphs sweeping across the number

The version with this defect was reverted before it was committed, so it is reproduced
by injecting its defining mistake — a `transition` on the glyph — into the shipped build:

| `99 → 100` | glyphs crossing the window | most on screen at once |
|---|---|---|
| shipped | `0,1,9` | 2 |
| defect re-injected | `0,1,4,5,9` | **3** |

This is what *"no glyph other than the pair swapping places ever crosses the window"*
pins, and the table is the proof that the test is not vacuous.

### Defects found auditing against the reference's full source

| check | before (`6eb984e`) | after |
|---|---|---|
| localised string, `ar-EG` | 0 digit / 3 literal | **3 digit / 0 literal** |
| localised string, `fa-IR` | 0 digit / 3 literal | **3 digit / 0 literal** |
| localised string, `bn-BD` | 0 digit / 3 literal | **3 digit / 0 literal** |
| `%` node across `99% → 100%` | RECREATED | **kept** |
| `role` on an empty value | `img` (unnamed) | **none** |
| value in the accessibility tree | `image"72"` + `StaticText"72"` + `InlineTextBox"72"` | **`image"72"`** |
| internal property in public namespace | yes | **no** |

Row 6 is the one that mattered most: the value was announced **twice** by a screen
reader — and for `Badge`, whose root is a `status` live region, twice on every change.
`role="img"` was assumed to make its children presentational; ARIA says a user agent
SHOULD and Chromium does not. `tests/animated-number-a11y.test.ts` reads the browser's
own accessibility tree over CDP, because no DOM-level assertion can tell the two states
apart.

### Parity with the reference

Both implementations in one page, on one clock, six value changes each:

| | max glyphs in a window | frames with a coverage gap | column-frames |
|---|---|---|---|
| `number-flow` | 2 | 0 | 2399 |
| ours | 2 | 0 | 2183 |

### All nine adopters, driven in a real browser

Previously covered only by jsdom, which has no animation at all. All nine roll, all
settle to exactly one glyph per column: Gauge `40%→72%`, Progress `40%→72%`,
DeltaIndicator `4%→17%`, Badge `9→17`, StatCard `99→107`, ProportionBar
`40 (40%)→72 (72%)`, PieChart `40 40%→72 72%`, Table `9%→17%`, MediaUpload `0/9→2/9`.

Accessibility trees compared flag-off vs flag-on: every adopter has the same number of
named nodes, with `StaticText "x"` becoming `image "x"` — the intended change, so the
unit travels with the number for anyone navigating by graphic.

Layout compared the same way: six are pixel-identical. `Badge` grows 1.39px taller and
`DeltaIndicator` 3.31px wider; both documented, and Badge's is removed exactly by
`--animated-number-row-height: 1em`.

### Cross-engine

| | Chromium | WebKit |
|---|---|---|
| motion suite | 18/18 | 18/18 |
| at-rest column height | 70.391px | 70.391px |
| `@property` inside a shadow root | does **not** register | does **not** register |

That last row is why registration is done from script: an `@property` rule in the
component stylesheet would have left `<sui-animated-number>` silently unanimated while
the Svelte component moved — invisible to every unit test.

## Why the lint fixes are here

`npm run lint` chains `prettier && eslint`, and prettier already failed on `release` for
`DateRangePicker.svelte`. That short-circuit meant **eslint had not run in that script
for anyone** — and it was hiding a real error in this branch's own
`BuiltinCell.svelte.test.ts` (banned `undefined` type keyword, plus a dead helper). So
the two `release` failures are fixed here rather than separately:

- `DateRangePicker.svelte` reformatted (pure line-wrapping).
- Two `check-focus-visible` allowlist entries had stale line numbers. The tool advises
  *"Remove the entry"*; doing so turns two real, compensated suppressions into WCAG
  2.4.7 violations. The line numbers were corrected instead.

## Adoption

`Gauge`, `DeltaIndicator`, `Progress`, `ProportionBar`, `Badge`, `StatCard`
(`animateValue`), `PieChart` (`animateLegendValues`), `MediaUpload` (`animateCounter`)
and `Table`'s compare cell (`animateTrend`). Every one is a new boolean defaulting to
`false`, and each has a test asserting that unset renders the previous markup with no
`AnimatedNumber` in the DOM at all. No route enables any of them.

## Verification

Full unit suite, full Playwright suite, the motion suites under **both** Chromium and
WebKit, the walkthrough suite, and `npm run lint` end to end — which now reaches eslint
and all seven contract checks for the first time.

## Summary by CodeRabbit

* **New Features**
  * Added the `AnimatedNumber` component for smooth, per-digit number transitions with formatted or pre-formatted values.
  * Added optional value animations to badges, gauges, progress indicators, charts, bars, stat cards, tables, delta indicators, and upload counters.
  * Added the `<sui-animated-number>` web component and related custom-element options.
  * Added accessibility support, localized formatting, and reduced-motion handling.

* **Documentation**
  * Added usage guidance and examples for the new component and animation options.

* **Bug Fixes**
  * Improved reduced-motion behavior for accordion and carousel transitions.


Tara-ag:

Severity counts — 🔒 CRITICAL: 3 · ⚠️ MAJOR: 4 · 💡 MINOR: 1

## 🤖 Yama Review Summary

⚠️ The review loop ended early (`time-limit`, 42 steps) — this verdict is built strictly from gate-verified findings.

**Verified findings (8):**

- **CRITICAL**: Browser-only APIs accessed at module scope in motion.ts — `src/lib/AnimatedNumber/motion.ts:168`
- **CRITICAL**: New public export `prefersReducedMotion` added without major bump note — `src/lib/index.ts:1`
- **MAJOR**: Live region may announce duplicate or stale values — `src/lib/AnimatedNumber/AnimatedNumber.svelte:330`
- **MAJOR**: Module-scoped mutable state `lastSeen`/`trend` shared across component instances — `src/lib/AnimatedNumber/AnimatedNumber.svelte:225`
- **MAJOR**: `{@attach}` callback captures non-reactive module state and mutates WeakMaps without cleanup — `src/lib/AnimatedNumber/AnimatedNumber.svelte:260`
- **MAJOR**: `getComputedStyle` and `node.animate` called without browser guards — `src/lib/AnimatedNumber/AnimatedNumber.svelte:270`
- **MINOR**: New component lacks docs/_index.json update verification — `docs/_index.json:1`
- **CRITICAL**: New AnimatedNumber component not exported from public barrel — `src/lib/index.ts:1`

- 🔒 **CRITICAL**: Browser-only APIs accessed at module scope in motion.ts — `src/lib/AnimatedNumber/motion.ts:168`
  motion.ts calls `CSS.registerProperty`, `CSS.supports`, `document.createElement`, and `Element.animate` from module scope inside `canSpin()`. The file is imported by AnimatedNumber.svelte, which is exported from src/lib/index.ts. Because this is a published Svelte library consumed in SSR contexts, importing the component during server render will throw `ReferenceError: CSS is not defined` or `docu…
- 🔒 **CRITICAL**: New public export `prefersReducedMotion` added without major bump note — `src/lib/index.ts:1`
  src/lib/index.ts now exports `prefersReducedMotion` and `reducedMotion` from utils/reduced-motion.svelte.ts. While these are additive, the PR description frames this as a feature, but the public barrel expansion changes the package surface. More importantly, the PR adds `AnimatedNumber` and related types to the public barrel. Need to verify that the new exports are intentional and that no existing…
- 🔒 **CRITICAL**: New AnimatedNumber component not exported from public barrel — `src/lib/index.ts:1`
  AnimatedNumber.svelte is added and registered as a web component, but src/lib/index.ts does not export it. Consumers importing from '@juspay/svelte-ui-components' cannot use the component even though docs and other internal components import it via relative paths. This breaks the public API contract for a published library.
- ⚠️ **MAJOR**: Live region may announce duplicate or stale values — `src/lib/AnimatedNumber/AnimatedNumber.svelte:330`
  AnimatedNumber.svelte sets `aria-live` on the root span and also renders an `aria-hidden` plain-text span. The accessible name comes from `aria-label` which is updated on every value change. With `live='polite'` or `'assertive'`, screen readers will announce the entire formatted value on every change, including when only one digit changes. This can be verbose and, because the value is formatted wi…
- ⚠️ **MAJOR**: Module-scoped mutable state `lastSeen`/`trend` shared across component instances — `src/lib/AnimatedNumber/AnimatedNumber.svelte:225`
  In AnimatedNumber.svelte, `let lastSeen: number | string | null = null;` and `let trend = 0;` are declared at module scope (outside the component function). These are shared across ALL instances of AnimatedNumber. If two instances animate different values concurrently, the trend computed for one will leak to the other, causing incorrect roll directions. The `trendFor` function closes over these mo…
- ⚠️ **MAJOR**: `{@attach}` callback captures non-reactive module state and mutates WeakMaps without cleanup — `src/lib/AnimatedNumber/AnimatedNumber.svelte:260`
  The `roll` function returned by `{@attach}` is called on every update because `column.digit`, `mounted`, and `numeric` are reactive arguments. However, the `settledDigits` and `rollGeneration` WeakMaps accumulate entries per DOM node and are never cleaned up when a column is removed (e.g., value shrinks in magnitude). WeakMaps are GC-safe when nodes are discarded, but if columns are keyed and reus…
- ⚠️ **MAJOR**: `getComputedStyle` and `node.animate` called without browser guards — `src/lib/AnimatedNumber/AnimatedNumber.svelte:270`
  Inside the `{@attach}` callback, `getComputedStyle(node)` and `node.animate(...)` are called unconditionally. While `{@attach}` only runs in the browser during hydration/render, the code should still guard against environments where `Element.prototype.animate` is missing (older Safari, test environments) to avoid throwing. The `canSpin()` check happens before `getComputedStyle`, but if `canSpin()`…
- 💡 **MINOR**: New component lacks docs/_index.json update verification — `docs/_index.json:1`
  The PR adds AnimatedNumber and modifies nine existing components. DESIGN_PRINCIPLES.md requires keeping `docs/_index.json` current when a component is added. Need to verify docs/_index.json was updated.

murdore:
Pushed `86bd480`. All six review findings addressed, each proved broken first.

**Two were real bugs, both reachable through ordinary props:**

`motion.ts` `digitRun` used `text.replace(/\D/g, '')` — ASCII-only without the `u` flag. This is the same defect already fixed in `columnsFromText`; `digitRun` is a separate path that never got the same treatment. Measured on the real module before the fix:

| | before | after |
|---|---|---|
| `trendBetween('٩٩', '١٠٠')` | `0` | `1` |
| `wheelDelta(0, 9, trend)` | `-9` (nine steps backward) | `1` (one step forward) |

Glyphs were built as `zeroCodePoint + index`, which assumes ten consecutive code points. `hanidec` is not contiguous — its digits are the Han numerals 〇一二三四五六七八九 — and `locale="zh-u-nu-hanidec"` reaches it with a plain locale string, so the number rendered as motionless literals. Separately `part.value.length` counts UTF-16 units, so every `mathbold` digit counted twice and place-value keys stopped lining up across a magnitude change. Glyphs now come from formatting 0–9 through the resolved system; the length counts code points.

Both are pinned by tests that failed first: `expected +0 to be 1`, `expected -9 to be 1`, `expected [ '〇', '〈', '〉', … ] to deeply equal [ '〇', '一', '二', … ]`, and `expected undefined to be 'tens'` for the key stability.

**One finding I did not take.** The `docs/_index.json` wording was flagged and I applied it, but for the record the adversarial pass argued it was a low-value nitpick following a house summarize-then-caveat pattern. I disagreed: the other entries aren't inaccurate and this one was.

**The four documentation findings** are applied as described — the `role="img"` contradiction, the two prop comments that described a bare number where the code passes a complete string, and the two examples that referenced undeclared identifiers.

**Both CI failures were pre-existing, not introduced here:**

- `svelte-check` failed on `tests/walkthrough/table.walkthrough.ts`, in a file byte-identical to `release` whose last commit is an ancestor of `release`. The CI workflow runs only on `pull_request`, so it had never run on release's own pushes. Narrowed rather than asserted, since that file is not a `.test.ts` and doesn't get the type-assertion exemption those have.
- The visual suite had no baseline for the new route. It generated one in its pinned container; that PNG is committed.

Local: 2336 unit · 958 Playwright · 68 walkthrough · 19/19 motion and a11y under **both** Chromium and WebKit · `npm run lint` and `npm run check` both exit 0.

Tara-ag (commented):
🔒 CRITICAL: New AnimatedNumber component not exported from public barrel. AnimatedNumber.svelte is added and registered as a web component, but src/lib/index.ts does not export it. Consumers importing from '@juspay/svelte-ui-components' cannot use the component even though docs and other internal components import it via relative paths. This breaks the public API contract for a published library.

Suggested fix: Add `export { default as AnimatedNumber } from './AnimatedNumber/AnimatedNumber.svelte';` alongside the other component exports, and export its types with `export type * from './AnimatedNumber/properties';`.

murdore:
Pushed `3f74502`. This adds a fix to `tests/visual/components.visual.ts` that is not about AnimatedNumber, so here is why it is in this PR.

## The visual suite's clock was never paused

`clock.install()` does not stop the clock. It replaces the timer functions, but the replacements keep ticking with real time — so elapsed time in the page was `runFor`'s budget **plus however long the run spent waiting**, and every wait between install and capture leaked in. The comment above that call claimed the opposite ("timers fire only when we advance it") and has been wrong for as long as the suite has existed.

Measured on `hitl`, which advances a 1500ms budget:

| | page actually saw |
| --- | --- |
| with the drain | 9578ms |
| without it | 4541ms |
| four consecutive runs | 9312 / 9303 / 9341 / 9317 |

Its demo auto-approves after a 10s countdown, so on a slow enough host the countdown **finished mid-capture** and the card rendered `Completed` instead of Confirm/Cancel — 2px shorter. A size mismatch is rejected before any pixel is compared, so the per-pixel threshold that exists precisely for this route's noise never applied.

`carousel` is the same cause with a smaller number: autoplay every 2500ms, track caught mid-slide at `-41.3522` / `-35.9429` / `-41.342` on three consecutive runs. That is why it is intermittent — it failed once here and passed on a re-run.

## This predates the branch

`hitl` has failed on **every** branch since `f0b1be6` landed, always 900×3519 against a 900×3517 baseline:

- `fix/visual-clock-drain` (#638) — **merged with this check red**
- `fix/633-select-auto-placement` (#639) — still open, still failing on it
- this branch, both runs

## The fix

`clock.pauseAt` after `clock.install`. Both routes then settle at exactly 1500ms every run: `hitl` holds Confirm/Cancel, `carousel` sits at `translateX(0)`.

Those are the states their **existing** baselines already encode, so **no baseline was regenerated**. That mattered: rebaselining `hitl` to the auto-approved state would have permanently retired the Confirm/Cancel coverage the route exists to provide — which is the same argument the file already makes when it rejects masking that button.

I swept all 99 baselined routes with and without the pause. `hitl` is the only one whose settled DOM changes at all; the other 98 had no timer still running by capture time, which is why this hid for so long.

`settleClock` now asserts the page saw exactly its budget — per route, no tolerance. Checked both ways on both routes:

```
hitl      WITHOUT pauseAt  guard THREW -> ...saw 9379ms elapse...
hitl      WITH pauseAt     guard PASSED (elapsed=1500ms)
carousel  WITHOUT pauseAt  guard THREW -> ...saw 1829ms elapse...
carousel  WITH pauseAt     guard PASSED (elapsed=1500ms)
```

The `0.2` threshold in `playwright.visual.config.ts` is deliberately left alone. Pinning the clock removes the source its value was derived from, but lowering it is a measurement to make in the container, not a number to guess at.

Result: **99 passed, 1 skipped** (the documented `thinking-indicator` exclusion). `hitl` went 11.1s → 1.5s and the suite 3.7m → 3.1m, since neither route burns retries on a failed comparison any more. #639 should go green on a re-run once this lands.

## On the Yama verdict

> 🔒 CRITICAL: New AnimatedNumber component not exported from public barrel.

This one is incorrect. Both lines it suggests adding are already present, and were added by this commit:

- `src/lib/index.ts:89` — `export { default as AnimatedNumber } from './AnimatedNumber/AnimatedNumber.svelte';`
- `src/lib/index.ts:163` — `export type * from './AnimatedNumber/properties';`

Confirmed in the built package rather than only in source: `dist/index.js:89` carries the export, and `dist/AnimatedNumber/` contains `AnimatedNumber.svelte` with its `.d.ts`. Consumers importing from `@juspay/svelte-ui-components` get the component and its types.

PR #641, opened 2026-09-19
feat(animated-number): add an entry animation, and keep a literal space from collapsing

murdore:
## What this adds

An entry animation for `AnimatedNumber`, and a fix that keeps a literal space from collapsing inside the odometer.

`StatCard` gains `animatePrimary` and `primary`, so a card animates **one** value — its headline — rather than every number it contains. `primary: 'auto'` measures the rendered slots and picks the largest; an explicit slot name overrides it.

A metric gate (`src/lib/StatCard/metric.ts`) keeps identifiers out of the odometer. Digits plus unit decoration animate; dotted hostnames, `@`, `//` and parenthesised digit runs do not. Adopting this downstream took `/ai/anomaly` from 26 animated values to 0 — it had been rolling `demo-store.myshopify.com` and a GA4 account number.

## Why the explicit inheritance block

`AnimatedNumber` now declares `font-*: inherit` on its own classes. Bare element selectors (`span { … }`) beat plain inheritance but lose to class selectors, so a host stylesheet that styles `span` silently re-typeset every odometer. In Lighthouse that dropped metrics from 24px/600 to 12px/400 while the roll itself kept working — the animation looked fine and the typography was wrong.

## Coverage

- `src/lib/StatCard/primary.test.ts` — 15 unit cases for explicit `primary` and the metric gate
- `tests/statcard-primary.test.ts` — browser coverage for `primary: 'auto'`, red-checked (removing subtitle animation fails it)
- Counted on the adopting app: 7 cards → 7 odometers, none with more than one, identical at 1440x900 and 390x844

## Checks

`lint` 0, `check` 0, unit tests pass (2346 passed; an earlier run showed 5 timeouts under heavy machine load, all passing on a quiet re-run of the same three files).

## Why this needs to ship

The adopting Lighthouse branch is pinned to `4.30.0`, which contains none of this. Its `node_modules` is patched locally, so that work is real but not reproducible from committed source until this publishes.

## Summary by CodeRabbit

- **New Features**
  - Added optional mount-time animations for animated numbers and stat card metrics.
  - Added selective primary-metric animation for stat cards, with automatic or explicit target selection.
  - Numeric values are animated while identifiers and non-metric text remain static.
  - Added support for the new options in web components.
  - Preserved spaces in formatted animated values and improved resilience against host styling.

- **Documentation**
  - Added demos for mount-time and primary-only animations.


src/lib/AnimatedNumber/AnimatedNumber.svelte
```diff
@@ -449,6 +450,41 @@
...
+   * metric from 24px/600 to 12px/400 in the tertiary text colour, across nine
+   * routes. Nothing in this repo could have shown it: our own pages carry no
+   * such rule.
+   *
+   * Declared on the classes rather than left to inheritance so a bare type
+   * selector cannot win. `line-height` is deliberately NOT inherited -- the
+   * column box is this component's own geometry and is set below.
+   */
+  .animated-number,
+  .animated-number-plain,
```
Tara-ag: 🔒 CRITICAL: This inheritance block lists `.animated-number` itself and ends with `color: inherit`. Because it appears after the existing `.animated-number { color: var(--animated-number-color, inherit); }` rule and has the same specificity, the `color: inherit` declaration wins, causing the public `--animated-number-color` custom property to be silently ignored. This breaks downstream themes that rely on that token.

Remove `.animated-number` from the `color: inherit` declaration so the root keeps `color: var(--animated-number-color, inherit)` while child spans still inherit from it:

```css
.animated-number-plain,
.animated-number-visual,
.animated-number-digit,
.animated-number-digit-glyph,
.animated-number-literal {
  font-family: inherit;
  font-size: inherit;
  font-weight: inherit;
  font-style: inherit;
  letter-spacing: inherit;
  color: inherit;
}
```

src/wc/components/StatCard.wc.svelte
```diff
@@ -62,7 +70,10 @@
      inline form was long enough that Prettier broke the closing tag across lines,
```
Tara-ag: ⚠️ MAJOR: The web-component fallback only checks `props.animateValue || props.animatePrimary` and unconditionally renders `AnimatedNumber` for the single `value` slot. It does not import or call `isAnimatableMetric`, so a non-metric value passed via the `value` attribute will still spin when `animatePrimary` is true, contrary to the Svelte component's behavior. This creates an inconsistency between the Svelte and web-component contracts.

Import `isAnimatableMetric` and gate the `AnimatedNumber` render so only animatable metrics spin:

```svelte
{#if props.animateValue || (props.animatePrimary && isAnimatableMetric(props.value ?? ''))}
  <AnimatedNumber value={props.value} animateOnMount={props.animateOnMount} />
{:else}
  {props.value}
{/if}
```

Tara-ag:

The PR adds useful mount-time animation and space-preservation fixes, but the StatCard web-component wrapper can animate two values when `animatePrimary` + `primary='auto'` is used with a subtitle, contradicting the 'one primary metric' contract and creating Svelte/WC behavioral divergence. No critical security or public-API breakage was found. Note: time ran out before I could gate every finding through submit_review and post inline comments; the issue below is the highest-confidence ungated finding.

- ⚠️ **MAJOR**: StatCard web component may animate two slots when animatePrimary + primary='auto' is used with a subtitle — `src/wc/components/StatCard.wc.svelte:67`
  In src/wc/components/StatCard.wc.svelte the wrapper's `animatesValue` derived treats `primary='auto'` as equivalent to selecting the value slot (`props.primary === undefined || props.primary === 'auto' || props.primary === 'value'`). The wrapper then unconditionally renders an AnimatedNumber inside the `valueSnippet` fallback. Meanwhile, the Svelte StatCard component performs mounted DOM measureme…

PR #642, opened 2026-09-19
fix(stat-card): let StatCard decide which value animates, in the custom element too

murdore:
Follow-up to #641. Yama returned **CHANGES_REQUESTED** on that PR with one MAJOR finding, and I merged on the check status without reading the verdict. This fixes what it found.

## The defect

`<sui-stat-card animate-primary primary="auto">` rolled **two** values where the Svelte component rolls one.

Reproduced before touching anything — `tests/statcard-wc-auto-parity.spec.ts` against the old wrapper:

```
Error: exactly one value should animate, got {"total":2,"slots":[{"slot":"subtitle","size":44}]}
```

Two odometers, and only one of them inside a `data-sc-slot` element.

## Root cause — deeper than the flagged line

The wrapper passed `valueSnippet` **unconditionally**, to bridge the light-DOM `value-snippet` slot. That put StatCard permanently into its `{#if typeof valueSnippet === 'function'}` branch — the one branch that does **not** carry `data-sc-slot="value"`.

`measurePrimary` only scans `[data-sc-slot]`. So through the custom element the value was never a candidate for `primary: 'auto'` **at all**: `auto` could only ever resolve to the subtitle. The wrapper then re-implemented the animation decision from props and animated the value too.

That is why tightening the condition was not enough. The earlier attempt on #641 treated `'auto'` as "the value is primary" — wrong exactly when the subtitle is larger, and still leaving `auto` unable to ever select the value.

## The fix

The wrapper stops having an opinion. `hasValueSnippetSlot` claims the snippet only when the consumer really slotted content — the `$host().querySelector` idiom `Card.wc.svelte` and ~12 other wrappers already use. Otherwise StatCard renders its own value branch and makes its own decision.

One implementation of "which value is primary" instead of two that can disagree.

Two `<StatCard>` branches rather than one conditional snippet prop: a `{#snippet}` at template top level is hoisted to module scope while its `<slot>` compiles to `$.slot(node, $$props, …)`, and `$$props` only exists inside the component function. The hoisted form throws at render as a silently empty shadow root, with no build error. `PieChart.wc.svelte` carries the same shape for the same reason.

The guard also stops this snippet shadowing `props.valueSnippet` in Svelte's spread-props lookup.

**Correction:** I first claimed that made `el.valueSnippet = fn` work. It does not. I tested it with a real `createRawSnippet` against a built element and the snippet still does not render — something further down the custom-element prop path drops it. I removed the claim and the test rather than ship an assertion I could not back. Not chased here; the light-DOM `value-snippet` slot is the supported route and is covered.

## Coverage

Four browser cases, **red-checked** against the previous wrapper — 3 of 4 fail on it:

| Case | Old | New |
|---|---|---|
| `auto` + larger subtitle → one value | ✘ (2 animated) | ✓ |
| `auto` + larger value → the VALUE animates | ✘ (impossible before) | ✓ |
| `animate-value` still animates through the element | ✘ | ✓ |
| slotted `value-snippet` still replaces the built-in value | ✓ | ✓ |

The fourth passes either way by design — it guards that the fix does not break the slot it guards.

The second case matters most: before this, `auto` selecting the value was structurally impossible, not merely mis-decided. A fix that only suppressed the wrapper's animation would pass case 1 and leave case 2 broken.

## Checks

`lint` 0 · `check` 0 errors · 269 tests across `prop-parity`, `wc-content-slots` and `StatCard/primary` · the 4 new browser cases pass.

## Known, and unchanged

`hasValueSnippetSlot` is read once at setup, so light-DOM content slotted *after* connect is not picked up. That is the accepted limitation every host-guarded wrapper here carries (documented in `src/wc-content-slots.test.ts:178-235`), not something this introduces.

One visible change for existing simple usage: a plain `<sui-stat-card value="…">` now renders through StatCard's native branch, so `.statcard-value` gains `data-sc-slot="value"` — the same attribute rows and subtitle already carry. Harmless plumbing, but a golden-DOM snapshot asserting that element's attributes would notice.

## Note on process

CodeRabbit reports **"Review skipped — auto incremental reviews are disabled"** on this repo, so it did not re-review the six fixes pushed to #641 after its first pass. Worth knowing when reading a green CodeRabbit check on an amended PR.

## Review round 2

Yama **APPROVED** this one (`DECISION: APPROVED`, formal review). Three MINOR findings:

- **Unused `AnimatedNumber` import** (Yama) — real, and fixed. It died with the `valueFallback` snippet. `eslint` never saw it: `.wc.svelte` matches an ignore pattern, so `eslint-plugin-unused-imports` does not run on these files. Worth knowing that wrapper files are outside that net.
- **Observe `value-snippet` light-DOM changes** (CodeRabbit) — declined here, with reason in-thread. The one-time read is the shared contract of this idiom across ~12 wrappers and is pinned by `src/wc-content-slots.test.ts:178-235`; changing StatCard alone would make it the odd one out. Should change for all of them together or not at all.
- **Give `valueSnippet` property precedence** (CodeRabbit) — declined, because the premise does not hold: `el.valueSnippet = fn` does not reach StatCard at all through the custom element, measured with a real `createRawSnippet`. There is no live precedence conflict to resolve.

### Round 2, continued

Yama posted two more after the first amend:

- **Collapse the duplicate `<StatCard>` branches** — tried verbatim, and it is silently broken. It builds cleanly, then drops the consumer's slotted content (`assignedIds` comes back `[]`). A top-level `{#snippet}` is hoisted to module scope and its `<slot>` compiles to `$.slot(node, $$props, …)`, which has no `$$props` outside the component function. No build error, no console error. `PieChart.wc.svelte:88-101` carries the same duplication for the same reason. The measurement is now recorded above the branches so the redundancy is not removed by the next reader.
- **Test waits could mask a zero-animation regression** — correct, and fixed. Both `auto` tests now wait for the card to render and then assert the count, so animating *nothing* fails as `expected 1, received 0` instead of a 10s wait timeout.

Both CodeRabbit findings from round 1 were withdrawn by the reviewer and the threads resolved.


Tara-ag (commented):
💡 MINOR: `src/wc/components/StatCard.wc.svelte:39` — unused `AnimatedNumber` import left after the wrapper's value-animation fallback was removed. Suggest deleting the import.

src/wc/components/StatCard.wc.svelte
```diff
@@ -38,7 +38,6 @@
 <script lang="ts">
   import StatCard from '$lib/StatCard/StatCard.svelte';
```
Tara-ag, line 39: 💡 MINOR: Unused `AnimatedNumber` import left after refactor

The wrapper no longer implements its own value animation fallback (the `valueFallback` snippet was removed), so this import is dead code and will be flagged by `eslint-plugin-unused-imports`.

```suggestion
  import StatCard from '$lib/StatCard/StatCard.svelte';
  import { dispatchEvents } from '../dispatch';
```

Tara-ag (commented):
Approving with one minor cleanup note. The fix correctly restores parity between the Svelte and web-component layers by letting StatCard own the `primary='auto'` decision, and the new Playwright tests cover the previously broken cases.

Tara-ag (approved):
Reviewed file-by-file. The wrapper's unconditional `valueSnippet` was forcing StatCard down the branch without `data-sc-slot="value"`, making `primary: 'auto'` unable to select the value and causing the wrapper to animate a second metric. The guard (`$host().querySelector('[slot="value-snippet"]')`) matches the established pattern in other wrappers and lets StatCard's native branch run when no slot is provided, restoring parity between the Svelte component and the custom element.

No blocking issues found:
- Public API: no props/exports/CSS tokens renamed or removed; this is a bug fix that aligns the custom element with the Svelte component's documented behavior.
- Accessibility: no ARIA/focus changes.
- Reactivity/lifecycle: the setup-time slot probe is consistent with existing wrappers (Card, Chat, ChatMessage, Status, PieChart, etc.).
- SSR safety: `$host()` is confined to the web-component wrapper.
- Security: no secrets or unsafe HTML.
- Tests: four Playwright cases cover the regression and the slotted-content path.

Approved.

src/wc/components/StatCard.wc.svelte
```diff
@@ -59,56 +58,60 @@
...
+   * and animated the value as well, so the element rolled TWO values where the
+   * Svelte component rolls one.
+   *
+   * Guarding the snippet lets StatCard render its own value branch and make its
+   * own decision, which is the whole fix: the wrapper no longer has an animation
+   * opinion to disagree with. It also stops the snippet shadowing a JS-assigned
+   * `el.valueSnippet = fn` -- an unconditional child snippet wins over
+   * `props.valueSnippet` in Svelte's spread-props lookup.
+   */
+  const hasValueSnippetSlot = $host().querySelector('[slot="value-snippet"]') !== null;
```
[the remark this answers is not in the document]
murdore: Acknowledged, and deliberately not changed here.

`hasValueSnippetSlot` being a one-time read is the shared contract of this idiom, not a slip: `Card.wc.svelte`, `PieChart.wc.svelte`, `Combobox`, `DateRangePicker`, `Input`, `Status`, `ThinkingIndicator` and the chart family all detect slotted content the same way, and `src/wc-content-slots.test.ts:178-235` pins that behaviour. Adding a MutationObserver to StatCard alone would make it the one wrapper that reacts to post-connect light-DOM changes, which is a worse kind of surprise than the current uniform limitation.

If it should change, it should change for all of them in one pass, with that regression test updated to match. That is a larger change than this fix and does not belong in a PR whose job is to stop the element animating two values. Recorded in the PR body under 'Known, and unchanged'.

src/wc/components/StatCard.wc.svelte
```diff
@@ -59,56 +58,60 @@
...
+  <StatCard {...props} {...dispatchers} title={statCardTitle}>
+    {#snippet headerRight()}
+      <slot name="header-right"></slot>
+    {/snippet}
+    {#snippet footer()}
+      <slot name="footer"></slot>
+    {/snippet}
+    {#snippet valueSnippet()}
+      <slot name="value-snippet"></slot>
+    {/snippet}
```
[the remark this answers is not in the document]
murdore: Checked this rather than reasoned about it, and the premise does not hold in practice.

`el.valueSnippet = fn` does not reach StatCard **at all** through the custom element — with or without this fix. I tested it with a real `createRawSnippet` against a built `sui-stat-card` (bundling it, because a bare `svelte` specifier will not resolve in a page), and the snippet never renders. Something further down the custom-element prop path drops it, so there is no live precedence conflict between a JS-assigned property and the light-DOM slot to resolve.

I had in fact claimed the opposite in the first version of this PR — that the guard made the JS-property path work. That was wrong, and the test is what caught it; the claim is now removed from the commit message, the PR body and the code comment.

Adding `props.valueSnippet()`-first precedence would therefore be writing a branch against a path that does not currently function, and would also invert the precedence every other guarded wrapper here uses (`hasXSlot ? impl : props.X`). Worth fixing, but as its own change once someone has established why the property is dropped.

src/wc/components/StatCard.wc.svelte
```diff
@@ -59,56 +57,66 @@
...
-  {/snippet}
-  {#snippet valueSnippet()}
-    <slot name="value-snippet">{@render valueFallback()}</slot>
-  {/snippet}
-  <slot></slot>
-</StatCard>
+<!--
+  Two branches rather than one conditional snippet prop, because a `{#snippet}`
+  declared at the top level of the template is hoisted to module scope while the
+  `<slot>` inside it compiles to `$.slot(node, $$props, ...)` -- and `$$props`
```
Tara-ag, line 90: 💡 MINOR: Duplicate StatCard invocation branches can be collapsed. The two blocks differ only by whether `valueSnippet` is passed, duplicating `headerRight`, `footer`, and default-slot wiring. `Card.wc.svelte` already uses the cleaner idiom: compute a snippet impl and pass it via a guarded prop (`valueSnippet={hasValueSnippetSlot ? valueSnippetImpl : props.valueSnippet}`), keeping one component call. Suggested change:

```svelte
{#snippet valueSnippetImpl()}<slot name="value-snippet"></slot>{/snippet}

<StatCard
  {...props}
  {...dispatchers}
  title={statCardTitle}
  valueSnippet={hasValueSnippetSlot ? valueSnippetImpl : props.valueSnippet}
>
  {#snippet headerRight()}<slot name="header-right"></slot>{/snippet}
  {#snippet footer()}<slot name="footer"></slot>{/snippet}
  <slot></slot>
</StatCard>
```

tests/statcard-wc-auto-parity.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+  }) => {
+    await loadBundle(page);
+
+    await page.evaluate(() => {
+      const el = document.createElement('sui-stat-card');
+      el.setAttribute('title', 'Total sales');
+      el.setAttribute('subtitle', '60000');
+      el.setAttribute('value', '194');
+      el.setAttribute('animate-primary', '');
+      el.setAttribute('primary', 'auto');
```
Tara-ag, line 62: 💬 SUGGESTION: Waiting until at least one `.animated-number` appears masks a zero-animation regression behind a timeout. Consider waiting for a stable state and then asserting `result.total.toBe(1)` directly, as the later tests do, so a failure reports the actual count rather than a generic wait timeout.

Tara-ag (commented):
Yama review: one MINOR and one SUGGESTION finding; no blocking issues.

src/wc/components/StatCard.wc.svelte
Tara-ag, line 90: 💡 MINOR: Duplicate StatCard invocation branches can be collapsed

The wrapper now has two nearly identical `<StatCard>` invocations that differ only by whether `valueSnippet` is passed. This duplicates the headerRight/footer/default-slot wiring and makes future edits error-prone. Card.wc.svelte already demonstrates the cleaner idiom: compute a snippet implementation per guarded slot and pass it via a single component call using a ternary (e.g. `valueSnippet={hasValueSnippetSlot ? valueSnippetImpl : props.valueSnippet}`). That keeps one source of truth for the component call and still lets StatCard render its own default branch when no slot is present.

**Suggested fix:**
```svelte
{#snippet valueSnippetImpl()}<slot name="value-snippet"></slot>{/snippet}

<StatCard
  {...props}
  {...dispatchers}
  title={statCardTitle}
  valueSnippet={hasValueSnippetSlot ? valueSnippetImpl : props.valueSnippet}
>
  {#snippet headerRight()}<slot name="header-right"></slot>{/snippet}
  {#snippet footer()}<slot name="footer"></slot>{/snippet}
  <slot></slot>
</StatCard>
```

This mirrors Card.wc.svelte, preserves the fix, and removes the duplicated component invocation.

tests/statcard-wc-auto-parity.spec.ts
Tara-ag, line 62: 💬 SUGGESTION: Test waits for any odometer before counting, which can mask zero-animation regressions

The first two tests wait until at least one `.animated-number` appears before counting. If a regression caused no values to animate at all, the test would time out rather than fail with a clear count assertion. A clearer pattern is to wait for the element to be connected/upgraded and then assert the final count directly, letting Playwright's default expectation timeout produce the failure. This is a test-style suggestion, not a correctness blocker.

**Suggested fix:**
Consider replacing the 'wait for >0 odometers' with a stable-state wait (e.g. `await page.waitForTimeout(800)` or a custom `waitForStableShadow`) and then assert `result.total.toBe(1)` directly, as the third and fourth tests already do.

Tara-ag (commented):
Reviewed the StatCard web-component parity fix. No blocking issues found; two non-blocking comments left.

src/wc/components/StatCard.wc.svelte
murdore: I tried the suggested shape verbatim rather than reasoning about it, and it is broken — silently.

Applying it exactly as written:

```svelte
{#snippet valueSnippetImpl()}<slot name="value-snippet"></slot>{/snippet}
<StatCard ... valueSnippet={hasValueSnippetSlot ? valueSnippetImpl : props.valueSnippet}>
```

`pnpm run build:wc` succeeds. Then `tests/statcard-wc-auto-parity.spec.ts` fails on 'slotted value-snippet content still replaces the built-in value':

```
Expected value: "slotted-value"
Received array: []
```

The consumer's slotted content is never assigned. A `{#snippet}` declared at template top level is hoisted to module scope, and the `<slot>` inside it compiles to `$.slot(node, $$props, …)` — `$$props` exists only inside the component function. No build error, no console error, just a card that quietly ignores what was slotted into it.

`PieChart.wc.svelte:88-101` carries the same branch duplication and documents the same reason. So the duplication is load-bearing, not accidental. I have put the measurement in the comment above the branches so the next person who sees the redundancy finds the evidence before removing it.

tests/statcard-wc-auto-parity.spec.ts
murdore: Agreed, and fixed in `ac4b374`.

Both `auto` tests now wait for the card to render (`[data-sc-slot]` present) rather than for an odometer to exist, then assert the count. You are right about the failure mode: gating on 'at least one animated' meant a regression that animated *nothing* would have surfaced as a generic 10s wait timeout instead of `expected 1, received 0` — the least informative possible failure for exactly the bug this file exists to catch.

Still 4 passed after the change.

Tara-ag (approved):
Reviewed the parity fix. The change correctly delegates the animation decision to the underlying `StatCard` component and only bridges `valueSnippet` when the consumer actually slotted content, restoring `primary="auto"` parity between Svelte and web-component builds. Tests cover the regression cases and the slot-guard interaction. No blocking issues found.

PR #643, opened 2026-09-19
chore(lint): catch dead imports in web-component wrappers

murdore:
## Summary

- lint `src/wc/**` for dead imports even though the main ESLint config excludes the wrapper tree
- keep this deliberately narrow instead of turning on the full main ruleset, which currently reports 106 unrelated wrapper errors
- run the focused check in the existing `lint` chain

## Why

The StatCard parity follow-up removed wrapper-side animation code but left an unused `AnimatedNumber` import. CI did not catch it because `src/wc/**` is globally ignored. This adds the smallest rule that would have caught that regression.

## Verification

- `pnpm run lint:wc-unused`
- `pnpm exec prettier --check eslint.wc.config.js package.json`
- `pnpm run check:wc` (0 errors; 8 pre-existing warnings)
- red-check: inserting an unused import makes `lint:wc-unused` fail with one error

## Summary by CodeRabbit

- **Chores**
  - Added linting coverage for unused imports in web component TypeScript and Svelte files.
  - Integrated the new check into the standard lint command.


Tara-ag (commented):
Review in progress

package.json
```diff
@@ -33,7 +33,7 @@
     "codemod": "node scripts/codemod/cli.ts",
     "check:wc": "svelte-check --tsconfig ./tsconfig.wc.json --compiler-warnings \"options_missing_custom_element:ignore\"",
     "check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
-    "lint": "prettier --check src scripts tests && eslint src scripts tests && npm run lint:event-casing && npm run lint:shadow-safety && npm run lint:css-contract && npm run lint:docs-contract && npm run lint:wc-contract && npm run lint:test-focus && npm run lint:focus-visible",
+    "lint": "prettier --check src scripts tests && eslint src scripts tests && npm run lint:event-casing && npm run lint:shadow-safety && npm run lint:css-contract && npm run lint:docs-contract && npm run lint:wc-contract && npm run lint:wc-unused && npm run lint:test-focus && npm run lint:focus-visible",
```
Tara-ag: 💬 SUGGESTION: Consider moving `npm run lint:wc-unused` earlier in the lint chain (e.g. right after `eslint src scripts tests`). Dead-import detection is cheap and failing fast avoids running the heavier contract checkers on code that already has hygiene issues.

package.json
```diff
@@ -50,6 +50,7 @@
     "lint:css-contract": "node scripts/check-css-contract.js",
     "lint:docs-contract": "node scripts/check-docs-contract.js",
     "lint:wc-contract": "node scripts/check-wc-contract.js",
+    "lint:wc-unused": "eslint --no-ignore --config eslint.wc.config.js 'src/wc/**/*.svelte' 'src/wc/**/*.ts'",
     "lint:test-focus": "node scripts/check-test-focus.js",
```
Tara-ag, line 54: 💡 MINOR: The `lint:wc-unused` script uses single-quoted globs (`'src/wc/**/*.svelte'`). On Windows shells, single quotes are not stripped, so ESLint receives literal quote characters in the path and fails to find files. Since this is a published cross-platform package, make the script portable.

Suggestion: rely on the config's own `files` array and run `eslint --no-ignore --config eslint.wc.config.js src/wc`, or use escaped double quotes for Windows compatibility.

Tara-ag (commented):
Reviewed. Two minor/suggestion items raised; no blocking issues found.

Tara-ag (approved):
Approved. The change adds a narrowly-scoped ESLint config that catches dead imports in `src/wc/**` without pulling in the full main ruleset, and wires it into the existing `lint` chain. The config correctly restricts itself to `unused-imports/no-unused-imports`, registers `@typescript-eslint` only to satisfy load-bearing disable comments, and uses `--no-ignore` to bypass the main config's wholesale `src/wc/**` ignore. No public API, accessibility, reactivity, SSR, or security concerns.

Tara-ag (approved):
Clean, focused tooling change. The separate ESLint config is the right way to un-ignore `src/wc/**` for a single rule without pulling in the 106 existing wrapper violations. Registering `@typescript-eslint` purely to satisfy existing disable comments, and turning off `reportUnusedDisableDirectives`, are both sensible given the narrow scope. The new `lint:wc-unused` step is correctly inserted into the existing lint chain and follows the established naming convention. No public API, component, accessibility, or security impact.

package.json
murdore: Implemented locally: `lint:wc-unused` now runs immediately after the main ESLint pass, before the contract checkers. Full `pnpm run lint` passes. This will be included in the amended single commit.

package.json
murdore: Implemented locally: the script now invokes `eslint --no-ignore --config eslint.wc.config.js src/wc`, relying on the config’s own file globs and avoiding shell quoting differences on Windows. Full `pnpm run lint` passes. This will be included in the amended single commit.

package.json
```diff
@@ -50,6 +50,7 @@
     "lint:css-contract": "node scripts/check-css-contract.js",
     "lint:docs-contract": "node scripts/check-docs-contract.js",
     "lint:wc-contract": "node scripts/check-wc-contract.js",
+    "lint:wc-unused": "eslint --no-ignore --config eslint.wc.config.js src/wc",
```
Tara-ag: 💡 MINOR: The `lint:wc-unused` script passes `--no-ignore` to ESLint. In flat config this disables all ignore patterns, including `.gitignore` exclusions loaded elsewhere. Since the config already scopes files to `src/wc/**`, the `--no-ignore` is broader than needed and could cause ESLint to descend into ignored directories under `src/wc`. Prefer a precise glob instead:

```json
"lint:wc-unused": "eslint --config eslint.wc.config.js \"src/wc/**/*.{ts,svelte}\""
```

This keeps `.gitignore` respected while still targeting only the wrapper files.

eslint.wc.config.js
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * So this config runs THAT ONE RULE, and nothing else, over the directory the
+ * main config skips. It is deliberately not a step toward linting `src/wc/**`
+ * properly; it is the narrow rule that pays for itself today.
+ */
+import ts from 'typescript-eslint';
+import svelte from 'eslint-plugin-svelte';
+import unusedImports from 'eslint-plugin-unused-imports';
+import svelteConfig from './svelte.config.js';
+
```
Tara-ag, line 28: 💬 SUGGESTION: Turning off `reportUnusedDisableDirectives` is correct for a single-rule config, but if this config ever expands, stale `eslint-disable` comments will silently accumulate. Consider adding a note that this should be revisited if additional rules are enabled.

Tara-ag (approved):
Light review: the change is a focused tooling addition that catches dead imports in web-component wrappers. I verified the new config structure, the `lint` chain integration, and spot-checked several wrapper files for unused imports (none found in the sample). CI build/visual checks passed; `checks` and Yama review are still running.

Two minor/tooling comments inline — neither blocks merge. Approving.

eslint.wc.config.js
Tara-ag, line 28: 💬 SUGGESTION: The config turns off `reportUnusedDisableDirectives` because only one rule is enabled. That's reasonable, but if the config is ever expanded, stale disable comments will silently accumulate. A note explaining this as a deliberate narrow-config choice would help future maintainers.

Suggestion: Add a one-line comment noting that `reportUnusedDisableDirectives: 'off'` should be revisited if additional rules are ever enabled in this config.

package.json
Tara-ag: 💡 MINOR: The `lint:wc-unused` script passes `--no-ignore` to ESLint. In flat config, `--no-ignore` disables all ignore patterns, including those loaded from `.gitignore` via `@eslint/compat/includeIgnoreFile`. While the config scopes files to `src/wc/**`, ESLint still walks the directory tree from `src/wc` and may descend into ignored folders. This is broader than necessary for catching dead imports in wrappers.

Suggestion: Replace `--no-ignore` with a precise glob, e.g. `"lint:wc-unused": "eslint --config eslint.wc.config.js \"src/wc/**/*.{ts,svelte}\""`, so `.gitignore` exclusions are still honored while the config's own `files` patterns handle scoping.

Tara-ag (approved):
Approved. The PR adds a focused dead-import lint check for `src/wc/**` and wires it into the lint chain. No blocking issues found. Two minor/tooling comments posted about `--no-ignore` usage and documenting the `reportUnusedDisableDirectives` choice.

package.json
murdore: Dropped, and the premise turned out to be even simpler than stated: `--config` REPLACES the main config, so its wholesale `src/wc/**` ignore is never loaded and there was nothing to bypass. Measured both forms against an injected dead import in `src/wc/index.ts` and `StatCard.wc.svelte` — byte-identical output, 2 errors each — so the flag only widened what ESLint would descend into. The script is now `eslint --config eslint.wc.config.js src/wc`, re-verified with the same injection, and the reason is recorded in the config header.

eslint.wc.config.js
murdore: Added, next to the setting rather than in the header so it is read at the point of change: the suppression holds only while this config stays single-rule, and anything else enabled here must turn `reportUnusedDisableDirectives` back on at the same time — otherwise a genuinely stale disable comment starts passing silently.

Tara-ag (approved):
Approved. The change adds a narrow, well-rationed ESLint config that catches dead imports in `src/wc/**` without attempting to lint the full wrapper tree. No component code, public API, exports, or CSS custom properties are touched, so there are no backward-compatibility concerns. The new script is wired into the existing `lint` chain and CI will exercise it.

PR #644, opened 2026-09-20
fix(stat-card): support value snippets through the WC bundle

murdore:
## The contradiction this resolves

4.31.1 shipped a comment in `StatCard.wc.svelte` saying a JS-assigned
`valueSnippet` "still does not render" and that "something further down the
custom-element prop path drops it", left unchased.

That was wrong, and the wrong half matters: the property path is fine. The
generated custom-element setter forwards the function exactly as written —

```js
set(n) {
  n = get_custom_element_value(e, n, t);
  this.$$d[e] = n;
  var i = this.$$c;
  i && (get_descriptor(i, e)?.get ? i[e] = n : i.$set({ [e]: n }));
}
```

What actually fails is narrower and more interesting: the standalone WC bundle
embeds **its own Svelte runtime**. A snippet built by `createRawSnippet` from a
consumer's *separate* `svelte` import closes over different runtime state, so
rendering it inside this bundle dies with

```
TypeError: Cannot read properties of undefined (reading 'call')
```

The earlier investigation missed this because it tested with a hand-written
function shaped like a compiled snippet — which proved the property is
forwarded, and exercised none of the runtime internals that actually break.

## The fix

The WC entry exports the factory bound to the runtime it ships:

```ts
export { createRawSnippet } from 'svelte';
```

so a consumer writes

```js
import { createRawSnippet } from '@juspay/svelte-ui-components/wc';
card.valueSnippet = createRawSnippet(() => ({ render: () => '<strong>₹1.23Cr</strong>' }));
```

`build:wc` now emits a declaration that says so, instead of `export {}`.

No wrapper logic changed. The two-branch `<StatCard>` shape stays exactly as it
is — collapsing it silently drops slotted content, and the parity spec that
catches that is untouched and still passing.

## Tests

Red first, for the right reason: both new cases failed on
`the WC entry must export its matching snippet factory` **before** the export
existed, against a real cross-bundle `createRawSnippet`.

Six cases now pass (`6 passed`):

- `auto` + larger subtitle animates ONE value
- `auto` + larger value animates the VALUE
- `animate-value` still animates through the element
- slotted `value-snippet` still replaces the built-in value
- **pre-connect** `valueSnippet` from the WC-owned factory renders
- **post-connect** `valueSnippet` from the WC-owned factory replaces the value

`pnpm run check` and `pnpm run lint` both exit 0, re-run after rebasing onto the
`#643` merge so the newly-added `lint:wc-unused` gate runs over the `src/wc`
files this branch touches.

## Also corrected

`docs/CHANGELOG.md` still asserted the old "does not render / not chased here"
claim and said there were four browser cases. Both are now accurate.

## Rows are unchanged, and that is deliberate

With `rows.length > 0`, StatCard renders its row layout instead of the
single-value area, so `valueSnippet` is not rendered there. That is existing
component behaviour, not a gap this introduces; `docs/StatCard.md` now states it
rather than leaving it to be rediscovered.

## Summary by CodeRabbit

- **New Features**
  - Web Component consumers can create JavaScript-assigned `valueSnippet` content using the exported snippet factory.
  - Snippets now render whether assigned before or after the StatCard connects.
  - The existing `value-snippet` light-DOM slot remains available for plain HTML content.

- **Documentation**
  - Added guidance for using JavaScript-assigned snippets and clarified that row layouts replace the single-value area.

- **Tests**
  - Added browser coverage for snippet assignment before and after connection.


Tara-ag (approved):
Approved. The change is additive (new `createRawSnippet` re-export from the WC entry), well-explained, covered by browser tests, and documented. No public API breakage, no accessibility or lifecycle regressions, and no security concerns.

PR #646, opened 2026-09-21
feat(motion): unified motion token system, tokenized overlay transitions

murdore:
## Summary

Builds the motion token system for this library: a named semantic tier (`--duration-*`, `--ease-*`, `--distance-*`) sitting between each component's own override token and the existing `--motion-duration`/`--motion-easing` root, anchored on values this library's own AI/chat components (ThinkingIndicator, ToolCallLog, TaskList, ChatBubble) already proved out, cross-checked against transitions.dev's published motion reference rather than copied from it.

- **Modal/Sheet/CommandMenu/Menu/ContextMenu**: tokenized via a new shared `tokenizedFly.ts` helper; entrance travel distance cut from 300–500px to **60px** (frame-scrubbed against several candidates before landing here — reads as "sliding into place," not "flying in"). CommandMenu/Menu/ContextMenu gain a tokenized entrance where none existed before. All five keep Svelte's `in:`/`out:` transition directives — they share a global scroll-lock reference count and dismissal-owner stack, so the outro-wait those directives provide is load-bearing for cleanup timing, not incidental.
- **Toast**: fully migrated to CSS-native `@starting-style` + `allow-discrete` (zero animation JavaScript). Fixes a real, empirically-confirmed bug as a side effect — the old reduced-motion check was a `$derived` that read a plain, non-reactive function, so toggling the OS preference mid-session could leave it stale. A CSS media query can't have that class of bug.
- **ThinkingIndicator/ToolCallLog/TaskList/ChatBubble**: converged onto one shared `--ease-smooth-out` curve (the two candidate curves were frame-scrubbed and found visually identical — least-churn convergence, not an arbitrary pick).
- **Chart family + ~18 more components**: adopted the new tier additively — every change here is a no-op until a consumer actually sets a token; no literal value changed.
- Two standalone bugs fixed: MediaUpload's reduced-motion guard targeted a class not in its markup; RatingGroup's hover scale had no `transition` at all.
- Three pre-existing repo-scanning tests given this repo's own established `REPO_SCAN_TIMEOUT_MS` headroom, and the visual suite's cold-build webServer timeout raised to 360s — both surfaced by this branch adding enough content to tip existing budgets over, not caused by the motion work itself.

Full design rationale lives in `DESIGN_PRINCIPLES.md`'s motion section (first commit).

## Test plan

- [x] `pnpm run check` — clean (svelte-check + WC + codemod + migrate + release + wc-parity)
- [x] `pnpm run lint` — clean (prettier, eslint, and every custom contract check)
- [x] `pnpm exec vitest run` — 2355 existing tests + 21 new (`tokenizedFly.test.ts`, covering the reduced-motion backstop and the bezier easing parser) — 0 failures
- [x] `pnpm run test:visual` (Docker-pinned Playwright) — 94 passed, 5 re-confirmed clean in an isolated re-run after flagging as end-of-run resource fatigue in the full sequential run (not pixel diffs), 1 pre-existing documented exclusion. Zero unexpected visual regressions across the whole library.
- [x] Live-browser spot checks (Modal, Sheet, Toast, CommandMenu) — correct open/close, correct scroll-lock engage/release, zero console errors, and a mid-transition capture on Toast directly confirming the 60px distance and direction resolve correctly at runtime.

## Summary by CodeRabbit

* **New Features**
  * Added centralized motion customization through CSS duration, distance, easing, and scale tokens.
  * Expanded transition theming across menus, dialogs, sheets, charts, forms, navigation, notifications, and other components.
  * Added configurable entry and exit animations for panels and overlays, including support for custom easing and travel distances.
  * Improved reduced-motion behavior with effectively instant or opacity-only transitions where appropriate.

* **Bug Fixes**
  * Toast hide events now fire after the closing transition completes.
  * Corrected reduced-motion styling for media upload cards.

* **Documentation**
  * Added motion guidelines and documented new transition customization variables across component references.


Tara-ag:

PR #646 introduces a unified motion-token tier and migrates overlay/toast transitions to CSS-native/tokenized transitions. No hardcoded secrets, unsafe HTML, or obvious breaking public-API removals were found in the time available. The new tokens fall back to the existing --motion-duration/--motion-easing roots, preserving downstream themes. Toast’s lifecycle was updated from Svelte outroend to transitionend and the matching WC test was adjusted. I could not fully gate every changed file (61 files) or run submit_review before time ran out, so the issues list is empty; the items below are observations, not gated findings. Consider adding dedicated unit tests for tokenizedFly/OverlayAnimation and confirming that the 0.001s reduced-motion duration reliably fires transitionend in all supported browsers.

src/lib/Toast/Toast.svelte
```diff
@@ -214,12 +204,112 @@
        area below, and consumers with actionable bottomContent can opt the
```
Tara-ag: 🔒 CRITICAL: Breaking change: Toast default animation distances silently changed. The old implementation defaulted to 500px (overlap) / 20px (non-overlap vertical) / 100px (top-to-bottom overlap out), but the new CSS-native rules fall back to `--distance-overlay` (60px). Consumers who did not set `inAnimationOffset`/`outAnimationOffset` will see a much shorter travel. This is a visible behavioral change to a public component without a major-version bump or migration note.

Fix: restore the previous default distances as literal fallbacks in the direction rules, e.g. `--toast-enter-x: calc(-1 * var(--toast-distance, 500px));`, or document this as a breaking change and bump the major version.

